Issue #44, opened 2021-02-01
Generic pipeline for importing Vulnerabilities.


oliverchang:
Extend the current OSS-Fuzz pipeline to support generic data sources.

- Input: Partially filled [Vulnerability](https://osv.dev/docs/#tag/vulnerability_schema) data from a repository.
- Output: Vulnerability with appended affected commit range and versions info pushed back to the same repository and location. 

More requirements:
- Parts of OSV will also be made to run as part of CI workflows (#51)
- Users can also submit PRs to fix up details in Vulnerability entries.
- OSS-Fuzz data will be dumped as well in the same format.

Current sketch (subject to change):

![Architecture](https://user-images.githubusercontent.com/759062/106834907-681c5600-66ea-11eb-85ce-eb359e32f84f.png)

westurner:
This may be a duplicate? 
"Existing artifact vuln scanners, databases, and specs?" #55

PR #65, opened 2021-02-16
Plumbing in preparation for processing user changes.

oliverchang:
- Start plumbing for detecting user changes and kicking off analysis tasks.
- Don't make empty commits when importing new OSS-Fuzz entries.

Part of #44.

### DIFF
--- a/docker/importer/importer.py
+++ b/docker/importer/importer.py
@@ -60,7 +60,7 @@ def _yaml_str_representer(dumper, data):
 def _repo_path(repo):
   """Return local disk path to repo."""
   # Remove '.git' component.
-  return os.path.dirname(repo.path.rstrip('/'))
+  return os.path.dirname(repo.path.rstrip(os.sep))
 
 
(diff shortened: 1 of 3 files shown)


docker/importer/importer.py
```diff
@@ -55,6 +57,22 @@ def _yaml_str_representer(dumper, data):
   return dumper.represent_scalar('tag:yaml.org,2002:str', data)
 
 
+def _repo_path(repo):
+  """Return local disk path to repo."""
+  # Remove '.git' component.
+  return os.path.dirname(repo.path.rstrip('/'))
```
jonathanmetzman, line 63: Is this rstrip needed even though we are using dirname?

jonathanmetzman (approved):
LGTM

docker/importer/importer.py
oliverchang: Yeah, pygit's Repository.path gives us `/path/.git/` and we need to strip it or we'll still have the `.git` in our path.

PR #67, opened 2021-02-18
Tweak the vulnerability schema.

oliverchang:
This is in progress and temporarily served from /v1new.

- Hoist out some fields for affected ranges to reduce duplication.
- Generalize "cves" to "aliases".
- Add an output-only lastModified field.


inferno-chromium (approved):
Thanks for your super quick response!

oliverchang:
FYI more changes were requested. See latest commit.

inferno-chromium (approved):
lgtm, 

i am even fine with just removing v1new and keeping just v1. can do later (tbr) whenever.

oliverchang:
> lgtm,
> 
> i am even fine with just removing v1new and keeping just v1. can do later (tbr) whenever.

Need to coordinate some migration details, so holding off on that for now.

PR #73, opened 2021-03-04
Factor out some common test setup functions.

oliverchang:


jonathanmetzman (approved):
LGTM

lib/osv/tests.py
```diff
@@ -12,17 +12,46 @@
...
+
+  def __init__(self, path):
+    self.path = path
+    self._repo = pygit2.init_repository(path, True)
+    tree = self._repo.TreeBuilder().write()
+    author = pygit2.Signature('OSV', 'infra@osv.dev')
+    self._repo.create_commit('HEAD', author, author, 'Initial commit', tree, [])
+
+  def add_file(self, path, contents):
+    """Add a file."""
```
jonathanmetzman, line 42: Super nit: I think the style guide says to make function docstrings in descriptive style ("Adds a file").

lib/osv/tests.py
```diff
@@ -12,17 +12,46 @@
...
+    oid = self._repo.write(pygit2.GIT_OBJ_BLOB, contents)
+    self._repo.index.add(pygit2.IndexEntry(path, oid, pygit2.GIT_FILEMODE_BLOB))
+    self._repo.index.write()
+
+  def commit(self, author_name, author_email):
+    """Commit."""
```
jonathanmetzman, line 48: nit: Maybe "makes a commit."

jonathanmetzman (approved):

lib/osv/tests.py
oliverchang: Done and fixed every case in this file.

lib/osv/tests.py
oliverchang: Done!

PR #75, opened 2021-03-05
Check for hash change when processing updates.

oliverchang:
If the hash changed, bail and let a later task perform the update.

### DIFF
--- a/lib/osv/sources.py
+++ b/lib/osv/sources.py
@@ -14,11 +14,14 @@
 # limitations under the License.
 """Importer sources."""
 
-from google.protobuf import json_format
+import hashlib
+import logging
 import os
 import pygit2
 import time
 import yaml
+
+from google.protobuf import json_format
 
 # pylint: disable=relative-beyond-top-level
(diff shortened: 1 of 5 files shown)


lib/osv/sources.py
```diff
@@ -173,3 +188,11 @@ def push_source_changes(repo, commit_message, git_callbacks):
 def git_author():
   """Get the git author for commits."""
   return pygit2.Signature('OSV', AUTHOR_EMAIL)
+
+
+def sha256(yaml_path):
+  """Computes sha256 sum."""
+  hasher = hashlib.sha256()
+  with open(yaml_path, 'rb') as f:
```
inferno-chromium, line 196: should it check if file exists, and return none, logerror if not ?

inferno-chromium (approved):

lib/osv/sources.py
oliverchang: This will be up to the caller. We should be getting an exception if this is getting called with a non-existent path.

PR #92, opened 2021-03-24
Move repo-specific functions to repos module.

oliverchang:
Also fix worker update dealing with existing source repo checkouts and deal
with repos with changed git URLs

### DIFF
--- a/lib/osv/repos.py
+++ b/lib/osv/repos.py
@@ -1,0 +1,77 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


lib/osv/repos.py
```diff
@@ -0,0 +1,77 @@
+# Copyright 2021 Google LLC
```
inferno-chromium, line 1: nit: Maybe repo_manger is a slightly better name ?

inferno-chromium (approved):

lib/osv/repos.py
oliverchang: This is more consistent with other existing modules, e.g. "sources.py" instead of "source_manager.py"

jonathanmetzman:
LGTM

PR #94, opened 2021-03-25
Export all Vulnerability JSON data to a GCS bucket.

oliverchang:
Including failed analyses.

Also include invalid bugs in bugs to make public as well.


jonathanmetzman (approved):
LGTM

PR #99, opened 2021-03-31
Allow empty introduced commit fields.

oliverchang:
Assume all commits prior to the fix are vulnerable.

### DIFF
--- a/docker/worker/testdata/expected_127.diff
+++ b/docker/worker/testdata/expected_127.diff
@@ -1,0 +1,19 @@
+diff --git a/BLAH-127.yaml b/BLAH-127.yaml
+index 915f291..c6a4223 100644
+--- a/BLAH-127.yaml
++++ b/BLAH-127.yaml
+@@ -12,5 +12,14 @@ affects:
+   - type: GIT
+     repo: https://osv-test/repo/url
+     fixed: 8d8242f545e9cec3e6d0d2e3f5bde8be1c659735
++  - type: GIT
++    repo: https://osv-test/repo/url
++    fixed: b9b3fd4732695b83c3068b7b6a14bb372ec31f98
++  versions:
++  - branch-v0.1.1
++  - branch_1_cherrypick_regress
++  - v0.1
++  - v0.1.1
+ references:
+ - https://ref.com/ref
++modified: '2021-01-01T00:00:00Z'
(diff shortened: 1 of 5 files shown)


lib/osv/impact.py
```diff
@@ -123,8 +123,12 @@ def get_affected(repo, regress_commit_or_range, fix_commit_or_range):
   for commit in regress_commits:
     tags_with_bug.update(get_tags_with_commits(repo, [commit]))
 
-  tags_with_fix = get_tags_with_commits(repo, fix_commits)
+  if not regress_commits:
+    # If no introduced commit provided, assume all commits prior to fix are
+    # vulnerable.
+    tags_with_bug.update(get_all_tags(repo))
```
inferno-chromium, line 129: aren't we adding all tags and not just tags before fix ?

docker/worker/testdata/expected_127.diff
```diff
@@ -0,0 +1,19 @@
+diff --git a/BLAH-127.yaml b/BLAH-127.yaml
```
inferno-chromium, line 1: At some point, we should think of renaming these test files to be more readable, like instead of _127, it could be _no_regress_commit ? can be another cl later at some point.

lib/osv/impact.py
```diff
@@ -163,7 +167,8 @@ def get_affected_range(repo, regress_commits, fix_commits):
       if equivalent_regress_commit:
         break
 
-    if not equivalent_regress_commit:
+    if not equivalent_regress_commit and regress_commits:
```
inferno-chromium, line 170: This condition is unclear, why add regress_commits ? maybe add more in comment.

inferno-chromium (approved):

docker/worker/testdata/expected_127.diff
oliverchang: ack.

lib/osv/impact.py
oliverchang: Yes, but then we compute `tags_with_bug - tags_with_fix` to get the actual list of tags.

lib/osv/impact.py
oliverchang: I did add comment here but I'll clarify it a bit more.

oliverchang (commented):
See https://github.com/google/oss-fuzz-vulns/commit/5a1e660f6e8ddd3d3db513f976f4987287fc258e for a real example.

jonathanmetzman (commented):
LGTM

PR #104, opened 2021-04-09
vulnfeeds: Get list of upstream versions and correlate extracted versions.

oliverchang:
Extract a list of upstream versions and use this to correlate with the
versions we extract from the CVE to ensure accuracy.

Currently just prints if we get a mismatch.


inferno-chromium:
Adding @asraa team member for review mix, please feel free to assign back if code is confusing. This code is for generating python vulnerability database from cve data feed.

vulnfeeds/cves/versions.go
```diff
@@ -62,7 +62,7 @@ func extractGitHubCommit(link string) *FixCommit {
 	}
 }
 
-func ExtractVersionInfo(cve CVEItem) VersionInfo {
+func ExtractVersionInfo(cve CVEItem, validVersions []string) VersionInfo {
```
asraa, line 65: is validVersions meat to be used later? (if so, TODO?)

vulnfeeds/pypi/pypi.go
```diff
@@ -89,14 +118,34 @@ func NewPyPI(pypiJSON string) *PyPI {
...
+// processVersions takes a pypi_versions.json and returns a map of packages to versions.
+func processVersions(versionsSource []pypiVersions) map[string][]string {
+	versions := map[string][]string{}
+	for _, data := range versionsSource {
+		versions[data.Name] = data.Versions
+	}
+	return versions
+}
+
+func New(pypiLinksJSON string, pypiVersionsJSON string) *PyPI {
```
asraa, line 133: love the cleanup!

asraa (commented):
I can't comment too much about the Go code, but thanks!

vulnfeeds/cves/versions.go
oliverchang: Yep! I had some TODOs already in this function but I'll clarify them to mention this variable explicitly.

asraa (approved):
LGTM!
(adding @inferno-chromium, because I am unsure of my Go style reviewability)

PR #105, opened 2021-04-12
Improve CVE version extraction.

oliverchang:
- Handle versionStartExcluding and versionEndIncluding by using
  the list of actual package versions to derive "introduced" and "fixed".
- Extract versions from description as a last resort.

### DIFF
--- a/vulnfeeds/cves/versions.go
+++ b/vulnfeeds/cves/versions.go
@@ -78,13 +78,19 @@ func versionIndex(validVersions []string, version string) int {
 }
 
 func nextVersion(validVersions []string, version string) string {
-	idx := versionIndex(validVersions, version) + 1
-	if idx < len(validVersions) {
-		return validVersions[idx]
+	idx := versionIndex(validVersions, version)
+	if idx == -1 {
+		log.Printf("Warning: %s is not a valid version", version)
+		return ""
 	}
 
-	log.Printf("Warning: %s is not a valid version", version)
-	return ""
+	idx += 1
+	if idx >= len(validVersions) {
+		log.Printf("Warning: %s does not have a version that comes after.", version)
+		return ""
+	}
+
+	return validVersions[idx]
 }
 
 func processExtractedVersion(version string) string {
--- a/vulnfeeds/cves/versions.go
+++ b/vulnfeeds/cves/versions.go
@@ -15,7 +15,9 @@
 package cves
 
 import (
+	"log"
 	"net/url"
+	"regexp"
 	"strings"
 )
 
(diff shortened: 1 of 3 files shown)


vulnfeeds/cves/versions.go
```diff
@@ -71,30 +141,55 @@ func ExtractVersionInfo(cve CVEItem, validVersions []string) VersionInfo {
...
-					// TODO: make sure these match validVersions.
-					Introduced: match.VersionStartIncluding,
-					Fixed:      match.VersionEndExcluding,
-				})
+			introduced := ""
+			fixed := ""
+			if match.VersionStartIncluding != "" {
+				introduced = match.VersionStartIncluding
+			} else if match.VersionStartExcluding != "" {
+				introduced = nextVersion(validVersions, match.VersionStartExcluding)
```
asraa, line 160: if the cpe match is not a valid version, then this returns the first version valid version. Want to confirm this is by design.

for e.g. the vuln started way before and now there are only new valid versions -> all are affected.

vulnfeeds/cves/versions.go
```diff
@@ -62,6 +64,74 @@ func extractGitHubCommit(link string) *FixCommit {
...
+	return -1
+}
+
+func nextVersion(validVersions []string, version string) string {
+	idx := versionIndex(validVersions, version) + 1
+	if idx < len(validVersions) {
+		return validVersions[idx]
+	}
+
+	log.Printf("Warning: %s is not a valid version", version)
```
asraa, line 86: this is a confusing error message, since versions that are not in `validVersions` have a `nextVersion` of `validVersions[0]`

Should this be something like "Warning: next version after %s is not a valid version"

vulnfeeds/cves/versions.go
```diff
@@ -71,30 +141,55 @@ func ExtractVersionInfo(cve CVEItem, validVersions []string) VersionInfo {
...
+			} else if match.VersionEndIncluding != "" {
+				fixed = nextVersion(validVersions, match.VersionEndIncluding)
 			}
+
+			if introduced == "" && fixed == "" {
+				continue
+			}
+
+			if introduced != "" && !hasVersion(validVersions, introduced) {
+				log.Printf("Warning: %s is not a valid introduced version", introduced)
```
asraa, line 174: in these cases should we still consider that we "Got the version"?

asraa (commented):
Thanks! A few questions on my understanding of what's going on

vulnfeeds/cves/versions.go
oliverchang: Woops, that's a mistake. If `version` does not exist, we should return `""` instead. I fixed this and made error message clearer.

Good catch, thanks!

vulnfeeds/cves/versions.go
oliverchang: As above this is wrong and fixed.

vulnfeeds/cves/versions.go
oliverchang: Yes, there will always still be a human element involved. This will flag cases in the future that require a set of eyes to review.

oliverchang (commented):
thanks!

inferno-chromium (approved):

PR #108, opened 2021-04-13
vulnfeeds: Write a .notes file when human intervention is needed.

oliverchang:
Convert previous warnings into an of notes which are accumulated.
These are then written alongside the .yaml to indicate that human eyes
are required.

Also implement a false positives input file to allow users to ignore
such cases in the future.

Document the workflow as well for human triagers.

### DIFF
--- a/vulnfeeds/README.md
+++ b/vulnfeeds/README.md
@@ -14,3 +14,58 @@ false positives.
 ## Extracting affected versions and commits
 Where possible, we try to extract affected version ranges from descriptions and
 other fields, and extract commit hashes from e.g. GitHub links.
+
+Extracted version numbers are cross referenced against the list of actual
+versions released in the package manager to ensure accuracy.
+
+## Workflow
+
+The intended workflow is for a human to periodically run the tools here against
+the latest NVD CVE feeds.
+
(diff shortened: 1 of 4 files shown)


oliverchang:
See the `README.md` changes in this PR for the proposed workflow for triagers.

vulnfeeds/README.md
```diff
@@ -14,3 +14,58 @@ false positives.
...
+False positives (i.e. CVEs that are incorrectly matched to a package) can be
+added into a `false_positives.txt` file and commited as part of the
+vulnerability repo for future runs.
+
+The format of this file is newline delimited CVE IDs.
+
+e.g.
+
+```
+CVE-2021-1337
```
asraa, line 68: maybe eventually this can hold corrected packages and a reverse map can be loaded and checked below?

asraa (approved):
LGTM!

vulnfeeds/README.md
oliverchang: Most of the time an incorrect match means we shouldn't be matching against that package at all, rather there being a more correct one.

We can definitely iterate here in the future though!

PR #111, opened 2021-04-14
vulnfeeds: Improve Python vuln matching.

oliverchang:
- Add a new heuristic to extract "vendor/product" from GitHub links
  and cross reference that against CPEs.

- Also fix a bug with PyPI link parsing.

### DIFF
--- a/vulnfeeds/cves/versions.go
+++ b/vulnfeeds/cves/versions.go
@@ -218,3 +218,18 @@ func ExtractVersionInfo(cve CVEItem, validVersions []string) (VersionInfo, []str
 	}
 	return v, notes
 }
+
+func CPEs(cve CVEItem) []string {
+	cpesMap := map[string]bool{}
+	for _, node := range cve.Configurations.Nodes {
+		for _, match := range node.CPEMatch {
+			cpesMap[match.CPE23URI] = true
+		}
+	}
+
+	var cpes []string
+	for cpe := range cpesMap {
+		cpes = append(cpes, cpe)
+	}
+	return cpes
+}
--- a/vulnfeeds/pypi/pypi.go
+++ b/vulnfeeds/pypi/pypi.go
@@ -198,6 +198,10 @@ func (p *PyPI) Matches(cve cves.CVEItem) string {
 	}
 
 	cpe := strings.Split(cpes[0], ":")
+	if len(cpe) < 5 {
+		return ""
+	}
+
 	vendorProduct := cpe[3] + "/" + cpe[4]
 	if pkgs, exists := p.vendorProductToPkg[vendorProduct]; exists {
 		for _, pkg := range pkgs {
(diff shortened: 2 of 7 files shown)


vulnfeeds/cves/versions.go
```diff
@@ -218,3 +218,18 @@ func ExtractVersionInfo(cve CVEItem, validVersions []string) (VersionInfo, []str
...
+
+func CPEs(cve CVEItem) []string {
+	cpesMap := map[string]bool{}
+	for _, node := range cve.Configurations.Nodes {
+		for _, match := range node.CPEMatch {
+			cpesMap[match.CPE23URI] = true
+		}
+	}
+
+	var cpes []string
```
asraa, line 230: you could pre-allocate the size of this slice with `len(cpesMap)`, i think (but avoid allocations vs readability! your choice!)

vulnfeeds/pypi/pypi.go
```diff
@@ -134,25 +167,45 @@ func New(pypiLinksPath string, pypiVersionsPath string) *PyPI {
...
-		if pkg := p.matchesPackage(reference.URL, desc); pkg != "" {
+		// Otherwise try to cross-reference the link against our set of known links.
+		if pkg := p.matchesPackage(reference.URL, cve.CVE); pkg != "" {
 			return pkg
 		}
 	}
+
+	// As a last resort, extract the vendor and product from the CPE and try to match that
+	// against vendor/product combinations extracted from e.g. GitHub links.
+	cpes := cves.CPEs(cve)
```
asraa, line 195: Do you anticipate using this function elsewhere, or maybe you only need a single CPE?

vulnfeeds/pypi/pypi.go
```diff
@@ -134,25 +167,45 @@ func New(pypiLinksPath string, pypiVersionsPath string) *PyPI {
...
+	if len(cpes) == 0 {
+		return ""
+	}
+
+	cpe := strings.Split(cpes[0], ":")
+	vendorProduct := cpe[3] + "/" + cpe[4]
```
asraa, line 201: maybe handle some kind of out of range? not sure if the likelihood processing went wrong is so low though

asraa (commented):
Thanks!

vulnfeeds/cves/versions.go
oliverchang: Right, but this isn't exactly performance critical so I think we can avoid this for readability :)

vulnfeeds/pypi/pypi.go
oliverchang: Done!

vulnfeeds/pypi/pypi.go
oliverchang: I want to keep the cves package general since that might be used by others, even if we don't use it anywhere else yet. 

A lot of this might end up also end up getting refactored later as we add support for more package ecosystems in the future to share more logic.

oliverchang (commented):
thanks!

asraa (approved):

PR #114, opened 2021-04-29
Add type field to reference URLS.

oliverchang:
Add a "reference_url_types" field to the Bug entity. This duplicates URL
information with the existing "reference_urls" field for now, but after
the migration we can remove "reference_urls".


inferno-chromium (approved):
any reason why all these type strings are capitalized, probably web instead of WEB, report instead of REPORT is better ?

oliverchang:
> any reason why all these type strings are capitalized, probably web instead of WEB, report instead of REPORT is better ?

This is just consistent with other current type fields (e.g. type=GIT,SEMVER).

PR #115, opened 2021-04-30
Update vulnfeeds to use new schema.

oliverchang:


inferno-chromium (approved):

inferno-chromium (approved):
i thought we were leaving severity as optional field?

oliverchang:
> i thought we were leaving severity as optional field?

This was moved to `database_specific` due to strong support in removing it.

PR #118, opened 2021-05-05
Generalize affected ranges in Datastore.

oliverchang:
Rather than only storing GIT ranges, we generalize the Datastore
properties to better reflect the actual vulnerability schema.

This is in preparation for storing ECOSYSTEM and SEMVER ranges.

### DIFF
--- a/docker/worker/oss_fuzz.py
+++ b/docker/worker/oss_fuzz.py
@@ -203,7 +203,6 @@ def set_bug_attributes(bug, regress_result, fix_result):
   severity = fix_result.severity or regress_result.severity
   reference_urls = fix_result.reference_urls or regress_result.reference_urls
 
-  bug.repo_url = repo_url
   bug.issue_id = issue_id
   bug.project = project
   bug.ecosystem = ecosystem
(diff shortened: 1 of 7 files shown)


docker/worker/oss_fuzz.py
```diff
@@ -223,6 +222,14 @@ def set_bug_attributes(bug, regress_result, fix_result):
   bug.regressed = regress_result.commit or ''
   bug.fixed = fix_result.commit or ''
 
+  bug.affected_ranges = [
+      osv.AffectedRange(
+          type='GIT',
```
inferno-chromium, line 227: can we define a enum for 'GIT' and other type.

inferno-chromium (approved):

docker/worker/oss_fuzz.py
oliverchang: enums are hard to deal with for strings (have to use Type.GIT.value) and we'll be taking them as a strings from the JSON anyway. 

I added a validator here to the Datastore type so that these are always valid.

PR #120, opened 2021-05-07
Fix cron exception.

oliverchang:


oliverchang:
tbring for exception fix.

inferno-chromium (commented):
lgtm

PR #128, opened 2021-05-21
Various fixes in preparation for loading additional sources.

oliverchang:
- Fix UI to show the correct IDs for non-OSV IDs.
- Fix UI to show non-git ranges.
- Use the Vulnerability JSON for rendering on the frontend, rather than
  building a custom JSON.
- Set Bug entities created from SourceRepositories to public by default.


oliverchang:
See example here: https://staging-dot-oss-vdb.appspot.com/vulnerability/PYSEC-2021-59

jonathanmetzman (approved):
LGTM

PR #142, opened 2021-06-07
Fix Datastore index.

oliverchang:


inferno-chromium (approved):

jonathanmetzman (commented):
LGTM.
Crazy this subtle difference results in a bug.

PR #144, opened 2021-06-07
Use "git" CLI for cloning.

oliverchang:
This is much faster than libgit2's clone for large repos.

### DIFF
--- a/docker/worker/worker_test.py
+++ b/docker/worker/worker_test.py
@@ -834,7 +834,7 @@ class UpdateTest(unittest.TestCase):
     self.maxDiff = None
     tests.reset_emulator()
 
-    self.original_clone = pygit2.clone_repository
+    self.original_clone = osv.clone
     tests.mock_clone(self, func=self.mock_clone)
 
     tests.mock_datetime(self)
--- a/lib/osv/repos.py
+++ b/lib/osv/repos.py
@@ -16,6 +16,7 @@
 import logging
 import os
 import shutil
+import subprocess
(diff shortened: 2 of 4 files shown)


lib/osv/repos.py
```diff
@@ -24,16 +25,50 @@
 RETRY_SLEEP_SECONDS = 5
 
 
+class GitRemoteCallback(pygit2.RemoteCallbacks):
```
inferno-chromium, line 28: where is GitRemoteCallback used, i dont see any callsite changes after move.

docker/worker/worker_test.py
```diff
@@ -834,7 +834,7 @@ def setUp(self):
     self.maxDiff = None
     tests.reset_emulator()
 
-    self.original_clone = pygit2.clone_repository
+    self.original_clone = osv.clone
```
inferno-chromium, line 837: osv.repos.clone ?

inferno-chromium (approved):

lib/osv/repos.py
oliverchang: In worker.py and importer.py, by users of the osv lib.

docker/worker/worker_test.py
oliverchang: osv.repos.clone is the internal import path. osv.clone is the external one.

jonathanmetzman:
> This is much faster than libgit2's clone for large repos.

Sad this is the case.
LGTM

PR #146, opened 2021-06-08
Use a git mirror for git.kernel.org.

oliverchang:


inferno-chromium (approved):

jonathanmetzman (commented):
LGTM.
I maybe would have made the name `_git_mirror` more verbal e.g. `_get_git_mirror`.

PR #149, opened 2021-06-09
Remove some indexes.

oliverchang:
"affected" can be quite large, so we should only index it once rather
than 4x previously.

Also fix package_info task to not rely on these indexes, and generalize
it to not require a Repo URL.


jonathanmetzman (approved):
LGTM

PR #153, opened 2021-06-10
Support branches for source repositories.

oliverchang:
Also clean up some checkout related code while we're here.


jonathanmetzman (approved):
LGTM

PR #154, opened 2021-06-10
vulnfeeds: Add a -without_notes flag.

oliverchang:
This flag will skip over any matched vulnerabilities with notes
associated. This can be used by automation to automatically import new
vulnerabilities with zero human intervention.

### DIFF
--- a/vulnfeeds/cmd/pypi/main.go
+++ b/vulnfeeds/cmd/pypi/main.go
@@ -78,6 +78,7 @@ func main() {
 	pypiLinksPath := flag.String("pypi_links", "", "Path to pypi_links.json.")
 	pypiVersionsPath := flag.String("pypi_versions", "", "Path to pypi_versions.json.")
 	falsePositivesPath := flag.String("false_positives", "", "Path to false positives file.")
+	withoutNotes := flag.Bool("without_notes", false, "Output vulnerabilities without notes only.")
 	outDir := flag.String("out_dir", "", "Path to output results.")
 
 	flag.Parse()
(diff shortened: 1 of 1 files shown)


vulnfeeds/cmd/pypi/main.go
```diff
@@ -147,6 +148,11 @@ func main() {
 			continue
 		}
 
+		if len(notes) > 0 && *withoutNotes {
+			log.Printf("Skipping %s as there are notes associated with it.", vulnPath)
```
inferno-chromium, line 152: nit: slightly confusing message, we are skipping since flag is set and not because notes are associated with.
maybe just
Skipping notes in %s ?
or Skipping notes in %s as skip notes flag is set.

vulnfeeds/cmd/pypi/main.go
```diff
@@ -78,6 +78,7 @@ func main() {
 	pypiLinksPath := flag.String("pypi_links", "", "Path to pypi_links.json.")
 	pypiVersionsPath := flag.String("pypi_versions", "", "Path to pypi_versions.json.")
 	falsePositivesPath := flag.String("false_positives", "", "Path to false positives file.")
+	withoutNotes := flag.Bool("without_notes", false, "Output vulnerabilities without notes only.")
```
inferno-chromium, line 81: maybe skip_notes ?

inferno-chromium (approved):

vulnfeeds/cmd/pypi/main.go
oliverchang: So a "note" means that a human has to look at it because automation ran into some problems.

This is saying that we are completely skipping all matched vulnerabilities with these notes, so that the only ones that are output are ones without problems. These we can safely auto-commit into the vulnerability database. 

I think the messages you are proposing are saying something different :)

vulnfeeds/cmd/pypi/main.go
oliverchang: See other comment above. I think skip_notes implies something different here.

vulnfeeds/cmd/pypi/main.go
inferno-chromium: Skip, i didnt know what note meant, maybe add a comment in code to explain this.

vulnfeeds/cmd/pypi/main.go
inferno-chromium: ignore this please, i didnt understand.

PR #159, opened 2021-06-16
Fix change detection in importer.

oliverchang:
- Handle changed files that were later deleted.
- Handle deleted files that were later re-added.


inferno-chromium (approved):

jonathanmetzman (commented):
LGTM

PR #164, opened 2021-06-17
Don't read severities from vulnerabilities

oliverchang:
First part of removing this from the protobuf definition.


oliverchang:
tbr

oliverchang:
(Sorry, this was wrong, ignore).

PR #167, opened 2021-06-18
Add SourceRepository.versions_from_repo

oliverchang:
This controls whether or not we populate the "versions" field from
input git ranges.


inferno-chromium (approved):
Who will set ignore_git and versions_from_repo ? if this for future cl ?

oliverchang:
> Who will set ignore_git and versions_from_repo ? if this for future cl ?

E.g. for python bugs.

We want to analyze git ranges so that `{"commit": "..."}` queries work, but we don't want the resulting git analysis to populate the "versions" because  we are populating that using the PyPI registry instead.

PR #181, opened 2021-06-28
Many commit impact calculation fixes.

oliverchang:
- Simplify impacted tag calculation, by finding them as we
  walk through commit ranges. This also fixes false positives with
  assuming all tags are affected when no "introduced" commit is available.

- Don't compute cherrypicks when the input ranges are incomplete (i.e.
  either no "introduced" or no "fixed"). This leads to inaccurate
  results and potential false positives.

- Throttle AffectedCommit writes to 20000/5 seconds per worker.


oliverchang:
Friendly ping on this review, thanks!

jonathanmetzman (approved):
LGTM

PR #187, opened 2021-07-06
Improve matching on links

### LINKED ISSUE #186: triage tool:  perform some normalization of reference links to further improve matching
Here are a few suggestions to improve matching based on links for pypi packages

- strip `.git` from the end of source control links.  For instance, in the `pypi_links.json` input file, for the `notebook` package, we have `https://github.com/jupyter/notebook.git`, but in the actual CVE references it is just `https://github.com/jupyter/notebook`, so a match is never found

- Convert older-style pypi urls to `https://pypi.org/project/` urls.  There are at least a couple of older ones: `https://pypi.python.org/pypi/`, `https://upload.pypi.org/legacy/`.  We could also transform `https://pypi.org/simple/`

westonsteimel:
Improves matching on urls for pypi packages.  Strips `.git` from the end of source control url's and extracts package names from some common variations of pypi urls

closes #186

### DIFF
--- a/vulnfeeds/pypi/pypi.go
+++ b/vulnfeeds/pypi/pypi.go
@@ -122,7 +122,7 @@ func extractVendorProduct(link string) string {
 		return ""
 	}
 
-	return strings.ToLower(parts[1]) + "/" + strings.ToLower(parts[2])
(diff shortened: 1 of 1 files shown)


oliverchang (approved):
thanks a lot!

PR #188, opened 2021-07-07
fixes for reference link matching

westonsteimel:
I realised I should have used `TrimSuffix` instead of `TrimRight` when trimming `.git` from the end of reference urls.  I also added several more very generic but often encountered links to the blocklist.


oliverchang (approved):
Thanks!

PR #189, opened 2021-07-08
Vulnfeeds fixes.

### LINKED ISSUE #185: triage tool: handle CVEs that map to multiple packages.

oliverchang:
- Allow matching of multiple packages for a single CVE. Fixes #185.
- Work with normalized package names internally in pypi package.
- Always append note with all valid versions if there are any version
  matching notes.
- Add documentation about continuously updated JSONs.


asraa (approved):
nice, lgtm!

PR #190, opened 2021-07-08
Allow shared IDs between DBs.

### LINKED ISSUE #174: Support duplicate IDs across DBs
In some cases, IDs may not be unique across databases.

oliverchang:
Some DBs don't allocate their own IDs. Store such `Bug`s with
Datastore IDs of the form `"source:ID"`.

If they do allocate their IDs, the Datastore ID is kept as is if the
prefix is what we expect.

Fixes #174.


jonathanmetzman (approved):
LGTM

PR #196, opened 2021-07-15
Simplify semver queries.

oliverchang:
Previously semver queries were only done if the ecosystem was specified
and was marked as semver.

Simplify this by doing the semver query anyway if the ecosystem isn't
specified.


jonathanmetzman (approved):
LGTM

PR #197, opened 2021-07-15
Add public key for pypa integration.

oliverchang:
This is static for now.

### DIFF
--- a/gcp/appengine/app.yaml
+++ b/gcp/appengine/app.yaml
@@ -35,6 +35,10 @@ handlers:
     script: auto
     secure: always
 
+  - url: /public_keys
+    static_dir: dist/public_keys
+    mime_type: text/plain
+
   - url: /.*
     static_files: dist/index.html
     upload: dist/index.html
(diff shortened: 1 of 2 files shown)


jonathanmetzman (approved):
LGTM

dekkagaijin (approved):

gcp/appengine/app.yaml
```diff
@@ -35,6 +35,10 @@ handlers:
     script: auto
     secure: always
 
+  - url: /public_keys
+    static_dir: dist/public_keys
+    mime_type: text/plain
```
dekkagaijin, line 40: `application/json`?

dekkagaijin: also the URL I have in the pypa pr is "https://osv.dev/public_keys/pypa", does this route handle that?

oliverchang: Mime type doesn't really matter here (and not everything might be json in this dir) since warehouse doesn't interpret it? This is just so that we you go there the browser shows it as plain text instead of trying to guess.

https://osv.dev/public_keys/pypa is also live already because I cheated and deployed this bit earlier.

PR #202, opened 2021-08-05
Prepare for supporting OSV schema 0.8

oliverchang:
This is the first stage of a migration to support 0.8 of 
https://github.com/ossf/osv-schema/blob/main/schema.md

This PR modifies OSV’s backend representation to store data in a similarly structured format.
The main differences here are:
- Multiple packages are supported instead of a single one.
- Ranges are represented using individual points ("events") rather than pairs. 

The JSON/YAML input/output OSV remains mostly unchanged at this stage.
Only the backend representation is changed.

Detailed changes:
Bug.project -> Bug.affected_packages[].package.name
Bug.ecosystem -> Bug.affected_packages[].package.ecosystem
Bug.purl -> Bug.affected_packages[].package.purl
Bug.ecosystem_specific -> Bug.affected_packages[].ecosystem_specific
Bug.database_specific -> Bug.affected_packages[].database_specific
Bug.affected_ranges -> Bug.affected_packages[].ranges (new events format)

Old properties are kept to avoid breakages during migration. They'll be
deleted/changed in a future PR.

Also modify tests to more easily enable auto-generation expected dictionary values.


oliverchang:
PTAL. This is ready for review.

For details on migration, see go/osv-0-8-migration (internal only).

oliverchang:
TBRing this as this migration really needs to happen soon.

PR #204, opened 2021-08-13
Re-enable branch/cherry pick analysis for incomplete ranges.

### LINKED ISSUE #203: Inaccurate impact analysis
https://github.com/google/oss-fuzz-vulns/pull/7#issuecomment-897529522

oliverchang:
Change things such that we analyze with all known introduced/fixed
commits at the same time, rather than independently.

Analyzing git ranges independently results in false positives when we do
cherry pick analysis, which is why we disabled this previously.

Fixes #203.


jonathanmetzman:
LGTM

PR #210, opened 2021-09-01
Fix a bug with 0.8 to 0.7 conversion.

oliverchang:
Distinguish between empty "introduced" and unset.

### DIFF
--- a/lib/osv/models.py
+++ b/lib/osv/models.py
@@ -505,44 +505,42 @@ class Bug(ndb.Model):
       # Convert flattened events to range pairs (pre-0.8 schema).
       # TODO(ochang): Remove this once all consumers are migrated.
       # pylint: disable=cell-var-from-loop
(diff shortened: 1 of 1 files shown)


lib/osv/models.py
```diff
@@ -505,44 +505,42 @@ def _get_pre_0_8_affects(self):
       # Convert flattened events to range pairs (pre-0.8 schema).
       # TODO(ochang): Remove this once all consumers are migrated.
       # pylint: disable=cell-var-from-loop
-      new_range = lambda: vulnerability_pb2.AffectedRange(
+      new_range = lambda x, y: vulnerability_pb2.AffectedRange(
```
jonathanmetzman, line 508: nit: I think it's preferred to define functions rather than assign variables to lambdas.

jonathanmetzman (commented):
LGTM

PR #217, opened 2021-09-02
Update docs

di:
Was trying to fix a typo in one example (d3b681c) but looks like there are spec changes that haven't been included yet (8b9338c). I also had to take some additional steps (06226ce) to get the docs to build.


oliverchang (approved):
awesome, thanks! we'll probably want to reference https://github.com/ossf/osv-schema/blob/main/schema.md instead but I can fix this later.

PR #222, opened 2021-09-13
vulnfeeds: Fix git range population.

oliverchang:
Git ranges should always have an "introduced": "0" entry.


jonathanmetzman (approved):
LGTM

PR #223, opened 2021-09-13
Always output "versions" field even if it's empty.

oliverchang:
This is to conform to the schema requirements and to to pass the
validator (https://github.com/ossf/osv-schema/pull/5) in all cases.


jonathanmetzman (approved):
LGTM

PR #238, opened 2021-10-13
Send more data to PyPI

di:
Include the `details` field, and the `introduced`/`fixed` ranges for ecosystem versions.

### DIFF
--- a/gcp/functions/pypi/main.py
+++ b/gcp/functions/pypi/main.py
@@ -46,22 +46,28 @@ def publish(event, context):
   private_key = serialization.load_pem_private_key(
       data=key_data['key'].encode(), password=None)
 
-  request = json.dumps([{
-      'id': vulnerability.id,
-      'project': vulnerability.package.name,
-      'versions': list(vulnerability.affects.versions),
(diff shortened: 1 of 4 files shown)


gcp/functions/pypi/main.py
```diff
@@ -52,6 +52,15 @@ def publish(event, context):
       'versions': list(vulnerability.affects.versions),
       'link': f'https://osv.dev/vulnerability/{vulnerability.id}',
       'aliases': list(vulnerability.aliases),
+      'details': vulnerability.details,
+      'ranges': [
+          {
+              'introduced': r.introduced,
+              'fixed': r.fixed,
+          }
+          for r in vulnerability.affects.ranges
```
oliverchang, line 61: `affects` is deprecated and will be removed soon. Could you get this from the `affected` field instead (https://ossf.github.io/osv-schema/#affected-fields)? (We also need to update the `'project'`  field above to use that as well because `package` was also moved into `affected`).

di:
Thanks! LGTM

oliverchang:
> Thanks! LGTM

Let me know when warehouse is ready for this and I can re-send everything we have.

PR #239, opened 2021-10-13
add some basic classification of reference url types

### LINKED ISSUE #235: vulnfeeds: Classification of reference types on NVD CVEs
Currently, all of the references on a CVE entry from NVD just get classified with type `WEB`.  We should be able to classify at least some types based on well-known URL patterns, similar to what is being done in the GHSA converter at https://github.com/ossf/osv-schema/blob/d4d2764253eeb002daa738440ae6333c8082d069/tools/ghsa/convert_ghsa.py#L115.

We could also potentially set the `FIX` type when extracting the git fix range.

westonsteimel:
Fixes #235

Signed-off-by: Weston Steimel <weston.steimel@gmail.com>

### DIFF
--- a/vulnfeeds/cves/versions.go
+++ b/vulnfeeds/cves/versions.go
@@ -49,6 +49,10 @@ func extractGitHubCommit(link string) *FixCommit {
 	}
 
 	pathParts := strings.Split(u.Path, "/")
+	if len(pathParts) < 2 {
+		return nil
+	}
+
 	if pathParts[len(pathParts)-2] != "commit" {
 		return nil
 	}
--- a/vulnfeeds/vulns/vulns.go
+++ b/vulnfeeds/vulns/vulns.go
@@ -17,6 +17,8 @@ package vulns
 import (
 	"fmt"
 	"io"
+	"net/url"
+	"strings"
 	"time"
 
 	"gopkg.in/yaml.v2"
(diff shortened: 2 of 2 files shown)


vulnfeeds/vulns/vulns.go
```diff
@@ -70,6 +72,59 @@ func timestampToRFC3339(timestamp string) (string, error) {
...
+	u, err := url.Parse(link)
+	if err != nil {
+		return "WEB"
+	}
+
+	pathParts := strings.Split(u.Path, "/")
+
+	if u.Host == "github.com" {
+		// Example: https://github.com/google/osv/commit/cd4e934d0527e5010e373e7fed54ef5daefba2f5
+		if pathParts[len(pathParts)-2] == "commit" {
```
oliverchang, line 85: Could we add a length check for pathParts here before indexing this?

vulnfeeds/vulns/vulns.go
```diff
@@ -70,6 +72,59 @@ func timestampToRFC3339(timestamp string) (string, error) {
...
+	if u.Host == "snyk.io" {
+		//Example: https://snyk.io/vuln/SNYK-PYTHON-TRYTOND-1730329
+		if pathParts[0] == "vuln" {
+			return "ADVISORY"
+		}
+	}
+
+	if u.Host == "nvd.nist.gov" {
+		//Example: https://nvd.nist.gov/vuln/detail/CVE-2021-23336
+		if pathParts[0] == "vuln" && pathParts[1] == "detail" {
```
oliverchang, line 105: same here with `pathParts[1]`

oliverchang (approved):
thanks a lot for this! Just one comment.

vulnfeeds/cves/versions.go
```diff
@@ -49,6 +49,10 @@ func extractGitHubCommit(link string) *FixCommit {
 	}
 
 	pathParts := strings.Split(u.Path, "/")
+	if len(pathParts) < 2 {
```
oliverchang, line 52: thanks for fixing this one as well :)

PR #240, opened 2021-10-14
Make frontend development easier.

oliverchang:
- Allow switching of backends URLs.
- Allow cross origin requests from http://localhost:8080


gcp/appengine/frontend_handlers.py
```diff
@@ -50,6 +51,26 @@ def check_rate_limit():
...
+
+  response = make_response()
+  response.headers.add('Access-Control-Allow-Origin', 'http://localhost:8080')
+  response.headers.add('Access-Control-Allow-Methods', '*')
+  response.headers.add('Access-Control-Allow-Headers', '*')
+  return response
+
+
+@blueprint.after_request
+def add_cors_headers(response):
```
inferno-chromium, line 68: where are these 2 functions called ?

inferno-chromium (approved):

gcp/appengine/frontend_handlers.py
oliverchang: It's set up to be automatically called as part of `before_request` and `after_request`.

PR #246, opened 2021-10-20
Properly take directory_path into account when importing vulnerabilities

oliverchang:


docker/importer/importer.py
```diff
@@ -44,6 +44,13 @@
 
 def _is_vulnerability_file(source_repo, file_path):
   """Return whether or not the file is a Vulnerability entry."""
+  if (source_repo.directory_path and
+      not file_path.startswith(source_repo.directory_path)):
```
inferno-chromium, line 48: nit: should it be matching against directory_path + path seperator ?

inferno-chromium (approved):

docker/importer/importer.py
oliverchang: good point, fixed!

PR #267, opened 2021-11-23
vulnfeeds: deduplicate pypi package matches when normalizing names

westonsteimel:
Signed-off-by: Weston Steimel <weston.steimel@gmail.com>

### DIFF
--- a/vulnfeeds/pypi/pypi.go
+++ b/vulnfeeds/pypi/pypi.go
@@ -135,8 +135,14 @@ func hasPrefix(list []string, item string) bool {
 func processMatches(names []string) []string {
 	// Normalize all PyPI package names.
 	normalized := make([]string, 0, len(names))
+	encountered := map[string]bool{}
 	for _, name := range names {
-		normalized = append(normalized, NormalizePackageName(name))
+		normalizedName := NormalizePackageName(name)
+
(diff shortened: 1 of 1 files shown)


vulnfeeds/pypi/pypi.go
```diff
@@ -131,8 +131,15 @@ func hasPrefix(list []string, item string) bool {
 func processMatches(names []string) []string {
 	// Normalize all PyPI package names.
 	normalized := make([]string, 0, len(names))
+	encountered := map[string]bool{}
 	for _, name := range names {
-		normalized = append(normalized, NormalizePackageName(name))
+		normalizedName := NormalizePackageName(name)
+		_, isFound := encountered[normalizedName]
+
+		if !isFound {
```
oliverchang, line 139: nit: in Go you can do

```go
if _, exists := encountered[normalizedName]; !exists {
  ...
}
```

oliverchang (approved):
thanks!

vulnfeeds/pypi/pypi.go
westonsteimel: Thanks Oliver!  I thought I remembered seeing something like that before, but I couldn't seem to find it last night.  I'll get it updated a bit later today.

oliverchang (approved):

PR #278, opened 2022-01-14
Set package URL if it wasn't already set.

oliverchang:


lib/osv/models.py
```diff
@@ -310,24 +311,29 @@ def _pre_put_hook(self):
...
-      self.purl = [
-          pkg.package.purl for pkg in self.affected_packages if pkg.package.purl
-      ]
-
-      for project in self.project:
-        search_indices.update(self._tokenize(project))
-
-      for ecosystem in self.ecosystem:
-        search_indices.update(self._tokenize(ecosystem))
+    for affected_package in self.affected_packages:
```
inferno-chromium, line 314: Do we want the bail out condition on self.affected_packages ? previously you had "if self.affected_packages:"

inferno-chromium (approved):

lib/osv/models.py
oliverchang: Not needed since this is a ndb repeated property so the default value is always `[]`. The if condition was previously used to distinguish between pre-1.0 and 1.0 schema versions during migraiton.

PR #280, opened 2022-01-19
Initial v2 UI

rzhw:
This PR starts work on #231 and partially implements the design. More work to be left for followup PRs.

Screenshots:

9rwgpfxUMPAkRpr

t8GeNuZhbueuRPN


gcp/appengine/frontend_handlers.py
```diff
@@ -77,6 +79,54 @@ def index():
...
+def index_v2():
+  return render_template('home.html')
+
+
+@blueprint.route('/v2/list')
+def list():
+  """Main page."""
+  # TODO: Can/should we do this as an internal query not a full on request to osv.dev?
+  page = request.args.get('page') if request.args.get('page') and request.args.get('page').isnumeric() else '1'
+  response = requests.get('https://osv.dev/backend/query?page=%s&search=&affected_only=true&ecosystem=' % page)
```
oliverchang, line 92: Instead of doing this, let's just factor out the querying part of `query_handler` in this file and call this here directly.

i.e. something similar to:

```python
def run_query(search_string, page, affected_only, ecosystem):
  query = osv.Bug.query(osv.Bug.status == osv.BugStatus.PROCESSED,                                                                                                                                                                                   
                        osv.Bug.public == True)  # pylint: disable=singleton-comparison)
   ...

@blueprint.route(_BACKEND_ROUTE + '/query')                                                                                                                                                                                                          
def query_handler():
  search_string = request.args.get('search')                                                                                                                                                                                                         
  page = int(request.args.get('page', 1))                                                                                                                                                                                                            
  affected_only = request.args.get('affected_only') == 'true'                                                                                                                                                                                        
  ecosystem = request.args.get('ecosystem')   
  return jsonify(run_query(search_string, page, affected_only, ecosystem))

@blueprint.route('/v2/list')
def list():
  results = run_query(...)
   ...
```

gcp/appengine/frontend_handlers.py
```diff
@@ -77,6 +79,54 @@ def index():
...
+
+
+@blueprint.route('/v2/vulnerability/<id>')
+def vulnerability(id):
+  """Vulnerability page."""
+  # TODO: Can/should we do this as an internal query not a full on request to osv.dev?
+  response = requests.get('https://osv.dev/backend/vulnerability?id=%s' % id)
+  item = json.loads(response.content)
+
+  vulnerability = {
```
oliverchang, line 114: Can we just return the vulnerability structure as-is, without doing this field mapping here? The structure format is stable and based on https://ossf.github.io/osv-schema/.

gcp/appengine/frontend_handlers.py
```diff
@@ -77,6 +79,54 @@ def index():
...
+    "id": item['id'],
+    "summary": item['summary'] if 'summary' in item else '',
+    "packages": item['affected'][0]['package']['ecosystem'],
```
oliverchang, line 117: This should be item['affected'][0]['package'] (not only taking the "ecosystem"), and we'd want to render it something like "{ecosystem}/{name}". e.g. "PyPI/django"

It's fine for now to only show the [0] item, but we'd also want to show every single entry in the `affected` entry in the final version.

gcp/appengine/frontend_handlers.py
```diff
@@ -77,6 +79,54 @@ def index():
...
+    })
+
+  return render_template('list.html', vulnerabilities = vulnerabilities)
+
+
+@blueprint.route('/v2/vulnerability/<id>')
+def vulnerability(id):
+  """Vulnerability page."""
+  # TODO: Can/should we do this as an internal query not a full on request to osv.dev?
+  response = requests.get('https://osv.dev/backend/vulnerability?id=%s' % id)
```
oliverchang, line 111: Same here. Let's reuse the code from `vulnerability_handler`.

oliverchang (commented):
Thanks a lot! This is AMAZING!!! 

Just some initial comments on the Python code around re-use, but otherwise this looks incredible already.

gcp/appengine/frontend_handlers.py
rzhw: Done.

gcp/appengine/frontend_handlers.py
rzhw: Done.

gcp/appengine/frontend_handlers.py
rzhw: SG, shall we do the same for the vuln list? From a look at the API it [does look the same](https://osv.dev/docs/#operation/OSV_QueryAffected)

gcp/appengine/frontend_handlers.py
rzhw: Yep this was a temporary hack to get a WIP table going, let's get the table looking right in a followup PR

oliverchang (approved):
nice!! thanks!!

PR #285, opened 2022-01-21
Support schema version 1.2.0

### LINKED ISSUE #284: Support schema version 1.2.0
Some new fields were added:
- schema_version
- severity
- credits
- top level database_specific

oliverchang:
Also delete legacy code dealing with pre-1.0 schema.

Fixes #284.


lib/osv/vulnerability.proto
```diff
@@ -148,14 +134,23 @@ message Affected {
...
+message Severity {
+  // Type of the severity.
+  enum Type {
+    UNSPECIFIED = 0;
+    CVSS_V3 = 1;
+  }
+  // The type of this severity entry.
+  Type type = 1;
+  // The quantitative score.
+  string score = 2;
```
inferno-chromium, line 146: if it is quantitative score, then shouldn't this be int or float ?

lib/osv/vulnerability.proto
```diff
@@ -148,14 +134,23 @@ message Affected {
...
+}
+
+message Credit {
+  // The name to give credit to.
+  string name = 1;
+  // Contact methods.
```
inferno-chromium, line 152: should the comment have example of methods ?

inferno-chromium (approved):

lib/osv/vulnerability.proto
oliverchang: This is a string for future extensibility, or if we relax the "quantitative" requirement.

lib/osv/vulnerability.proto
oliverchang: Clarified this to be URLs. This is all better documented at https://ossf.github.io/osv-schema/ :)

PR #288, opened 2022-02-01
Initial v2 package info UI

rzhw:
This PR introduces an initial pass at the package info UI on the v2 vulnerability page.

## Notes

While `@material/mwc-tab-bar` and `@material/tab-bar` were considered, they were ultimately discarded:

- Close adherence to Material Design, which this design heavily diverges from, makes design adjustments more trouble than it's worth (e.g. compared to the search textbox implemented in #280).
- Lack of tab panel switching functionality means we would still need to DIY some JavaScript to switch panels anyway.

This PR proposes the tab UI is implemented by adopting [spicy-sections](https://github.com/tabvengers/spicy-sections), an *experimental* creation of the Open UI community group (hosted by the W3C [but run](https://www.w3.org/community/open-ui/) by the community), that is *not published on npm*.

Despite its experimental status, it has made implementation of the design incredibly straightforward, with no extra JS required (beyond the `import` statement).

YekoTfXoJQ2LMTa

The progressive enhancement approach adopted by `<spicy-sections>` makes markup minimal, requiring us to only write markup consisting of pairs of `<h2>` and sibling elements.

Styling is also minimal, with only lines 227-244 of styles.scss being specific to `<spicy-sections>`.

This minimal approach is expected to make adopting an alternative tab solution, should it be necessary either in this PR or at a later point, much easier.

As the `spicy-sections` project [aims to standardize tabs](https://github.com/tabvengers/spicy-sections/blob/40cb5a5ab38b463cfe55b3f2bc2204cfa3bb0029/README.md#standards-path) as part of the HTML standard, I believe this PR can also serve as a data point for real-world usage, so long as we don't consider any other considerations in its adoption to be deal-breaking.

### Dependency not published to npm ⚠️

This PR pins to a specific commit in `packages.json` in order to guarantee API and functional stability. However, if the underlying repo is moved, this build will break.

### Why not DIY?

The complexity of the [spicy-sections writeup](https://open-ui.org/components/tabs.research.parts) shows this isn't a trivial component to roll our own solution for. We also get extra stuff for free; `spicy-sections` allows us to use accordions or even normal sections in different viewports if desired.

### Impact on filesize

(Custom CSS used to measure is the same as this PR.)

  - Uncompressed static/vendors.js 798 KiB before, **817 KiB after**, static/main.css 48.4 KiB no change
  - If using `@material/mwc-tab-bar` instead: static/vendors.js **958 KiB after**, static/main.css 48.4 KiB no change
  - If using `@material/tab-bar` instead: static/vendors.js **868 KiB after**, static/main.css **61 KiB after**

### Accessibility

While this OSV v2 design is in early stages, [preliminary research into accessibility](https://github.com/tabvengers/spicy-sections/search?q=accessibility) with spicy-sections shows some work put in, though I haven't checked this in-depth.

A cursory test using VoiceOver also shows spicy-sections works at minimum at a rudimentary level with screen readers:

https://user-images.githubusercontent.com/79461/151932524-6471665e-ca82-4b75-a6c7-848fb76e685c.mov

### A personal nit

A supposed more HTML5-ish way of writing the pattern of:

```html
    <h2>Package 1</h2>
    <div>
      <p class="subtitle">package-name</p>
      <div>content</div>
    </div>
```

Would've been to write something like instead:

```html
    <header>
      <h2>Package 1</h2>
      <p>package-name</p>
    </header>
    <div>content</div>
```

(As per http://html5doctor.com/howto-subheadings/)

But spicy-sections doesn't expect this and does nothing when given such markup. (i.e. a normal bunch of HTML is rendered without being moved into tabs)

## Further reading

- https://css-tricks.com/spicy-sections/
- https://css-tricks.com/newsletter/281-tabs-and-spicy-drama/

### DIFF
--- a/gcp/appengine/frontend3/package-lock.json
+++ b/gcp/appengine/frontend3/package-lock.json
@@ -13,7 +13,8 @@
         "@material/mwc-icon": "^0.25.3",
         "@material/mwc-icon-button": "^0.25.3",
         "@material/mwc-textfield": "^0.25.3",
-        "@material/theme": "^13.0.0"
+        "@material/theme": "^13.0.0",
+        "spicy-sections": "git+https://github.com/tabvengers/spicy-sections.git#1b0f4850bdc5e87a5272b709f353472ed0a05284"
       },
       "devDependencies": {
         "copy-webpack-plugin": "^10.2.0",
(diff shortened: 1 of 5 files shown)


oliverchang (approved):
Amazing work and detailed design considerations!!! Pretty magic that this doesn't need any additional JS.

PR #289, opened 2022-02-04
Initial v2 infinite load implementation

rzhw:
This PR introduces an initial "infinite load" (as opposed to "infinite scroll") implementation on the vulnerability table, replacing the previous placeholder pagination.

It does not attempt to fix other known issues with the table e.g. columns aren't complete, width goes out of bounds, etc.

79q4QkjzVejBbhQ

## Notes

This implementation adds no additional custom JS by taking advantage of `<turbo-frame>`, **part of the pre-existing dependency on Turbo**, which replaces only itself—and not the full page—with an updated version upon request load.

For example, the HTML in `/list?page=1` looks like the following:

```html
<turbo-frame id="vulnerability-page-1">
  vuln 1
  vuln 2
  ..
  vuln n
  <turbo-frame id="vulnerability-page-2">
    <a href="/list?page=2">Load more</a>
  </turbo-frame>
</turbo-frame>
```

Similarly, the HTML in `/list?page=2` looks like the following:

```html
<turbo-frame id="vulnerability-page-2">
  vuln 1
  vuln 2
  ..
  vuln n
  <turbo-frame id="vulnerability-page-3">
    <a href="/list?page=3">Load more</a>
  </turbo-frame>
</turbo-frame>
```

Note how everything is nearly the same, except `id`s attached to the `<turbo-frame>`.

When clicking the link on page 1, this is intercepted by `<turbo-frame id="vulnerability-page-2">`, which fetches `/list?page=2` and looks for a frame with the same id. It then replaces itself with those contents.

Hence the resulting rendered HTML of:

```html
<turbo-frame id="vulnerability-page-1">
  vuln 1
  vuln 2
  ..
  vuln n
  <turbo-frame id="vulnerability-page-2">
    vuln 1
    vuln 2
    ..
    vuln n
    <turbo-frame id="vulnerability-page-3">
      <a href="/list?page=3">Load more</a>
    </turbo-frame>
  </turbo-frame>
</turbo-frame>
```

Ignoring the existence of `<turbo-frame>`, we can also see how this HTML would enable navigation even in web browsers [without custom elements support](https://caniuse.com/?search=Custom%20Elements).

## Known issues

- Unfortunately, `<table>` does not support `<turbo-frame>` wrapping table rows. This is invalid HTML. Adopting this solution was only possible by first rewriting the vulnerability table using `<div>` and `<span>` elements, and using CSS to display it as a table, and ARIA markup for proper screen reader support.
  - In order for the `<turbo-frame>` to not affect the structure of the table, we take advantage of [`display: contents;`](https://caniuse.com/css-display-contents) so that the *rendered* structure of the page is as if all table rows had the same parent.
- The case of the last page isn't handled.

## Further reading

- Turbo Frames documentation: https://turbo.hotwired.dev/handbook/frames
- Prior art for infinite loading/scroll using Turbo: https://lewisyoul.github.io/infinitely-scrolling-lists-with-hotwire-and-zero-javascript

### DIFF
--- a/gcp/appengine/frontend3/package-lock.json
+++ b/gcp/appengine/frontend3/package-lock.json
@@ -10,6 +10,7 @@
         "@hotwired/turbo": "^7.1.0",
         "@material/data-table": "^13.0.0",
         "@material/layout-grid": "^13.0.0",
+        "@material/mwc-circular-progress": "^0.25.3",
         "@material/mwc-icon": "^0.25.3",
         "@material/mwc-icon-button": "^0.25.3",
         "@material/mwc-textfield": "^0.25.3",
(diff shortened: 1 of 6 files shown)


oliverchang (approved):
nice, thanks!!

PR #291, opened 2022-02-09
Initial v2 search and ecosystem filtering UI

rzhw:
This PR gives functionality to the previously non-working search box, as well as adds a list of ecosystems underneath the searchbox that can be used to filter the results by ecosystem.

It also adds a blur effect when searches and filters are loading (if a full page load is not being performed).

It does not fix the known issue where a version not being set, which is common in many ecosystems including crates.io, NuGet, Go, etc., will cause the page to 500. This will be fixed in an upcoming PR.

image
image

## Notes

Alas, with today's PR, the Dream of No Custom JS (apart from the code to init `@material/data-table`) has crumbled into pieces.

There are two things we add custom JS to address:

- Creating a new custom element to **wrap the `<mwc-textfield>` custom element**, and behave exactly the same, except to add back native-like behavior of submitting a form upon pressing the Enter key.
  - While this could be a fix to contribute upstream to Material Web, I current expect it would only be possible to land after the M3 release in late 2022. See material-components/material-web#3050.
- Creating a new custom element to **contain `<input type=radio>` elements**, which causes them to submit the form when changed.
  - This allows the *visual* behavior of looking and behaving like chips that reload the table, but *semantically* implemented as a bog standard series of radio inputs in the search form.

Both of these are implemented as custom elements. We pull in the pre-existing `lit@^2.0.0` dependency that came from Material components, which doesn't increase the size of vendors.js. This cuts down the boilerplate of writing a custom element by a lot.

Use of custom elements lets us avoid worrying about the re-initialization and teardown lifecycle whenever a page load is done via JS, as well as not having to specifically target HTML elements e.g. the pair of `document.getElementById('add-radio-behavior')` and `<div id="add-radio-behavior">`.

Note how both of these uses of custom JS are to address minor behaviors only.

The actual search works without requiring a full page refresh.

### DIFF
--- a/gcp/appengine/frontend3/package-lock.json
+++ b/gcp/appengine/frontend3/package-lock.json
@@ -15,6 +15,7 @@
         "@material/mwc-icon-button": "^0.25.3",
         "@material/mwc-textfield": "^0.25.3",
         "@material/theme": "^13.0.0",
+        "lit": "^2.0.0",
         "spicy-sections": "git+https://github.com/tabvengers/spicy-sections.git#1b0f4850bdc5e87a5272b709f353472ed0a05284"
       },
       "devDependencies": {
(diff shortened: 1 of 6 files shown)


oliverchang (approved):
Awesome work as always!!! No custom JS was nice while it lasted. RIP.

PR #293, opened 2022-02-10
Tweaks for v2 home and list UI

rzhw:
This PR makes some visual tweaks to the home and list pages. (Preliminary work on homepage. Vulnerability list table has received border restyling. Last modified column has been added.)

This PR also makes some fixes to the list page. (Package names are now properly listed albeit with no character limit. Versions will display "See details" like v1 instead of 500ing the page if there are none available.)

Screen Shot 2022-02-10 at 9 10 01 pm


oliverchang (approved):
nice!! this is starting to really take shape and become usable :)

PR #296, opened 2022-02-14
Support NuGet versions and expansion

### LINKED ISSUE #229: Version expansion for NuGet
NuGet is *almost* semver, but not exactly: https://docs.microsoft.com/en-us/nuget/concepts/package-versioning#where-nugetversion-diverges-from-semantic-versioning

oliverchang:
Fixes #229.

### DIFF
--- a/lib/osv/nuget.py
+++ b/lib/osv/nuget.py
@@ -1,4 +1,3 @@
-"""NuGet version parser."""
 # Copyright 2022 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -12,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+"""NuGet version parser."""
 
 import functools
 import re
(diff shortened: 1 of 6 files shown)


lib/osv/nuget.py
```diff
@@ -0,0 +1,71 @@
...
+# https://docs.microsoft.com/en-us/nuget/concepts/package-versioning
+#  - Optional 4th component (x.y.z.R).
+#  - Prerelease components are compared case insensitively.
+#  - Non-major version segments are optional. e.g. "1" is a valid version
+#    number.
+
+
+def _extract_revision(str_version):
+  """Extract revision (4th component) from version number (if any)."""
+  pattern = re.compile(r'^(\d+)(\.\d+)(\.\d+)(\.\d+)(.*)')
```
inferno-chromium, line 32: nit: maybe add example in a comment or use named groups.

inferno-chromium (approved):

lib/osv/nuget.py
```diff
@@ -0,0 +1,71 @@
+"""NuGet version parser."""
```
jonathanmetzman, line 1: This should be below the license.

lib/osv/nuget_test.py
```diff
@@ -0,0 +1,85 @@
+"""NuGet version parser tests."""
```
jonathanmetzman, line 1: This should be below the license.

jonathanmetzman (approved):
LGTM

lib/osv/nuget.py
oliverchang: done!

lib/osv/nuget_test.py
oliverchang: done!

PR #302, opened 2022-02-17
Clear existing versions when adding OSS-Fuzz fix.

oliverchang:
Otherwise we could have a race condition where we include versions that
may include the fix.


oliverchang:
tbr

inferno-chromium (commented):
LGTM

PR #304, opened 2022-02-17
Add a worker test case for Andorid CVEs.

baekhyunwook:
This change adds a test case for Android ecosystem CVEs, which will be updated through a GCS bucket.


oliverchang:
Thanks @baekhyunwook (and sorry for the delay!)

This looks great, but could you please fix the lint error here: https://github.com/google/osv/runs/5269436849?check_suite_focus=true

baekhyunwook:
Ouch, sorry for the long delay @oliverchang; I missed your last comment.
I quickly fixed the lint error. Please take a look.

oliverchang (approved):
thanks!!

PR #308, opened 2022-02-21
Make OSS-Fuzz FixResult setting more accurate.

oliverchang:
Use the repo URL to key on if a fix commit needs to be added to a git
range.


inferno-chromium (approved):

jonathanmetzman (commented):
LGTM

PR #311, opened 2022-02-22
Update pypa/advisory-db to pypa/advisory-database

di:
Ref: https://github.com/pypa/advisory-database/issues/62


gcp/functions/pypi/t.json
```diff
@@ -0,0 +1,12 @@
+{"vulns": {
```
oliverchang, line 1: unintentionally included file?

oliverchang (approved):
thanks!!

PR #314, opened 2022-02-23
Reduce duplicates in indexes.

oliverchang:
To keep index sizes small.

### DIFF
--- a/lib/osv/models.py
+++ b/lib/osv/models.py
@@ -331,21 +331,21 @@ class Bug(ndb.Model):
         affected_package.package.purl = purl_helpers.package_to_purl(
             affected_package.package.ecosystem, affected_package.package.name)
 
-    self.project = list(set([
+    self.project = list({
         pkg.package.name for pkg in self.affected_packages if pkg.package.name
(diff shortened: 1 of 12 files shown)


lib/osv/models.py
```diff
@@ -331,25 +331,31 @@ def _pre_put_hook(self):
         affected_package.package.purl = purl_helpers.package_to_purl(
             affected_package.package.ecosystem, affected_package.package.name)
 
-    self.project = [
+    self.project = list(set([
```
inferno-chromium, line 334: nit: sorted(set( would avoid list and later .sort call ?

inferno-chromium (approved):
LGTM with minor change to simplify

lib/osv/models.py
oliverchang: Just trying to avoid a list copy and doing an in-place .sort() instead.

inferno-chromium: smart, i keep forgetting python is a memory hog!

jonathanmetzman (commented):
LGTM

PR #316, opened 2022-02-24
Reduce Datastore index fanout.

### LINKED ISSUE #315: Make version indexes more scalable.
Currently, all indexes for package+version are stored in the same `Bug` entity. 

This has large benefits in terms of read/write latency and consistency, but we are now seeing some entries with large fanout of these indexes (high package count x version count).

Some options here
- Use partial indexes (package count + version count) and rely on Datastore's zig zag merge. Will need to evaluate performance here.
- Break out indexes into its own separate entity.

oliverchang:
Redefine composite indexes to rely on [zig zag merge](https://cloud.google.com/datastore/docs/concepts/optimize-indexes#index_merge_performance) for queries.
This results in `O(version + package)` indexes instead of
`O(version x package)` indexes per vulnerability.

Also fix a bug with querying by package+version on rare vulnerabilities
with multiple different packages.

Previously we were incorrectly returning matches if a query matches any
listed package and any listed version. Now we filter to make sure the
version actually matches the package.

Fixes #315.

### DIFF
--- a/gcp/api/server.py
+++ b/gcp/api/server.py
@@ -195,6 +195,36 @@ def _is_semver_affected(affected_packages, package_name, ecosystem, purl,
   return affected
 
 
+def _is_version_affected(affected_packages,
+                         package_name,
+                         ecosystem,
+                         purl,
+                         version,
+                         normalize=False):
+  """Returns whether or not the given version is within an affected ECOSYSTEM
+  range."""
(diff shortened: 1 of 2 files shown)


oliverchang:
TBR

jonathanmetzman (commented):
LGTM
Sorry for the delay on this.

PR #327, opened 2022-03-04
Tweaks to v2 vuln page UI

rzhw:
Adds external link icons, increases spacing between text, fixes the most egregious bug of vulnerability details overflowing way past the page bounds

Screenshot:
vuln


oliverchang (approved):
thanks!!

Issue #257, opened 2021-11-09
Support vulnerability scanning from SBOMs


oliverchang:
Support both [CycloneDX](https://cyclonedx.org/) and [SPDX](https://spdx.dev/).

PR #328, opened 2022-03-14
tools/scanner: Add support for SPDX scanning.

oliverchang:
Also refactor the existing scanner to conform to standard Go project
structure.

First part of #257


oliverchang:
@calebbrown mind reviewing this Go PR for OSV?

tools/scanner/cmd/scanner.go
```diff
@@ -0,0 +1,136 @@
...
+			log.Printf("Scanned %s SBOM", provider.Name())
+			return nil
+		}
+
+		if errors.Is(err, sbom.InvalidFormat) {
+			continue
+		}
+
+		return err
+	}
```
calebbrown, line 63: Moving this logic into its own function make make the code easier to read (e.g. `scanFile()`)

tools/scanner/cmd/scanner.go
```diff
@@ -0,0 +1,136 @@
...
+	cmd.Stdout = &out
+
+	err := cmd.Run()
+	if err != nil {
+		return "", err
+	}
+	return strings.TrimSpace(out.String()), nil
+}
+
+func doScan(repoDir string) error {
```
calebbrown, line 99: How does `doScan()` differ to `scan()`? The names imply they are the same - is there a better name for this one?

calebbrown (approved):

tools/scanner/cmd/scanner.go
oliverchang: Done!

tools/scanner/cmd/scanner.go
oliverchang: Good point. renamed `doScan` to `scanGit`.

PR #333, opened 2022-03-24
Pin all GCP Python dependencies.

oliverchang:
Use a Pipfile.lock to generate the requirements.txt for App Engine/API/Functions
deployment.


jonathanmetzman (approved):
lgtm

PR #360, opened 2022-03-30
Fix missing search box on vulnerability list.

oliverchang:
This happens when we start at a different page (e.g. the home page)
and navigate to the vulnerability list page.

This seems to be because the place where we were instantiating the
MDCDataTable (index.js) can be run in a context where the actual element
hasn't been loaded in if we're not on the right page. 

This causes the `mwc-textfield-with-enter` element to not be defined properly,
leading to the missing search box.


rzhw (approved):
Thanks for figuring this out!

PR #364, opened 2022-03-30
Create codeql-analysis.yml

oliverchang:


gcp/appengine/main.py
```diff
@@ -58,4 +58,4 @@ def create_app():
 cache.instance.init_app(app)
 
 if __name__ == '__main__':
-  app.run(host='127.0.0.1', port=8000, debug=True)
+  app.run(host='127.0.0.1', port=8000, debug=False)
```
oliverchang, line 61: This was from a CodeQL finding. It's not actually used in production, so it's a false positive, but changing this just in case this ever changes.

jonathanmetzman: Hmm...but isn't debug=True the correct behavior for local development?

jonathanmetzman (approved):
lgtm

gcp/appengine/main.py
jonathanmetzman: If you don't need it then that's the final word, but I think this is pretty normal and I think 127.0.0.1 is accessible only within the same host so it would have to be another user on your machine.

PR #396, opened 2022-04-11
Make vulnerability list page render better on mobile.

### LINKED ISSUE #380: V2 UI: Improve mobile layout for list page
image

Currently:
- The ecosystems list takes up almost all of the screen on my phone.
- We can probably hide some columns, and just keep the ID + package column (and maybe summary if we can fit that somehow).

oliverchang:
- Make ecosystems list expandable on mobile (collapsed by default).
- Only show ID, Packages columns.

Fixes #380.

image

### DIFF
--- a/gcp/appengine/frontend3/src/index.js
+++ b/gcp/appengine/frontend3/src/index.js
@@ -34,9 +34,58 @@ export class SubmitRadiosOnClickContainer extends LitElement {
   constructor() {
     super();
     this.addEventListener('change', () => submitForm(this.closest('form')))
+    this.expanded = false;
+  }
+
+  toggleExpanded() {
+    this.expanded = !this.expanded;
+    const expander = this.renderRoot.querySelector('#ecosystem-expander');
+    if (this.expanded) {
+      expander.classList.add('expanded');
+    } else {
+      expander.classList.remove('expanded');
+    }
+
+    for (const label of this.querySelectorAll('.expandable')) {
+      if (this.expanded) {
+        label.classList.add('expanded');
+      } else {
+        label.classList.remove('expanded');
+      }
+    }
+  }
+
+  firstUpdated() {
+    const expander = this.renderRoot.querySelector('#ecosystem-expander');
+    expander.addEventListener('click', this.toggleExpanded.bind(this));
+  }
+
+  static get styles() {
+    return [
+      css`
+        #ecosystem-expander {
+          display: none;
+        }
+
+        @media (max-width: 500px) {
+          #ecosystem-expander {
+            display: block;
+            background: url(/static/img/arrow.png) no-repeat;
+            background-position: center;
+            width: 16px;
+            height: 16px;
+          }
+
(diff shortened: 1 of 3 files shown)


rzhw:
Would using [`<details>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details) work better in this case?

rzhw (changes requested):
See above

oliverchang:
Interesting! I didn't know about `<details`>. Is there an easy way (without using additional JS) to use this such that it's expanded by default on wider screens, but collapsed by default on mobile?

oliverchang:
There's also some nuance here in that the "All ecosystems" button shouldn't expand when clicked. It's the little arrow next to it that does this. We'd have to also add some hackery to work around that if we use the "All ecosystems" button as the `<summary>`?

rzhw:
Ah sorry you're right, `<details>` wouldn't handle the responsive case - I completely missed that aspect at first glance.

I wonder if it'd be possible to use spicy-sections for this use case - see the [demo](https://codepen.io/bkardell/pen/VwpJGGL?editors=1100) where the "collapse" affordance is applied only at a certain screen size.

Instead of `<h2>` etc. we could use the `<spicy-h>` element per [this documentation](https://github.com/tabvengers/spicy-sections/blob/959b42cbd1379d84565c40a15b2a0ff926046286/README.md#usage).

oliverchang:
> Ah sorry you're right, `<details>` wouldn't handle the responsive case - I completely missed that aspect at first glance.
> 
> I wonder if it'd be possible to use spicy-sections for this use case - see the [demo](https://codepen.io/bkardell/pen/VwpJGGL?editors=1100) where the "collapse" affordance is applied only at a certain screen size.
> 
> Instead of `<h2>` etc. we could use the `<spicy-h>` element per [this documentation](https://github.com/tabvengers/spicy-sections/blob/959b42cbd1379d84565c40a15b2a0ff926046286/README.md#usage).

I tried that at first, but it broke the flex layout. Not sure if I was holding it wrong though.

rzhw:
> There's also some nuance here in that the "All ecosystems" button shouldn't expand when clicked. It's the little arrow next to it that does this. We'd have to also add some hackery to work around that if we use the "All ecosystems" button as the `<summary>`?

That's a great point too - I believe that might be workaround-able by changing

`this.addEventListener('change', () => submitForm(this.closest('form')))`

to

    this.addEventListener('change', () => {
      submitForm(this.closest('form')));
      return false;
    }

In order to prevent the click event bubbling up out of the element.

jonathanmetzman (approved):
lgtm

oliverchang:
> After discussing the above offline I've tried looking into whether we can leverage `<spicy-sections>` without breaking the flex layout.
> 
> Seems like it's possible! @oliverchang please take a look at the above commit, WDYT?

@rzhw Awesome!!!! This is so much simpler. Thanks for doing this.

PR #399, opened 2022-04-12
Allow ecosystem bubbles on homepage to overflow horizontally

### LINKED ISSUE #387: V2 UI: Improve ecosystems wrapping on home page
image

Maybe we can make the bottom row centered, and extend the background to the whole row.

rzhw:
You need to have a small enough screen width for the ecosystems to overflow, which is more likely on touch devices where the scrollbar isn't visually jarring and it's more obvious that cut-off items can be scrolled.

Also updated styling to more closely match mocks.

Fixes #387

AmepCoXoERs4WpT


oliverchang (approved):
thanks!!

PR #402, opened 2022-04-13
Support Markdown rendering for vulnerability details.

oliverchang:
image


gcp/appengine/frontend3/src/templates/vulnerability.html
```diff
@@ -27,7 +27,10 @@ <h1 class="title">{{ vulnerability.id }}</h1>
         <dt>Modified</dt>
         <dd>{{ vulnerability.modified }}</dd>
         <dt>Details</dt>
-        <dd><pre class="details">{{ vulnerability.details }}</pre></dd>
+        <!-- <dd><pre class="details">{{ vulnerability.details }}</pre></dd> -->
```
rzhw, line 30: Probably can remove this?

gcp/appengine/frontend3/src/styles.scss
```diff
@@ -422,6 +422,53 @@ dl.vulnerability-details,
     grid-column: 2;
   }
 
+
+  dd.details {
+    overflow-wrap: break-word;
+    li {
```
rzhw, line 428: I wonder if the styles here and below we should be pulling out into a more top-level position.

rzhw (changes requested):

gcp/appengine/frontend3/src/templates/vulnerability.html
oliverchang: Good catch, removed!

gcp/appengine/frontend3/src/styles.scss
oliverchang: Hmm, this is to set specific styles for the user provided markdown, which probably don't make sense to be more top level? 

Do you have a suggestion for how to make these all cleaner?

rzhw: Good point - I felt like I saw duplication but when I took a closer look it's just this bit that I'm concerned with:

```css

    li {
      $active-spacing: 6px;
      margin: -$active-spacing 0;
      padding: $active-spacing 0;
    }
```

It's duplicated from `.top-bar` - this makes sense in a horizontal list of links context but doesn't seem to be styling that makes sense for typical paragraph content?

rzhw (approved):

gcp/appengine/frontend3/src/styles.scss
oliverchang: removed as discussed!

Issue #321, opened 2022-02-24
Bulk query API


JamieMagee:
A bulk query API would allow developers to more easily query the API without hitting rate limits. It would also help with scenarios like #257, where an SBOM will contain many dependencies.

maitrayshah:
@JamieMagee  Not sure if this helps you answer question but I had a similar requirement (Issue - https://github.com/google/osv/issues/306)
FWIW, we went with using the GCS Bucket option and do version comparison on our end. Would be nice to leverage the API but for now just downloading the zip and using that.

JamieMagee:
Yes, I went a similar route. I wrote a package, https://github.com/JamieMagee/osv-offline, to do this in Node.js. It's to support [this issue](https://github.com/renovatebot/renovate/issues/6562) in Renovate

maitrayshah:
Sweet, will check it out.
Also, renovate looks pretty interesting thanks for linking it as well.

JamieMagee:
Yeah, returning a minimal, dehydrated, response is reasonable.

Thanks for clarifying about the rate limits. I've gone the route of creating an offline copy of the database for use in Renovate for now, because individual queries for each dependency in a project would be too much of a performance hit. You know how large the dependency graphs can get for node projects 😅  We could probably migrate to a bulk query API, but we can cross that bridge when it comes to it.

oliverchang:
The implementation of this has started in #418.

oliverchang:
This is deployed to our staging API instance at `https://api-staging.osv.dev/v1/querybatch`.

Sample query:

```json                                                                                                                                                                                                                                                                        
{
  "queries": [
    {
      "package": {
        "purl": "pkg:pypi/antlr4-python3-runtime@4.7.2"
      }
    },
    {
      "commit": "6879efc2c1596d11a6a6ad296f80063b558d5e0f"
    },
    {
      "package": {
        "ecosystem": "PyPI",
        "name": "jinja2"
      },
      "version": "2.4.1"
    }
  ]
}
```

oliverchang:
@JamieMagee and others: Would you be willing to try the staging API to see if this would fit your needs?

JosiahOne:
I tried the API in staging and it definitely fits our needs assuming the response ordering is guaranteed to match the request. It seems like it is.

oliverchang:
> I tried the API in staging and it definitely fits our needs assuming the response ordering is guaranteed to match the request. It seems like it is.

That's right. The response ordering will be guaranteed to match the input.

PR #424, opened 2022-05-11
Optimize batch API performance.

oliverchang:
Use asynchronous Datastore queries so individual queries happen in parallel.

Part of #321.


Alan32Liu (approved):
LGTM

PR #429, opened 2022-05-18
Move PyPI bridge call to worker from importer.

oliverchang:
Workers add additional information to vulnerabilities (e.g. by expanding
ranges). With GHSA advisories, they needed this step to be properly
propagated to warehouse.

Also make the source repo processing faster when loading it from
scratch, rather than enumerating through every single commit.

This will help https://github.com/trailofbits/pip-audit/issues/274.


Alan32Liu (approved):
LGTM

PR #433, opened 2022-05-20
Remove unused "log" package from osv.go

sbassett29:
Remove unused "log" package, as go run, etc. throws 'internal/osv/osv.go:6:2: imported and not used: "log"' error.


oliverchang (approved):
thanks a lot!

PR #447, opened 2022-06-03
Add clarification on max versions for batch query API.

### LINKED ISSUE #446: Clarify max query limit on bulk query endpoint
Looks like there's [a limit to the number of queries](https://github.com/google/osv/blob/168a6b2c593d4fac2defbc8546e634256cdfa5b0/gcp/api/server.py#L92-L94) that can be submitted at once with the bulk query endpoint, which is [set to 1000](https://github.com/google/osv/blob/168a6b2c593d4fac2defbc8546e634256cdfa5b0/gcp/api/server.py#L42).

It's understandable but not documented in the api docs as far as I can see.

oliverchang:
Fixes #446


another-rex (approved):
Looks good to me!

PR #463, opened 2022-06-15
feat: Add ability to enumerate Debian package versions

another-rex:
Add the ability to enumerate versions for debian packages via their snapshot api.


lib/osv/ecosystems.py
```diff
@@ -310,6 +310,36 @@ def enumerate_versions(self, package, introduced, fixed, limits=None):
...
+    response = requests.get(url)
+    if response.status_code == 404:
+      raise EnumerateError(f'Package {package} not found')
+    if response.status_code != 200:
+      raise RuntimeError(
+          f'Failed to get Debian versions for {package} with: {response.text}')
+
+    response = response.json()
+
+    versions = list(map(lambda x: x['version'], response['result']))
```
oliverchang, line 336: Prefer list comprehension for this case. 

i.e.

```python
versions = [x['version'] for x in response['result']]
```

is much easier to read.

oliverchang (commented):
very nice!!! 

This looks great, but can we also add a test (similar to the pypi, maven ones) here: https://github.com/google/osv/blob/367d977409547d62713d15b59dec50f0bf1228d5/docker/worker/worker_test.py#L930 ?

lib/osv/ecosystems.py
```diff
@@ -318,6 +348,7 @@ def enumerate_versions(self, package, introduced, fixed, limits=None):
     'NuGet': NuGet(),
     'PyPI': PyPI(),
     'RubyGems': RubyGems(),
+    'Debian': Debian(),
```
oliverchang, line 351: Oh, one more thing. Since we are splitting by `Debian:<release version>`, we will need to improve the matching logic here. 

`get` below will need to return the right result for e.g. 'Debian:11'. The release version will need to be tied to the resulting object too since that will affect the enumeration.

lib/osv/ecosystems.py
```diff
@@ -310,6 +310,36 @@ def enumerate_versions(self, package, introduced, fixed, limits=None):
...
+  _API_PACKAGE_URL = 'https://snapshot.debian.org/mr/package/{package}/'
+  versions_to_idx = None
+
+  def sort_key(self, version):
+    if self.versions_to_idx is None:
+      raise Exception('versions to idx not initiated')
+
+    return self.versions_to_idx[version]
+
+  def enumerate_versions(self, package, introduced, fixed, limits=None):
```
oliverchang, line 325: We will need to add the "introduced" version detection in this module here instead of in the converter.

We should keep Debian advisories as close to the source as possible, and put the heavy lifting into our "osv" module/infra.

lib/osv/debian_version_cache.py
```diff
@@ -0,0 +1,18 @@
+"""Caches the debian package first version dataframe"""
+
+import pandas as pd
+
+debian_version_cache = pd.read_pickle(
```
oliverchang, line 5: Is there a large benefit to using pandas here to deserialize and index this? We should aim to minimize the number of third party dependencies we require. 

What if we did the heavy lifting in the tool that generates this index, which just dumps a JSON file that we download and deserialize here as a dict?

oliverchang (commented):
thanks!

lib/osv/debian_version_cache.py
another-rex: There's not really a large benefit, so yeah can do the json option.

lib/osv/debian_version_cache.py
```diff
@@ -0,0 +1,30 @@
+"""Caches the debian package first version dataframe"""
```
oliverchang, line 1: Please add a license header.

lib/osv/ecosystems.py
```diff
@@ -23,6 +23,7 @@
 from . import maven
 from . import nuget
 from . import semver_index
+from . import debian_version_cache
```
oliverchang, line 26: Please put this before line 23 to preserve alphabetic order

lib/osv/debian_version_cache.py
```diff
@@ -0,0 +1,30 @@
...
+
+debian_version_cache = None
+
+
+def initiate_from_cloud_cache(force_reload: bool = False) -> dict[str, dict]:
+  """Load the debian version cache from the Google cloud link, if it hasn't been loaded yet"""
+  global debian_version_cache
+  if debian_version_cache is None or force_reload:
+    res = request.urlopen(
+        'https://storage.googleapis.com/debian-osv/first_package_cache.json.gz')
```
oliverchang, line 14: Please put this in a constan.

lib/osv/ecosystems.py
```diff
@@ -310,23 +310,74 @@ def enumerate_versions(self, package, introduced, fixed, limits=None):
...
+
+    if introduced == '0':
+      # Update introduced to the first version of the debian version
+      introduced = debian_version_cache.get_first_package_version(
+          package, self.debian_release_ver)
+
+    return self._get_affected_versions(versions, introduced, fixed, limits)
+
+
+# All lowercase ecosystem names
```
oliverchang, line 357: technical ecosystem names are case sensitive. could we revert this? it doesn't make our matching any easier right?

lib/osv/ecosystems_test.py
```diff
@@ -82,6 +82,20 @@ def test_nuget(self):
     with self.assertRaises(ecosystems.EnumerateError):
       ecosystem.next_version('doesnotexist123456', '1')
 
+  def test_debian(self):
```
oliverchang, line 85: Thanks for adding this test! Could you also add a more end-to-end test here: https://github.com/google/osv/blob/367d977409547d62713d15b59dec50f0bf1228d5/docker/worker/worker_test.py#L930

lib/osv/debian_version_cache.py
```diff
@@ -0,0 +1,30 @@
...
+    debian_version_cache = json.loads(gzip.decompress(res.read()))
+
+
+def get_first_package_version(package_name: str, release_number: str) -> str:
+  """Get first package version"""
+
```
oliverchang, line 20: nit: unnecesaary newline.

lib/osv/ecosystems.py
```diff
@@ -310,23 +310,74 @@ def enumerate_versions(self, package, introduced, fixed, limits=None):
...
+  _API_PACKAGE_URL = 'https://snapshot.debian.org/mr/package/{package}/'
+  versions_to_idx = None
+  debian_release_ver: str
+
+  def __init__(self, debian_release_ver: str):
+    self.debian_release_ver = debian_release_ver
+    debian_version_cache.initiate_from_cloud_cache()
+
+  def sort_key(self, version):
+    if self.versions_to_idx is None:
```
oliverchang, line 325: We actually need to implement version sorting independently from the `enumerate_versions` method after all. This can come in a separate PR. 

Our OSV evaluation algorithm relies on this for expanding the individual versions from the events list: https://github.com/google/osv/blob/367d977409547d62713d15b59dec50f0bf1228d5/lib/osv/impact.py#L390

oliverchang (commented):
nice!! This looks almost good :)

lib/osv/ecosystems.py
another-rex: Will revert this. Is Debian's ecosystem name capitalized? i.e. `Debian:9` or `debian:9`?

lib/osv/ecosystems.py
another-rex: Resolved using univer's debian.py library

lib/osv/ecosystems.py
another-rex: Using capitalized version

lib/osv/ecosystems.py
```diff
@@ -310,14 +311,52 @@ def enumerate_versions(self, package, introduced, fixed, limits=None):
...
+    if response.status_code != 200:
+      raise RuntimeError(
+          f'Failed to get Debian versions for {package} with: {response.text}')
+
+    response = response.json()
+
+    versions = [x['version'] for x in response['result']]
+    # Reverse so versions is in ascending order
+    versions.reverse()
+    self.versions_to_idx = dict(zip(versions, range(len(versions))))
```
oliverchang, line 342: This is no longer needed right?

lib/osv/ecosystems.py
```diff
@@ -310,14 +311,52 @@ def enumerate_versions(self, package, introduced, fixed, limits=None):
...
+    if response.status_code == 404:
+      raise EnumerateError(f'Package {package} not found')
+    if response.status_code != 200:
+      raise RuntimeError(
+          f'Failed to get Debian versions for {package} with: {response.text}')
+
+    response = response.json()
+
+    versions = [x['version'] for x in response['result']]
+    # Reverse so versions is in ascending order
```
oliverchang, line 340: Let's just sort ourselves to avoid this assumption altogether.

lib/osv/debian_version_cache.py
```diff
@@ -0,0 +1,42 @@
...
+"""Caches the debian package first version dataframe"""
+import gzip
+from urllib import request
+import json
+
+CLOUD_API_CACHE_URL = 'https://storage.googleapis.com/debian-osv/first_package_cache.json.gz'
+debian_version_cache = None
+
+
+def initiate_from_cloud_cache(force_reload: bool = False) -> dict[str, dict]:
```
oliverchang, line 23: nit: prefix this with '_' to mark this as a private method.

lib/osv/debian_version_cache.py
```diff
@@ -0,0 +1,42 @@
...
+  """Load the debian version cache from the Google cloud link, if it hasn't been loaded yet"""
+  global debian_version_cache
+  if debian_version_cache is None or force_reload:
+    res = request.urlopen(CLOUD_API_CACHE_URL)
+    debian_version_cache = json.loads(gzip.decompress(res.read()))
+
+
+def get_first_package_version(package_name: str, release_number: str) -> str:
+  """Get first package version"""
+  if debian_version_cache is None:
```
oliverchang, line 33: Just initialize the cache here.

oliverchang (approved):
very nice!! some more minor comments. otherwise this looks pretty good.

docker/worker/Pipfile
```diff
@@ -8,7 +8,7 @@ google-cloud-pubsub = "*"
 google-cloud-datastore = "*"
 google-cloud-logging = "*"
 google-cloud-ndb = "*"
-pygit2 = "*"
+pygit2 = "==1.5.0"
```
oliverchang, line 11: Can you add a comment here explaining why we need this? And possibly also a TODO.

another-rex:
Oh accidentally pressed rereview request, please ignore

lib/osv/ecosystems.py
```diff
@@ -310,6 +311,44 @@ def enumerate_versions(self, package, introduced, fixed, limits=None):
...
+    versions.reverse()
```
oliverchang, line 340: No need for the reverse call at all?

oliverchang (approved):
very nice work!! just one minor comment before we merge this.

lib/osv/ecosystems.py
another-rex: Well I'm thinking it probably sorts faster if it's already in the correct order so might as well do it. The actual speed improvement is almost certainly negligible, so yeah I'll remove it for less clutter.

oliverchang: Right, any performance gains are negligible :) Much better for the code to be simpler here.

PR #480, opened 2022-07-01
Debian support for frontend and indexing

another-rex:
- Support ecosystem with sub ecosystem names using :
- Aggregate sub-ecosystems on the frontend 
- Frontend ecosystem_count unit tests


lib/osv/models.py
```diff
@@ -336,11 +336,17 @@ def _pre_put_hook(self):
     })
     self.project.sort()
 
-    self.ecosystem = list({
-        pkg.package.ecosystem
+    ecosystems_set = {
+        pkg.package.ecosystem.split(':')
```
oliverchang, line 340: This shouln't be split here right?

gcp/appengine/frontend_handlers_test.py
```diff
@@ -0,0 +1,67 @@
...
+    self.tmp_dir = tempfile.mkdtemp()
+
+    tests.mock_datetime(self)
+
+  def tearDown(self):
+    shutil.rmtree(self.tmp_dir, ignore_errors=True)
+
+  @mock.patch('osv.Bug.query')
+  @mock.patch('frontend_handlers.osv_get_ecosystems')
+  def test_ecosystem_counts(self, mock_osv_get_ecosystems: mock.Mock,
```
oliverchang, line 42: thanks for adding this :)

lib/osv/models.py
```diff
@@ -336,11 +336,17 @@ def _pre_put_hook(self):
...
         for pkg in self.affected_packages
         if pkg.package.ecosystem
-    })
+    }
+
+    # For all ecosystems that specify a specific version with colon,
+    # also add the base name
+    ecosystems_set.update({x.split(':')[0] for x in ecosystems_set if ':' in x})
```
oliverchang, line 347: We can skip the `if ':' in x`, since split() will always at least return an array of size 1.

oliverchang (commented):
very nice!

oliverchang (approved):

PR #481, opened 2022-07-04
Automatic retry and backoff

another-rex:


lib/osv/ecosystems.py
```diff
@@ -326,7 +327,14 @@ def sort_key(self, version):
 
   def enumerate_versions(self, package, introduced, fixed, limits=None):
     url = self._API_PACKAGE_URL.format(package=package.lower())
-    response = requests.get(url)
+    session = requests.session()
+    retries = Retry(
+        backoff_factor=1,
```
oliverchang, line 332: Can you move these into class constants? Then we can use `self.BACKOFF_FACTOR` and `self.RETRY_TOTAL` here. 

Also, let's be a bit more conservative here and set backoff_factor to 4. 

If i understand the calculations correctly, this gives us `4 * 2^(6 - 1) = 128 seconds)` of delay in the last retry, as opposed to 32 seconds which seems insufficient.

oliverchang (approved):

PR #482, opened 2022-07-04
URL encode generated PURLs.

### LINKED ISSUE #473: PURLs for scoped NPM packages are invalid
The package URLs for scoped NPM packages (e.g. [`@hapi/hoek`](https://osv.dev/vulnerability/GHSA-22h7-7wwg-qmgg)) are invalid. Parsing them with any package URL library fails.

According to [the spec](https://github.com/package-url/purl-spec/blob/master/PURL-SPECIFICATION.rst#rules-for-each-purl-component), segments in the namespace (here: `@hapi`) must be percent-encdoded.

So

```
pkg:npm/@hapi/hoek
```

should be

```
pkg:npm/%40hapi/hoek
```

On the same note, the name segment must be percent-encoded, too. I haven't encountered a PURL in OSV where the package name contains characters that'd need encoding, but if this is done for the namespace, it should be considered for the name as well.

oliverchang:
Fixes #473.

### DIFF
--- a/lib/osv/purl_helpers.py
+++ b/lib/osv/purl_helpers.py
@@ -13,6 +13,8 @@
 # limitations under the License.
 """PURL conversion utilities."""
 
+from urllib.parse import quote
+
 PURL_ECOSYSTEMS = {
     'crates.io': 'cargo',
     'Hex': 'hex',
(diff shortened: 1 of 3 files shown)


lib/osv/purl_helpers.py
```diff
@@ -27,6 +29,12 @@
 }
 
 
+def _url_encode(package_name):
+  """URL encode a PURL `namespace/name` or `name`."""
+  parts = package_name.split('/')
+  return '/'.join(quote(p) for p in parts)
```
another-rex, line 35: Are there possibly qualifiers or subpath in the package name purls? If so this will also % encode the ? and # separators.

another-rex (approved):
Just one question, otherwise LGTM!

lib/osv/purl_helpers.py
oliverchang: Good question! PURLs do support qualifiers (https://github.com/package-url/purl-spec/blob/master/PURL-SPECIFICATION.rst#package-url-specification-v10x)

However, we don't generate them currently, so there's no need to account for them.

PR #483, opened 2022-07-04
Improve App Engine deployment script.

oliverchang:
- Set -ex in bash script.
- Use pipenv through python module instead of the system default.


another-rex (approved):
LGTM!

Issue #488, opened 2022-07-06
Add a caching layer to `ecosystems`


oliverchang:
To reduce API load when we retrieve lists of versions for packages in our ecosystems code.

oliverchang:
One way to do this might be to add a `set_cache` method to https://github.com/google/osv.dev/blob/57ac34b2f15a80c584a8fa02a50d16e6fc4e10fc/lib/osv/ecosystems.py. This would need an interface defined for what methods that cache object needs to support.

If set, `Ecosystem` implementations can that cache.

another-rex:
Debian caching completed, implementing additional caching for other ecosystems is continued in #498

PR #490, opened 2022-07-06
Redis caching ecosystems

another-rex:
Solves #488 by implementing redis caching for debian ecosystems. Other ecosystems have not yet been implemented.


lib/osv/ecosystems.py
```diff
@@ -381,3 +399,9 @@ def get(name: str) -> Ecosystem:
     return Debian(name.split(':')[1])
   else:
     return _ecosystems.get(name)
+
+
+def set_cache(redis_host: str, redis_port: str):
```
oliverchang, line 404: Let's make this a bit more extensible rather than tying this too much to Redis. This also makes testing easier. 

We can create something like the following:

```python
class Cache:
  """Cache interface."""
  
  def get(key):
    raise NotImplementedError

  def set(key, value, expiry):
    raise NotImplementedError


class RedisCache(Cache):
  """Redis cache implementation."""

  def __init__(self, redis_host, redis_port):
   ....
  
  def get(...)

# If we need to in another PR. 
class InMemoryCache(Cache):
  ...
```


Then we can have set_cache take in instances of `Cache`.

lib/osv/ecosystems.py
```diff
@@ -327,23 +332,36 @@ def __init__(self, debian_release_ver: str):
   def sort_key(self, version):
     return DebianVersion.from_string(version)
 
-  def enumerate_versions(self, package, introduced, fixed, limits=None):
+  def _get_package_version_response(self, package):
+    """Return the response from debian snapshot api for the given package"""
     url = self._API_PACKAGE_URL.format(package=package.lower())
+
+    if redis_cache is not None:
```
oliverchang, line 339: Let's factor this all out into a reusable `RequestHelper` class (or something similarly named).

```python
class RequestHelper:
  retry_total: int = 0
  backoff_factor: int = 0
  cache: Cache = None
  cache_ttl:  int = 0

  def get(url):
    ...
```

Then we can just use this like:

```python
request_helper = RequestHelper(
  self.RETRY_TOTAL, self._BACKOFF_FACTOR, 
  cache=global_cache,
  cache_ttl=self._CACHE_TTL)

response = request_helper.get(...)
```

lib/osv/ecosystems.py
```diff
@@ -318,6 +322,7 @@ class Debian(Ecosystem):
   _API_PACKAGE_URL = 'https://snapshot.debian.org/mr/package/{package}/'
   _BACKOFF_FACTOR = 5
   _RETRY_TOTAL = 7
+  _REDIS_TTL_SECONDS = 3600
```
oliverchang, line 325: We can make this even longer, like 6 hours.

Also let's break this up into

```python
6 * 60 * 60  # 6 hours.
```

To make the duration clearer to readers.

oliverchang (changes requested):
nice! some comments.

another-rex:
Added a test in ecosystem_test to make sure cache is being called, should I also add a unit test for request_helpers?

oliverchang:
> Added a test in ecosystem_test to make sure cache is being called, should I also add a unit test for request_helpers?

Yes please! Let's add a test with a mock/in-memory cache to make sure that the cache is being used properly.

lib/osv/cache.py
```diff
@@ -0,0 +1,75 @@
...
+  """Cache Interface"""
+
+  def get(self, key):
+    raise NotImplementedError
+
+  def set(self, key, value, ttl):
+    raise NotImplementedError
+
+
+class RedisCache(Cache):
```
oliverchang, line 31: Can we add this implementation to the worker.py instead? That way, not all users of the "osv" library has to install the redis package.

lib/osv/request_helper.py
```diff
@@ -0,0 +1,73 @@
...
+from requests.adapters import HTTPAdapter, Retry
+
+from .cache import Cache
+
+DEFAULT_BACKOFF_FACTOR = 5
+DEFAULT_RETRY_TOTAL = 7
+DEFAULT_REDIS_TTL_SECONDS = 6 * 60 * 60
+
+
+class RequestException(Exception):
```
oliverchang, line 26: nit: per https://google.github.io/styleguide/pyguide.html#244-decision, exception class names should end with "Error".

lib/Pipfile
```diff
@@ -11,5 +11,6 @@ pyyaml = "*"
 # TODO: Update Dockerfile to use newer libgit2 version
 pygit2 = "==1.5.0"
 attrs = "*"
+redis = "*"
```
oliverchang, line 14: See comment below on keeping the redis cache implementation in the worker instead.

oliverchang (commented):
nice! this almost LGTM, with just some minor remaining comments.

lib/osv/cache.py
another-rex: Moved to worker, though it seems like something else is also depending on redis, because the pipfile.lock file still has the redis dependency even though it's been removed from the pipfile

another-rex: pipenv graph shows `google-cloud-ndb` pulling in redis as a dependency

oliverchang: Interesting. It's not ideal in the first place that `google-cloud-ndb` is included as a dependency for OSV, so let's not make this worse and keep the explicit redis dependency in the worker.

oliverchang (approved):

oliverchang:
Looks like something broke with the worker tests (https://github.com/google/osv.dev/pull/490/checks?check_run_id=7209020773). Could you please take a look?

another-rex:
> Looks like something broke with the worker tests (https://github.com/google/osv.dev/pull/490/checks?check_run_id=7209020773). Could you please take a look?

Maybe the issue is the name of the dependency being weird. Adding the dependency as:
```
osv = {editable = true, path = "./lib"}
```
or
```
[packages.osv]
path = "./lib"
editable = true
```
seem to solve it.

oliverchang:
> > Looks like something broke with the worker tests (https://github.com/google/osv.dev/pull/490/checks?check_run_id=7209020773). Could you please take a look?
> 
> Maybe the issue is the name of the dependency being weird. Adding the dependency as:
> 
> ```
> osv = {editable = true, path = "./lib"}
> ```
> 
> or
> 
> ```
> [packages.osv]
> path = "./lib"
> editable = true
> ```
> 
> seem to solve it.

very nice!!

lib/osv/ecosystems.py
```diff
@@ -327,24 +329,22 @@ def __init__(self, debian_release_ver: str):
   def sort_key(self, version):
     return DebianVersion.from_string(version)
 
+  def _get_package_version_response(self, package):
```
oliverchang, line 332: Is this method still needed?

lib/osv/request_helper.py
```diff
@@ -0,0 +1,73 @@
...
+               backoff_factor=DEFAULT_BACKOFF_FACTOR,
+               retry_total=DEFAULT_RETRY_TOTAL,
+               cache_ttl=DEFAULT_REDIS_TTL_SECONDS):
+    self.backoff_factor = backoff_factor
+    self.retry_total = retry_total
+    self.cache_ttl = cache_ttl
+    self.cache = cache
+
+  def get(self, url):
+    if self.cache is not None:
```
oliverchang, line 53: nit: per https://google.github.io/styleguide/pyguide.html#214-truefalse-evaluations

Let use implicit evaluation rather than comparing against None here.

lib/osv/request_helper.py
```diff
@@ -0,0 +1,73 @@
...
+        total=self.retry_total,
+    )
+    session.mount('https://', HTTPAdapter(max_retries=retries))
+    response = session.get(url)
+
+    if response.status_code != 200:
+      raise RequestError(response)
+
+    text_response = response.text
+    if self.cache is not None:
```
oliverchang, line 70: same here.

lib/osv/cache.py
```diff
@@ -0,0 +1,58 @@
...
+  """In memory cache implementation."""
+
+  key_val_map: typing.Dict[str, _CacheEntry]
+
+  def __init__(self):
+    self.key_val_map = dict()
+
+  def get(self, key):
+    entry = self.key_val_map.get(key)
+    if entry:
```
oliverchang, line 48: nit: prefer early exit over additional intentation.


```python
if not entry: 
  return None

if entry.expiry >= datetime.datetime.now().timestamp():
  return entry.data
else:
  self.key_val_map.pop(key)

return None
```

oliverchang (approved):

PR #496, opened 2022-07-07
Pin gVisor version to workaround issue with latest version.

oliverchang:


another-rex (approved):
LGTM, also able to build on my machine with this change

Issue #486, opened 2022-07-05
GCS importer: watch for changes.


oliverchang:
Currently importing from GCS buckets is done from scratch every day. 

We should be smarter about scheduling imports for OSV entries that have not changed since the last time we imported them. 

Currently, GCS bucket imports are done once per day here:

https://github.com/google/osv.dev/blob/57ac34b2f15a80c584a8fa02a50d16e6fc4e10fc/docker/importer/importer.py#L288

And we simply iterate through each file and re-schedule an import: 

https://github.com/google/osv.dev/blob/57ac34b2f15a80c584a8fa02a50d16e6fc4e10fc/docker/importer/importer.py#L297

We should instead track the `modified` date of these entries, and save it as a new property `import_last_modified` here: https://github.com/google/osv.dev/blob/57ac34b2f15a80c584a8fa02a50d16e6fc4e10fc/lib/osv/models.py#L225

We can't reuse `last_modified` since we update it ourselves if we add data to it (e.g. by expanding versions). 

Once we do this, we can increase the frequency of imports (and reduce the latency from ingesting new advisories).

oliverchang:
Note this will need `modified` support for the Debian converter in the OSSF repo.

PR #499, opened 2022-07-08
Skip identical imports by looking at the last modified date

another-rex:
Not sure how to test this locally, so I don't know if this actually works or not. 
Fixes: #486


docker/worker/worker.py
```diff
@@ -472,6 +472,9 @@ def _do_update(self, source_repo, repo, vulnerability, relative_path,
           timestamp=osv.utcnow(),
           status=osv.BugStatus.PROCESSED,
           source_of_truth=osv.SourceOfTruth.SOURCE_REPO)
+    elif bug.import_last_modified == vulnerability.import_last_modified:
```
oliverchang, line 475: Rather than add this complexity here, could we do this in the importer instead? It's better to keep this logic closer to the importer as that's where most of the decisions are being made about what to pocess.  

We're already downloading entries prior to requesting workers to process them:.

https://github.com/google/osv.dev/blob/57ac34b2f15a80c584a8fa02a50d16e6fc4e10fc/docker/importer/importer.py#L303

So we just need to extend this part to parse out the `modified` field for a comparison.

oliverchang (commented):
thanks! for testing, if we move this logic to the importer, we can just extend https://github.com/google/osv.dev/blob/57ac34b2f15a80c584a8fa02a50d16e6fc4e10fc/docker/importer/importer_test.py#L339

docker/worker/worker.py
another-rex: Done, moved the check into importer.

another-rex:
Adding the extra tests are still in progress

docker/importer/importer.py
```diff
@@ -300,7 +300,23 @@ def _process_updates_bucket(self, source_repo):
 
       logging.info('Bucket entry triggered for for %s/%s', source_repo.bucket,
                    blob.name)
-      original_sha256 = osv.sha256_bytes(blob.download_as_bytes())
+      blob_bytes = blob.download_as_bytes()
+
+      # Might be an array, might not
+      vulnerabilities = json.loads(blob_bytes)
+
```
oliverchang, line 307: nit: no need for this newline.

docker/importer/importer.py
```diff
@@ -300,7 +300,23 @@ def _process_updates_bucket(self, source_repo):
 
       logging.info('Bucket entry triggered for for %s/%s', source_repo.bucket,
                    blob.name)
-      original_sha256 = osv.sha256_bytes(blob.download_as_bytes())
+      blob_bytes = blob.download_as_bytes()
+
+      # Might be an array, might not
+      vulnerabilities = json.loads(blob_bytes)
```
oliverchang, line 306: Use `osv.parse_vulnerabilities()` here.

docker/importer/importer.py
```diff
@@ -300,7 +300,23 @@ def _process_updates_bucket(self, source_repo):
...
+
+      if not isinstance(vulnerabilities, list):
+        vulnerabilities = [vulnerabilities]
+
+      def identical_vuln_exist(vuln):
+        bug = osv.Bug.get_by_id(vuln['id'])
+        return bug and bug.import_last_modified == datetime.datetime.fromisoformat(
+            vuln['modified'].replace('Z', '+00:00'))
```
oliverchang, line 314: You can avoid this hack if you use `osv.parse_vulnerabilities()` which properly handles these. 

You just need to call `vuln.modified.ToDatetime()`

docker/importer/importer_test.py
```diff
@@ -358,10 +358,25 @@ def tearDown(self):
     shutil.rmtree(self.tmp_dir, ignore_errors=True)
 
   @mock.patch('google.cloud.pubsub_v1.PublisherClient.publish')
-  def test_bucket(self, mock_publish):
+  @mock.patch('osv.Bug.get_by_id')
```
oliverchang, line 361: Since we have the Datastore emulator set up, let's write real entities instead.

See https://github.com/google/osv.dev/commit/57ac34b2f15a80c584a8fa02a50d16e6fc4e10fc#diff-b2ca69ffb645923d373a2f387c9388ced1d8bf75c3d9137ea56344824a8d5a03L44 

for an example of how I did this.

docker/importer/importer_test.py
```diff
@@ -358,10 +358,25 @@ def tearDown(self):
...
+  def test_bucket(self, mock_get_by_id: mock.MagicMock,
+                  mock_publish: mock.MagicMock):
     """Test bucket updates."""
     imp = importer.Importer('fake_public_key', 'fake_private_key', self.tmp_dir,
                             'bucket')
+
+    def get_by_id_mock_side_effect(id_value):
```
oliverchang, line 368: per above, let's use the datastore emulator for this.

docker/worker/testdata/BLAH-123.yaml
```diff
@@ -27,3 +27,5 @@ affected:
   - branch-v0.1.1
 database_specific:
   specific: 1337
+modified: '2020-01-01T00:00:00Z'
+import_last_modified: '2020-01-01T00:00:00Z'
```
oliverchang, line 31: This is an internal property that shouldn't be externally visible here.

lib/osv/models.py
```diff
@@ -442,6 +444,7 @@ def update_from_vulnerability(self, vulnerability):
 
     if vulnerability.HasField('modified'):
       self.last_modified = vulnerability.modified.ToDatetime()
+      self.import_last_modified = vulnerability.modified.ToDatetime()
```
oliverchang, line 447: This unfortunately won't work. 

This is usually set after we do this: 
https://github.com/google/osv.dev/blob/61afe8fc2aa9b1a0bc17aad4b4f517a96b29c526/lib/osv/impact.py#L538

So it'll just end up being the same as `modified` after we modify it

We should instead save the modified date here (before we call `_analyze_vulnerability`)
https://github.com/google/osv.dev/blob/27eab62add6e0893b724c7fa2e963ca1e6c07494/docker/worker/worker.py#L451

and then set it here on the Bug directly:

https://github.com/google/osv.dev/blob/27eab62add6e0893b724c7fa2e963ca1e6c07494/docker/worker/worker.py#L477

oliverchang (changes requested):
nice! some comments.

docker/importer/importer.py
```diff
@@ -300,7 +300,21 @@ def _process_updates_bucket(self, source_repo):
 
       logging.info('Bucket entry triggered for for %s/%s', source_repo.bucket,
                    blob.name)
-      original_sha256 = osv.sha256_bytes(blob.download_as_bytes())
+      blob_bytes = blob.download_as_bytes()
+      # Might be an array, might not
```
oliverchang, line 304: No need for this comment. It'll always be an array.

docker/worker/worker.py
```diff
@@ -453,6 +453,7 @@ def _do_update(self, source_repo, repo, vulnerability, relative_path,
       logging.warning('%s has an encoding error, skipping.', vulnerability.id)
       return
 
+    og_modified_date = vulnerability.modified
```
oliverchang, line 456: s/og/orig/

oliverchang (approved):

PR #500, opened 2022-07-08
Add "modified" field to bulk query results.

### LINKED ISSUE #492: Include `modified` timestamp in bulk query results.
This allows clients to cache results reliably by using ID+modified as the key.

oliverchang:
Fixes #492.


another-rex (approved):
LGTM!

PR #509, opened 2022-07-12
Much more frequent imports

another-rex:
Remove limitation of once a day imports

### DIFF
--- a/docker/importer/importer.py
+++ b/docker/importer/importer.py
@@ -174,6 +174,10 @@ class Importer:
 
   def schedule_regular_updates(self, repo, source_repo):
     """Schedule regular updates."""
+    if (source_repo.last_update_date and
+        source_repo.last_update_date >= utcnow().date()):
+      return
+
     for bug in osv.Bug.query(
         osv.Bug.status == osv.BugStatus.PROCESSED,
         osv.Bug.is_fixed == False,  # pylint: disable=singleton-comparison
--- a/docker/importer/importer.py
+++ b/docker/importer/importer.py
@@ -174,10 +174,6 @@ class Importer:
 
   def schedule_regular_updates(self, repo, source_repo):
     """Schedule regular updates."""
-    if (source_repo.last_update_date and
-        source_repo.last_update_date >= utcnow().date()):
-      return
-
     for bug in osv.Bug.query(
         osv.Bug.status == osv.BugStatus.PROCESSED,
         osv.Bug.is_fixed == False,  # pylint: disable=singleton-comparison
--- a/docker/importer/importer.py
+++ b/docker/importer/importer.py
@@ -287,10 +287,6 @@ class Importer:
 
   def _process_updates_bucket(self, source_repo):
     """Process updates from bucket."""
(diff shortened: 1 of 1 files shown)


docker/importer/importer.py
```diff
@@ -174,10 +174,6 @@ def import_new_oss_fuzz_entries(self, repo, oss_fuzz_source):
 
   def schedule_regular_updates(self, repo, source_repo):
     """Schedule regular updates."""
-    if (source_repo.last_update_date and
```
oliverchang, line 177: This was the only place this property was being used. Let's just get rid of it completely (from models.py too).

oliverchang (commented):
thanks!

osv/impact.py
```diff
@@ -91,7 +91,9 @@ def get_affected(self,
                    fix_commits,
                    limit_commits=None):
     """"Get list of affected tags and commits for a bug given regressed and
-    fixed commits."""
+
```
oliverchang, line 94: unintentional change?

osv/impact.py
```diff
@@ -196,8 +198,11 @@ def _get_equivalent_commit(self,
                              to_search,
                              target_commit,
                              detect_cherrypicks=True):
-    """Find an equivalent commit at to_search, or None. The equivalent commit
-    can be equal to target_commit."""
+    """Find an equivalent commit at to_search, or None.
+
+    The equivalent commit
```
oliverchang, line 203: unintentional change?

osv/semver_index.py
```diff
@@ -64,7 +64,9 @@ def parse(version):
 
 def normalize(version):
   """Normalize semver version for indexing (to allow for lexical
-  sorting/filtering)."""
+
+  sorting/filtering).
```
oliverchang, line 68: unintentional change?

osv/vulnerability_pb2.py
```diff
@@ -10,740 +10,1203 @@
```
oliverchang, line 10: was this meant to be changed here?

osv/third_party/univers/gem.py
```diff
@@ -21,15 +21,15 @@
 
 
 class InvalidRequirementError(AttributeError):
-    pass
```
oliverchang, line 24: no need to format third_party code.

docker/importer/importer.py
oliverchang, line 177: Sorry, I completely forgot about this case where we still need it. Let's revert to your old way for now.

osv/impact.py
another-rex: Ah... I set up a git hook to run pyformat on all files committed. Not sure why pyformat likes adding blank lines to multiline doc comments though

docker/importer/importer.py
oliverchang: The condition we want to delete is the one in  `_process_updates_bucket`. This one is the the one we need to keep.

another-rex: Ah ok, should be fixed now

oliverchang (approved):

PR #516, opened 2022-07-13
Fix App Engine and API deployment.

oliverchang:
API: Use pipenv to install dependnecies rather than generating
requirements.

App Engine: Exclude the '-e ' dependency from the generated
requirements.txt

Also,
- Update the API server to use Python 3.8 to be consistent with other
  environments.
- Unpin pygit2 in the root Pipfile.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -7,9 +7,7 @@ name = "pypi"
 google-cloud-ndb = "*"
 semver = "*"
 pyyaml = "*"
-# Has to be locked to 1.5.0 because docker pulls down specific libgit2 version that only works with 1.5.0
-# TODO: Update Dockerfile to use newer libgit2 version
-pygit2 = "==1.5.0"
+pygit2 = "*"
 attrs = "*"
 
 [dev-packages]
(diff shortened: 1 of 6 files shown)


Pipfile
```diff
@@ -7,9 +7,7 @@ name = "pypi"
 google-cloud-ndb = "*"
 semver = "*"
 pyyaml = "*"
-# Has to be locked to 1.5.0 because docker pulls down specific libgit2 version that only works with 1.5.0
-# TODO: Update Dockerfile to use newer libgit2 version
-pygit2 = "==1.5.0"
+pygit2 = "*"
```
another-rex, line 10: This should probably be `= "1.9.2"` to match the version used by the Pipfile in worker.

another-rex (changes requested):

Pipfile
oliverchang: This is the root Pipfile which will be used for dev environments (and eventually setup.py). We can't make any assumptions there.

another-rex (approved):

PR #519, opened 2022-07-14
Api fixes and updates

### LINKED ISSUE #518: Make API queries work against generic "Debian" ecosystem
Currently, 

```
curl -X POST -d '{"version": "2.2.12-1+deb10u1",
    "package": {"name": "gnupg2", "ecosystem": "Debian:10"}}' "https://api.osv.dev/v1/query"
````

Works, but just specifying "Debian" doesn't, i.e.

```
curl -X POST -d '{"version": "2.2.12-1+deb10u1",      
    "package": {"name": "gnupg2", "ecosystem": "Debian"}}' "https://api.osv.dev/v1/query"
```

We should make the latter work as well, given that all Debian packages live in the same global version namespace.

### PR DESCRIPTION
**Pull request by another-rex**
Fixes #518

Also updates grpcio, and support absolute paths for service account json

### CODE REVIEW

File: gcp/api/server.py [hunk 1]
```diff
@@ -222,14 +219,16 @@ def _is_version_affected(affected_packages,
...
+  """Returns whether or not the given version is within an affected ECOSYSTEM range."""
   for affected_package in affected_packages:
     if package_name and package_name != affected_package.package.name:
       continue
 
-    if ecosystem and ecosystem != affected_package.package.ecosystem:
-      continue
+    if ecosystem:
+      # If package ecosystem has a :, also try ignoring parts after it.
+      if affected_package.package.ecosystem != ecosystem and affected_package.package.ecosystem.split(
```
oliverchang, line 229: nit: I think this line is long 

side note: I think our linting is broken and not picking this up properly.

oliverchang: we can simplify this a bit:

```python
normalized_ecosystem = affected_package.package.ecosystem.split(':')[0]
if normalized_ecosystem != ecosystem:
  continue
```

And actually, since we have a number of places where we are doing this ':' splitting (e.g. in `models.py`), let's just add a helper function to `ecosystems.py` called `normalize` which doese this splitting.

oliverchang (commented):
nice! some comments

gcp/api/server.py
another-rex: Wouldn't this fail if someone searches for `Debian:11`, instead of just `Debian`?

oliverchang: Good point! 

We should be doing this then:

```python
if ecosystems.normalize(affected_package.package.ecosystem) != ecosystems.normalize(ecosystem):
  continue
```

gcp/api/server.py
```diff
@@ -223,13 +224,17 @@ def _is_version_affected(affected_packages,
...
+  """
   for affected_package in affected_packages:
     if package_name and package_name != affected_package.package.name:
       continue
 
-    if ecosystem and ecosystem != affected_package.package.ecosystem:
-      continue
+    if ecosystem:
+      # If package ecosystem has a :, also try ignoring parts after it.
+      if affected_package.package.ecosystem != ecosystem and\
```
oliverchang, line 235: nit: prefer ( ) for line continuations per https://google.github.io/styleguide/pyguide.html#32-line-length

Also can we just do:

```python
if (ecosystems.normalize(affected_package.package.ecosystem) !=
    ecosystems.normalize(ecosystem)): 
  continue
```

This makes the logic a bit easier to follow.

another-rex: Yes, but that behavior will also be different, if someone is searching specifically for Debian:4.0 for example, that will show packages with Debian:11. With the current implementation, only if they search for the general Debian ecosystem, will it show all debian entries

oliverchang: Thanks for clarifying. That makes sense. Let's keep your approach then.

Can you add some integration tests for this?

oliverchang (approved):

PR #528, opened 2022-07-18
Update ecosystem to refresh first_version_cache at least every 24 hours

another-rex:


osv/debian_version_cache.py
```diff
@@ -17,28 +17,42 @@
...
-  global debian_version_cache
-  if debian_version_cache is None or force_reload:
-    res = request.urlopen(CLOUD_API_CACHE_URL)
-    debian_version_cache = json.loads(gzip.decompress(res.read()))
+    for package, version in sources.items():
+      debian_version_cache.set((package, release), version,
+                               CACHE_DURATION_SECONDS)
 
 
 def get_first_package_version(package_name: str, release_number: str) -> str:
```
oliverchang, line 42: Let's just create a Python decorator for these cases.

And let's restructure the caching like so:

e.g. 

```python
@cache.Cached(cache=debian_version_cache, ttl=CACHE_DURATION_SECONDS)
def get_debian_first_versions():
  ...

def get_first_package_version(...):
 first_versions = get_debian_first_versions()
 return first_versions.get((package_name, release_number))
```

cache.Cached can be implemented something similar to

```python
def Cached(cache: ..., ttl: ...)
  def decorator(func):
    @functools.wraps(func)
    def wrapper(*args, **kwargs):
       if cache_hasnt_expired:
         return cached_value
       
       value = func()
       set_cache(value)
       return value
    return wrapper
  
  return decorator
```

The key value can be constructed by doing some reflection (e.g. taking __file__ and the name of the function)

another-rex: This will change the the behavior of the cache, the current implementation will also try to update the cache when it can't find the package it is looking for. The proposed implementation from my understanding is essentially having a cache with one key, which is either expired or not expired.

oliverchang: Actually, taking a step back here, we might not need to cache this at all. 

If instead, we structure our first packages version data like so in the bucket:

```json
gs://debian-osv/first_package_versions/<package>/<release>
```

Then we can just make a http request on-demand to get the latest without any risk of stale data.

another-rex: Good idea, added both the caching decorator and the new format, so each package can be retrieved individually, but since once received it is very unlikely to change, it's cached for 24 hours.

osv/cache.py
```diff
@@ -56,3 +59,30 @@ def get(self, key):
 
   def set(self, key, value, ttl):
     self.key_val_map[key] = _CacheEntry(value, ttl)
+
+
+def Cached(cache: Cache, ttl: int = 60 * 60, cache_blacklist=frozenset()):
```
oliverchang, line 64: nit: s/blacklist/blocklist/

osv/cache.py
```diff
@@ -56,3 +59,30 @@ def get(self, key):
...
+  :param cache_blacklist: Avoid caching these results
+  :param cache: Cache object to store information
+  :param ttl: Time to live in seconds (default 1 hour)
+  """
+
+  def decorator(func):
+
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs):
+      key = frozenset(getcallargs(func, *args, **kwargs).items())
```
oliverchang, line 76: I think we should include the function module + name here too. It seems pretty likely that one may re-use the same cache object for multiple functions.

oliverchang (approved):
nice!

osv/debian_version_cache.py
```diff
@@ -14,31 +14,29 @@
...
-  """Load the debian version cache from the Google cloud link, if it hasn't been loaded yet"""
-  global debian_version_cache
-  if debian_version_cache is None or force_reload:
-    res = request.urlopen(CLOUD_API_CACHE_URL)
-    debian_version_cache = json.loads(gzip.decompress(res.read()))
+debian_version_cache = cache.InMemoryCache()
 
 
+@cache.Cached(
+    debian_version_cache, 24 * 60 * 60, cache_blacklist=frozenset({'0'}))
```
oliverchang, line 29: Also, we're not using `cache_blacklist` right? Would prefer to remove it add and it later if we ever need it (YAGNI).

another-rex: Well this is using it, when the web returns 404, we don't want to cache '0' response right? since it means it's a new package so that information will be outdated very soon. 

I suppose 24 hours cache is not very long, especially since the first_package_finder only runs once every 24 hours anyway, so it can be removed.

osv/cache.py
another-rex: I thought that was the benefit of passing in the cache object, so that the caller can control what sharing of cache they want to have... So if they pass the same cache object into two functions, we assume that's intentional?

osv/debian_version_cache.py
oliverchang: Oops I somehow missed that. My bad! 

We should just move this logic to the caller instead, and raise an exception in this case to make this behaviour more transparent.
i.e. in `ecosystems.py`

```python
try:
     introduced = debian_version_cache.get_first_package_version(                                                                                                                                                                                                                                                                   
          package, self.debian_release_ver) 
except VersionNotExist:
    pass  # introduced is already '0' at this point
```

osv/cache.py
oliverchang: That seems like it would be more surprising than not. Less magic is always better than more, and if something truly needs to be shared between two different functions, then that should be refactored out.

It also makes it difficult to e.g. share a single redis cache across multiple functions. There's no easy mechanism to use the same redis instance for multiple functions right now.

osv/debian_version_cache.py
```diff
@@ -12,33 +12,28 @@
...
-    force_reload: bool = False) -> typing.Dict[str, typing.Dict]:
-  """Load the debian version cache from the Google cloud link, if it hasn't been loaded yet"""
-  global debian_version_cache
-  if debian_version_cache is None or force_reload:
-    res = request.urlopen(CLOUD_API_CACHE_URL)
-    debian_version_cache = json.loads(gzip.decompress(res.read()))
+debian_version_cache = cache.InMemoryCache()
 
 
+class VersionNotFoundError(Exception):
```
oliverchang, line 25: nit: add a brief docstring.

osv/cache.py
```diff
@@ -56,3 +59,30 @@ def get(self, key):
 
   def set(self, key, value, ttl):
     self.key_val_map[key] = _CacheEntry(value, ttl)
+
+
+def Cached(cache: Cache, ttl: int = 60 * 60):
+  """
+  Decorates your function to cache results
+  :param cache_blocklist: Avoid caching these results
```
oliverchang, line 67: nit: this is no longer needed right?

osv/cache.py
another-rex: Ah yeah that makes sense

oliverchang (approved):

PR #531, opened 2022-07-18
Add ability to query by package.

oliverchang:
Also add pagination support (currently only for this query mode).


gcp/api/integration_tests.py
```diff
@@ -354,6 +354,62 @@ def test_query_batch(self):
...
+                'ecosystem': 'Maven',
+                'name': 'org.apache.tomcat:tomcat',
+            },
+            'page_token': result['next_page_token'],
+        }))
+
+    result = response.json()
+    vulns_second = set(v['id'] for v in result['vulns'])
+
+    self.assertEqual(set(), vulns_first & vulns_second)
```
andrewpollock, line 385: Could we use `vulns_first.intersection(vulns_second)` here instead of `&` for greater readability? See also the discussion at https://docs.python.org/3/library/stdtypes.html#frozenset.intersection

gcp/api/integration_tests.py
```diff
@@ -354,6 +354,62 @@ def test_query_batch(self):
...
+            'package': {
+                'purl': 'pkg:maven/org.apache.tomcat/tomcat',
+            },
+            'page_token': result['next_page_token'],
+        }))
+
+    result = response.json()
+    vulns_second = set(v['id'] for v in result['vulns'])
+
+    self.assertEqual(set(), vulns_first & vulns_second)
```
andrewpollock, line 411: Same as above.

andrewpollock (commented):
I've made a stylistic readability comment on the integration tests.

another-rex (approved):

gcp/api/integration_tests.py
oliverchang: good call, done!

gcp/api/integration_tests.py
oliverchang: done!

andrewpollock (approved):

PR #538, opened 2022-07-19
Exclude patches for previous versions when enumerating versions

### LINKED ISSUE #529: Debian: overlapping versions
e.g. on https://osv.dev/vulnerability/DSA-5087-1

"2.1.27+dfsg-1+deb10u2" is the fixed version for Debian 10. However, Debian 11 includes "2.1.27+dfsg-1+deb10u2" in the list of affected versions. 

We may need to do some filtering based on the +debXXuX" string here while enumerating for a particular Debian release.

another-rex:
Exclude +deb versions that are not for the current debian version. 

Fixes #529 partially. Because of:
> Some of the older debian releases actually don't use the +debXXuX syntax, but use +<codename>uX. For example:
2.1.22.dfsg1-23+squeeze1.

There might still be some overlapping issues for very old advisories


oliverchang (approved):

oliverchang:
LGTM, but CC @andrewpollock in case he has any input re enumerating package versions for a particular Debian release.

andrewpollock:
Ah, Debian versions. I'll refer you to https://www.debian.org/doc/debian-policy/ch-controlfields.html#s-f-version for the authoritative information

PR #551, opened 2022-07-22
Ignore import time once, if column is set to True in SourceRepository

another-rex:
Not 100% sure this will work the way I think it will.


osv/models.py
```diff
@@ -668,6 +668,8 @@ class SourceRepository(ndb.Model):
   detect_cherrypicks = ndb.BooleanProperty(default=True)
   # Whether to populate "versions" from git ranges.
   versions_from_repo = ndb.BooleanProperty(default=True)
+  # Ignore last import time once
```
oliverchang, line 671: nit add a period to this comment to be consistent with others.

oliverchang (commented):
thanks! mind adding a quick test to importer_test ? that would ensure this works :)

oliverchang (approved):

PR #572, opened 2022-08-02
Add purl fixes for debian packages

another-rex:
Will need to do a reimport after deploying


osv/purl_helpers.py
```diff
@@ -45,4 +46,7 @@ def package_to_purl(ecosystem, package_name):
     # PURLs use / to separate the group ID and the artifact ID.
     package_name = package_name.replace(':', '/', 1)
 
+  if purl_type == 'deb':
+    package_name = 'debian/' + package_name
```
oliverchang, line 50: Should we add `?arch=source` ?

oliverchang (commented):
nice!

osv/purl_helpers.py
another-rex: It probably should? However the way purl is put into the database is just as a string, and queried by identical string matching, to properly match parameters will need more logic.

oliverchang: That's a good point, but we should solve that separately -- possibly by just adding an index which includes the common permutations we'd expect (either no parameters or with "arch=").

osv/models.py
```diff
@@ -352,9 +352,19 @@ def _pre_put_hook(self):
...
     self.ecosystem.sort()
 
-    self.purl = list({
-        pkg.package.purl for pkg in self.affected_packages if pkg.package.purl
-    })
+    def with_and_without_qualifiers(affected_packages) -> list[str]:
+      resulting_set = set()
+      for pkg in affected_packages:
+        if pkg.package.purl:
+          res = pkg.package.purl.split('?')
```
oliverchang, line 359: nit: we can reduce some duplication:

```python
  if pkg.package.purl:
    indexes.append(pkg.package.purl)
    if '?' in pkg.package.purl:
      # Remove qualifiers.
      indexes.append(pkg.package.purl.split('?')[0])
```

osv/models.py
```diff
@@ -352,9 +352,19 @@ def _pre_put_hook(self):
     self.ecosystem = list(ecosystems_set)
     self.ecosystem.sort()
 
-    self.purl = list({
-        pkg.package.purl for pkg in self.affected_packages if pkg.package.purl
-    })
+    def with_and_without_qualifiers(affected_packages) -> list[str]:
+      resulting_set = set()
```
oliverchang, line 356: Does this need to be a set at all?

osv/models.py
```diff
@@ -352,9 +352,19 @@ def _pre_put_hook(self):
     self.ecosystem = list(ecosystems_set)
     self.ecosystem.sort()
 
-    self.purl = list({
-        pkg.package.purl for pkg in self.affected_packages if pkg.package.purl
-    })
+    def with_and_without_qualifiers(affected_packages) -> list[str]:
```
oliverchang, line 355: Can you pull this up into a module level function? 

And name it `_get_purl_indexes` to make this more clear what it's for.

osv/models.py
another-rex: Not sure, it was a set in the previous implementation, so I just added it here. I'm guessing it would only matter if two affected packages have the same purl

oliverchang: Ah, ack. This should indeed still be a set.

osv/models.py
```diff
@@ -62,6 +62,16 @@ def utcnow():
   return datetime.datetime.utcnow()
 
 
+def _get_purl_indexes(affected_packages):
```
oliverchang, line 65: Please add a docstring.

oliverchang (approved):
LGTM with one last comment. Thanks!

PR #577, opened 2022-08-03
Adds alias links and reverse links

another-rex:
Fixes: #452 by adding alias links and reverse links.

Also fixes pressing Load More will remove your query, causing the newly loaded results to be random.


andrewpollock (approved):

gcp/appengine/frontend_handlers.py
```diff
@@ -190,6 +191,28 @@ def add_source_info(bug, response):
...
 
 
+def add_related_aliases(bug: osv.Bug, response):
+  """Add links to other osv entries that's related through aliases"""
+  # Add links to other entires if they exist
+  aliases = []
+  for alias in bug.aliases:
+    result = bug.get_by_id(alias)
+    if result:
+      aliases.append((alias, True))
```
oliverchang, line 201: Can we put this in dicts instead? So that the template can refer to fields like:

`alias.id` and `alias.exists`

gcp/appengine/frontend3/src/templates/vulnerability.html
```diff
@@ -12,13 +12,22 @@ <h1 class="title">{{ vulnerability.id }}</h1>
...
+              <li>
+              {% if alias[1] -%}
+                <a href="/vulnerability/{{ alias[0] }}">{{ alias[0] }}</a>
+              {% else -%}
+                {{ alias[0] }}
+              {% endif %}
+              </li>
             {% endfor %}
+              {% for referenced in vulnerability.alias_referenced %}
+                <li>(<a href="/vulnerability/{{ referenced }}">{{ referenced }}</a>)</li>
```
oliverchang, line 29: Can we associate these with the alias? i.e.

```
CVE-2022-11231 (GHSA-BLAH, PYSEC-BLAH)
CVE-2021-444 (GHSA-123)
```

to make it clearer which one belongs which alias.

oliverchang (commented):
nice!

gcp/appengine/frontend3/src/templates/vulnerability.html
another-rex: This list does not display entries with the same aliased CVE's as the current entry entry, but direct references. I can add a separate search to do the same CVE references as well though.

another-rex: Maybe it's better to add another field called `Referenced By`?

oliverchang: Ah, got it, I misunderstood what this was doing. 

In that case, maybe it doesn't make much sense to distinguish these at all. They can all be part of the "aliases" row since they all reference one another in some way. We'd just want to deduplicate them.

oliverchang (approved):

gcp/appengine/frontend_handlers.py
```diff
@@ -190,6 +191,28 @@ def add_source_info(bug, response):
...
+  response['aliases'] = aliases
+
+  # Add links to other entries if they have this element as an alias
+  if bug.aliases:
+    query = osv.Bug.query(
+        osv.Bug.aliases.IN(bug.aliases), osv.Bug.aliases == bug.db_id)
+  else:
+    query = osv.Bug.query(osv.Bug.aliases == bug.db_id)
+
+  response['alias_referenced'] = [x.id() for x in query if x.id() != bug.id()]
```
oliverchang, line 213: Can this result in duplicates in our UI? If we have an entry in aliases, and then a reverse link here?

gcp/appengine/frontend_handlers.py
```diff
@@ -190,6 +191,30 @@ def add_source_info(bug, response):
...
+  response['aliases'] = aliases
+
+  # Add links to other entries if they have this element as an alias
+  if bug.aliases:
+    query = osv.Bug.query(
+        osv.Bug.aliases.IN(bug.aliases), osv.Bug.aliases == bug.db_id)
+  else:
+    query = osv.Bug.query(osv.Bug.aliases == bug.db_id)
+
+  response['alias_referenced'] = [
```
oliverchang, line 213: Since we're not distinguishing between `alias` and `alias_referenced`, can we merge both into `alias`? It would also simplify the UI logic a bit.

gcp/appengine/frontend_handlers.py
another-rex: Good point, added an additional filter to filter out entries that is already in aliases

gcp/appengine/frontend_handlers.py
another-rex: Merged and now uses a dictionary to avoid duplicates

another-rex:
> **Oliver:**
> Can we associate these with the alias? i.e.
> 
> ```
> CVE-2022-11231 (GHSA-BLAH, PYSEC-BLAH)
> CVE-2021-444 (GHSA-123)
> ```
> 
> to make it clearer which one belongs which alias.

> **Rex:**
> This list does not display entries with the same aliased CVE's as the current entry entry, but direct references. I can add a separate search to do the same CVE references as well though.

Turns out I forgot what my own code did :sweat_smile:. In addition to the direct references, it was also suppose to show two entries with same aliases (though it didn't work because of a bug in the original code).

I now fixed it in the latest commit, and added the brackets when two share the same alias. This does mean potential duplicates for the bracketed entries though.

gcp/appengine/frontend_handlers.py
```diff
@@ -190,6 +191,39 @@ def add_source_info(bug, response):
   response['source_link'] = response['source']
 
 
+def add_related_aliases(bug: osv.Bug, response):
+  """Add links to other osv entries that's related through aliases"""
+  # Add links to other entires if they exist
```
oliverchang, line 196: s/entires/entries/

gcp/appengine/frontend_handlers.py
```diff
@@ -190,6 +191,39 @@ def add_source_info(bug, response):
...
+  aliases = dict()
```
oliverchang, line 197: nit: just use `{}` instead of `dict()`.

gcp/appengine/frontend_handlers.py
```diff
@@ -190,6 +191,39 @@ def add_source_info(bug, response):
...
+  for alias in bug.aliases:
+    result = bug.get_by_id(alias)
+    aliases[alias] = {'exists': result, 'same_alias_entries': list()}
+
+  # Add links to other entries that have the same alias
+  if bug.aliases:
+    query = osv.Bug.query(osv.Bug.aliases.IN(bug.aliases))
+    for x in query:
```
oliverchang, line 205: nit: s/x/other/ to be clearer.

Same with `x_alias` -> `other_alias`.

gcp/appengine/frontend3/src/templates/vulnerability.html
```diff
@@ -12,12 +12,25 @@ <h1 class="title">{{ vulnerability.id }}</h1>
       <dl class="vulnerability-details">
         <dt>Source</dt>
         <dd><a href="{{ vulnerability.source_link }}">{{ vulnerability.source }}</a></dd>
-        {% if vulnerability.aliases %}
+        {% if vulnerability.aliases or vulnerability.alias_referenced %}
```
oliverchang, line 15: delete `alias_referenced` ?

gcp/appengine/frontend_handlers.py
```diff
@@ -190,6 +191,39 @@ def add_source_info(bug, response):
   response['source_link'] = response['source']
 
 
+def add_related_aliases(bug: osv.Bug, response):
+  """Add links to other osv entries that's related through aliases"""
+  # Add links to other entires if they exist
+  aliases = dict()
+  for alias in bug.aliases:
+    result = bug.get_by_id(alias)
+    aliases[alias] = {'exists': result, 'same_alias_entries': list()}
```
oliverchang, line 200: s/list()/[]/

gcp/appengine/frontend_handlers.py
```diff
@@ -190,6 +191,39 @@ def add_source_info(bug, response):
...
+      if x.id() == bug.id():
+        continue
+      for x_alias in x.aliases:
+        if x_alias in aliases:
+          aliases[x_alias]['same_alias_entries'].append(x.id())
+
+  # Add links to other entries if they have this element as an alias
+  query = osv.Bug.query(osv.Bug.aliases == bug.db_id)
+  for x in query:
+    aliases[x.id()] = {'exists': True, 'same_alias_entries': list()}
```
oliverchang, line 215: s/list()/[]/

gcp/appengine/frontend_handlers.py
```diff
@@ -190,6 +191,39 @@ def add_source_info(bug, response):
...
+  """Add links to other osv entries that's related through aliases"""
+  # Add links to other entires if they exist
+  aliases = dict()
+  for alias in bug.aliases:
+    result = bug.get_by_id(alias)
+    aliases[alias] = {'exists': result, 'same_alias_entries': list()}
+
+  # Add links to other entries that have the same alias
+  if bug.aliases:
+    query = osv.Bug.query(osv.Bug.aliases.IN(bug.aliases))
```
oliverchang, line 204: Can we just merge this case with the case on line 214 by doing the IN on bug.aliases + [bug.id()] ?

another-rex:
This update could cause issues with entries like:
https://osv.dev/vulnerability/DSA-4824-1

Running locally it's really slow, but probably will be faster in the cloud.

gcp/appengine/frontend_handlers.py
```diff
@@ -190,6 +191,39 @@ def add_source_info(bug, response):
...
+    query = osv.Bug.query(osv.Bug.aliases.IN(bug.aliases + [bug.id()]))
+    for other in query:
+      if other.id() == bug.id():
+        continue
+      for other_alias in other.aliases:
+        if other_alias in aliases:
+          aliases[other_alias]['same_alias_entries'].append(other.id())
+    # Add links to other entries if they have this element as an alias
+    # This has to be a separate loop to avoid the new aliases entries interfering
+    for other in query:
```
oliverchang, line 213: I think this actually runs the query twice. 

Could you explain what the interference here would be in more detail?

oliverchang:
> This update could cause issues with entries like: https://osv.dev/vulnerability/DSA-4824-1
> 
> Running locally it's really slow, but probably will be faster in the cloud.

We could set a threshold on the maximum number of alias entries to perform this more heavyweight querying. If say >8 entries, we just render everything as is. 

We could probably solve all of this with smarter indexing, but that's probably not worth it for what we're getting here.

gcp/appengine/frontend_handlers.py
another-rex: Oh... in that case we might want to just put it in the same loop.

The main thing that this is avoiding is inconsistently showing the `(<same-alias>)` links in brackets for the newly added direct references. Because let's say a direct reference `x` is added on loop 4 / 10, if there was another entry that happened to have an alias of `x`, it will not show up if it has already been processed in an earlier loop. 

This is probably pretty rare though, so we can start by putting it into the same loop until it becomes a problem in the future. (We can also pull the entire query into a list in memory and loop twice that way?)

gcp/appengine/frontend_handlers.py
```diff
@@ -190,6 +191,40 @@ def add_source_info(bug, response):
   response['source_link'] = response['source']
 
 
+def add_related_aliases(bug: osv.Bug, response):
+  """Add links to other osv entries that's related through aliases"""
+  # Add links to other entries if they exist
+  aliases = {}
+  for alias in bug.aliases:
+    if len(
+        bug.aliases) <= 8:  # only if there aren't too many, otherwise skip this
```
oliverchang, line 200: nit: can you put 8 into a global CONSTANT above?

oliverchang: Also the line break is a little strange. If this is because of the comment, we could put the comment before line 201 instead.

gcp/appengine/frontend_handlers.py
oliverchang: Hmm.. not sure i fully follow. Could we not just set `exists` to be True in the direct reference case and not touch the existing `same_alias_entries` case?

oliverchang (approved):

Issue #508, opened 2022-07-12
Normalize PyPI packages from sources. 


xu1119:
### Description
PyPI package names are case insensitive. When search vulnerability by ID GHSA-jwqp-28gf-p498, here are two  affected packages: Scrapy and scrapy, but they are the same one package.

https://osv.dev/list?ecosystem=&q=GHSA-jwqp-28gf-p498
![image](https://user-images.githubusercontent.com/26019024/178423477-0fdbb1f4-35a0-443c-b6bb-6951718ea542.png)

oliverchang:
Hi, 

Thanks for reporting this. This is a known issue (https://github.com/ossf/osv-schema/pull/42) on the GitHub advisories side. My understanding is that this may take some time to fix on their end. 

We could also potentially normalize this ourselves, so I'll keep this open as a FR for that.

Issue #521, opened 2022-07-14
Fix lint action


oliverchang:
It's currently broken and not showing up as red even though there are a number of failures.

PR #578, opened 2022-08-05
Normalize PyPI package names

andrewpollock:
Fixes: #508


osv/models.py
```diff
@@ -332,18 +332,25 @@ def _tokenize(self, value):
     value_lower = value.lower()
     return re.split(r'\W+', value_lower) + [value_lower]
 
+  def _normalize(self, name):
+    """Normalize package name per https://peps.python.org/pep-0503/#normalized-names"""
+
```
oliverchang, line 337: nit: remove this newline.

osv/models.py
```diff
@@ -332,18 +332,25 @@ def _tokenize(self, value):
...
+    return re.sub(r"[-_.]+", "-", name).lower()
```
oliverchang, line 338: nit: use single quotes rather than double quotes.

osv/models.py
```diff
@@ -332,18 +332,25 @@ def _tokenize(self, value):
     value_lower = value.lower()
     return re.split(r'\W+', value_lower) + [value_lower]
 
+  def _normalize(self, name):
```
oliverchang, line 335: nit: s/_normalize/_normalize_pypi/ to be clearer. (but see below for another comment which makes this no longer relevant).

osv/models.py
```diff
@@ -332,18 +332,25 @@ def _tokenize(self, value):
...
+    for pkg in self.affected_packages:
       # Set PURL if it wasn't provided.
-      if not affected_package.package.purl:
-        affected_package.package.purl = purl_helpers.package_to_purl(
+      if not pkg.package.purl:
+        pkg.package.purl = purl_helpers.package_to_purl(
             ecosystems.normalize(affected_package.package.ecosystem),
-            affected_package.package.name)
+            pkg.package.name)
+      if pkg.package.ecosystem and pkg.package.ecosystem == 'PyPI':
```
oliverchang, line 352: nit: no need for the first part of the if, as if it's not set the second comparison will fail anyway. 

That said, I would suggest refactoring this into its own function like so:

```python
def _normalize_package(package):
  """docstring here"""
  if package.ecosystem == 'PyPI':
     # per link...
    return re.sub(r"[-_.]+", "-", package.name).lower()

#  ... in the pre_put_hook 
for pkg in self.affected_packages:
  pkg.package = _normalize_package(pkg.package)
```

osv/models.py
```diff
@@ -332,18 +332,25 @@ def _tokenize(self, value):
...
     search_indices = set()
 
     search_indices.update(self._tokenize(self.id()))
 
-    for affected_package in self.affected_packages:
+    for pkg in self.affected_packages:
       # Set PURL if it wasn't provided.
-      if not affected_package.package.purl:
-        affected_package.package.purl = purl_helpers.package_to_purl(
+      if not pkg.package.purl:
```
oliverchang, line 348: This should go after your normalization, as otherwise the PURL won't get the correct normalized value.

oliverchang (commented):
nice! some comments.

Please also update the test here: https://github.com/google/osv.dev/blob/a4b682a32575cc3314a5ef83c8e91b70c60f7b77/docker/worker/worker_test.py#L891

to make the test data have an un-normalized name.

osv/models.py
andrewpollock: Addressed in 47438b76ae7c3b24ee97d85478bc01325d888bb6

osv/models.py
andrewpollock: Addressed in 47438b76ae7c3b24ee97d85478bc01325d888bb6

osv/models.py
andrewpollock: Ack. Addressed in 47438b76ae7c3b24ee97d85478bc01325d888bb6

osv/models.py
andrewpollock: Addressed in 47438b76ae7c3b24ee97d85478bc01325d888bb6

osv/models.py
andrewpollock: Ah, good catch, thanks. Addressed in 47438b76ae7c3b24ee97d85478bc01325d888bb6

oliverchang (commented):
Nice! Looking good. We just need that test now :)

andrewpollock:
By George, I do believe I have tests. PTAL

oliverchang:
Very nice! I think you just need to format your code per https://github.com/google/osv.dev/runs/7782343232?check_suite_focus=true.

(Ignore the lint errors -- this wasn't blocking our check and we need to fix them per https://github.com/google/osv.dev/issues/521)

andrewpollock:
My bad, I knew I'd introduced changes that needed that and forgot to run it. PTAL.

oliverchang (approved):
very nice!! LGTM

PR #585, opened 2022-08-09
Add instructions for running UI and API servers locally.

oliverchang:
Also:
- Add instructions to CONTRIBUTING.md for how to invoke the targets.
- Make the API test_server.py a bit more robust by killing existing
  ESPv2 instances prior to starting.
- Rename the ESPv2 docker containers to something less generic to avoid potential
  collisions.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -89,3 +89,25 @@ To format your code, run
 ```shell
 $ yapf -i <file>.py
 ```
+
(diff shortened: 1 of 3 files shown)


another-rex (approved):

another-rex:
Seems like the docker name change broke some tests using the old esp name

oliverchang:
> Seems like the docker name change broke some tests using the old esp name

Yep, fixing. Thanks!

PR #586, opened 2022-08-10
Repo-tracking initial functionality

susperius:
This PR adds initial functionality to read and hash repositories. 

1) the configs are read
2) the repositories are cloned and/or updated to retrieve branches, tags and commit
3) the different branches, tags and commits are checked out and the files are hashed.


docker/indexer/config/config.go
```diff
@@ -0,0 +1,62 @@
...
+	for {
+		attrs, err := iter.Next()
+		if err == iterator.Done {
+			break
+		}
+		if err != nil {
+			return nil, err
+		}
+
+		if strings.HasSuffix(attrs.Name, "textproto") {
```
oliverchang, line 32: nit: invert condition and continue instead to avoid unnecessary indentation here for the rest of the code here.

docker/indexer/config/config.go
```diff
@@ -0,0 +1,62 @@
...
+			obj := cfgBucket.Object(attrs.Name)
+			r, err := obj.NewReader(ctx)
+			if err != nil {
+				log.Errorf("failed to receive object %s: %v", err, attrs.Name)
```
oliverchang, line 36: Should we be just returning propagating these errors back to the caller? 

It seems dangerous to continue if most of these fail.

docker/indexer/stages/preparation/preparation.go
```diff
@@ -0,0 +1,294 @@
...
+	wg.Wait()
+	wCancel()
+	return err
+}
+
+func (s *Stage) objectExists(ctx context.Context, name string) bool {
+	objItr := s.RepoHdl.Objects(ctx, &storage.Query{Prefix: name + shared.TarExt})
+	_, err := objItr.Next()
+	return err == nil
+
```
oliverchang, line 107: unnecessary newline.

docker/indexer/stages/processing/processing.go
```diff
@@ -0,0 +1,132 @@
...
+
+// Storer is used to permanently store the results.
+type Storer interface {
+	Store(ctx context.Context, repoInfo *preparation.Result, fileResults []FileResult) error
+}
+
+// FileResult holds the per file hash and path information.
+type FileResult struct {
+	Path string
+	Hash [md5.Size]byte
```
oliverchang, line 33: Why md5 instead of say sha1? Does it make a huge difference in performance here?

docker/indexer/stages/preparation/preparation.go
```diff
@@ -0,0 +1,294 @@
...
+)
+
+const workers = 5
+
+// Result is the data structure returned by the stage.
+type Result struct {
+	Name            string
+	BaseCPE         string
+	VersionRE       *regexp.Regexp
+	TagMatching     bool
```
oliverchang, line 33: When would this ever be not true?

docker/indexer/stages/preparation/preparation.go
```diff
@@ -0,0 +1,294 @@
...
+	BranchMatching  bool
```
oliverchang, line 34: same here.

docker/indexer/proto/config.proto
```diff
@@ -0,0 +1,24 @@
...
+
+message Repository {
+    // the origin address
+    string address = 1;
+    string name = 2;
+    RepositoryType type = 3;
+    string base_cpe = 4;
+    // the regex is used to extract the version number from the branch/tag name
+    string version_regex = 5;
+    bool tag_versioning = 6;
```
oliverchang, line 20: Do we ever expect `tag_versioning` and `branch_versioning` to be false?

docker/indexer/stages/preparation/preparation.go
```diff
@@ -0,0 +1,294 @@
...
+		results <- &Result{
+			Name:        repoCfg.Name,
+			BaseCPE:     repoCfg.BaseCpe,
+			VersionRE:   versRE,
+			TagMatching: true,
+			CheckoutOptions: &git.CheckoutOptions{
+				Branch: ref.Name(),
+				Force:  true,
+			},
+			When:     when,
```
oliverchang, line 171: What will this field be used for?

docker/indexer/stages/preparation/preparation.go
```diff
@@ -0,0 +1,294 @@
...
+			return err
+		}
+		results <- &Result{
+			Name:        repoCfg.Name,
+			BaseCPE:     repoCfg.BaseCpe,
+			VersionRE:   versRE,
+			TagMatching: true,
+			CheckoutOptions: &git.CheckoutOptions{
+				Branch: ref.Name(),
+				Force:  true,
```
oliverchang, line 169: What will this field be used for?

docker/indexer/proto/example.textproto
```diff
@@ -0,0 +1,15 @@
+# proto-message: Repository
+
+address: "https://github.com/protocolbuffers/protobuf.git"
+name: "protobuf"
+type: GIT
+base_cpe: "cpe:2.3:a:google:protobuf:"
+version_regex: "v({\\d}+\\.{\\d}+)"
```
oliverchang, line 7: Would it work at all to just have more generic logic to try to pull these out? e.g. we can detect runs of digits and extract them. 

We have some similar logic implemented in Python here: https://github.com/google/osv.dev/blob/e84e63b9b8ae1b89992ec5ca750a4d2ae6175459/osv/bug.py#L31

There might be inconsistent formatting even within the same repo, and a more general approach might help and reduce manual effort when we want to scale this up.

docker/indexer/indexer.go
```diff
@@ -0,0 +1,73 @@
...
+	procChan := make(chan *preparation.Result, 10)
+	prepGrp := errgroup.Group{}
+	prepWg := sync.WaitGroup{}
+	prepGrp.SetLimit(2)
+	prepStage := preparation.Stage{
+		RepoHdl: repoBucketHdl,
+		Checker: &idxStorage.Store{},
+	}
+
+	prepWg.Add(1)
```
oliverchang, line 52: Could you please add some general high level comments here to explain the flow/pipeline?

i.e. the preparation stage clones/updates repos, caches them, and prepares the list of tags/branches to analyze.

Then each processing stage/worker gets a single ref to analyze/index. 

(Assuming I got that right?)

docker/indexer/stages/preparation/preparation.go
```diff
@@ -0,0 +1,294 @@
...
+// Run runs the stage and outputs Result data types to the results channel. 
+func (s *Stage) Run(ctx context.Context, cfgs []*pb.Repository, results chan *Result) error {
+	var err error
+	wErr := make(chan error, workers)
+	wg := sync.WaitGroup{}
+	routineCtr := 0
+	wCtx, wCancel := context.WithCancel(ctx)
+	for _, repoCfg := range cfgs {
+		if routineCtr >= workers {
+			wg.Wait()
```
oliverchang, line 62: Can we just use a semaphore here (i.e. https://pkg.go.dev/golang.org/x/sync/semaphore) ?

docker/indexer/stages/processing/processing.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+// Run runs the stages and hashes all files for each incoming request.
+func (s *Stage) Run(ctx context.Context, input chan *preparation.Result) error {
+	wErr := make(chan error, workers)
+	routineCtr := 0
+	wg := sync.WaitGroup{}
+	for {
+		if routineCtr >= workers {
+			wg.Wait()
```
oliverchang, line 49: Same here: can we use a semaphore?

docker/indexer/stages/processing/processing.go
oliverchang, line 33: Maybe we should support multiple hashes here. 

E.g. if we want to extend this to support multiple hashing methods (e.g. one where we "normalize" the input file by formatting/removing comments), we can just add it here without changing any underlying data structures.

oliverchang (commented):
amazing!!!! some comments.

docker/indexer/stages/processing/processing.go
susperius: With my initial test md5 was faster to compute and with the amount of hashes that we need to compute it seemed like the right decision to use the faster one.

docker/indexer/stages/preparation/preparation.go
susperius: It can be used to place an emphasis on how out dated a dependency version is.

docker/indexer/stages/preparation/preparation.go
susperius: This is used in processing to check out the corresponding branch/commit.

docker/indexer/stages/preparation/preparation.go
susperius: Not sure I'm following you on this. Some repos do tag based versioning others based on branch names and then some only use commit hashes.

We could also decide to just hash everything and do a best effort detection of the version as you mentioned in a later comment. Thus eliminating the need for TagMatching, BranchMatching and HashAllCommits.

docker/indexer/stages/preparation/preparation.go
susperius: please see comment above

docker/indexer/stages/processing/processing.go
susperius: changed `Hash` to a slice and added a field to define the hash type

docker/indexer/proto/example.textproto
susperius: I added the generic version regex, which is used if the specific one fails.

docker/indexer/stages/preparation/preparation.go
oliverchang: Sorry for being unclear. I think TagMatching would be the most common, and it seems very unlikely this would ever need to be true. Even if tags aren't used, by either using the version regex specified or doing the best effort detection we can skip tags that we can't get extract version information from. 

This can be extended to branch matching too, but I'm unsure we can always just do HashAllCommits because repos can get very big and that would get slow very quickly.

docker/indexer/stages/preparation/preparation.go
```diff
@@ -0,0 +1,308 @@
...
+		}
+		if found {
+			return nil
+		}
+
+		var when time.Time
+		if c, ok := allCommits[ref.Hash()]; ok {
+			when = c.Author.When
+		}
+		versRE, err := regexp.Compile(repoCfg.VersionRE)
```
oliverchang, line 169: Can we make VersionRE optional? As this an unspecified VersionRE will result in an error right?

oliverchang (commented):
thanks! This LG with some final comments.

susperius:
* removed tag versioning option and made tag versioning the default
* made version regex optional
* moved force enabling to the processing stage

docker/indexer/stages/preparation/preparation.go
```diff
@@ -0,0 +1,313 @@
...
+		if repoCfg.VersionRE != "" {
+			versRE, err := regexp.Compile(repoCfg.VersionRE)
+			if err != nil {
+				return err
+			}
+			version = versRE.FindString(ref.Name().String())
+		}
+		if version == "" {
+			version = genericVersionRE.FindString(ref.Name().String())
+		}
```
oliverchang, line 180: If `version` doesn't match any of the provided regexes, should we just bail instead of sending to the channel?

docker/indexer/config/example.yaml
```diff
@@ -0,0 +1,14 @@
+address: "https://github.com/protocolbuffers/protobuf.git"
+name: "protobuf"
+type: "GIT"
+base_cpe: "cpe:2.3:a:google:protobuf:"
+version_regex: "v({\\d}+\\.{\\d}+)"
+tag_versioning: true
```
oliverchang, line 6: Is this still required?

oliverchang (approved):
LGTM with some remaining comments -- can be addressed in a subsequent PR. I'll merge this to reduce the back and forth and make it easier to work on subsequent PRs.

Thanks again!

docker/indexer/stages/preparation/preparation.go
susperius: Yes, I think that makes sense.

docker/indexer/config/example.yaml
susperius: No. Thanks!

PR #589, opened 2022-08-11
Osv scanning tooling

another-rex:
A start on the osv scanning tooling. This includes:

- Fix for purl matching for debian packages
- Debian docker image scanner in Go
- Utilizes osv-detector lockfile package to scan many lockfiles
- Fix/Update to allow querying of over 1000 packages by chunking the queries


gcp/api/server.py
```diff
@@ -197,6 +197,24 @@ def query_by_commit(commit, to_response=bug_to_response):
   return _get_bugs(bug_ids, to_response=to_response)
 
 
+def _match_purl(purl_query, purl_db) -> bool:
```
andrewpollock, line 200: I'm not entirely clear on purls and what you're trying to accomplish here, but I was wondering if utilising the likes of https://urllib3.readthedocs.io/en/stable/reference/urllib3.util.html#urllib3.util.parse_url might be a more robust option?

tools/osv-scanner/internal/lockfiles/cargo.go
```diff
@@ -0,0 +1,34 @@
+package lockfiles
+
+import (
+	"log"
+	"os"
+)
+import "github.com/BurntSushi/toml"
```
oliverchang, line 7: As discussed, let's reuse https://github.com/G-Rath/osv-detector/tree/main/pkg/lockfile

gcp/api/server.py
oliverchang: That's a good point about making the parsing more robust, but we should use the offiicial package URL module here if we can (which we already import in this file in https://github.com/google/osv.dev/blob/3f8b4067bb83dfe9e872a8b7bae584f9e8821f12/gcp/api/server.py#L26)

tools/osv-scanner/cmd/scanner.go
```diff
@@ -23,6 +25,9 @@ func usage() {
 func scan(query *osv.BatchedQuery, arg string) error {
 	info, err := os.Stat(arg)
 	if err != nil {
+		// Assume it's a docker image if file can't be found
+		// TODO: Have actual commands to differentiate these functions
```
oliverchang, line 29: Let's just add a new flag for this. e.g. `--docker=image/name`

tools/query_converter.py
```diff
@@ -0,0 +1,57 @@
+import argparse
```
oliverchang, line 1: If you already implemented this in Go, no need to include this :)

oliverchang (commented):
nice! some initial comments

tools/osv-scanner/cmd/scanner.go
```diff
@@ -119,18 +155,78 @@ func printResults(query osv.BatchedQuery, resp *osv.BatchedResponse) {
...
-		if err := scan(&query, arg); err != nil {
-			log.Printf("scan failed: %v\n", err)
-			return
-		}
+	app := &cli.App{
+		Name:  "osv-scanner",
+		Usage: "scans various mediums for dependencies and matches it against the OSV database",
+		Commands: []*cli.Command{
+			{
+				Name:  "docker",
```
oliverchang, line 166: I'm assuming that these let us do:

```bash
$ osv-scanner docker ...
$ osv-scanner lockfile ...
```
etc

Maybe instead we build a more flexible and extensible interface like so:

```bash
$ osv-scanner scan  -docker docker/image -docker docker/image2 -lockfile blah -sbom blah.sbom  <dir/path>
````

That is, we can pass multiple types in a single command. Paths can be passed with either an explicit type (e.g. "-sbom"), or where a type **isn't** supplied, we try to do our automated directory walking/scanning.

Comment (another-rex) [reply tools/osv-scanner/cmd/scanner.go:166]: Ah good idea, I'll change it to this now

File: gcp/api/server.py (hunk 1 above)
Comment (another-rex) [reply gcp/api/server.py]: [f1b4443](https://github.com/google/osv.dev/pull/589/commits/f1b4443d240bfff5f1463ba03fbbd5d0b4972bba) Now fixes this by using PackageURL for the purl variable being passed around.

File: gcp/api/server.py [hunk 6]
```diff
@@ -176,12 +177,20 @@ def _get_bugs(bug_ids, to_response=bug_to_response):
 
 
 def _clean_purl(purl):
-  """Clean a purl object."""
+  """
+  Clean a purl object.
+
+  Removes version, subpath, and qualifiers with the exeption of
```
oliverchang, line 183: nit: spelling of "exception"

gcp/api/server.py
```diff
@@ -198,8 +207,22 @@ def query_by_commit(commit, to_response=bug_to_response):
...
-def _is_semver_affected(affected_packages, package_name, ecosystem, purl,
-                        version):
+def _match_purl(purl_query: PackageURL, purl_db: PackageURL) -> bool:
+  """Check if purl match at the specifity level of purl_query
+
+  If purl_query doesn't have qualifiers, then we will match against purl_db
+  without qualifiers, otherwise match with qualifiers
+  """
+
+  if len(purl_query.qualifiers) == 0:
```
oliverchang, line 217: nit: just do

```python
if not purl_query.qualifiers
```

See https://google.github.io/styleguide/pyguide.html#214-truefalse-evaluations

gcp/api/server.py
```diff
@@ -254,7 +278,8 @@ def _is_version_affected(affected_packages,
               affected_package.package.ecosystem) != ecosystem):
         continue
 
-    if purl and purl != affected_package.package.purl:
+    if purl and not (affected_package.package.purl and _match_purl(
```
oliverchang, line 281: nit: can we just push the logic of handling an empty purl into `_match_purl`? This makes this logic a bit easier to comprehend.

tools/osv-scanner/cmd/scanner.go
```diff
@@ -103,6 +103,44 @@ func scanGit(repoDir string) (*osv.Query, error) {
...
+		text := scanner.Text()
+		text = strings.TrimSpace(text)
+		if len(text) == 0 {
+			continue
+		}
+		splitText := strings.Split(text, "###")
+		pkgDetails := osv.MakePkgDetailsRequest(lockfile.PackageDetails{
+			Name:    splitText[0],
+			Version: splitText[1],
+			// TODO(rexpan): Get and specify exact debian release version
```
oliverchang, line 136: This is not strictly necessary right?

tools/osv-scanner/cmd/scanner.go
```diff
@@ -119,18 +157,80 @@ func printResults(query osv.BatchedQuery, resp *osv.BatchedResponse) {
...
+			for _, lockfile := range lockfiles {
+				err := scanLockfile(&query, lockfile)
+				if err != nil {
+					return err
+				}
+			}
+
+			sboms := context.StringSlice("sbom")
+			for _, sbom := range sboms {
+				err := scanSbomFile(&query, sbom)
```
oliverchang, line 211: nit: in Go we typically name acronyms like SBOMs in all caps.

See https://github.com/golang/go/wiki/CodeReviewComments#initialisms

tools/osv-scanner/cmd/scanner.go
```diff
@@ -103,6 +103,44 @@ func scanGit(repoDir string) (*osv.Query, error) {
 	return osv.MakeCommitRequest(commit), nil
 }
 
+type DockerPackageVersion struct {
```
oliverchang, line 106: Can you add some comments here? See the "Types" section on https://go.dev/doc/comment

tools/osv-scanner/cmd/scanner.go
```diff
@@ -103,6 +103,44 @@ func scanGit(repoDir string) (*osv.Query, error) {
...
+	for scanner.Scan() {
+		text := scanner.Text()
+		text = strings.TrimSpace(text)
+		if len(text) == 0 {
+			continue
+		}
+		splitText := strings.Split(text, "###")
+		pkgDetails := osv.MakePkgDetailsRequest(lockfile.PackageDetails{
+			Name:    splitText[0],
+			Version: splitText[1],
```
oliverchang, line 135: Can we check for the length of `splitText` here and report the error accordingly?

tools/osv-scanner/internal/osv/osv.go
```diff
@@ -57,25 +64,60 @@ func MakePURLRequest(purl string) *Query {
...
+	for chunkSize < len(items) {
+		items, _chunks = items[chunkSize:], append(_chunks, items[0:chunkSize:chunkSize])
 	}
-	defer resp.Body.Close()
+	return append(_chunks, items)
+}
+
+func MakeRequest(request BatchedQuery) (*BatchedResponse, error) {
+	// API has a limit of 1000 bulk query per request
+	queryChunks := chunkBy(request.Queries, 990)
```
oliverchang, line 89: nit: can you put 990 in a global const, and explain why this isn't 1000?

oliverchang (commented):
amazing work!!! sorry for the slow review here.

oliverchang:
Can you also update the homepage docs on https://github.com/google/osv.dev/blob/55d70d05bc41ef5063bfc8c9e8fc61afa27091ef/gcp/appengine/frontend3/src/templates/home.html#L167 ? 

Also the README in https://github.com/google/osv.dev/tree/master/tools/osv-scanner.

Probably wait for @andrewpollock to first rename the module properly.

andrewpollock:
> Probably wait for @andrewpollock to first rename the module properly.

FYI #606 is in flight

tools/osv-scanner/internal/osv/osv.go
another-rex: Well it probably can be 1000, just wanted to avoid potential edge cases so used 990. Changed it to 1000 in the constant now.

tools/osv-scanner/cmd/scanner.go
another-rex: Probably not, but having it might avoid some false positives. Can't really think of a concrete example though.

gcp/api/server.py
oliverchang: Ah I just realised I skimmed this too quickly and the types here don't match. Can skip this comment.

tools/osv-scanner/main.go
```diff
@@ -119,18 +161,80 @@ func printResults(query osv.BatchedQuery, resp *osv.BatchedResponse) {
...
+				TakesFile: true,
+			},
+			&cli.StringSliceFlag{
+				Name:      "git",
+				Aliases:   []string{"G"},
+				Usage:     "scan for git repository in this directory",
+				TakesFile: true,
+			},
+		},
+		ArgsUsage: "[directory1 directory2...]",
```
oliverchang, line 196: Does this refer to arbitrary directories being passed that don't come with a prefixing type flag? 

How is this handled right now?

tools/osv-scanner/main.go
```diff
@@ -1,38 +1,22 @@
...
-
-	if info.IsDir() {
-		return scanDir(query, arg)
-	}
-
-	return scanFile(query, arg)
-}
-
 func scanDir(query *osv.BatchedQuery, dir string) error {
 	log.Printf("Scanning dir %s\n", dir)
```
oliverchang, line 21: Can you add a todo here to recursively detect and scan all files too, not just nested git repos?

tools/osv-scanner/main.go
```diff
@@ -119,18 +161,80 @@ func printResults(query osv.BatchedQuery, resp *osv.BatchedResponse) {
...
+
+			sboms := context.StringSlice("sbom")
+			for _, sbom := range sboms {
+				err := scanSBOMFile(&query, sbom)
+				if err != nil {
+					return err
+				}
+			}
+
+			gitDirs := context.StringSlice("git")
```
oliverchang, line 221: should this just be handled as part of an untyped directory scanning? (per the "[directory1 directory2...]" part above).

oliverchang:
(and please also make that homepage/README change to add the -sbom flag to the relevant examples).

tools/osv-scanner/main.go
another-rex: Yes, but it turns out I forgot to implement it. Added now in 030f8da

tools/osv-scanner/main.go
another-rex: Now scans all files in 030f8da

tools/osv-scanner/internal/osv/osv.go
```diff
@@ -57,25 +67,61 @@ func MakePURLRequest(purl string) *Query {
 	}
 }
 
-func MakeRequest(request BatchedQuery) (*BatchedResponse, error) {
-	requestBytes, err := json.Marshal(request)
-	if err != nil {
-		return nil, err
+func MakePkgDetailsRequest(pkgDetails lockfile.PackageDetails) *Query {
```
oliverchang, line 70: nit: MakePkgRequest ? 

"Details" part seems a bit confusing.

tools/osv-scanner/README.md
```diff
@@ -20,19 +20,56 @@ auto-detected based on the input file contents.
...
+
+```bash
+$ go run main.go --lockfile=/path/to/your/package-lock.json -L /path/to/another/Cargo.lock
+```
+
+## Scanning a Debian based docker image packages
+
+This tool will scrape the list of installed packages in a debian image and query for vulnerabilities on them.
+
+Currently only debian based docker image scanning is supported.
```
oliverchang, line 49: s/debian/Debian/

oliverchang (approved):
very nice! can you also edit the home page to add the -sbom flag to the existing SBOM example, and add a lockfile example (say using npm) ?

oliverchang (approved):

PR #599, opened 2022-08-15
Fix lint findings

andrewpollock:
This fixes the vast majority of the outstanding lint findings. 

I need to discuss/get some assistance with the outstanding ones.

```
************* Module importer
docker/importer/importer.py:213:4: R1702: Too many nested blocks (6/5) (too-many-nested-blocks)
************* Module osv.cache
osv/cache.py:87:14: W1505: Using deprecated method getcallargs() (deprecated-method)
osv/cache.py:70:0: C0103: Function name "Cached" doesn't conform to the `snake_case` group in the '^(?:(?P<exempt>setUp|tearDown|setUpModule|tearDownModule)|(?P<camel_case>_?[A-Z][a-zA-Z0-9]*)|(?P<snake_case>_?[a-z][a-z0-9_]*))$' pattern (invalid-name)

------------------------------------------------------------------
Your code has been rated at 9.99/10 (previous run: 9.99/10, +0.00)
```

### DIFF
--- a/docker/importer/importer.py
+++ b/docker/importer/importer.py
@@ -259,8 +259,7 @@ class Importer:
         continue
 
       try:
-        vulnerability = osv.parse_vulnerability(
-            path, key_path=source_repo.key_path)
+        _ = osv.parse_vulnerability(path, key_path=source_repo.key_path)
       except Exception as e:
         logging.error('Failed to parse %s: %s', changed_entry, str(e))
         continue
--- a/docker/importer/importer.py
+++ b/docker/importer/importer.py
@@ -210,8 +210,10 @@ class Importer:
     changed_entries = set()
     deleted_entries = set()
 
-    if source_repo.last_synced_hash:
-      # Syncing from a previous commit.
+    def _sync_from_previous_commit(repo):
+      entries_changed = set()
+      entries_deleted = set()
+
       walker = repo.walk(repo.head.target, pygit2.GIT_SORT_TOPOLOGICAL)
       walker.hide(source_repo.last_synced_hash)
 
@@ -233,14 +235,21 @@ class Importer:
             if delta.old_file and _is_vulnerability_file(
                 source_repo, delta.old_file.path):
               if delta.status == pygit2.GIT_DELTA_DELETED:
-                deleted_entries.add(delta.old_file.path)
+                entries_deleted.add(delta.old_file.path)
                 continue
 
-              changed_entries.add(delta.old_file.path)
+              entries_changed.add(delta.old_file.path)
 
             if delta.new_file and _is_vulnerability_file(
                 source_repo, delta.new_file.path):
-              changed_entries.add(delta.new_file.path)
+              entries_changed.add(delta.new_file.path)
+
+      return entries_changed, entries_deleted
+
+    if source_repo.last_synced_hash:
+      # Syncing from a previous commit.
+      changed_entries, deleted_entries = _sync_from_previous_commit(repo)
+
     else:
       # First sync from scratch.
       logging.info('Syncing repo from scratch')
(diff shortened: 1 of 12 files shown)


oliverchang:
Nice!! Can you also fix the lint action here: https://github.com/google/osv.dev/blob/master/.github/workflows/lint.yaml ? This just uses the referenced bash script, which may not be re-surfacing all the necessary errors. 

For `docker/importer/importer.py`, can you refactor some of the nested code into smaller helper functions? 

For 

>osv/cache.py:87:14: W1505: Using deprecated method getcallargs() (deprecated-method)

Are we able to use the recommended replacements on https://docs.python.org/3/library/inspect.html#inspect.getcallargs ? 

> osv/cache.py:70:0: C0103: Function name "Cached" doesn't conform to the `snake_case` group in the '^(?:(?P<exempt>setUp|tearDown|setUpModule|tearDownModule)|(?P<camel_case>_?[A-Z][a-zA-Z0-9]*)|(?P<snake_case>_?[a-z][a-z0-9_]*))$' pattern (invalid-name) 

Let's just rename this one. I think we just missed this through code review previously.

andrewpollock:
> Nice!! Can you also fix the lint action here: https://github.com/google/osv.dev/blob/master/.github/workflows/lint.yaml ? This just uses the referenced bash script, which may not be re-surfacing all the necessary errors.

I think you're referring to what I've done in https://github.com/google/osv.dev/issues/521 so the answer is yes, in a separate PR. The fix is going to start causing hard fails on legacy lint errors, hence this PR.

> For docker/importer/importer.py, can you refactor some of the nested code into smaller helper functions?

Yep, that was slightly higher hanging fruit than the rest.

> For 
>
> osv/cache.py:87:14: W1505: Using deprecated method getcallargs() (deprecated-method)
>
> Are we able to use the recommended replacements on https://docs.python.org/3/library/inspect.html#inspect.getcallargs ?

I'll need a bit of help, this function is doing things beyond my immediate skill level.

> osv/cache.py:70:0: C0103: Function name "Cached" doesn't conform to the snake_case group in the '^(?:(?PsetUp|tearDown|setUpModule|tearDownModule)|(?P<camel_case>?[A-Z][a-zA-Z0-9]*)|(?P<snake_case>?[a-z][a-z0-9_]*))$' pattern (invalid-name)
>
> Let's just rename this one. I think we just missed this through code review previously.

Ack, done.

another-rex:
> > For
> > osv/cache.py:87:14: W1505: Using deprecated method getcallargs() (deprecated-method)
> > Are we able to use the recommended replacements on https://docs.python.org/3/library/inspect.html#inspect.getcallargs ?
> 
> I'll need a bit of help, this function is doing things beyond my immediate skill level.

Looking at the example here: https://docs.python.org/3/library/inspect.html#inspect.BoundArguments
I think for this, you'll need to split it out into a few more lines, from:
```
      key = (*getcallargs(func, *args, **kwargs).items(), unique_f_key)
```
to
```
sig = inspect.signature(func)
# Passing through the arguments made to the function
bound_args = sig.bind(*args, **kwargs) 
# Making it hashable and combining it with the function name
key = (*bound_args.arguments.items(), unique_f_key) 
```

andrewpollock:
Thanks so much for your assistance @another-rex 

This is now 100% lint-clean @oliverchang

docker/importer/importer.py
```diff
@@ -210,8 +210,10 @@ def _process_updates_git(self, source_repo):
     changed_entries = set()
     deleted_entries = set()
 
-    if source_repo.last_synced_hash:
-      # Syncing from a previous commit.
+    def _sync_from_previous_commit(repo):
```
oliverchang, line 213: Can we pull this out into a top level method? I **think** we just need to pass `source_repo` here as an extra arg?

oliverchang (approved):
thank you so much for cleaning this up! Just one last comment.

docker/importer/importer.py
andrewpollock: Done in bfc7965

oliverchang (approved):

PR #600, opened 2022-08-15
Fix lint_and_format action so lint findings are not ignored

andrewpollock:
Fixes: #521

### DIFF
--- a/tools/lint_and_format.sh
+++ b/tools/lint_and_format.sh
@@ -16,17 +16,17 @@
 script_dir=$(dirname "$0")
 IN_SCOPE_FILES="$(git ls-files | grep '\.py$' | grep -v -E '(_pb2|third_party)')"
 
-lint_findings=0
+lint_findings=""
 if ! echo "$IN_SCOPE_FILES" | xargs pylint --rcfile="$script_dir/../.pylintrc"; then
-  lint_findings=1
+  lint_findings="lint_findings"
 fi
 
-format_findings=0
+format_findings=""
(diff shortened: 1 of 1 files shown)


oliverchang (approved):
thanks!

tools/lint_and_format.sh
```diff
@@ -14,5 +14,19 @@
 # limitations under the License.
 
 script_dir=$(dirname "$0")
-git ls-files | grep '\.py$' | grep -v '_pb2' | grep -v 'third_party' | xargs pylint --rcfile="$script_dir/../.pylintrc"
-git ls-files | grep '\.py$' | grep -v '_pb2' | grep -v 'third_party' | xargs yapf -d --style "$script_dir/../.style.yapf"
+IN_SCOPE_FILES="$(git ls-files | grep '\.py$' | grep -v -E '(_pb2|third_party)')"
+
+lint_findings=""
+if ! echo "$IN_SCOPE_FILES" | xargs pylint --rcfile="$script_dir/../.pylintrc"; then
+  lint_findings="lint_findings"
```
oliverchang, line 21: nit: this seems a bit confusing to read at first wrt intention. I thought this was reading the same variable at first.

Maybe just stick with the number here, and use `[ $lint_findings -ne 0 ]` ?

andrewpollock: I've done this so it's a little easier to comprehend when running under `sh -x`

You'll see the string "lint_findings" and/or "format_findings" so it's obvious with check has failed here.

oliverchang: ack, that's a great point. Let's keep this then!

oliverchang (approved):

PR #601, opened 2022-08-15
Repo-Indexer: Adding functionality to store results in Datastore

susperius:
This PR adds the functionality to store the repo hashing results as documents in datastore.


docker/indexer/storage/storage.go
```diff
@@ -18,22 +18,82 @@ package storage
...
+// document represents a single repository entry in datastore.
+type document struct {
+	Name        string
+	BaseCPE     string
+	Version     string
+	Commit      []byte
+	When        time.Time
+	RepoType    string
+	FileExts    []string
+	FileResults []processing.FileResult
```
oliverchang, line 45: Thinking more on extension: We should pull out the hash type from the `FileResult` and put it as a property here instead. 

This saves a few redundant bytes per hash entry, and makes it easier to add more hashing algorithms, as there is an entity size limit of 1MiB that we may hit if we shove all of them in the same entity. 

Then we should key these documents by repo + hash type + git ref.

docker/indexer/storage/storage.go
```diff
@@ -18,22 +18,82 @@ package storage
...
+	if err != nil {
+		return nil, err
+	}
+	return &Store{dsCl: client, cache: sync.Map{}}, nil
 }
 
+// Exists checks whether a name/hash pair already exists in datastore.
 func (s *Store) Exists(ctx context.Context, name string, hash plumbing.Hash) (bool, error) {
-	return false, nil
+	if _, ok := s.cache.Load(fmt.Sprintf(keyFmt, name, hash)); ok {
```
oliverchang, line 79: Can we key based on the repo URL instead of name? 

We want to eventually run this on all git repos for C/C++ projects that we see in OSV.

oliverchang (commented):
nice, thanks!! some comments.

docker/indexer/indexer.go
```diff
@@ -1,17 +1,17 @@
 /*
 Copyright 2022 Google LLC
 
- Licensed under the Apache License, Version 2.0 (the "License");
- you may not use this file except in compliance with the License.
- You may obtain a copy of the License at
+	 Licensed under the Apache License, Version 2.0 (the "License");
```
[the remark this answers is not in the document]
susperius: Yes, thanks. Auto-formatting ...

docker/indexer/storage/storage.go
```diff
@@ -18,22 +18,82 @@ package storage
...
 
-	log "github.com/golang/glog"
+const (
+	kind   = "source-repository"
+	keyFmt = "%s-%x"
 )
 
+// document represents a single repository entry in datastore.
+type document struct {
+	Name        string
```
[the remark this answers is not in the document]
susperius: Your assumption is correct. Added tags.

docker/indexer/storage/storage.go
```diff
@@ -18,22 +18,86 @@ package storage
...
 
+	"cloud.google.com/go/datastore"
 	"github.com/go-git/go-git/v5/plumbing"
 	"github.com/google/osv.dev/docker/indexer/stages/preparation"
 	"github.com/google/osv.dev/docker/indexer/stages/processing"
+)
 
-	log "github.com/golang/glog"
+const (
+	kind   = "source-repository"
```
oliverchang, line 32: Can you name this RepoIndex or something similar. We name all of our entities in that casing as well.

oliverchang: This is also a bit too similar to our existing SourceRepository entity.

oliverchang (approved):
Just one last comment. Otherwise this LGTM!

oliverchang (approved):

PR #604, opened 2022-08-17
Fix lint findings

andrewpollock:
#599 omitted a few lint finding fixes because I was running an older pylint from outside of pipenv

### DIFF
--- a/gcp/api/server.py
+++ b/gcp/api/server.py
@@ -132,7 +132,9 @@ def do_query(query, context, include_details=True):
       context.abort(grpc.StatusCode.INVALID_ARGUMENT, 'Invalid Package URL.')
       return None
 
-  to_response = lambda b: bug_to_response(b, include_details)  # pylint: disable=unnecessary-lambda-assignment
+  def to_response(b):
+    return bug_to_response(b, include_details)
+
   next_page_token = None
 
   if query.WhichOneof('param') == 'commit':
--- a/gcp/api/server.py
+++ b/gcp/api/server.py
@@ -132,7 +132,7 @@ def do_query(query, context, include_details=True):
       context.abort(grpc.StatusCode.INVALID_ARGUMENT, 'Invalid Package URL.')
       return None
 
-  to_response = lambda b: bug_to_response(b, include_details)
+  to_response = lambda b: bug_to_response(b, include_details)  # pylint: disable=unnecessary-lambda-assignment
   next_page_token = None
 
   if query.WhichOneof('param') == 'commit':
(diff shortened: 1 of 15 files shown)


gcp/api/server.py
```diff
@@ -131,7 +132,7 @@ def do_query(query, context, include_details=True):
       context.abort(grpc.StatusCode.INVALID_ARGUMENT, 'Invalid Package URL.')
       return None
 
-  to_response = lambda b: bug_to_response(b, include_details)
+  to_response = lambda b: bug_to_response(b, include_details)  # pylint: disable=unnecessary-lambda-assignment
```
another-rex, line 135: I just figured out what the lint here is trying to say. I think it wants you to do this:
```
...
      context.abort(grpc.StatusCode.INVALID_ARGUMENT, 'Invalid Package URL.')
      return None

      def to_response(b):
          return bug_to_response(b, include_details)
      
      next_page_token = None
...

```

This way it prevents people from accidentally modifying the `to_response` variable when it shouldn't be modified

oliverchang, line 135: Let's just add this to https://github.com/google/osv.dev/blob/f660fa6a7dc02337f1944cd276181a3bf4abd0e1/.pylintrc#L10

oliverchang (approved):
thank you very much!!

LGTM with just one last comment

gcp/api/server.py
oliverchang: I reckon this is fine. Up to you to make the judgement call if you think there's value to keeping this check.

gcp/api/server.py
```diff
@@ -200,6 +201,7 @@ def query_by_commit(commit, to_response=bug_to_response):
 def _is_semver_affected(affected_packages, package_name, ecosystem, purl,
                         version):
   """Returns whether or not the given version is within an affected SEMVER
+
```
another-rex, line 204: Odd formatting here with the new line. I had this issue before when running `pyformat` to format files, `yapf` seems to not have the same problem for me.

gcp/api/server.py
andrewpollock: Thank you, that makes perfect sense now that you say it :-)

andrewpollock: Ideally I'd like to better understand why it's being done in the first place...

andrewpollock: This is now a redundant conversation thanks to @another-rex and 1ab0afb

another-rex: It's done here because they wanted to add set the `include_details` parameter. If they just passed in bug_to_response, they will also need to add another argument to set whether to `include_details` or not

gcp/api/server.py
andrewpollock: I'll get to the bottom of this at some point, it's only cosmetic...

Issue #603, opened 2022-08-16
Installation of osv-scanner does not work ( ubuntu / osx )


liorkesos:
Hi guys , I'm a newbie to Go so have installed go just to run the scanner.
I've tried both in Linux and OSX and it does not work for me in either environment.
I might be missing something trivial and I would be  glad to send a PR with the proper documentation when I figure out what I'm doing wrong...
I'm pastng the outputs for both environments
Linux - Ubuntu 20.04.4 LTS
```
> snap install go --classic
> go install github.com/google/osv/tools/osv-scanner@latest
go: downloading github.com/google/osv v0.0.14
go: downloading github.com/google/osv/tools/osv-scanner v0.0.0-20220816044059-f660fa6a7dc0
go: github.com/google/osv/tools/osv-scanner@latest: module github.com/google/osv/tools/osv-scanner@latest found (v0.0.0-20220816044059-f660fa6a7dc0), but does not contain package github.com/google/osv/tools/osv-scanner
```

OSX
```
liorkesos@mbp-14 ~ % go install github.com/google/osv/tools/osv-scanner@latest
go: downloading github.com/google/osv v0.0.14
go: downloading github.com/google/osv/tools/osv-scanner v0.0.0-20220816044059-f660fa6a7dc0
go: github.com/google/osv/tools/osv-scanner@latest: module github.com/google/osv/tools/osv-scanner@latest found (v0.0.0-20220816044059-f660fa6a7dc0), but does not contain package github.com/google/osv/tools/osv-scanner
```

andrewpollock:
Thanks for surfacing this!

It looks like https://github.com/google/osv.dev/pull/377 only partially implemented a rename, and in hindsight wasn't ideal. We'll rename this directory back and fix up a few references.

andrewpollock:
I think I fixed this with https://github.com/google/osv.dev/pull/785 and #784 is basically a duplicate of this issue.

PR #605, opened 2022-08-17
Rename package name to reflect directory name change

andrewpollock:
I think https://github.com/google/osv.dev/pull/377 is what's causing
https://github.com/google/osv.dev/issues/603

Fixes: #603


andrewpollock:
Discussed out of band, I'll abandon this PR in favour of one that essentially undoes https://github.com/google/osv.dev/pull/377

PR #610, opened 2022-08-17
Rename directory back to osv-scanner

andrewpollock:
Turns out we want the resultant binary to have "osv" in the name.

Fixes: #603


oliverchang (commented):
I think you need to add the newly named files :)

tools/osv-scanner/cmd/main.go
```diff
@@ -11,8 +11,8 @@ import (
 	"path/filepath"
```
oliverchang, line 11: Rename this to `tools/osv-scanner/main.go` ?

andrewpollock: Isn't this https://pkg.go.dev/path/filepath ?

andrewpollock:
> I think you need to add the newly named files :)

Yeah got a little bit trigger happy and omitted a `git add` post-rename of the directory. Fixed in 
3e83538

tools/osv-scanner/cmd/main.go
oliverchang: I meant the file itself.

andrewpollock: Oh, got it, move it out of the cmd directory.

andrewpollock: Addressed in 49ea940

oliverchang (approved):

PR #615, opened 2022-08-17
Repo-Indexer: add pubsub as work queue

susperius:
This PR adds pubsub as mechanism to distribute work to the workers. It increases scalability by separating the preparation and processing stages into two distinct execution paths.


docker/indexer/indexer.go
```diff
@@ -34,67 +33,70 @@ var (
...
-	go func() {
-		defer prepWg.Done()
-		defer close(procChan)
-
-		err := prepStage.Run(stageCtx, cfgs, procChan)
-		if err != nil {
-			log.Errorf("preparation stage error: %v", err)
-		}
-	}()
+	cfgBucketHdl := gcsClient.Bucket(*configsBucket)
```
oliverchang, line 86: nit: put the rest of these here into `runController()` or something similar.

docker/indexer/indexer.go
```diff
@@ -34,67 +33,70 @@ var (
...
-	defer cancel()
-	procChan := make(chan *preparation.Result, 10)
-	prepGrp := errgroup.Group{}
-	prepWg := sync.WaitGroup{}
-	prepGrp.SetLimit(2)
-	prepStage := preparation.Stage{
-		RepoHdl: repoBucketHdl,
-		Checker: storer,
+	if *worker {
+		procStage := processing.Stage{
```
oliverchang, line 70: nit: can we put this into a `runWorker()` function?

docker/indexer/indexer.go
```diff
@@ -34,67 +33,70 @@ var (
...
 
 	ctx := context.Background()
+
+	psCl, err := pubsub.NewClient(ctx, *projectID)
+	if err != nil {
+		log.Exitf("failed to initialize pubsub client: %v", err)
+	}
+	defer psCl.Close()
+
+	topic := psCl.Topic(*pubsubTopic)
```
oliverchang, line 52: Move this below into the non-worker part since it's not needed by the worker? (ideally into a `runController` function which I suggested below)

docker/indexer/stages/processing/processing.go
```diff
@@ -51,48 +50,32 @@ type FileResult struct {
...
 
 // Run runs the stages and hashes all files for each incoming request.
-func (s *Stage) Run(ctx context.Context, input chan *preparation.Result) error {
-	wErr := make(chan error, workers)
-	sem := semaphore.NewWeighted(workers)
-	for {
-		if err := sem.Acquire(ctx, 1); err != nil {
-			return fmt.Errorf("failed to acquire semaphore: %v", err)
+func (s *Stage) Run(ctx context.Context) error {
+	s.Input.ReceiveSettings.MaxOutstandingMessages = 5
```
oliverchang, line 58: nit: put 5 in a constant.

oliverchang (commented):
very nice!

docker/indexer/stages/processing/processing.go
oliverchang: Also, one thing to note is that in the past I found this was not respected unless `Synchronous` is set to `true`. What ended up happening is one worker would queue up most of the tasks while the other remaining available workers got nothing. 

However, according to the latest docs at https://pkg.go.dev/cloud.google.com/go/pubsub#ReceiveSettings, this seems to be fixed. Still worth double checking though to make sure it works as attended.

susperius: I tried this on a GCE instance and the number of go routines reduced drastically to not setting it. Also with two receiving clients both got work to do.

oliverchang (approved):

PR #616, opened 2022-08-18
Fix dependabot alerts

another-rex:
This PR fixes most of the dependabot alerts by running npm audit fix, and update Flask packages in Pipfile (though not to the latest version (2.2.2) because of some dependency conflict).

Now also updated other dependencies in appengine backend. And added staging deployment configuration

Still runs fine locally.


oliverchang (approved):
Thanks for doing this! 

Let's be a bit careful here in testing this, since there are some major version bumps.

andrewpollock:
I'm not familiar enough with the build system to perform an informed review of this. I'll make some time with you to walk me through the change for my own edification though.

PR #622, opened 2022-08-22
Tooling fixes

another-rex:
Some minor changes to how SBOM scanning works to avoid go panics and segmentation fault when doing a directory walk.

### DIFF
--- a/tools/osv-scanner/main.go
+++ b/tools/osv-scanner/main.go
@@ -37,14 +37,14 @@ func scanDir(query *osv.BatchedQuery, dir string) error {
 
 		if !info.IsDir() {
 			if parser, _ := lockfile.FindParser(path, ""); parser != nil {
-				scanLockfile(query, path)
+				err := scanLockfile(query, path)
 				if err != nil {
 					log.Println("Attempted to scan lockfile but failed: " + path)
 				}
 			}
 			extension := strings.ToLower(filepath.Ext(info.Name()))
-			if extension == ".json" || extension == ".spdx" {
-				scanSBOMFile(query, path)
+			if extension == ".json" || strings.Contains(info.Name(), ".spdx") {
+				_ := scanSBOMFile(query, path)
 			}
 		}
 
--- a/tools/osv-scanner/main.go
+++ b/tools/osv-scanner/main.go
@@ -44,7 +44,7 @@ func scanDir(query *osv.BatchedQuery, dir string) error {
 			}
 			extension := strings.ToLower(filepath.Ext(info.Name()))
 			if extension == ".json" || strings.Contains(info.Name(), ".spdx") {
-				_ := scanSBOMFile(query, path)
+				_ = scanSBOMFile(query, path)
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


tools/osv-scanner/main.go
```diff
@@ -37,12 +37,15 @@ func scanDir(query *osv.BatchedQuery, dir string) error {
...
 			if parser, _ := lockfile.FindParser(path, ""); parser != nil {
-				scanLockfile(query, path)
+				err := scanLockfile(query, path)
 				if err != nil {
 					log.Println("Attempted to scan lockfile but failed: " + path)
 				}
 			}
-			scanSBOMFile(query, path)
+			extension := strings.ToLower(filepath.Ext(info.Name()))
+			if extension == ".json" || strings.Contains(info.Name(), ".spdx") {
```
andrewpollock, line 46: I'm not understanding why this can't be `extension == ".spdx"`

Could you either change it to check the same way as the `.json` or add a comment explaining the need for this?

tools/osv-scanner/main.go
```diff
@@ -37,12 +37,15 @@ func scanDir(query *osv.BatchedQuery, dir string) error {
...
+				_ = scanSBOMFile(query, path)
```
andrewpollock, line 47: What's the purpose of performing this scan if not checking the error?

tools/osv-scanner/main.go
andrewpollock: Based on the comment in `scanSBOMFile` it sounds like checking for the `.json` extension is redundant?

andrewpollock (commented):
I've made some in-line comments (excuse my ignorance)

tools/osv-scanner/main.go
another-rex: > Based on the comment in scanSBOMFile it sounds like checking for the .json extension is redundant?

Ah yes, I think you're right I can remove this check out here.

> I'm not understanding why this can't be extension == ".spdx"

According to this: https://spdx.github.io/spdx-spec/v2.3/conformance/
SPDX files can have multiple different formats:

Format | Extension
-- | --
tag:value | *.spdx
RDF | *.spdx.rdf
JSON | *.spdx.json
XML | *.spdx.xml
YAML | *.spdx.yaml or *.spdx.yml

I'll add a comment on the check with the link.

tools/osv-scanner/main.go
another-rex: If this succeeds then the dependencies are added to the `query` array. If this fails, then this is not an SBOM file, so nothing happens.

tools/osv-scanner/main.go
andrewpollock: Ah, thanks for the explanation.

andrewpollock (approved):
It's been a while since I've done any Go, I think this LGTM.

tools/osv-scanner/main.go
```diff
@@ -37,12 +37,12 @@ func scanDir(query *osv.BatchedQuery, dir string) error {
...
 		if !info.IsDir() {
 			if parser, _ := lockfile.FindParser(path, ""); parser != nil {
-				scanLockfile(query, path)
+				err := scanLockfile(query, path)
 				if err != nil {
 					log.Println("Attempted to scan lockfile but failed: " + path)
 				}
 			}
-			scanSBOMFile(query, path)
+			_ = scanSBOMFile(query, path)
```
oliverchang, line 45: should we be checking for an error here too instead?

oliverchang (approved):

tools/osv-scanner/main.go
another-rex: I don't think so, because if this succeeds then the dependencies are added to the query array. If this fails, then this is not an SBOM file, so nothing happens.

I'll add a comment here explaining why error checking is skipped.

andrewpollock (approved):

PR #630, opened 2022-08-22
Stop enumerating debian end of life advisories

another-rex:
Stop enumerating Debian end of life advisories, as they are quite rare (as of writing only 7 EOL advisory entries, and around 3 packages affected). If enumerated it is likely to cause false positives, since the EOL generally only affects the specific Debian release, not following releases.

### DIFF
--- a/osv/ecosystems.py
+++ b/osv/ecosystems.py
@@ -388,6 +388,10 @@ class Debian(Ecosystem):
       introduced = debian_version_cache.get_first_package_version(
           package, self.debian_release_ver)
 
+    if fixed == "<end-of-life>":
+      # Special case for eol, we do not enumerate
+      versions = []
+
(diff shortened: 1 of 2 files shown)


osv/ecosystems.py
```diff
@@ -388,6 +389,10 @@ def version_is_valid(v):
       introduced = debian_version_cache.get_first_package_version(
           package, self.debian_release_ver)
 
+    if fixed == self._END_OF_LIFE_VER:
+      # Special case for eol, we do not enumerate
+      versions = []
```
oliverchang, line 394: Can we just `return []` here?

oliverchang (approved):

osv/ecosystems.py
another-rex: Good point, will change to that now

PR #638, opened 2022-08-26
Debian edge cases + logging fix for importer

another-rex:
- Currently importer/exporter logs are all categorized as error logs, making actual errors hard to find. This patch should fix this issue.

- Fixes worker error logs so that exceptions are sent to error reporting. 
- Log edge cases for debian imports as warnings
- Redirect git clone stderr to stdout instead to do info logs instead of error logs.

### DIFF
--- a/docker/worker/worker.py
+++ b/docker/worker/worker.py
@@ -177,7 +177,7 @@ class _PubSubLeaserThread(threading.Thread):
           logging.info('Task complete, stopping renewal.')
           break
       except Exception as e:
-        logging.error('Leaser thread failed: %s', str(e))
+        logging.exception('Leaser thread failed: ')
 
 
 def clean_artifacts(oss_fuzz_dir):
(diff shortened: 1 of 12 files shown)


docker/worker/worker.py
```diff
@@ -557,7 +557,7 @@ def _do_process_task(self, subscriber, subscription, ack_id, message,
         _state.source_id = None
         subscriber.acknowledge(subscription=subscription, ack_ids=[ack_id])
     except Exception:
-      logging.error('Unexpected exception while processing task: %s',
+      logging.error('Unexpected exception while processing task:\n%s',
```
oliverchang, line 560: Maybe there's something we can do here where: if we're ever in an exception handling context, automatically attach the exception traceback after a newline.

oliverchang (approved):
thanks a lot!

docker/worker/worker.py
oliverchang: oh wait, I think this is exactly what `logging.exception` is for: https://docs.python.org/3/library/logging.html#logging.Logger.exception

oliverchang: can we change up all of these cases to do that instead?

oliverchang (approved):
LGTM

PR #639, opened 2022-08-26
Repo-Indexer: Change storage model and improvements

susperius:
The storage model now supports pagination of hashing results to avoid running into the 1MB entity boundary. Furthermore, a flag was added to set the number of pubsub receiver go-routines dynamically.


docker/indexer/indexer.go
```diff
@@ -37,6 +37,7 @@ var (
 	worker        = flag.Bool("worker", false, "makes this a worker node reading from pubsub to process the data")
 	pubsubTopic   = flag.String("topic", "", "sets the pubsub topic to publish to or to read from")
 	subName       = flag.String("subscription", "", "sets the pubsub subscription name for workers")
+	subMessages   = flag.Int("messages", 1, "pubsub outstanding messages")
```
oliverchang, line 40: Why do we need this to be configurable here? Does 1 not work well in general?

docker/indexer/storage/storage.go
```diff
@@ -80,26 +112,40 @@ func New(ctx context.Context, projectID string) (*Store, error) {
...
 	return true, nil
 }
 
 // Store stores a new entry in datastore.
-func (s *Store) Store(ctx context.Context, repoInfo *preparation.Result, hashType string, fileResults []processing.FileResult) error {
-	key := datastore.NameKey(kind, fmt.Sprintf(keyFmt, repoInfo.Addr, hashType, repoInfo.Commit[:]), nil)
-	_, err := s.dsCl.Put(ctx, key, newDoc(repoInfo, hashType, fileResults))
-	return err
+func (s *Store) Store(ctx context.Context, repoInfo *preparation.Result, hashType string, fileResults []*processing.FileResult) error {
+	docKey := datastore.NameKey(docKind, fmt.Sprintf(docKeyFmt, repoInfo.Addr, hashType, repoInfo.Commit[:]), nil)
```
oliverchang, line 132: Can we wrap this all into a Transaction ?

docker/indexer/storage/storage.go
```diff
@@ -80,26 +112,40 @@ func New(ctx context.Context, projectID string) (*Store, error) {
...
+	doc, results := newDoc(repoInfo, hashType, fileResults)
+	_, err := s.dsCl.Put(ctx, docKey, doc)
+	if err != nil {
+		return err
+	}
+	for _, r := range results {
```
oliverchang, line 138: nit: add a comment saying stale pages beyond the new page count aren't deleted?

oliverchang (commented):
nice!

docker/indexer/indexer.go
susperius: It avoids recompilation if it needs to be readjusted. I used it for local testing on a VM instead of a K8s cluster. Furthermore, if we see to little CPU utilization on workers it's a just a flag change.

docker/indexer/storage/storage.go
susperius: Done.

docker/indexer/storage/storage.go
susperius: The RunInTransaction rolls back automatically on failure now.

oliverchang: Was more referring to the case where:

in Run 1: We create 5 pages
in Run 2: We somehow only create 4 pages, leaving page 5 dangling. 

Not a big deal in practice, but worth mentioning as a comment.

PR #652, opened 2022-08-31
Alpine/general CVE conversion pipeline implementation

another-rex:
Updated vulnfeeds project to support alpine conversion, and split up the pieces to have alpine export a intermediate format (format still to be finalized). 

Possible TODOs before being merged:
- Remove constant paths and take paths from arguments
- Combiner to use NVD API instead of files to get the latest data
- Update diagram in design document to match

It should be following the structure discussed last week, with one change which is CVE's are not loaded and converted into the intermediate step, but done at the end, as it seems like creating the intermediate OSV/CVE hybrid does not provide any benefits.


vulnfeeds/cmd/alpine/alpine_secdb.go
```diff
@@ -0,0 +1,15 @@
+package main
+
+type AlpineSecDB struct {
+	Apkurl        string   `json:"apkurl"`
```
oliverchang, line 4: nit: ApkURL for this name?

vulnfeeds/cmd/alpine/alpine_secdb.go
```diff
@@ -0,0 +1,15 @@
...
+	Archs         []string `json:"archs"`
+	Reponame      string   `json:"reponame"`
+	Urlprefix     string   `json:"urlprefix"`
```
oliverchang, line 7: nit: URLPrefix

vulnfeeds/cmd/alpine/alpine_secdb.go
```diff
@@ -0,0 +1,15 @@
+package main
+
+type AlpineSecDB struct {
+	Apkurl        string   `json:"apkurl"`
+	Archs         []string `json:"archs"`
+	Reponame      string   `json:"reponame"`
```
oliverchang, line 6: Please use UpperCase for all variable names.

vulnfeeds/cmd/alpine/alpine_secdb.go
```diff
@@ -0,0 +1,15 @@
...
+	Urlprefix     string   `json:"urlprefix"`
+	Distroversion string   `json:"distroversion"`
```
oliverchang, line 8: Please use UpperCase for all variable names.

vulnfeeds/cmd/alpine/main.go
```diff
@@ -0,0 +1,130 @@
...
+	"net/http"
+	"os"
+	"regexp"
+	"strings"
+
+	"github.com/google/osv/vulnfeeds/vulns"
+)
+
+const (
+	alpineUrlBase             = "https://secdb.alpinelinux.org/%s/main.json"
```
oliverchang, line 18: nit: s/Url/URL/ per https://github.com/golang/go/wiki/CodeReviewComments#initialisms

vulnfeeds/vulns/vulns.go
```diff
@@ -259,6 +268,47 @@ func FromCVE(id string, cve cves.CVEItem, pkg, ecosystem, purl, versionType stri
...
+					{Fixed: info.FixedCommit},
+				},
+			}
+			affected.Ranges = append(affected.Ranges, versionRange)
+		}
+		v.Affected = append(v.Affected, affected)
+	}
+}
+
+// FromCVEWithVersionExtraction Create an OSV object from a given CVEItem and package name
```
oliverchang, line 306: nit: s/Create/creates/. See https://go.dev/doc/comment

vulnfeeds/vulns/vulns.go
```diff
@@ -229,17 +230,25 @@ func extractAliases(id string, cve cves.CVE) []string {
 	return aliases
 }
 
-func FromCVE(id string, cve cves.CVEItem, pkg, ecosystem, purl, versionType string, validVersions []string) (*Vulnerability, []string) {
+type PackageInfo struct {
+	PkgName      string
+	Ecosystem    string
+	Purl         string
```
oliverchang, line 236: s/Purl/PURL/

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -0,0 +1,208 @@
...
+		for _, item := range nvdcve.CVEItems {
+			result[item.CVE.CVEDataMeta.ID] = item
+		}
+		log.Printf("Loaded CVE: %s", entry.Name())
+		file.Close()
+	}
+	return result
+}
+
+func downloadCveAsNeeded(version string) {
```
oliverchang, line 146: s/Cve/CVE/

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -0,0 +1,208 @@
...
+			output[cveId] = append(output[cveId], pkgInfos...)
+
+			log.Printf("Loaded Alpine Item: %s", entryInner.Name())
+			file.Close()
+		}
+	}
+	return output
+}
+
+func combineIntoOsv(loadedCves map[string]cves.CVEItem, allParts map[string][]vulns.PackageInfo) {
```
oliverchang, line 93: s/Osv/OSV/

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -0,0 +1,208 @@
...
+	cveUrlBase     = "https://nvd.nist.gov/feeds/json/cve/1.1/"
+	fileNameBase   = "nvdcve-1.1-"
+	startingYear   = 2002
+	cveOutputPath  = "output2"
+	partsInputPath = "parts"
+	osvOutputPath  = "osv_output"
+)
+
+func main() {
+	//jsonPath := flag.String("nvd_json", "", "Path to NVD CVE JSON.")
```
oliverchang, line 32: Let's remove this commented code. It's easy enough to add when we need it.

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+	err = os.MkdirAll(osvOutputPath, 0755)
+	if err != nil {
+		log.Fatalf("Can't create output path: %s", err)
+	}
+
+	for i := startingYear; i <= currentYear; i++ {
+		downloadCveAsNeeded(strconv.Itoa(i))
+	}
+	downloadCveAsNeeded("modified")
```
oliverchang, line 48: In general it's better to build more modular programs here, i.e. breaking up the downloading into a separate thing and passing that as a flag. To start with this could even be in bash right? 

This makes this easier to test (i.e. we can just provide a file) and extensible in the future.

vulnfeeds/cmd/alpine/main.go
```diff
@@ -0,0 +1,130 @@
...
+		for _, verPkg := range verPkgs {
+			pkgInfo := vulns.PackageInfo{
+				PkgName:      verPkg.Pkg,
+				FixedVersion: verPkg.Ver,
+				Ecosystem:    "Alpine:" + verPkg.AlpineVer,
+				Purl:         "pkg:alpine/" + verPkg.Pkg,
+			}
+			pkgInfos = append(pkgInfos, pkgInfo)
+		}
+		//if len(pkgInfos) > 1 {
```
oliverchang, line 98: Let's avoid committing commented code in general unless it's complicated and difficult to replicate.

oliverchang (commented):
Very nice!! some initial high level and stylistic comments

vulnfeeds/cmd/combine-to-osv/main.go
another-rex: That makes sense, I'll pull it out into a separate program

vulnfeeds/vulns/vulns.go
```diff
@@ -329,3 +378,8 @@ func (v *Vulnerability) ToYAML(w io.Writer) error {
 	encoder := yaml.NewEncoder(w)
 	return encoder.Encode(v)
 }
+
+func (v *Vulnerability) ToJSON(w io.Writer) error {
+	encoder := json.NewEncoder(w)
+	return encoder.Encode(v)
```
oliverchang, line 384: nit: can we just avoid the `encoder` variable completely?

vulnfeeds/cmd/download_cves/main.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+func downloadCVEAsNeeded(version string) {
+	file, err := os.OpenFile(cvePath+"/"+fileNameBase+version+".json", os.O_CREATE|os.O_RDWR, 0644)
+	defer file.Close()
+	if err != nil { // There's an existing file, check if it matches server file
+		log.Fatalf("Something's went wrong when creating/opening file %s, %s", version, err)
+	}
+
+	hasher := sha256.New()
```
oliverchang, line 40: We can just simplify **a lot** by not checking hashes and dumbly downloading all the time. 

There's probably not many scenarios where the complexity here justifies the savings? Without migrating to the NVD API, we'd just be running a cron job at some regular interval that's unlikely to hit any rate limiting.

vulnfeeds/cmd/download_cves/main.go
```diff
@@ -0,0 +1,95 @@
...
+					// File has been changed, exit loop and continue
+					break
+				}
+			}
+		}
+	}
+
+	res, err := http.Get(cveURLBase + fileNameBase + version + ".json.gz")
+	if err != nil {
+		log.Fatalf("Failed to get meta file for version '%s' with error %s", version, err)
```
oliverchang, line 71: Incorrect error message?

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -0,0 +1,124 @@
...
+	if err != nil {
+		log.Fatalf("Can't create output path: %s", err)
+	}
+
+	allCves := loadAllCVEs()
+	allParts := loadParts()
+	combineIntoOSV(allCves, allParts)
+}
+
+func loadParts() map[string][]vulns.PackageInfo {
```
oliverchang, line 36: Can you please write some tests for this?

vulnfeeds/vulns/vulns.go
```diff
@@ -229,17 +230,25 @@ func extractAliases(id string, cve cves.CVE) []string {
...
+type PackageInfo struct {
+	PkgName      string
+	Ecosystem    string
+	PURL         string
+	FixedVersion string
+	FixedCommit  string
+	Repo         string
+}
+
+// FromCVE Create an OSV object from a given CVEItem and package name
```
oliverchang, line 242: nit: s/Create/creates/

vulnfeeds/vulns/vulns.go
```diff
@@ -259,6 +268,47 @@ func FromCVE(id string, cve cves.CVEItem, pkg, ecosystem, purl, versionType stri
...
+		v.Affected = append(v.Affected, affected)
+	}
+}
+
+// FromCVEWithVersionExtraction create an OSV object from a given CVEItem and package name
+// Tries to extract version information from the given CVEItem
+func FromCVEWithVersionExtraction(id string, cve cves.CVEItem, pkgInfo []PackageInfo, validVersions []string) (*Vulnerability, []string) {
+	v, notes := FromCVE(id, cve, pkgInfo)
+	// TODO(rexpan): Check this doesn't change behaviour
+	affected := &v.Affected[0]
```
oliverchang, line 311: Rather than do this can you just add a boolean to control whether if we do version extraction? this can be part of the private interface i.e. `doFromCVE()`, and we can still have the public `FromCVE...` methods which take only the necessary args.

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -0,0 +1,124 @@
...
+	"flag"
+	"github.com/google/osv/vulnfeeds/cves"
+	"github.com/google/osv/vulnfeeds/vulns"
+	"log"
+	"os"
+	"strings"
+)
+
+const (
+	cvePath        = "cve_jsons"
```
oliverchang, line 14: Can you make these flags instead?

vulnfeeds/cmd/download_cves/main.go
another-rex: Yeah, I originally implemented that because it made local development much faster, but now that it's split up into two programs, I think it's safe to just remove it here. It does take a while to download (~5 minutes or so ) compared to <10 seconds to check, but it isn't long enough to have all this complexity.

vulnfeeds/vulns/vulns.go
another-rex: Removed the function entirely as it's not used anymore.

vulnfeeds/cmd/alpine/main.go
```diff
@@ -0,0 +1,126 @@
...
+
+	err := os.MkdirAll(alpineOutputPath, 0755)
+	if err != nil {
+		log.Fatalf("Can't create output path: %s", err)
+	}
+
+	generateAlpineOSV()
+}
+
+func getAllAlpineVersions() []string {
```
andrewpollock, line 34: Could you please add a descriptive comment per function?

See also https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences

vulnfeeds/cmd/alpine/run_alpine_convert.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash -e
```
andrewpollock, line 1: I'm not exactly sure where this is intended to be run (adding some comments wouldn't hurt to help clarify this).

It would be safer to use `/bin/sh` instead. Disregarding what https://google.github.io/styleguide/shellguide.html#which-shell-to-use says about Bash, it does say to use `set -e` rather than options on the shebang line...

vulnfeeds/cmd/alpine/run_alpine_convert.sh
```diff
@@ -0,0 +1,11 @@
...
+
+OSV_PARTS_OUTPUT=parts/alpine_output
```
andrewpollock, line 3: Quote this string: https://google.github.io/styleguide/shellguide.html#quoting and also quote all of your variable usage to make it space-safe.

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -0,0 +1,135 @@
...
+		log.Fatalf("Can't create output path: %s", err)
+	}
+
+	allCves := loadAllCVEs(*cvePath)
+	allParts := loadParts(*partsInputPath)
+	combinedData := combineIntoOSV(allCves, allParts)
+	writeOsvFile(combinedData, *osvOutputPath)
+}
+
+func loadParts(partsInputPath string) map[string][]vulns.PackageInfo {
```
andrewpollock, line 41: Could you please add a descriptive comment per function?

See also https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences

vulnfeeds/cmd/combine-to-osv/main_test.go
```diff
@@ -0,0 +1,84 @@
...
+
+func TestLoadParts(t *testing.T) {
+	allParts := loadParts("../../test_data/parts")
+
+	if len(allParts) != 10 {
+		t.Errorf("Expected 10 entries, got %d entries", len(allParts["Alpine"]))
+	}
+
+	hasCve := 0
+	for id, v := range allParts {
```
andrewpollock, line 48: Is this a good candidate for https://github.com/golang/go/wiki/TableDrivenTests?

vulnfeeds/cmd/combine-to-osv/main_test.go
```diff
@@ -0,0 +1,84 @@
...
+			hasCve++
+		}
+	}
+
+	if hasCve != 2 {
+		t.Errorf("Expected CVEs does not exist")
+	}
+}
+
+func TestCombineIntoOSV(t *testing.T) {
```
andrewpollock, line 71: Is this a good candidate for https://github.com/golang/go/wiki/TableDrivenTests?

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -0,0 +1,139 @@
...
+			log.Println("Unexpected file entry in " + partsInputPath)
+			continue
+		}
+		dirInner, err := os.ReadDir(partsInputPath + "/" + entry.Name())
+		if err != nil {
+			log.Fatalf("Failed to read dir? %s", err)
+		}
+
+		for _, entryInner := range dirInner {
+			file, err := os.Open(partsInputPath + "/" + entry.Name() + "/" + entryInner.Name())
```
oliverchang, line 59: Can you use https://pkg.go.dev/path#Join instead?

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -0,0 +1,139 @@
...
+	if err != nil {
+		log.Fatalf("Failed to read dir? %s", err)
+	}
+	output := map[string][]vulns.PackageInfo{}
+	for _, entry := range dir {
+		if !entry.IsDir() {
+			log.Println("Unexpected file entry in " + partsInputPath)
+			continue
+		}
+		dirInner, err := os.ReadDir(partsInputPath + "/" + entry.Name())
```
oliverchang, line 53: Can you use https://pkg.go.dev/path#Join instead?

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -0,0 +1,139 @@
...
+			if err != nil {
+				log.Fatalf("Failed to open cve json: %s", err)
+			}
+			var pkgInfos []vulns.PackageInfo
+			err = json.NewDecoder(file).Decode(&pkgInfos)
+			if err != nil {
+				log.Fatalf("Failed to decode json: %s", err)
+			}
+
+			cveId := strings.Split(entryInner.Name(), ".")[0]
```
oliverchang, line 69: Can you use https://pkg.go.dev/path#Base instead?

vulnfeeds/cmd/combine-to-osv/main.go
oliverchang, line 53: nit: create another function to process the inner dirs.

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -0,0 +1,139 @@
...
+			output[cveId] = append(output[cveId], pkgInfos...)
+
+			log.Printf("Loaded Alpine Item: %s", entryInner.Name())
+			file.Close()
+		}
+	}
+	return output
+}
+
+// combineIntoOsv creates OSV entry by combining loaded CVEs from NVD and PackageInfo information from security advisories.
```
oliverchang, line 79: s/Osv/OSV/

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -0,0 +1,139 @@
...
+			continue
+		}
+		cve, _ := vulns.FromCVE(vId, v, allParts[vId])
+		convertedCves[vId] = cve
+	}
+	return convertedCves
+}
+
+// writeOsvFile writes out the given osv objects into individual json files
+func writeOsvFile(osvData map[string]*vulns.Vulnerability, osvOutputPath string) {
```
oliverchang, line 94: s/Osv/OSV/

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -0,0 +1,139 @@
...
+	for vId, osv := range osvData {
+		file, err := os.OpenFile(osvOutputPath+"/"+vId+".json", os.O_CREATE|os.O_RDWR, 0644)
```
oliverchang, line 96: nit: use  https://pkg.go.dev/path#Join instead.

vulnfeeds/vulns/vulns.go
```diff
@@ -229,17 +229,25 @@ func extractAliases(id string, cve cves.CVE) []string {
 	return aliases
 }
 
-func FromCVE(id string, cve cves.CVEItem, pkg, ecosystem, purl, versionType string, validVersions []string) (*Vulnerability, []string) {
+type PackageInfo struct {
+	PkgName      string
```
oliverchang, line 233: Can you add struct tags per https://pkg.go.dev/encoding/json#Marshal to make these use `snake_case` instead to match the OSV schema's casing?

vulnfeeds/cmd/alpine/run_alpine_convert.sh
another-rex: Moved the -e into a `set -e` line. Let's stick with bash for now since all the other scripts are running bash as well. 

This is intended to be the entrypoint of the docker image. I've added a comment at the top of the two script files clarifying this.

vulnfeeds/vulns/vulns.go
another-rex: Not too sure if that will fix the issue though, the new changes allows adding multiple PackageInfo, which means we still need the &v.Affected[0] line. It also shortens the very long FromCVE function into more modular parts.

oliverchang: I see -- in that case, best to refactor this functionality into a separate function that's clearer about what it's doing. e.g.

```go
func ExtractVersionInformation(affected Affected, cve cves.CVEItem, ..)
```

We can also refactor `AddPkgInfo` to take a single PkgInfo instead, and return the converted `Affected` pointer, and remove that argument from `FromCVE`. 

That way, callers can be very explicit about the cases where they care about version extraction, rather than encoding this weird limitation here.

Something like...

```go
v, notes := vulns.FromCVE(id, cve)
affected := v.AddPkgInfo(vulns.PackageInfo{
  PkgName:   pkg,
  Ecosystem: "PyPI",
  PURL:      purl,
})
notes = append(notes, ExtractVersionInformation(affected, cve, ..., validVersions)...)
```


WDYT?

another-rex: That looks good to me, I'll refactor it now.

another-rex: It now looks something like this:
```
v, notes := vulns.FromCVE(id, cve)
v.AddPkgInfo(pkgInfo)

versions, versionNotes := cves.ExtractVersionInfo(cve, validVersions)
notes = append(notes, versionNotes...)

v.Affected[0].AttachExtractedVersionInfo(versions)
    log.Printf("No affected versions detected.")
}
```

A possibly cleaner version would be to have `addPkgInfo` return the affected field it added, but that would require changing the affected array in `Vulnerability` to an array of `*Affected` instead of `Affected`

oliverchang: Thanks! This is fine for now. We can iterate on this later if you have better ideas here :)

vulnfeeds/test_data/parts/alpine/CVE-2015-9251.alpine.json
```diff
@@ -0,0 +1,10 @@
+[
+  {
+    "PkgName": "ruby",
```
oliverchang, line 3: Do we need to update all of these test files with the new casing?

oliverchang (approved):
nice! LGTM with one last comment.

vulnfeeds/cmd/combine-to-osv/main.go
another-rex: Turns out path.base doesn't actually give the file name without extension, it's just for getting the file name in a long path.

vulnfeeds/cmd/combine-to-osv/main_test.go
another-rex: Yep, converted into a more table driven version now

oliverchang:
LGTM! Is this ready to merge?

another-rex:
Almost, just going to add a run_test.sh file to automate running the go tests

another-rex:
Actually that might be better to put into a separate PR, let's merge this for now.

oliverchang:
> Actually that might be better to put into a separate PR, let's merge this for now.

Yep, we need to start running our Go tests in CI!

PR #653, opened 2022-08-31
Better output printing for osv-scanner

another-rex:
Also assign source to each scanned object to know where a vulnerability is from when scanning multiple files. 

This PR is focused on the human readable output. Machine readable output will be in a separate commit once format is decided. The output should look something like this:

Also fixes a bug/typo where --lockfile and --sbom flags doesn't actually work


oliverchang:
holy smokes, this looks amazing!

tools/osv-scanner/main.go
```diff
@@ -252,5 +299,22 @@ func main() {
 		return
 	}
 
-	printResults(query, resp)
+	genericPrintResults(query, resp, func(query *osv.Query) bool {
```
oliverchang, line 302: Can we push this complicated logic into `genericPrintResults` instead? There's a bit of complicated indirection and logic getting passed around with the current setup.

tools/osv-scanner/main.go
```diff
@@ -149,30 +157,65 @@ func scanDebianDocker(query *osv.BatchedQuery, dockerImageName string) {
...
+	query osv.BatchedQuery,
+	resp *osv.BatchedResponse,
+	condition func(*osv.Query) bool,
+	dataExtractor func(*osv.Query) []interface{},
+	headers []interface{})  {
+	outputTable := table.NewWriter()
+	outputTable.SetOutputMirror(os.Stdout)
+	headerRows := table.Row{"Source"}
+	headerRows = append(headerRows, headers...)
+	headerRows = append(headerRows, "Vulnerability ID", "OSV Url")
```
oliverchang, line 186: nit: s/Url/URL/

tools/osv-scanner/main.go
```diff
@@ -252,5 +299,22 @@ func main() {
 		return
 	}
 
-	printResults(query, resp)
+	genericPrintResults(query, resp, func(query *osv.Query) bool {
+		return query.Commit == "" && query.Package.PURL == ""
+	}, func(query *osv.Query) []interface{} {
+		return table.Row{query.Package.Ecosystem, query.Package.Name, query.Version}
+	}, table.Row{"Affected Ecosystem", "Affected Package", "Installed Version"})
```
oliverchang, line 306: Can we just unify the header columns here? 

e.g. we can print somethign like

| Ecosystem | Package | Version | 
| Debian | foo | 1.0.0 |
| GIT | <commit> | ..sha1 hash..  |

Ideally we can unpack the Package URL and present a consistent Ecosystem/Package label for these too. That can come in another PR. In the meantime maybe something like:

| Package URL | pkg:url without version | version |

oliverchang (commented):
very nice! some comments

tools/osv-scanner/main.go
```diff
@@ -4,13 +4,16 @@ import (
 	"bufio"
 	"bytes"
 	"errors"
+	"golang.org/x/crypto/ssh/terminal"
```
oliverchang, line 7: This should go into the third party block from line 14 too, since it's not a standard library.

tools/osv-scanner/main.go
```diff
@@ -149,30 +157,79 @@ func scanDebianDocker(query *osv.BatchedQuery, dockerImageName string) {
...
 		for _, vuln := range resp.Results[i].Vulns {
-			urls = append(urls, osv.BaseVulnerabilityURL+vuln.ID)
+			outputRow := table.Row{query.Source}
+			shouldMerge := false
+			if query.Commit != "" {
+				outputRow = append(outputRow, "GIT", query.Commit, query.Commit)
+				maxCharacters = max(maxCharacters, len(query.Commit))
+				shouldMerge = true
+			} else if query.Package.PURL != "" {
+				outputRow = append(outputRow, "PURL", query.Package.PURL, query.Package.PURL)
```
oliverchang, line 199: Rather than duplicate the PURL, let's put it in a single column (or split it up into base PURL + version).

another-rex: So this actually will merge them into one, which looks like this:
image
Well the duplicate commits will result in something like this, for PURLs I am now parsing it to split it up as suggested.

oliverchang: Wow, that's pretty magical. That's fine for now then!

oliverchang: Or if you did it already, let's just go with that. I think we eventually want to abstract away PURLS here and show a consistent "Ecosystem + Package + version"

oliverchang (approved):

oliverchang (approved):
let's get this merged!

PR #674, opened 2022-09-07
Group osv-scanner results by aliases.

### LINKED ISSUE #668: osv-scanner: Group results by aliases

oliverchang:
- Hydrate results of batch query (by making individual GET requests) to get the aliases for each matched vulnerability.
- Add a internal/grouper package for performing grouping by aliases.
- Output grouped vulnerabilities and links in the same row in the human readable output from the osv-scanner tool.

Fixes #668.


tools/osv-scanner/internal/osv/osv.go
```diff
@@ -40,11 +43,15 @@ type BatchedQuery struct {
 	Queries []*Query `json:"queries"`
 }
 
+// Vulnerability represents a vulnerability entry from OSV.
+type Vulnerability struct {
```
another-rex, line 47: Prefer having the batched vulnerability and the hydrated version be two different types. This way it's clear that the Group  only takes already hydrated elements.

oliverchang: Good suggestion, done!

another-rex (approved):

PR #675, opened 2022-09-07
Packagist version enumeration support

### LINKED ISSUE #230: Version expansion for Packagist
Packagist recommends, but does not enforce Semver

another-rex:
Fixes #230 

Potential improvements before merging:
- From the Best Practices here: https://packagist.org/apidoc, it is recommended to send a User-agent header to allow them to contact us to resolve any bandwidth issues.
- Also can utilize `If-Modified-Since` headers to avoid expiring caches too early when calling the API
- The version enumeration is not quite identical to the C implementation in PHP, though matches the slightly ambiguous specification given in the documentation. Should be fine for the vast majority of cases.

We should also think about pushing the packagist comparison upstream to the univers library.


osv/packagist_version.py
```diff
@@ -0,0 +1,137 @@
+import re
+
+
+class PackagistVersion(object):
```
oliverchang, line 4: Nit: no need for `object` here in Python3.

Incredible work! Are there any upstream tests we can port, from either Packagist or php itself to test version_compare and the packagist versioning?

osv/packagist_version.py
```diff
@@ -0,0 +1,137 @@
+import re
```
oliverchang, line 1: Nit: add license header.

oliverchang (commented):
incredible work!

oliverchang:
Reposting comment since it may have gotten lost in a previously resolved comment: 

Are there any upstream tests we can port, from either Packagist or php itself to test version_compare and the packagist versioning?

another-rex:
Ah, yes, found some here, I'll try to port them in to see if the implementation is the same:
https://github.com/php/php-src/blob/master/ext/standard/tests/versioning/version_compare.phpt

osv/ecosystems_test.py
```diff
@@ -146,12 +146,29 @@ def test_packagist(self):
...
-    self.assertIn("3.3.1", enumerated_versions)
-    self.assertIn("3.3.0", enumerated_versions)
+    enumerated_versions = ecosystem.enumerate_versions('neos/neos', '3.3.0',
+                                                       '4.4.0')
+    self.assertIn('4.3.19', enumerated_versions)
+    self.assertIn('4.2.18', enumerated_versions)
+    self.assertIn('3.3.1', enumerated_versions)
+    self.assertIn('3.3.0', enumerated_versions)
+
+    with open('osv/testdata/packagist_test_cases.txt') as file:
```
oliverchang, line 156: nit: create a _TEST_DATA_DIR constant and make it relative to the file. also use os.path.join

something like

```python
_TEST_DATA_DIR = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'testdata'))

File: osv/testdata/packagist_test_cases.txt [hunk 4]
```diff
@@ -0,0 +1,53 @@
+1 < 2
```
oliverchang, line 1: Can we add the tests from https://github.com/composer/semver/blob/main/tests/ComparatorTest.php as well?

another-rex: Added now

osv/testdata/packagist_test_cases.txt
```diff
@@ -0,0 +1,61 @@
...
+1.0pl1 > 1.0-dev
+1.0pl1 > 1.0a1
+1.0pl1 > 1.0b1
+1.0pl1 > 1.0RC1
+1.0pl1 > 1.0rc1
+1.0pl1 > 1.0
+1.0pl1 = 1.0pl1
+1.25.0 < 1.26.0
+1.0.0 < 1.2-dev
+dev-foo < 1.26.0
```
oliverchang, line 56: How about dev-foo, dev-master etc? 

https://github.com/composer/semver/blob/main/tests/ComparatorTest.php#L166. Can we just copy every single one from here rather then picking select ones? 

Can we also add a note somewhere (can be just the python code that references this file) about where we pulled these testcases from?

osv/ecosystems.py
```diff
@@ -24,6 +24,7 @@
 from .third_party.univers.debian import Version as DebianVersion
 from .third_party.univers.gem import GemVersion
 
+from . import packagist_version
```
oliverchang, line 27: Oh one more nit: Please move this to after line 30 to preserve the alphabetic ordering.

osv/packagist_version.py
```diff
@@ -0,0 +1,154 @@
...
+  def compare_special_versions(version_part_a: str, version_part_b: str) -> int:
+    """
+    Compares the order of special characters against the order specified in php
+    docs.
+
+    any string not found in this list < dev < alpha = a < beta = b < RC = rc < # < pl = p.
+
+    :return: 1 if a > b, -1 if b > a, 0 if a == b
+    """
+    # This isn't quite the behaviour of the c implementation of packagist
```
oliverchang, line 145: c implementation in PHP? or packagist?

osv/packagist_version.py
```diff
@@ -0,0 +1,154 @@
...
+
+  The function first replaces _, - and + with a dot . in the version strings and
+  also inserts dots . before and after any non number so that for example
+  '4.3.2RC1' becomes '4.3.2.RC.1'. Then it compares the parts starting from
+  left to right.
+
+  If a part contains special version strings these are handled in the following order:
+  any string not found in this list < dev < alpha = a < beta = b < RC = rc < # < pl = p.
+  This way not only versions with different levels like '4.1' and '4.1.2' can be
+  compared but also any PHP specific version containing development state.
```
oliverchang, line 32: Any other ambiguous/undocumented things to call out in the implementation here?

osv/testdata/packagist_test_cases.txt
another-rex: Most of them are duplicates because they are also testing the different `> < = <>`...etc. Which means they also need to test when they are `false` instead of `true`. 

The dev-foo to dev-master seems to be a weird case, according to the tests it is neither less than, greater than, nor equal. Not sure what the desired behavior is here.

oliverchang (commented):
amazing work as always! a couple more comments

> From the Best Practices here: https://packagist.org/apidoc, it is recommended to send a User-agent header to allow them to contact us to resolve any bandwidth issues.

Yes please do this. Just set osv.dev or something similar. 

> Also can utilize If-Modified-Since headers to avoid expiring caches too early when calling the API

Can we at least start with our own redis cache? We can build out some more sophisticated logic to expire our own cache etc later. Just add a TODO.

> The version enumeration is not quite identical to the C implementation in PHP, though matches the slightly ambiguous specification given in the documentation. Should be fine for the vast majority of cases.

Please document any known cases in the code.

osv/testdata/packagist_test_cases.txt
oliverchang: We should match that behaviour, if it's easy to do so. 

For enumeration though, we should filter these out.

osv/packagist_version.py
another-rex: There was one other issue with how it dealt with `#`, but I believe I have now fixed that so that it should have the same behavior

another-rex:
> > From the Best Practices here: https://packagist.org/apidoc, it is recommended to send a User-agent header to allow them to contact us to resolve any bandwidth issues.
> 
> Yes please do this. Just set osv.dev or something similar.

User-agent header now sends osv.dev

> > Also can utilize If-Modified-Since headers to avoid expiring caches too early when calling the API
> 
> Can we at least start with our own redis cache? We can build out some more sophisticated logic to expire our own cache etc later. Just add a TODO.

Yep, it is already using the redis cache, added a TODO

> The version enumeration is not quite identical to the C implementation in PHP, though matches the slightly ambiguous specification given in the documentation. Should be fine for the vast majority of cases.
> 
> Please document any known cases in the code.

Done

osv/testdata/packagist_test_cases.txt
another-rex: Not 100% sure how to match that behavior and have it make sense in how we are using the `sort_version` function. For enumeration I added a TODO as I'm not sure exactly what these versions will look like for actual packages, so let's just monitor it for now and implement filtering when we find some on actual packages.

oliverchang (approved):

PR #687, opened 2022-09-12
Add a new build and deploy script, add google cloud logging

another-rex:
To avoid spamming error messages into logs.
Also log errors with stacktrace to propagate the error to error reporting.


another-rex:
TODO: Fix go tests because logger is not initiated

another-rex:
> TODO: Fix go tests because logger is not initiated

Fixed now by creating a LoggerWrapper type

vulnfeeds/utility/logginghelpers.go
```diff
@@ -0,0 +1,60 @@
...
+)
+
+// LoggerWrapper wraps the Logger provided by google cloud
+// Will default to the go stdout and stderr logging if GCP logger is not set
+type LoggerWrapper struct {
+	Logger *logging.Logger
+}
+
+// InfoLogf prints Info level log
+func (wrapper LoggerWrapper) InfoLogf(format string, a ...any) {
```
oliverchang, line 19: nit: just Infof, Warnf, Fatalf etc.

The "Log" part is redundant because you're calling this from a logger instance that's going to be named "logger" or something similar anyway.

oliverchang (approved):

PR #689, opened 2022-09-12
DetermineVersion endpoint

susperius:
Adding a DetermineVersion gRPC endpoint to the API server. The endpoint requires a package name for now. It then compares hashes in a brute force way.

This PR also adds the possibility to define a local run in the `server.py` to allow using the gRPC service with `grpcurl`.


susperius:
Oliver could you please double check [gcp/api/osv_service_v1_pb2.py](https://github.com/google/osv.dev/pull/689/files#diff-9d2c6da56a0ac1a4052cdbcd40e7db806fc30665fa810d390e318ff5520be0dd). I just used the plain proto compiler command without additional flags except for gRPC output.

gcp/api/v1/osv_service_v1.proto
```diff
@@ -67,6 +67,50 @@ message QueryAffectedBatchParameters {
...
+    UNSPECIFIED = 0;
+    OSV = 1;
+    CPE = 2;
+    VERSION = 3;
+    GIT_REPO = 4;
+  }
+  IDType type = 1;
+  // Identifier value.
+  string value = 2;
+  // Score in the intervall (0.0 1.0] with 1.0
```
oliverchang, line 109: nit: s/intervall/interval/

Also missing comma in the range?

gcp/api/v1/osv_service_v1.proto
```diff
@@ -67,6 +67,50 @@ message QueryAffectedBatchParameters {
...
+  // being a perfect match.
```
oliverchang, line 110: nit: no need for this to be on a newline.

gcp/api/v1/osv_service_v1.proto
```diff
@@ -93,4 +137,12 @@ service OSV {
       body:"query"
     };
   }
+
+  // Determine the version of the provided hash values.
+  rpc DetermineVersion(DetermineVersionParameters) returns (VersionMatchList) {
+    option (google.api.http) = {
+      post:"/v1/determineversion"
```
oliverchang, line 144: can we put this under `/v1experimental/identifycode` for now?

gcp/api/server.py
```diff
@@ -105,6 +108,58 @@ def QueryAffectedBatch(self, request, context):
 
     return osv_service_v1_pb2.BatchVulnerabilityList(results=batch_results)
 
+  @ndb_context
+  def DetermineVersion(self, request, context):
+    """Determine the version of the provided hashes."""
+    if request.query.name == "":
```
oliverchang, line 114: nit:

```python
if not request.query.name:
 ...
```

See https://google.github.io/styleguide/pyguide.html#214-truefalse-evaluations

gcp/api/server.py
```diff
@@ -105,6 +108,58 @@ def QueryAffectedBatch(self, request, context):
...
+                    'Querying only by file hash is not implemented yet.')
+      return None
+    return get_version_by_name(request.query, context).result()
+
+
+@ndb.tasklet
+def get_version_by_name(query: osv_service_v1_pb2.VersionQuery,
+                        context: grpc.ServicerContext) -> ndb.Future:
+  """Identifies the version based on the provided name."""
+  q = osv.RepoIndex.query(osv.RepoIndex.name == query.name)
```
oliverchang, line 125: s/q/query/. We can rename the parameter `version_query` instead.

gcp/api/server.py
```diff
@@ -105,6 +108,58 @@ def QueryAffectedBatch(self, request, context):
...
+
+@ndb.tasklet
+def compare_hashes_from_commit(
+    idx: osv.RepoIndex,
+    hashes: List[osv_service_v1_pb2.FileHash]) -> ndb.Future:
+  """"Retrieves the hashes from the provided index and compares them to the input hashes."""
+  total_files = 0
+  matching_hashes = 0
+  for i in range(idx.pages):
+    key = ndb.Key(idx.key.kind(), idx.key.id(), osv.RepoIndexResult,
```
oliverchang, line 148: put this in a helper function to generate the key.

gcp/api/server.py
```diff
@@ -105,6 +108,58 @@ def QueryAffectedBatch(self, request, context):
...
+    while (yield it.has_next_async()):
+      res = it.next()
+      for idxHash in res.file_results:
+        for inHash in hashes:
+          if inHash.hash == idxHash:
+            matching_hashes += 1
+            break
+        total_files += 1
+  score = matching_hashes / total_files if total_files != 0 else 0.0
+  return osv_service_v1_pb2.VersionMatch(type=3, value=idx.version, score=score)
```
oliverchang, line 161: Instead of 3, can you use the generated enum instead?

gcp/api/server.py
```diff
@@ -105,6 +108,58 @@ def QueryAffectedBatch(self, request, context):
...
+  total_files = 0
+  matching_hashes = 0
+  for i in range(idx.pages):
+    key = ndb.Key(idx.key.kind(), idx.key.id(), osv.RepoIndexResult,
+                  f"{idx.commit.hex()}-{idx.file_hash_type}-{i}")
+    q = osv.RepoIndexResult.query(osv.RepoIndexResult.key == key)
+    it = q.iter()
+    while (yield it.has_next_async()):
+      res = it.next()
+      for idxHash in res.file_results:
```
oliverchang, line 154: nit: s/idxHash/idx_hash/

gcp/api/server.py
```diff
@@ -105,6 +108,58 @@ def QueryAffectedBatch(self, request, context):
...
+        for inHash in hashes:
```
oliverchang, line 155: nit: s/inHash/in_hash/

gcp/api/server.py
```diff
@@ -105,6 +108,58 @@ def QueryAffectedBatch(self, request, context):
...
+@ndb.tasklet
+def get_version_by_name(query: osv_service_v1_pb2.VersionQuery,
+                        context: grpc.ServicerContext) -> ndb.Future:
+  """Identifies the version based on the provided name."""
+  q = osv.RepoIndex.query(osv.RepoIndex.name == query.name)
+  it = q.iter()
+  futures = []
+  while (yield it.has_next_async()):
+    idx = it.next()
+    v = compare_hashes_from_commit(idx, query.file_hashes)
```
oliverchang, line 130: s/v/match/

gcp/api/server.py
```diff
@@ -105,6 +108,58 @@ def QueryAffectedBatch(self, request, context):
...
+    futures.append(v)
+  results = []
+  for f in futures:
+    r = f.result()
```
oliverchang, line 134: s/r/match/

gcp/api/server.py
```diff
@@ -105,6 +108,58 @@ def QueryAffectedBatch(self, request, context):
...
+
+@ndb.tasklet
+def get_version_by_name(query: osv_service_v1_pb2.VersionQuery,
+                        context: grpc.ServicerContext) -> ndb.Future:
+  """Identifies the version based on the provided name."""
+  q = osv.RepoIndex.query(osv.RepoIndex.name == query.name)
+  it = q.iter()
+  futures = []
+  while (yield it.has_next_async()):
+    idx = it.next()
```
oliverchang, line 129: s/idx/repo_index/

oliverchang (commented):
nice! do you have any numbers re performance?

oliverchang:
Re the generated protobuf code, did you run the command from [https://github.com/google/osv.dev/tree/master/gcp/api/v1 ?](https://github.com/google/osv.dev/blob/master/gcp/api/v1/README.md) ?

susperius:
> 

thanks! no just some poking around. Since, we just receive the data sequentially it might take a time especially for large repos with many versions.

osv/models.py
```diff
@@ -637,6 +637,36 @@ def to_vulnerability(self, include_source=False):
...
+  # The hash algorithm used
+  file_hash_type = ndb.StringProperty()
+  # The repository type
+  repo_type = ndb.StringProperty()
+
+
+class RepoIndexResult(ndb.Model):
+  """RepoIndexResult entries containing the actual hash values"""
+  # The hash values
+  file_results = ndb.BlobProperty(repeated=True)
```
oliverchang, line 665: This should be: 
```
+class FileResult(ndb.Model):
+  """Indexed file result."""
+  # The hash of the file.
+  hash = ndb.BlobProperty()
+  # The path of the file.
+  path = ndb.StringProperty()
+
+
 class RepoIndexResult(ndb.Model):
   """RepoIndexResult entries containing the actual hash values"""
-  # The hash values
-  file_results = ndb.BlobProperty(repeated=True)
-  # The actual page
+  # The hash values.
+  file_results = ndb.StructuredProperty(FileResult, repeated=True)
+  # The actual page.
   page = ndb.IntegerProperty()
 
```

gcp/api/server.py
```diff
@@ -105,6 +108,66 @@ def QueryAffectedBatch(self, request, context):
...
+  matching_hashes = 0
+  for i in range(idx.pages):
+    key = version_hashes_key(idx.key, idx.commit, idx.file_hash_type, i)
+    q = osv.RepoIndexResult.query(osv.RepoIndexResult.key == key)
+    it = q.iter()
+    while (yield it.has_next_async()):
+      res = it.next()
+      for idx_hash in res.file_results:
+        for in_hash in hashes:
+          if in_hash.hash == idx_hash:
```
oliverchang, line 155: This would need to be changed to `idx_hash.hash` after the models.py change below.

gcp/api/server.py
```diff
@@ -105,6 +108,66 @@ def QueryAffectedBatch(self, request, context):
...
+@ndb.tasklet
+def compare_hashes_from_commit(
+    idx: osv.RepoIndex,
+    hashes: List[osv_service_v1_pb2.FileHash]) -> ndb.Future:
+  """"Retrieves the hashes from the provided index and compares them to the input hashes."""
+  total_files = 0
+  matching_hashes = 0
+  for i in range(idx.pages):
+    key = version_hashes_key(idx.key, idx.commit, idx.file_hash_type, i)
+    q = osv.RepoIndexResult.query(osv.RepoIndexResult.key == key)
```
oliverchang, line 149: You can just do `key.get()`. keys are also unique, so there will always be at most one entity matching it.

oliverchang (approved):
Thanks! This LGTM, but trying this locally it's quite slow and I think we need to do a bit of optimization here.

I also pushed some fixes to this PR:
- Added the grpioc-reflection dependenc
-  Increased the ESPv2 proxy deadline for the test API config so that REST requests complete. 

(see https://github.com/google/osv.dev/pull/689/commits/0f5c08d8e43f1f3d37801c2378fe5873ec082e05)

gcp/api/v1/osv_service_v1.proto
```diff
@@ -67,6 +67,49 @@ message QueryAffectedBatchParameters {
...
+// Result of DetmineVersion.
+message VersionMatchList {
+  repeated VersionMatch matches = 1;
+}
+
+// Match information for the provided VersionQuery.
+message VersionMatch {
+  enum IDType {
+    UNSPECIFIED = 0;
+    OSV = 1;
```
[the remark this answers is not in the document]
susperius: With the suggested added version field couldn't that be achieved by using the value field for the ecosystem?

gcp/api/v1/osv_service_v1.proto
```diff
@@ -67,6 +67,49 @@ message QueryAffectedBatchParameters {
...
+    CPE = 2;
+    VERSION = 3;
```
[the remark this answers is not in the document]
susperius: The idea was to have a single version type instead of providing different ones based on the type. So, a GIT type returns repository information, OSV an OSV based identifier (I think I mixed this up with OSV vuln ids) and CPE the full cpe tag.

I can see it working the way you suggest it but we need to change the data collection slightly because we don't store the tag in datastore.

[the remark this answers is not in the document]
oliverchang: Right, I think we should store the original tag either way. It will give us more flexibility if we need to change our version parsing algorithm/regex too.

gcp/api/v1/osv_service_v1.proto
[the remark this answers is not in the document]
oliverchang: We need: 
- ecosystem name (e.g. PyPI, npm etc)
- package name (e.g. "django")
- version.

Even if we add version, we still need a place to encode "ecosystem".

oliverchang (approved):
merging this for now, but leaving some replies to your questions :)

PR #698, opened 2022-09-14
OSV scanner github action

### LINKED ISSUE #664: Build GitHub actions
For running osv-scanner

another-rex:
Creates docker image for osv scanner to allow it to run as a github action, and the appropriate action.yml.

The current implementation is not particularly useful, only printing out the found vulnerabilities to the logs. It should be running under this PR.

TODO: Decide what the goal of the github action will be.


tools/osv-scanner/action.yml
```diff
@@ -0,0 +1,13 @@
...
+inputs:
+  to-scan: 
+    description: 'Directory to scan'
+    required: true
+    default: '/github/workspace'
+runs:
+  using: 'docker'
+  image: 'Dockerfile'
+  args:
+    - '--skip-git'
```
oliverchang, line 12: Why do we want to skip git? Is it causing performance issues?

oliverchang (commented):
nice! this is a really good starting point, but I wouldn't have "Closes #664" in the PR description just yet :)

tools/osv-scanner/action.yml
another-rex: Skipping git because github actions seem to do something weird with safe directories:
https://github.com/actions/checkout/issues/760

The issue has been closed, but the issue is still there and the fixes and workarounds proposed doesn't work. (I think github actions inject it's own safe directories to the docker container?). I'm not 100% sure that is the issue either, since osv-scanner doesn't pipe out the subcommand output. 

However I don't think it's worth debugging more at least for the moment, since this action is being ran on the PR, the scanning is scanning a brand new commit hash, which will never get any hits in OSV.  It would be useful for searching vendored in git subdirectories/submodules though, so maybe we should look into fixing this some more. 

Other related issues:
https://github.com/actions/checkout/issues/766
https://github.com/actions/runner/issues/2033

oliverchang (approved):

Issue #700, opened 2022-09-14
support `last_affected`


oliverchang:
We've been putting this off since no producers have been using this yet, but GitHub will be doing this soon so we need to add support for it: https://github.com/github/advisory-database/issues/470

oliverchang:
Remaining work: support `last_affected` for git ranges.

PR #702, opened 2022-09-14
Support last_affected events for versions.

oliverchang:
Supporting git commits is still a TODO.

Also clean up some remaining code that handled old/removed fields in the OSV schema.

Part of #700.


another-rex:
It seems like we are passing around these together a lot, with most ecosystems not actually needing to use them and is directly passed from `enumerate_functions` into `_get_affected_versions`. 
```
introduced,
fixed=None,
last_affected=None,
limits=None):
```

It might be worth it to bundle all of these into one object to make it cleaner and easier to extend in the future. (i.e. we don't have to change the signature of `enumerate_versions` when we add an additional field.)

oliverchang:
> It seems like we are passing around these together a lot, with most ecosystems not actually needing to use them and is directly passed from `enumerate_functions` into `_get_affected_versions`.
> 
> ```
> introduced,
> fixed=None,
> last_affected=None,
> limits=None):
> ```
> 
> It might be worth it to bundle all of these into one object to make it cleaner and easier to extend in the future. (i.e. we don't have to change the signature of `enumerate_versions` when we add an additional field.)

Hmm, not sure if it's worth the additional indirection. Changing this should be a very rare occurrence (there very likely at most one more time we do this, and the chances of that are currently fairly low).

another-rex (approved):

PR #707, opened 2022-09-15
Add Hex to SEMVER_ECOSYSTEMS

oliverchang:
Hex enforces semver according to
https://hex.pm/docs/publish#adding-metadata-to-code-classinlinemixexscode


andrewpollock (approved):

another-rex (commented):
I think you also need to add Hex to _ecosystems similar to Go or npm.

oliverchang:
> I think you also need to add Hex to _ecosystems similar to Go or npm.

Good catch, done!

another-rex (approved):

PR #711, opened 2022-09-15
Update diagram to reflect exporter, importer and sources more specifi…

andrewpollock:
…cally


oliverchang:
nice! can you make the background white instead of transparent? otherwise it's a bit unreadable on dark mode.

oliverchang (approved):

PR #727, opened 2022-09-19
Machine output for osv-scanner

another-rex:
Follows a similar structure to G-Rath/osv-detector. 

Possible improvement would be to figure out how to better display git commits in the output.


tools/osv-scanner/internal/output/machinejson.go
```diff
@@ -0,0 +1,71 @@
...
+			FilePath: source,
+			Packages: packages,
+		})
+	}
+
+	marshal, err := json.MarshalIndent(output, "", "  ")
+	if err != nil {
+		return
+	}
+	print(string(marshal))
```
oliverchang, line 70: rather than do this, can we just create an `Encoder` (https://pkg.go.dev/encoding/json#Encoder) and set the output stream to `os.Stdout`.

tools/osv-scanner/internal/output/purl_to_package.go
```diff
@@ -0,0 +1,18 @@
+package output
+
+import (
+	"github.com/package-url/packageurl-go"
+)
+
+func PURLToPackage(purl string) (Package, error) {
+	parsedPURL, err := packageurl.FromString(purl)
+	if err != nil {
+		// log.Printf("Failed to parse purl: %s, with error: %s", purl, err)
```
oliverchang, line 10: Remove  debug log if we don't intend to turn it on?

tools/osv-scanner/internal/output/machinejson.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+// PrintJSONResults writes results to the provided writer in JSON format
+func PrintJSONResults(query osv.BatchedQuery, resp *osv.HydratedBatchedResponse, outputWriter io.Writer) {
+	output := Output{}
+	groupedBySource := map[string][]Package{}
+
+	for i, query := range query.Queries {
+		response := resp.Results[i]
+		if len(response.Vulns) == 0 {
```
oliverchang, line 34: I wonder if we need a flag where we output everything we scanned, just to provide assurance that we did in fact consider all the dependencies that were provided. Can come later if people ask for it.

oliverchang (approved):

tools/osv-scanner/internal/output/machinejson.go
```diff
@@ -0,0 +1,71 @@
...
+	for source, packages := range groupedBySource {
+		output.Results = append(output.Results, Result{
+			FilePath: source,
+			Packages: packages,
+		})
+	}
+
+	encoder := json.NewEncoder(outputWriter)
+	encoder.SetIndent("", "  ")
+	encoder.Encode(output)
```
oliverchang, line 70: return error? or at least check and panic if this fails?

tools/osv-scanner/internal/output/purl_to_package.go
```diff
@@ -0,0 +1,18 @@
...
+		return Package{}, err
+	}
+	return Package{
+		Name:      parsedPURL.Name,
+		Ecosystem: parsedPURL.Type, // TODO: Might want some mapping here to properly cased ecosystems
```
oliverchang, line 15: Missed this earlier: yeah we will need to fix this. See https://github.com/google/osv.dev/blob/2b01f0db2a37b8628db18a94cdfeaa2eb5e0735c/osv/purl_helpers.py#L18 for some mappings.

tools/osv-scanner/internal/output/purl_to_package.go
```diff
@@ -0,0 +1,35 @@
+package output
+
+import (
+	"github.com/package-url/packageurl-go"
+)
+
+var purl_ecosystems = map[string]string{
```
oliverchang, line 7: @another-rex this needs to be `purlEcosystems` :)

PR #729, opened 2022-09-20
RepoIndexResult: Fix type of hash.

oliverchang:
This should be a BlobProperty, not BlobKeyProperty.

Since this is in the Python model where writing isn't done, we don't have to migrate
any existing entities.


susperius (approved):
Thank you!

PR #731, opened 2022-09-20
Fix XSS with markdown rendering.

### LINKED ISSUE #730: Missing HTML escaping in advisory description
See https://osv.dev/vulnerability/GHSA-prc3-vjfx-vhm9 for example, the XSS example is actually interpreted as HTML and breaks the page.

oliverchang:
Fixes #730

### DIFF
--- a/gcp/appengine/frontend_handlers.py
+++ b/gcp/appengine/frontend_handlers.py
@@ -403,8 +403,8 @@ def group_versions(versions):
 def markdown(text):
   """Render markdown."""
   if text:
-    text = text.replace('<', '&lt;').replace('>', '&gt;')
-    return markdown2.markdown(text, extras=['fenced-code-blocks'])
+    return markdown2.markdown(
+        text, safe_mode='escape', extras=['fenced-code-blocks'])
 
   return ''
 
--- a/gcp/appengine/frontend_handlers.py
+++ b/gcp/appengine/frontend_handlers.py
@@ -403,6 +403,7 @@ def group_versions(versions):
 def markdown(text):
   """Render markdown."""
   if text:
+    text = text.replace('<', '&lt;').replace('>', '&gt;')
     return markdown2.markdown(text, extras=['fenced-code-blocks'])
 
   return ''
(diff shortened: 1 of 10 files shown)


gcp/appengine/frontend_handlers.py
```diff
@@ -403,6 +403,7 @@ def group_versions(versions):
 def markdown(text):
   """Render markdown."""
   if text:
+    text = text.replace('<', '&lt;').replace('>', '&gt;')
     return markdown2.markdown(text, extras=['fenced-code-blocks'])
```
another-rex, line 407: Rather than manually replacing the tags, use markdown2's safe_mode option:
```
markdown2.markdown(text, safe_mode = 'escape', extras=...)
```

another-rex: I can't really find any documentation for it, but here it is in the source code:
https://github.com/trentm/python-markdown2/blob/master/lib/markdown2.py#L1335

another-rex (changes requested):

gcp/appengine/frontend_handlers.py
andrewpollock: +1, this seems preferable and slightly more comprehensive, looking at the code. Good find @another-rex

oliverchang: Thanks, couldn't find any documentation on this before.

I had avoided things like html.escape before because I wanted to preserve things like `&amp;` wrt rendering (and not getting that double escaped), but maybe that's also unsafe in some way I haven't considered.

oliverchang: Done!

another-rex (approved):

gcp/appengine/frontend_handlers.py
another-rex: It seems like it's safe for `&` to not be escaped as long as we are not putting it in any javascript lines, which we are not. Are there examples of the potential double escaping?

oliverchang: No examples I can find. We can revisit this if this ever comes up, but either way it's not a huge deal.

PR #732, opened 2022-09-20
Server: DetermineVersion changes

susperius:
This change adds commit preselection functionality to avoid comparing against all commits for the dependency and also enable lookups without a package.

For the protobuf version 21.0 hashes the change reduces the lookup time from ~ 1:14 minutes to 54 seconds.


gcp/api/server.py
```diff
@@ -111,29 +111,45 @@ def QueryAffectedBatch(self, request, context):
...
-def get_version_by_name(
-    version_query: osv_service_v1_pb2.VersionQuery) -> ndb.Future:
-  """Identifies the version based on the provided name."""
-  query = osv.RepoIndex.query(osv.RepoIndex.name == version_query.name)
-  it = query.iter()
+def determine_version(version_query: osv_service_v1_pb2.VersionQuery,
+                      context: grpc.ServicerContext) -> ndb.Future:
+  """Identify fitting commits based on a subset of hashes"""
+  hashes = []
+  for i in range(min(20, len(version_query.file_hashes))):
```
oliverchang, line 122: can you put 20 into a constant? 

also, this can be simpler as:

```python
hashes = [f.hash in f in version_query.file_hashes[:_MAX_HASHES_TO_TRY]]
```

Or we can use `random.sample` as well: https://docs.python.org/3/library/random.html#random.sample instead of just taking the first 20.

gcp/api/server.py
```diff
@@ -111,29 +111,45 @@ def QueryAffectedBatch(self, request, context):
...
+      r = it.next()
+      idx_keys_set.add(r.key.parent())
+  if len(idx_keys_set) == 0:
+    context.abort(grpc.StatusCode.NOT_FOUND, 'no matches found')
+    return None
+
   futures = []
-  while (yield it.has_next_async()):
-    idx = it.next()
+  for r in idx_keys_set:
```
oliverchang, line 138: Nit: we can use ndb.get_multi() to save on some queries.

oliverchang, line 138: Can we filter this down even more? i.e. to consider a `RepoIndex` at all, it needs to match **all** 20 hashes, rather than a single one. 

To do this, for the above, I think we can just track an intersection of candidates, rather than just `add`ing to a set.

We can also make this 20 number dynamic depending on the number of provided hashes.

gcp/api/server.py
```diff
@@ -111,29 +111,45 @@ def QueryAffectedBatch(self, request, context):
...
+def determine_version(version_query: osv_service_v1_pb2.VersionQuery,
+                      context: grpc.ServicerContext) -> ndb.Future:
+  """Identify fitting commits based on a subset of hashes"""
+  hashes = []
+  for i in range(min(20, len(version_query.file_hashes))):
+    hashes.append(version_query.file_hashes[i].hash)
+  idx_keys_set = set()
+  for h in hashes:
+    query = osv.RepoIndexResult.query(
+        osv.RepoIndexResult.file_results.hash == h)
```
oliverchang, line 127: We can likely structure this in a better way to take advantage of the async querying such that we can actually run these queries in parallel.  

Right now we're doing this sequentially because we're waiting for each query to complete before we proceed to the next hash.

oliverchang (commented):
nice!! some suggestions. We need to shave this down more. 

Also, it may be good at some point to start some profiling to see where the time is being spent the most on in these queries.

Do you also have some command line utility you're using to test this locally?

gcp/api/server.py
susperius: Sure, thanks!

I played around with random sampling. It doesn't seem like it is improving the time needed. However it strongly depends on the ordering of hashes I assume and with some of the other changes could be an improvement. However,  it might have negative effects when picking common 3p deps adding a lot of overlap.

It looks like it provides a good balance. Since always taking the first X hashes can bring problems too.

Everything below is for protobuf v21.0:

First 20 hashes:
```
zsh grpc_cmd.sh  0.01s user 0.03s system 0% cpu 12.126 total
zsh grpc_cmd.sh  0.02s user 0.03s system 0% cpu 11.516 total
zsh grpc_cmd.sh  0.03s user 0.01s system 0% cpu 10.613 total
zsh grpc_cmd.sh  0.02s user 0.02s system 0% cpu 11.051 total
zsh grpc_cmd.sh  0.02s user 0.03s system 0% cpu 10.430 total
zsh grpc_cmd.sh  0.00s user 0.05s system 0% cpu 10.405 total
zsh grpc_cmd.sh  0.01s user 0.05s system 0% cpu 10.534 total
zsh grpc_cmd.sh  0.02s user 0.02s system 0% cpu 10.765 total
zsh grpc_cmd.sh  0.02s user 0.02s system 0% cpu 10.961 total
zsh grpc_cmd.sh  0.01s user 0.04s system 0% cpu 10.784 total
```

100 random hashes:
```
zsh grpc_cmd.sh  0.01s user 0.03s system 0% cpu 7.781 total
zsh grpc_cmd.sh  0.01s user 0.04s system 0% cpu 23.819 total
zsh grpc_cmd.sh  0.02s user 0.04s system 1% cpu 5.258 total
zsh grpc_cmd.sh  0.04s user 0.01s system 0% cpu 24.728 total
zsh grpc_cmd.sh  0.03s user 0.02s system 0% cpu 5.334 total
zsh grpc_cmd.sh  0.01s user 0.04s system 0% cpu 29.863 total
zsh grpc_cmd.sh  0.02s user 0.03s system 0% cpu 29.197 total
zsh grpc_cmd.sh  0.03s user 0.03s system 0% cpu 28.342 total
zsh grpc_cmd.sh  0.01s user 0.04s system 1% cpu 5.052 total
zsh grpc_cmd.sh  0.03s user 0.02s system 1% cpu 5.147 total
```

50 random hashes:
```
zsh grpc_cmd.sh  0.02s user 0.03s system 0% cpu 26.184 total
zsh grpc_cmd.sh  0.03s user 0.02s system 0% cpu 21.110 total
zsh grpc_cmd.sh  0.01s user 0.05s system 1% cpu 4.614 total
zsh grpc_cmd.sh  0.01s user 0.05s system 1% cpu 4.447 total
zsh grpc_cmd.sh  0.03s user 0.03s system 1% cpu 4.237 total
zsh grpc_cmd.sh  0.03s user 0.02s system 0% cpu 24.868 total
zsh grpc_cmd.sh  0.03s user 0.02s system 1% cpu 3.625 total
zsh grpc_cmd.sh  0.01s user 0.05s system 0% cpu 24.234 total
zsh grpc_cmd.sh  0.02s user 0.03s system 0% cpu 21.577 total
zsh grpc_cmd.sh  0.02s user 0.03s system 0% cpu 22.538 total
```

gcp/api/server.py
susperius: Yes, sorry this was only a first iteration of improvements.

gcp/api/server.py
susperius: PTAL

oliverchang (approved):

oliverchang:
Thanks! Could you please share `grpc_cmd.sh` ?

PR #738, opened 2022-09-23
Add OSV generation to NVD CVE parsing code

andrewpollock:
This is another in-progress prototype snapshot.

We're at a point here where OSV output is successfully generated. Additional validation to determine if it's "good enough" needs to be performed.

Next steps:

Refactor the functions into a library and add tests
Add repository language inference heuristics for Github repositories
Add repository language inference heuristics for non-Github repositories
~~Derive a "Fixed" version from patch references (where the repository's language is determined to be C/C++)~~
~~Derive a "Fixed" version from inspecting discovered repositories (where the repository's language is determined to be C/C++)~~


another-rex (commented):
Nice, added some general comments. We might also want to look at whether we can combine the combine-to-osv/main.go and cpp/main.go into one application to avoid pulling in all the CVEs twice.

vulnfeeds/cpp/main.go
```diff
@@ -32,6 +38,28 @@ func IsRepoURL(url string) bool {
 	return re.MatchString(url)
 }
 
+// Returns the base repository URL
+func Repo(u string) (string, bool) {
+	parsedURL, err := url.Parse(u)
+	if err != nil {
+		Logger.Fatalf("%v", err)
+	}
+	// GitHub and GitLab URLs are structured one way
```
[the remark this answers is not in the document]
andrewpollock: Done.

vulnfeeds/cpp/main.go
```diff
@@ -32,6 +38,28 @@ func IsRepoURL(url string) bool {
...
+	if strings.Contains(parsedURL.Path, "commit") {
+		return fmt.Sprintf("%s://%s%s", parsedURL.Scheme, parsedURL.Hostname(), strings.Join(strings.Split(parsedURL.Path, "/")[0:3], "/")), true
+	}
+
+	// GitWeb URLs are structured another way
+	if parsedURL.Path == "/" && strings.Contains(parsedURL.RawQuery, "commit") {
```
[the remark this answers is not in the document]
andrewpollock: Done

vulnfeeds/cpp/main.go
```diff
@@ -140,9 +168,75 @@ func MaybeGetSourceRepoFromDebian(mdir string, pkg string) string {
...
+func InScopeGitRepo(repoURL string) bool {
+	// TODO(apollock): Implement
+	return true
+}
+
+// Looks at what the repo to determine if it contains code using an in-scope language
+func InScopeRepo(repoURL string) bool {
+	parsedURL, err := url.Parse(repoURL)
+	if err != nil {
+		return false
```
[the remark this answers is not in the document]
andrewpollock: Done

vulnfeeds/cpp/main.go
```diff
@@ -140,9 +168,75 @@ func MaybeGetSourceRepoFromDebian(mdir string, pkg string) string {
...
+		Logger.Fatalf("Failed to create dir: %v", err)
+	}
+	outputFile := filepath.Join(vulnDir, v.ID+extension)
+
+	f, err := os.Create(outputFile)
+	if err != nil {
+		Logger.Fatalf("Failed to open %s for writing: %v", outputFile, err)
+	}
+	defer f.Close()
+	err = v.ToYAML(f)
```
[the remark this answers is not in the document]
andrewpollock: @oliverchang what are your thoughts here? I was plagiarising heavily from https://github.com/google/osv.dev/blob/728fdaaebb60609f3df3efbbf23af0f238d3e262/vulnfeeds/cmd/pypi/main.go#L171 and https://github.com/google/osv.dev/blob/05d2a3d3add1dc8ceffcc5da3f69f0030c43b70f/vulnfeeds/cmd/ids/main.go#L108

vulnfeeds/cpp/main.go
```diff
@@ -140,9 +168,75 @@ func MaybeGetSourceRepoFromDebian(mdir string, pkg string) string {
...
+		Logger.Fatalf("Can't generate an OSV record for %s without CPE data", cve.CVE.CVEDataMeta.ID)
+	}
+	versions, versionNotes := cves.ExtractVersion(cve)
+	v, notes := vulns.FromCVE(cve.CVE.CVEDataMeta.ID, cve)
+	pkgInfo := vulns.PackageInfo{
+		PkgName:   CPE.Product,
+		Ecosystem: CPE.Product,
+	}
+	v.AddPkgInfo(pkgInfo)
+	notes = append(notes, versionNotes...)
```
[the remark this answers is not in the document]
andrewpollock: Done

vulnfeeds/cpp/main.go
```diff
@@ -167,8 +261,9 @@ func main() {
 
 	for _, cve := range parsed.CVEItems {
 		refs := cve.CVE.References.ReferenceData
```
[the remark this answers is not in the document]
andrewpollock: Agreed. It's getting a bit long. I'll refactor this this week.

vulnfeeds/cves/versions.go
```diff
@@ -257,6 +257,78 @@ func ExtractVersionInfo(cve CVEItem, validVersions []string) (VersionInfo, []str
 	return v, notes
 }
 
+func ExtractVersion(cve CVEItem) (VersionInfo, []string) {
```
oliverchang, line 260: As discussed, rather than duplicating all of these, let's just change the existing `ExtractVersionInfo` function to allow passing an empty `validVersions`.

vulnfeeds/cpp/main.go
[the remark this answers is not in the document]
oliverchang: Yes let's go with JSON for this use case. YAML is used for Python because it's intended to be human editable.

vulnfeeds/cpp/main.go
```diff
@@ -178,54 +278,100 @@ func main() {
 		}
 
 		// Does it have any application CPEs?
-		appCpes := 0
+		appCpeCount := 0
```
oliverchang, line 281: s/Cpe/CPE/

vulnfeeds/cpp/main.go
```diff
@@ -32,6 +38,33 @@ func IsRepoURL(url string) bool {
...
+// Returns the base repository URL
+func Repo(u string) (string, bool) {
+	parsedURL, err := url.Parse(u)
+	if err != nil {
+		Logger.Fatalf("%v", err)
+	}
+	// GitHub and GitLab URLs are structured one way, e.g.
+	// https://github.com/MariaDB/server/commit/b1351c15946349f9daa7e5297fb2ac6f3139e4a8
+	// https://gitlab.freedesktop.org/virgl/virglrenderer/-/commit/b05bb61f454eeb8a85164c8a31510aeb9d79129c
+	// https://gitlab.com/qemu-project/qemu/-/commit/4367a20cc4
```
oliverchang, line 50: I think there's also ones like `https://github.com/google/osv.dev/pull/738` in these sometimes. We should make this work too.

vulnfeeds/cpp/main.go
```diff
@@ -140,9 +173,75 @@ func MaybeGetSourceRepoFromDebian(mdir string, pkg string) string {
...
+// Takes an NVD CVE record and outputs an OSV file in the specified directory.
+func CVEToOSV(cve cves.CVEItem, directory string) {
+	CPEs := cves.CPEs(cve)
+	CPE, ok := cves.ParseCPE(CPEs[0])
+	if !ok {
+		Logger.Fatalf("Can't generate an OSV record for %s without CPE data", cve.CVE.CVEDataMeta.ID)
+	}
+	versions, _ := cves.ExtractVersion(cve)
+	v, _ := vulns.FromCVE(cve.CVE.CVEDataMeta.ID, cve)
+	pkgInfo := vulns.PackageInfo{
```
oliverchang, line 213: Let's just not have this package information at all.

vulnfeeds/cpp/main.go
andrewpollock: Done.

vulnfeeds/cpp/main.go
andrewpollock: Done.

vulnfeeds/cpp/main.go
andrewpollock: Done.

vulnfeeds/cpp/main.go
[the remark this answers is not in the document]
andrewpollock: Ack. Doing this in https://github.com/google/osv.dev/pull/748 to stop this PR getting too hairy.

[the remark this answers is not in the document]
andrewpollock: Done: https://github.com/google/osv.dev/pull/738/commits/6a0a65bfaeece02fce821ba68140e239e75ea433

vulnfeeds/cves/versions.go
andrewpollock: I've completely excised the call to this because we're focusing on commits and not versions, so this need is now removed.

vulnfeeds/cpp/main.go
andrewpollock: Oh I remember why I was doing this. AddPkgInfo has a side-effect of adding the Affected field to the vulnerability...

It seems to have a lot of commit-related logic as well. Am I better off modifying it to support the PackageInfo struct not having ecosystem and PURL specified?

oliverchang: Yep, let's refactor things here to have the end result we want.

andrewpollock: Yeah now I have my head around the schema creation, I've managed to get the outcome I needed.

vulnfeeds/cpp/main.go
```diff
@@ -140,9 +194,72 @@ func MaybeGetSourceRepoFromDebian(mdir string, pkg string) string {
...
+		return InScopeGitRepo(repoURL)
+	}
+}
+
+// Takes an NVD CVE record and outputs an OSV file in the specified directory.
+func CVEToOSV(CVE cves.CVEItem, repo, directory string) {
+	CPEs := cves.CPEs(CVE)
+	CPE, ok := cves.ParseCPE(CPEs[0])
+	if !ok {
+		Logger.Fatalf("Can't generate an OSV record for %s without CPE data", CVE.CVE.CVEDataMeta.ID)
```
oliverchang, line 230: Can address in a follow up PR: Generally we prefer returning errors and letting the caller handle this rather than erroring out in Go.

vulnfeeds/cpp/main.go
```diff
@@ -167,65 +284,112 @@ func main() {
...
-				if tag == "Patch" && IsRepoURL(ref.URL) {
+				if tag == "Third Party Advisory" {
+					continue
+				}
+				// Alternative to the above:
+				// if tag == "Patch" && IsRepoURL(ref.URL) {
+				if IsRepoURL(ref.URL) {
 					Logger.Infof("\t * %s", ref.URL)
-					patchRefs += 1
+					// CVE entries have one set of references, but can have multiple CPEs
```
oliverchang, line 327: This is a lot of nesting and getting hard to read.. Consider putting some of this stuff (either the code here or the entire for loop) into its own function.

vulnfeeds/cpp/main.go
```diff
@@ -167,65 +284,112 @@ func main() {
...
+					for _, CPEstr := range cves.CPEs(cve) {
+						CPE, ok := cves.ParseCPE(CPEstr)
+						if !ok {
+							Logger.Infof("Failed to parse CPE %s: %v", CPEstr, err)
+							continue
+						}
+						// Avoid unnecessary calls to Repo() if we already have the repo
+						if _, ok := repos[CPE.Product]; !ok {
```
oliverchang, line 335: I wonder if we ever get clashes here if we only key on the product. . Why don't we use the full Vendor + Product combo here?

oliverchang (approved):
nice! LGTM with some followups that we can address in another PR

vulnfeeds/cpp/main.go
andrewpollock: Ack. Bear in mind that this is still proof-of-concept and from talking with @another-rex the plan of record is to move a lot of the functions from this into a library and then extend `vulnfeeds/cmd/combine-to-osv/main.go` to also do what's being done with this.

vulnfeeds/cpp/main.go
andrewpollock: Yep, I'll look into that.

PR #742, opened 2022-09-26
Slight change to schedule wording in renovate.json

another-rex:
I believe the old wording is actually incorrect

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -4,7 +4,7 @@
     "config:base"
   ],
   "timezone": "Australia/Sydney",
-  "schedule": ["before 6am every monday"],
+  "schedule": ["before 6am on monday"],
   "packageRules": [
     {
       "matchUpdateTypes": ["major"],


andrewpollock:
I'm assuming that the "why" is that the previous wording was invalid and wasn't working. Is there some way to validate this configuration?

andrewpollock (approved):

another-rex:
Found this linked in the renovate docs (though not in the scheduling section): https://codepen.io/rationaltiger24/full/ZExQEgK

It believes every to mean everyday, no matter what day is after the word `every`.

However, it also seems to consider configurations valid that the renovate bot validator does not, so I'm not sure if I should be relying on this. 

e.g.  `before 6am every abcdefgh` is valid in the codepen, but not with the renovate bot validator.

oliverchang (approved):

oliverchang (approved):

PR #754, opened 2022-09-28
A few documentation updates

andrewpollock:
- Use a better outlink for the OSV schema
- Add some more sources OSV uses, and sort the list
- General reformatting

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,9 @@ Together, these include vulnerabilities from:
 -   Android
 -   crates.io
 -   Debian GNU/Linux
+-   GitHub Actions
 -   Go
+-   Hex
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -48,16 +48,20 @@ The following ecosystems have vulnerabilities encoded in this format:
...
+-   Android
+-   crates.io
+-   Debian GNU/Linux
+-   GitHub Actions
 -   Go
+-   Hex
+-   Linux kernel
+-   Maven
+-   npm
 -   NuGet
```
oliverchang, line 60: I think we're missing "Pub" ?

andrewpollock: Added

PR #756, opened 2022-09-28
Use the tags as the first port of call to infer the type of the link

andrewpollock:
The lovely folks at the NVD are already giving us hints, so only resort to URL shape-based guesswork if we don't have a more explicit signal.

### DIFF
--- a/vulnfeeds/cves/cve.go
+++ b/vulnfeeds/cves/cve.go
@@ -24,20 +24,23 @@ type CVE struct {
 	CVEDataMeta struct {
 		ID string
 	} `json:"CVE_data_meta"`
-	References struct {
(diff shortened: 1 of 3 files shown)


andrewpollock:
For scenarios where one URL has multiple tags, is it appropriate for the same URL to appear multiple times with different types? At the moment, it's the first match that wins, which is a little random and order-dependent as to what the result is, which isn't going to be ideal.

oliverchang (approved):
thanks!

andrewpollock:
@oliverchang this changed pretty significantly since you LGTMed it, would you mind taking another look?

oliverchang (approved):

PR #759, opened 2022-09-29
Better importer performance

### LINKED ISSUE #576: Improve importer performance.

another-rex:
Partially fixes #576, still need to solve the initial GHSA cloning speed issue for new instances. 

This improves the incremental updates by multithreading the retrieval of bucket entries from google cloud store, and batching the datastore requests.

### DIFF
--- a/docker/importer/importer.py
+++ b/docker/importer/importer.py
@@ -251,8 +251,10 @@ class Importer:
 
     return changed_entries, deleted_entries
 
-  def _process_updates_git(self, source_repo):
+  def _process_updates_git(self, source_repo: osv.SourceRepository):
     """Process updates for a git source_repo."""
+    logging.info("Begin processing git: %s", source_repo.name)
+
     repo = self.checkout(source_repo)
 
     # Get list of changed files since last sync.
@@ -307,10 +309,13 @@ class Importer:
     source_repo.last_synced_hash = str(repo.head.target)
     source_repo.put()
 
-  def _process_updates_bucket(self, source_repo):
+    logging.info("Finish processing git: %s", source_repo.name)
+
+  def _process_updates_bucket(self, source_repo: osv.SourceRepository):
     """Process updates from bucket."""
     # TODO(ochang): Use Pub/Sub change notifications for more efficient
     # processing.
+    logging.info("Begin processing bucket: %s", source_repo.name)
 
     ignore_last_import_time = source_repo.ignore_last_import_time
     if ignore_last_import_time:
@@ -322,8 +327,8 @@ class Importer:
       if not _is_vulnerability_file(source_repo, blob.name):
         continue
 
-      logging.info('Bucket entry triggered for for %s/%s', source_repo.bucket,
-                   blob.name)
+      logging.debug('Bucket entry triggered for %s/%s', source_repo.bucket,
+                    blob.name)
       blob_bytes = blob.download_as_bytes()
       try:
         vulnerabilities = osv.parse_vulnerabilities_from_data(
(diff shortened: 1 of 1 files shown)


docker/importer/importer.py
```diff
@@ -307,50 +315,82 @@ def _process_updates_git(self, source_repo):
...
+        conv_func_partial = functools.partial(convert_blob_to_vuln, source_repo)
+        converted_vulns = processing_pool.map(conv_func_partial, batch_blob)
+        # Flatten list[list[]] to list[]
+        vulnerabilities = [
+            inner for outer in converted_vulns for inner in outer
+        ]
+
+        # Batch query the bugs
+        vuln_ids = [x.id for x, _, _ in vulnerabilities]
+        existing_vulns = osv.Bug.query(osv.Bug.db_id.IN(vuln_ids))
```
oliverchang, line 359: What about ndb initialization? According to this: https://googleapis.dev/python/python-ndb/latest/migrating.html#setting-up-a-connection

it's not thread safe.

docker/importer/importer.py
```diff
@@ -307,50 +315,82 @@ def _process_updates_git(self, source_repo):
...
-      if not ignore_last_import_time and all(
-          unchanged_vuln_exist(vuln) for vuln in vulnerabilities):
-        logging.info('Skipping updates for %s as modified date unchanged.',
-                     blob.name)
-        continue
-
-      original_sha256 = osv.sha256_bytes(blob_bytes)
-      self._request_analysis_external(source_repo, original_sha256, blob.name)
+    listed_blob_names = [
+        blob.name for blob in (storage_client.list_blobs(source_repo.bucket))
```
oliverchang, line 334: Do you need the parathenses around the `list_blobs` call ?

docker/importer/importer.py
```diff
@@ -37,6 +41,8 @@
 _OSS_FUZZ_EXPORT_BUCKET = 'oss-fuzz-osv-vulns'
 _EXPORT_WORKERS = 32
 _NO_UPDATE_MARKER = 'OSV-NO-UPDATE'
+_BUCKET_BATCH_SIZE = 50
+_BUCKET_THREAD_POOL = 10
```
oliverchang, line 45: s/_BUCKET_THREAD_POOL/_BUCKET_THREAD_COUNT/ instead to make it clearer this is a count/number and not an actual thread pool object.

docker/importer/importer.py
```diff
@@ -307,50 +315,82 @@ def _process_updates_git(self, source_repo):
...
-        logging.info('Skipping updates for %s as modified date unchanged.',
-                     blob.name)
-        continue
-
-      original_sha256 = osv.sha256_bytes(blob_bytes)
-      self._request_analysis_external(source_repo, original_sha256, blob.name)
+    listed_blob_names = [
+        blob.name for blob in (storage_client.list_blobs(source_repo.bucket))
+    ]
+    batched_blob_names = [
```
oliverchang, line 336: How much does batching get us if we're doing multithreading anyway? Can we just increase the multithreading count to get the same perf? 

Also I'm pretty sure a `.IN` query in ndb just runs parallel queries under the hood, so we're not actually saving on # of requests if that's what we were after. 

If it doesn't buy us much, I reckon let's just do the multithreading and keep the code a lot simpler.

oliverchang (commented):
nice! some comments.

docker/importer/importer.py
another-rex: This is still ran on the main thread I believe. The only part that is multithreaded is downloading blobs from the google cloud bucket. Each client there is initialized per thread.

docker/importer/importer.py
another-rex: Did a test, for about 50 entries, it takes roughly double the time of .IN, but it's still a lot faster than running one by one.

oliverchang: Weird, there must be some additional optimization happening under the hood for .IN. Let's start with this since it's already a big improvement.

another-rex: Now using the multithreading code instead of .IN, it does simplify the code quite a bit, so it's probably worth it in terms of complexity to performance tradeoff.

another-rex:
Made some substantial changes, including now using `concurrent.futures.ThreadPoolExecutor` instead of `multiprocessor.Pool`, improving efficiency and performance since we are no longer creating 10 separate python processes. 

This also allowed me to move some functions around so that the relevant logic is grouped in one place. Having to initialize `ndb client `and `storage client` for each thread does take a few seconds (3-5 seconds for 10 threads), so that's probably where using .IN makes it a lot faster by avoiding the setup, but overall this should be a huge performance improvement (guessing about 8x?) for importing bucket entries. We will have to see actual numbers in production.

docker/importer/importer.py
```diff
@@ -307,50 +314,75 @@ def _process_updates_git(self, source_repo):
...
+    # Setup storage client
+    def thread_init():
+      _client_store.storage_client = storage.Client()
+      _client_store.ndb_client = ndb.Client()
+
+    with ThreadPoolExecutor(
+        _BUCKET_THREAD_COUNT, initializer=thread_init) as executor:
+      converted_vulns = executor.map(convert_blob_to_vuln, listed_blob_names)
+      for cv in converted_vulns:
+        if cv is not None:
```
oliverchang, line 377: do we need the explicit None check as opposed to implicit false?

oliverchang (approved):
nice!

PR #765, opened 2022-10-04
Git clone shallow to improve new importer instance performance

### LINKED ISSUE #576: Improve importer performance.

another-rex:
A shallow clone that only clones commits up to the previous synced date. This avoids having to clone the entire repository history when unneeded. A pull/fetch will only fetch future commits, without needing to fill up the entire history. 

A potential issue is if default branch changes to a completely different branch (not just renamed), this will mean potentially needing more history than what is available since the last sync time. This is very rare and I am not aware of any reason for doing this.

Fixes #576

### DIFF
--- a/osv/repos.py
+++ b/osv/repos.py
@@ -93,7 +93,6 @@ def clone(git_url: str,
     :param git_callbacks: TODO:
     :param last_update_date: Optional python datetime object used to specify
       the date of the shallow clone.
-
   """
   # Use 'git' CLI here as it's much faster than libgit2's clone.
   env = {}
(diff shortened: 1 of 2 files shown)


osv/repos.py
```diff
@@ -81,7 +81,7 @@ def _checkout_branch(repo, branch):
   repo.reset(remote_branch.target, pygit2.GIT_RESET_HARD)
 
 
-def clone(git_url, checkout_dir, git_callbacks=None):
+def clone(git_url, checkout_dir, git_callbacks=None, last_update_date=None):
```
andrewpollock, line 84: Could you:

at a minimum, add a docstring that describes the parameters, including the expected format of this date
preferably, add some input validation for `last_update_date`

another-rex: Done, added type definitions to the clone function and it's callers, also added docstrings describing each parameter.

another-rex: Can you clarify what type of input validation you think are good here? I'm not sure what I can implement for this.

oliverchang (approved):
thanks!

osv/repos.py
andrewpollock: I think the type checking you now have will be sufficient

andrewpollock (approved):

PR #773, opened 2022-10-11
Add pagination support

andrewpollock:
Also fixed a few initialisms consistently with existing code

### DIFF
--- a/vulnfeeds/cmd/download-cves/main.go
+++ b/vulnfeeds/cmd/download-cves/main.go
@@ -100,12 +100,12 @@ func downloadCVE2(APIKey string, CVEPath string) {
 	if err != nil { // There's an existing file, check if it matches server file
 		Logger.Fatalf("Something went wrong when creating/opening file: %+v", err)
 	}
-	completeNVDDataset := cves.NVDCVE2{}
+	var vulnerabilities []json.RawMessage
 	page := cves.NVDCVE2{}
 	offset := 0
 	for {
 		page = downloadCVE2WithOffset(APIKey, offset)
-		completeNVDDataset.Vulnerabilities = append(completeNVDDataset.Vulnerabilities, page.Vulnerabilities...)
+		vulnerabilities = append(vulnerabilities, page.Vulnerabilities...)
 		offset += PageSize
 		if offset > *page.TotalResults {
 			break
@@ -113,7 +113,7 @@ func downloadCVE2(APIKey string, CVEPath string) {
 		time.Sleep(6)
 	}
 	// Make this look like one giant page of results from the API call
-	page.Vulnerabilities = completeNVDDataset.Vulnerabilities
+	page.Vulnerabilities = vulnerabilities
 	*page.StartIndex = 0
 	page.ResultsPerPage = page.TotalResults
 	err = page.ToJSON(file)
--- a/vulnfeeds/cmd/download-cves/main.go
+++ b/vulnfeeds/cmd/download-cves/main.go
@@ -106,11 +106,11 @@ func downloadCVE2(APIKey string, CVEPath string) {
 	offset := 0
 	for {
 		page = downloadCVE2WithOffset(APIKey, offset)
-		if len(page.Vulnerabilities) == 0 {
-			break
-		}
 		completeNVDDataset.Vulnerabilities = append(completeNVDDataset.Vulnerabilities, page.Vulnerabilities...)
 		offset += PageSize
+		if offset > page.TotalResults {
+			break
+		}
 		time.Sleep(6)
 	}
 	// completeNVDDataset starts out completely uninitialised
--- a/vulnfeeds/cmd/download-cves/main.go
+++ b/vulnfeeds/cmd/download-cves/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	"compress/gzip"
 	"context"
+	"encoding/json"
 	"flag"
 	"io"
 	"log"
@@ -12,6 +13,7 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/google/osv/vulnfeeds/cves"
 	"github.com/google/osv/vulnfeeds/utility"
 
 	"cloud.google.com/go/logging"
@@ -19,6 +21,7 @@ import (
 
 const (
 	cveURLBase     = "https://nvd.nist.gov/feeds/json/cve/1.1/"
+	nvdAPI         = "https://services.nvd.nist.gov/rest/json/cves/2.0"
 	fileNameBase   = "nvdcve-1.1-"
 	startingYear   = 2002
 	cvePathDefault = "cve_jsons"
@@ -26,6 +29,8 @@ const (
 )
 
 var Logger utility.LoggerWrapper
+var apiKey = flag.String("api_key", "", "API key for accessing NVD API 2.0")
+var cvePath = flag.String("cvePath", cvePathDefault, "Where to download CVEs to")
 
 func main() {
 	client, err := logging.NewClient(context.Background(), projectId)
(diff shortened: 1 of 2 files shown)


vulnfeeds/cmd/download-cves/main.go
```diff
@@ -74,26 +83,64 @@ func downloadCVE2(apiKey string, cvePath string) {
...
+	completeNVDdataset := cves.NVDCVE2{}
+	page := cves.NVDCVE2{}
+	offset := 0
+	for {
+		page = downloadCVE2WithOffset(APIKey, offset)
+		if len(page.Vulnerabilities) == 0 {
+			break
+		}
+		completeNVDdataset.Vulnerabilities = append(completeNVDdataset.Vulnerabilities, page.Vulnerabilities...)
+		offset += 2000
```
another-rex, line 112: Place page size in a const value

vulnfeeds/cmd/download-cves/main.go
```diff
@@ -74,26 +83,64 @@ func downloadCVE2(apiKey string, cvePath string) {
...
+	defer file.Close()
+	if err != nil { // There's an existing file, check if it matches server file
+		Logger.Fatalf("Something went wrong when creating/opening file: %+v", err)
+	}
+	completeNVDdataset := cves.NVDCVE2{}
+	page := cves.NVDCVE2{}
+	offset := 0
+	for {
+		page = downloadCVE2WithOffset(APIKey, offset)
+		if len(page.Vulnerabilities) == 0 {
```
another-rex, line 108: We can save a request here by placing the if break after 112, and checking to see if `offset` is > than the `page.TotalResults`

vulnfeeds/cmd/download-cves/main.go
```diff
@@ -74,26 +83,64 @@ func downloadCVE2(apiKey string, cvePath string) {
...
+			break
+		}
+		completeNVDdataset.Vulnerabilities = append(completeNVDdataset.Vulnerabilities, page.Vulnerabilities...)
+		offset += 2000
+		time.Sleep(6)
+	}
+	// completeNVDdataset starts out completely uninitialised
```
another-rex, line 115: We can simplify this by changing the type of `completeNVDdataset` to an array of vulnerabilities. So code from 115 to 127 can just consist of setting `page.Vulnerabilities = completeNVDdataset`, and calling `page.ToJSON(file)` instead.

vulnfeeds/cmd/download-cves/main.go
```diff
@@ -74,26 +83,64 @@ func downloadCVE2(apiKey string, cvePath string) {
...
+
+// Download all of the CVE data using the 2.0 API
+// See https://nvd.nist.gov/developers/vulnerabilities
+func downloadCVE2(APIKey string, CVEPath string) {
+	file, err := os.OpenFile(path.Join(CVEPath, "nvdcve-2.0.json.new"), os.O_CREATE|os.O_RDWR|os.O_TRUNC, 0644)
+	defer file.Close()
+	if err != nil { // There's an existing file, check if it matches server file
+		Logger.Fatalf("Something went wrong when creating/opening file: %+v", err)
+	}
+	completeNVDdataset := cves.NVDCVE2{}
```
another-rex, line 103: nit: completeNVDdataset should be completeNVDDataset

another-rex (changes requested):
Nice! just a few comments.

vulnfeeds/cmd/download-cves/main.go
andrewpollock: Fixed in https://github.com/google/osv.dev/pull/773/commits/14aa9d18967aa74c0a3e2ff8670d038603e3c47c

vulnfeeds/cmd/download-cves/main.go
andrewpollock: Fixed in https://github.com/google/osv.dev/pull/773/commits/22653e9cf824f96eb7098b48e782f598bc6e8f01

vulnfeeds/cmd/download-cves/main.go
andrewpollock: I'm not quite following, sorry.

Every page returned from the API call (and currently by `downloadCVE2WithOffset`) is a `cves.NVDCVE2` struct. Some of the metadata fields change with each subsequent page (i.e. the `StartIndex`, potentially the `TotalResults` if things are updated between calls). Definitely the `timestamp` changes.

The current approach disregards those values for the intermediate pages, only focusing on what we wind up with   after we've reached the end, because the final values are all that really matter.

I don't want each page's metadata winding up in the output, that way what gets emitted looks exactly like one page's worth of output, it's just one very big page...

vulnfeeds/cmd/download-cves/main.go
andrewpollock: Thanks. Fixed in https://github.com/google/osv.dev/pull/773/commits/1ccbd66e02c6bbb955918765d67e6c1c264c4afd and https://github.com/google/osv.dev/pull/773/commits/28fdb2ce9316469e1ef1d850ac4f372c8519d638

andrewpollock: Actually, this is now going to break prematurely, it's not going to run for the final partial page of results, right?

andrewpollock: No, never mind, I'm not thinking straight.

vulnfeeds/cmd/download-cves/main.go
another-rex: Yes, that makes sense. In the current code you are already replacing page every loop, so when the final loop finishes, the page variable will only contain the metadata for the very last page, (plus all the vulnerabilities for the final page in `page.Vulnerabilities). 

Now if we replace `page.Vulnerabilities` with the `completeNVDDataset`, which contains all the vulnerabilities concatenated together, you will have the final page's metadata, plus one very big page of vulnerabilities, similarly to what you are doing right now. 

Put another way, what is happening right now is you are filling up `completeNVDDataset.Vulnerabilities`, leaving everything else empty, then copying everything from `page` **except** the `Vulnerabilities` field over to `completeNVDDataset`. What I am suggesting is we can achieve the exact same result, by copying in the other direction, i.e. copying `completeNVDDataset.Vulnerabilties` into `page.Vulnerabilities`, and calling toJSON on `page` instead.

andrewpollock: Ah, thanks for clarifying, yes, that makes much more sense and reduces code. Good thinking!

Changed in https://github.com/google/osv.dev/pull/773/commits/ba10e2ab0c1677288177a158a6f59b01a5a0828a

vulnfeeds/cmd/download-cves/main.go
```diff
@@ -74,26 +84,68 @@ func downloadCVE2(apiKey string, cvePath string) {
...
+	for {
+		page = downloadCVE2WithOffset(APIKey, offset)
+		completeNVDDataset.Vulnerabilities = append(completeNVDDataset.Vulnerabilities, page.Vulnerabilities...)
+		offset += PageSize
+		if offset > *page.TotalResults {
+			break
+		}
+		time.Sleep(6)
+	}
+	// completeNVDDataset starts out completely uninitialised
```
another-rex, line 116: This section down to line 128 now can be removed I believe

vulnfeeds/cmd/download-cves/main.go
```diff
@@ -74,26 +84,68 @@ func downloadCVE2(apiKey string, cvePath string) {
...
+
+// Download all of the CVE data using the 2.0 API
+// See https://nvd.nist.gov/developers/vulnerabilities
+func downloadCVE2(APIKey string, CVEPath string) {
+	file, err := os.OpenFile(path.Join(CVEPath, "nvdcve-2.0.json.new"), os.O_CREATE|os.O_RDWR|os.O_TRUNC, 0644)
+	defer file.Close()
+	if err != nil { // There's an existing file, check if it matches server file
+		Logger.Fatalf("Something went wrong when creating/opening file: %+v", err)
+	}
+	completeNVDDataset := cves.NVDCVE2{}
```
another-rex, line 104: Since we are using just one field of NVDCVE2 (i.e. `Vulnerabilities`), might as well just make it an array of `Vulnerabilities`

another-rex (changes requested):

vulnfeeds/cmd/download-cves/main.go
andrewpollock: I could have sworn I had, but clearly I hadn't... Done in https://github.com/google/osv.dev/pull/773/commits/c2c8fc4de8e16e405a127ab7bc47ef7095ad3c22

vulnfeeds/cmd/download-cves/main.go
andrewpollock: Ah, now I understand what you were saying earlier. Done in https://github.com/google/osv.dev/pull/773/commits/cb8262e93c0cd4d084f887e1effc050b285d9202

another-rex (approved):
LGTM!

Issue #619, opened 2022-08-19
Generate the ecosystems.txt in the root


VinodAnandan:
Could you please generate the ecosystems.txt in the root that would contain a list of all the ecosystems currently in OSV, from which users can construct URLs to download all advisories for the ecosystems?

oliverchang:
We now have a `https://osv-vulnerabilities.storage.googleapis.com/ecosystems.txt` in the root. 

This still needs to be generated dynamically, but our ecosystems change infrequently so it should be fine to depend on this in the short/medium term.

PR #777, opened 2022-10-13
Make the exporter also export ecosystems.txt

andrewpollock:
Adjust the exporter so it can be passed a non-production bucket to export to

Make it possible to run locally under pipenv for manual testing purposes

Fixes: #619


another-rex (approved):
LGTM!

Issue #662, opened 2022-09-01
osv-scanner: Build exclusions/ignore format


oliverchang:
To allow vulnerabilities to be ignored.

PR #778, opened 2022-10-14
Add ability to filter/ignore specific vulnerabilities

another-rex:
Solves #662 

> By placing a `osv-scanner.toml` file in any parent directory of the file being
> scanned will be used to configure scanning of that file. This can be overridden
> by passing the `--config=/path/to/config` flag.
> 
> Currently, there is only 1 option to configure:
> ### Ignore vulnerabilities by ID
> Vulnerabilities can be marked as ignored by putting the vuln ID in an array
> under the `IgnoreVulnIds` key. 
> ```
> IgnoredVulnIds = [
>     "GO-2022-0968",
>     "GO-2022-1059"
> ]
> ```


tools/osv-scanner/cmd/osv-scanner/config.go
```diff
@@ -0,0 +1,49 @@
...
+)
+
+type Config struct {
+	IgnoredVulnIds []string
+}
+
+// TryLoadConfig tries to load config in `target` or any of it's parent dirs
+// `target` will be the key for the entry in configMap
+// Will shortcut and return "" if globalConfig is not nil
+func TryLoadConfig(target string, configMap map[string]Config) string {
```
andrewpollock, line 19: Looking at https://go.dev/doc/effective_go#errors (and go/go-style/best-practices#program-init internally), I think it's preferable to return an error or an ok boolean to main, rather than abending within this function.

Please adjust this to either return the configuration and an error or boolean indicating success.

tools/osv-scanner/cmd/osv-scanner/config.go
```diff
@@ -0,0 +1,49 @@
...
+		return ""
+	}
+	stat, err := os.Stat(target)
+	if err != nil {
+		log.Fatalf("Failed to stat path: %s", err)
+	}
+
+	if stat.IsDir() && !strings.HasSuffix(target, "/") {
+		// Make sure directories ends with '/'
+		target += "/"
```
andrewpollock, line 30: I can see benefit in aiming to make the scanner platform agnostic. To this end, consider using https://pkg.go.dev/path/filepath#pkg-constants rather than hardcoding Unix-specific filesystem constructs.

tools/osv-scanner/cmd/osv-scanner/config.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+
+	currentDir := target
+	for currentDir != "/" {
```
andrewpollock, line 34: I'm wondering if walking the tree to the root could result in a situation where another party with write access to an intermediate directory could insert a broader, maliciously configuration with the malicious intent of hiding a vulnerability? I may be overthinking things, and I suspect there's no specific threat model for what environments this scanner is intended to be run in...

andrewpollock: Furthermore, now that I've made it to main.go, it looks like this is intended to be run on a Git repo, so realistically walking past the Git repo root shouldn't be necessary, right?

tools/osv-scanner/cmd/osv-scanner/main.go
```diff
@@ -10,56 +10,67 @@ import (
...
-	"github.com/g-rath/osv-detector/pkg/lockfile"
 	"github.com/urfave/cli/v2"
+	"golang.org/x/exp/slices"
 )
 
+const osvScannerConfigName = "osv-scanner.toml"
+
+var globalConfig *Config
+
 // scanDir walks through the given directory to try to find any relevant files
```
andrewpollock, line 26: Please elaborate a little more in the docstring as to what's considered relevant.

tools/osv-scanner/cmd/osv-scanner/main.go
```diff
@@ -10,56 +10,67 @@ import (
...
-			_ = scanSBOMFile(query, path)
+			_ = scanSBOMFile(query, path, configMap)
 		}
 
 		return nil
 	})
 }
 
-func scanLockfile(query *osv.BatchedQuery, path string) error {
+func scanLockfile(query *osv.BatchedQuery, path string, configMap map[string]Config) error {
```
andrewpollock, line 64: Please add a docstring

tools/osv-scanner/cmd/osv-scanner/main.go
```diff
@@ -69,7 +80,8 @@ func scanLockfile(query *osv.BatchedQuery, path string) error {
 	return nil
 }
 
-func scanSBOMFile(query *osv.BatchedQuery, path string) error {
+func scanSBOMFile(query *osv.BatchedQuery, path string, configMap map[string]Config) error {
```
andrewpollock, line 83: Please add a docstring

andrewpollock (changes requested):
This is getting pretty long, consider refactoring all of the helper functions into a standalone library. Added benefit is other integrations can be made.

Please add an invalid configuration file to your test cases.

tools/osv-scanner/cmd/osv-scanner/config.go
another-rex: That's a good point, I originally had it going all the way up to the root because I thought someone might want to have some global configuration of osv-scanner in their home directory or similar, but that would be pretty niche and better accommodated by having some specific paths set up for where the config files will be located, rather than walking up to root. 

I'll change the behavior to scan only up to the directory initially asked to be scanned, or if no directory has been specified, only a lockfile, just look at the immediate directory that the lockfile is in.

How osv-scanner should load configs is probably worthy of meeting + design doc to clarify the goals.

osv-scanner.toml
```diff
@@ -0,0 +1,6 @@
+IgnoredVulnIds = [
```
oliverchang, line 1: nit: maybe VulnIDs for casing?

tools/osv-scanner/cmd/osv-scanner/config.go
```diff
@@ -0,0 +1,71 @@
...
+	"path/filepath"
+	"strings"
+
+	"github.com/BurntSushi/toml"
+)
+
+type LoadConfigErrorType int64
+
+const (
+	GlobalConfigSet LoadConfigErrorType = iota
```
oliverchang, line 15: these all seem to only be internally used. Can we either:

- make these private, or 
- just use plain strings directly via `errors.New`/`fmt.Errorf` instead? It seems like this distinction is not needed programatically anywhere anyway, so we can keep this simple.

tools/osv-scanner/cmd/osv-scanner/main.go
```diff
@@ -10,56 +10,73 @@ import (
...
+// scanLockfile will load, identify, and parse the lockfile path passed in, and add the dependencies specified
+// within to `query`, and the config files in the parent folders found to configMap
+func scanLockfile(query *osv.BatchedQuery, path string, configMap map[string]Config) error {
+	configPath, configErr := TryLoadConfig(path, configMap)
 	parsedLockfile, err := lockfile.Parse(path, "")
 	if err != nil {
 		return err
 	}
 	log.Printf("Scanned %s file and found %d packages", path, len(parsedLockfile.Packages))
+	if configErr == nil {
```
oliverchang, line 77: nit: should we move this to right after line 71? 

It makes the code easier to follow, and perhaps it's a more logical order for the printed output as well. 

That said, can we avoid complicating all of these `scan*` functions completely, and not have any of this code here. We can just load the necessary configs in `filterResponse` since we have all the needed source information there anyway. This makes the code less tied together and more modular.

tools/osv-scanner/cmd/osv-scanner/main.go
```diff
@@ -10,56 +10,73 @@ import (
...
 	"github.com/google/osv.dev/tools/osv-scanner/internal/sbom"
-
-	"github.com/g-rath/osv-detector/pkg/lockfile"
 	"github.com/urfave/cli/v2"
+	"golang.org/x/exp/slices"
 )
 
+const osvScannerConfigName = "osv-scanner.toml"
+
+var globalConfig *Config
```
oliverchang, line 24: Maybe rename this to `configOverride` or something similar? It wasn't clear to me what exactly this meant at first read, and I thought this referred to a config that exists in say `$HOME/.osv.config` or something similar rather than an explicit config someone passed in.

See below though, for a possible refactor to make all of this cleaner and avoid this global var completely.

tools/osv-scanner/cmd/osv-scanner/main.go
```diff
@@ -160,17 +194,52 @@ func scanDebianDocker(query *osv.BatchedQuery, dockerImageName string) {
...
 		query.Queries = append(query.Queries, pkgDetailsQuery)
+		packages += 1
 	}
-	log.Printf("Scanned docker image")
+	log.Printf("Scanned docker image with %d packages", packages)
+}
+
+// Filters response according to config, returns number of responses removed
+func filterResponse(query osv.BatchedQuery, resp *osv.BatchedResponse, globalConfig *Config, configMap map[string]Config) int {
+	//response := []osv.MinimalResponse{}
```
oliverchang, line 204: delete?

tools/osv-scanner/cmd/osv-scanner/main.go
```diff
@@ -160,17 +194,52 @@ func scanDebianDocker(query *osv.BatchedQuery, dockerImageName string) {
...
+			if ok {
+				configToUse = &configToUseTemp
+			}
+		}
+		if configToUse != nil {
+			for _, vuln := range result.Vulns {
+				if slices.Contains(configToUse.IgnoredVulnIds, vuln.ID) {
+					hiddenVulns[vuln.ID] = struct{}{}
+				} else {
+					filteredVulns = append(filteredVulns, vuln)
```
oliverchang, line 224: nit: log a line saying it was ignored?

tools/osv-scanner/cmd/osv-scanner/config.go
```diff
@@ -0,0 +1,71 @@
...
+	case NoConfigFound:
+		return "No config file found on this or any ancestor path: " + e.targetPath
+	}
+	panic("Invalid error type")
+}
+
+// TryLoadConfig tries to load config in `target` or any of it's parent dirs
+// `target` will be the key for the entry in configMap
+// Will shortcut and return "" if globalConfig is not nil
+func TryLoadConfig(target string, configMap map[string]Config) (string, error) {
```
oliverchang, line 41: Can we hide the details of this `configMap` completely? And make this a more modular struct with methods? As is, users (e.g. the `filterResponse` function in the other file) have to do complicated logic to use and understand this structure, when ideally they should just call a single function. 

i.e. something like: 

```golang 

type Config struct {
 ...
}

type ConfigManager struct {
  configMap map[...]...
   ...
}

func (c *Config) ShouldIgnore(vuln) bool {
}

func (c *ConfigManager) UseOverride(configPath string) {
   ...
}

func (c *ConfigManager) Get(targetPath string) Config {
   ... // caches things in configMap if needed. 
}
```

oliverchang (commented):
nice! some refactoring comments :)

G-Rath:
Just want to flag that this functionality already exists in `osv-detector`, since @oliverchang and I have been talking about getting that merged into here, and I'm keen to maintain parity with the detector since its being used by automated tooling.

I've not dug too heavily into this PR, but it doesn't seem like anything radically different is being proposed other than that you're using TOML rather than YAML which I'd like to support since languages like Ruby can parse that by default (but its easy to support both), but you might find it useful to check the detectors implementation if you haven't already :)

tools/osv-scanner/cmd/osv-scanner/main.go
another-rex: Initially the idea is that the config will have other information that will affect the scanning part. But you're right we should add that when we need it, I've moved it all to `filterResponse`, which simplifies the code quite a bit.

another-rex:
@G-Rath Thanks for pointing this out!

Other than the YAML and TOML difference, the only main difference for feature parity I can see is that this allows additionally specifying an expiry date + some reason string for each ignored entry.

tools/osv-scanner/cmd/osv-scanner/config.go
```diff
@@ -0,0 +1,115 @@
...
+	if index == -1 {
+		return false, IgnoreLine{}
+	}
+	ignoredLine := c.IgnoredVulns[index]
+	if ignoredLine.IgnoreUntil.IsZero() {
+		// If IgnoreUntil is not set, should ignore.
+		return true, ignoredLine
+	}
+	// Should ignore if IgnoreUntil is still after current time
+	return ignoredLine.IgnoreUntil.After(time.Now().UTC()), ignoredLine
```
oliverchang, line 45: `IgnoreUntil` also gets parsed in as UTC right? what happens when someone provides a non-UTC timestamp?

tools/osv-scanner/cmd/osv-scanner/config.go
```diff
@@ -0,0 +1,115 @@
...
+
+// Attempts to get the config
+func (c *ConfigManager) Get(targetPath string) Config {
+	if c.overrideConfig != nil {
+		return *c.overrideConfig
+	}
+
+	configPath := normalizeConfigLoadPath(targetPath)
+	config, alreadyExists := c.configMap[configPath]
+	if !alreadyExists {
```
oliverchang, line 70: nit: prefer 

```python
if alreadyExists {
  return config
```

To avoid indenting a larger block of code.

tools/osv-scanner/cmd/osv-scanner/config.go
```diff
@@ -0,0 +1,115 @@
...
+	// Cache to store loaded configs
+	configMap map[string]Config
+}
+
+type Config struct {
+	IgnoredVulns []IgnoreLine
+	LoadPath     string
+}
+
+type IgnoreLine struct {
```
oliverchang, line 28: nit: "Line" is a bit ambiguous -- I had thought it was at first referring to a line number or a single line, when it doesn't. Can we rename it "IgnoreEntry" or something similar?

oliverchang (approved):

tools/osv-scanner/cmd/osv-scanner/config.go
another-rex: IgnoreUntil should be an ISO 8601 datestring, which can contain an offset for non-UTC timestamp. If they don't pass in an offset I think this will assume it is UTC, though I haven't tested this.

I don't think this is a big concern since at most it's off by 12 hours, but the expiry is designed to be used with a granularity of days, not hours.

andrewpollock: If the objective is to be purely date-based, maybe make this more unambiguous by only taking a date. https://toml.io/en/v1.0.0#local-date seems to support this.

oliverchang: We should still get it right because it can be confusing to people. 

I wonder if  just calling `IgnoreUntil.UTC()` is sufficient here.

oliverchang: +1 to just limiting this to a date as @andrewpollock suggested, but there's still the UTC question to resolve.

another-rex: Just tested the time UTC/not UTC issue. The `.After` comparison takes into account the timezone specified, so calling `.UTC()` in either part of the comparison doesn't change the result. 

If no timezone is specified, it assumes that it is **local** time, not UTC time as I guessed above. 

> If the objective is to be purely date-based, maybe make this more unambiguous by only taking a date. https://toml.io/en/v1.0.0#local-date seems to support this.

I don't see any way to restrict the type of date specified for the toml parser without using a completely custom type. I can change the examples in the readme and the osv-scanner files we use to use local dates instead though.

oliverchang: Sounds good, thanks!

oliverchang:
Is this ready to merge?

andrewpollock (approved):

Issue #771, opened 2022-10-11
Define behaviour for invalid updates


andrewpollock:
#761 identified a need to define behaviour around existing record updates:

1. when the source record is retracted, how should this be reflected in the record in OSV?
2. when the source record is completely deleted (but the file still exists, e.g. what happened with https://github.com/cloudsecurityalliance/gsd-database/commit/19eb9623e7c21662f8e61c3b775983c2a8b14187#diff-ae4add4297501a5ffd209adca63c6c45837ef8ac416ac276048b778e5613de5f), how should this be reflected in the record in OSV?

andrewpollock:
**Status update**: we have an internal document defining what is invalid behaviour and we're working on a public blog post and FAQ update relating to this to communicate it.

**Next steps**:

- [ ] Publish document as a blog post on OSV.dev’s blog and add an FAQ entry
- [ ] Define a machine readable feedback format and modify [replace_importer_log()](https://github.com/another-rex/osv/blob/4ed4040f268b03d9659cf4222ed43fc44602902d/docker/importer/importer.py#L65) to output using it
- [ ] [Correct export behavior](https://github.com/google/osv.dev/issues/1007) for bugs marked as INVALID in Datastore
- [ ] Enable JSON validation
- [ ] Extend [_process_updates_git()](https://github.com/google/osv.dev/blob/26050deb42785bc5a4dc7d802eac8e7f95135509/docker/importer/importer.py#L272) to do something with Git records identified as deleted
- [ ] Extend or compliment [_process_updates_bucket()](https://github.com/google/osv.dev/blob/26050deb42785bc5a4dc7d802eac8e7f95135509/docker/importer/importer.py#L330) to do a reverse check to identify deleted records
- [ ] Extend [_process_updates_git()](https://github.com/google/osv.dev/blob/26050deb42785bc5a4dc7d802eac8e7f95135509/docker/importer/importer.py#L272) and [_process_updates_bucket()](https://github.com/google/osv.dev/blob/26050deb42785bc5a4dc7d802eac8e7f95135509/docker/importer/importer.py#L330) to validate package names and versions
- [ ] Directly communicate with existing data source providers

andrewpollock:
End of year update for interested parties:

This hasn't seen any significant further progress due to higher priorities in 2023. 

It will be actively worked on in 2024, and the approach will be refreshed based on observations and subsequent learnings from real-world data quality issues in 2023.

[...]

PR #788, opened 2022-10-26
JSON schema validation

### LINKED ISSUE #770: Improve validation of OSV records before ingesting them
#761 identified that there's some room for improvement with validation of records before we ingest them in the first place:

https://github.com/cloudsecurityalliance/gsd-database/commit/12cef3506c418e9dea902c8a4c78c1423c72cb70
and
https://github.com/cloudsecurityalliance/gsd-database/commit/22635a956a09eb01569a4092343f127613e0aeb9
are examples of invalid records that should not have been ingested in the first place

https://github.com/google/osv.dev/blob/e869bd799ee6e1e5683cfa6f67b1e7bd7903a13e/docker/importer/importer.py#L293 is where the importer is currently parsing to make a decision on whether to treat a record as valid in the first place.

https://github.com/google/osv.dev/blob/ff50a96d47416e52019b50f9fd95b133bca9d9bb/osv/sources.py#L65 is the code that needs to be improved.

another-rex:
Fixes #770 
When loading a osv schema, be more strict on what is allowed by first checking the imported osv.json against the schema. 

- Do this validation when loading
- Fix issue where EVIDENCE is not an entry in vulnerability.proto
- Update worker and importer tests to actually test against valid osv entries
  - Add modified date to yaml test cases
    - YAML has weird importing where datetime is converted into datetime.datetime python object instead of str. Add code to account for that.
- Add osv-schema as a submodule
- Add symbolic link to osv validation schema
- Manually copy over validation schema in docker container to avoid issues with symbolic links

### Before merging:
A potentially large number of bucket entries might not be valid osv, we probably need to make a decision on how to deal with them.

- Probably spin up the testing environment to see how many entries are actually rejected.
- Determine behavior for what to do to handle invalid entries that's already in osv's database #771


osv/sources.py
```diff
@@ -52,6 +53,8 @@ def repo_path(repo):
 
 
 class NoDatesSafeLoader(yaml.SafeLoader):
+  """Safe yaml loader that removes datetime autoparsing"""
```
andrewpollock, line 56: Nit: s/yaml/YAML/

osv/sources.py
```diff
@@ -52,6 +53,8 @@ def repo_path(repo):
 
 
 class NoDatesSafeLoader(yaml.SafeLoader):
```
andrewpollock, line 55: there's obviously sometime about YAML parsing that I'm ignorant about here. Can you add some sort of hint for readers like me?

andrewpollock:
This seems to be only changing tests and not actual code that imports?

osv/sources.py
another-rex: Added more detailed docstring

another-rex:
> This seems to be only changing tests and not actual code that imports?

The actual validation just happens on line 149 in `sources.py`, and some pipfile changes to import in the jsonschema library. All the rest is just test file change.

another-rex:
The test is failing because pipenv lock on python3.10 and on python3.8 actually results in different pipfile.lock files, because different transient deps that's needed changes based on python version.

osv/sources.py
```diff
@@ -53,18 +53,20 @@ def repo_path(repo):
...
-  """Safe yaml loader that removes datetime autoparsing"""
+  """
+  Safe YAML loader that removes datetime autoparsing
+
+  pyYAML automatically parses date strings into python datetime.datetime, which
```
andrewpollock, line 59: Nit: s/python/Python/
Typo: s/failling/failing/

osv/sources.py
andrewpollock: Thanks, that was very helpful

andrewpollock:
> > This seems to be only changing tests and not actual code that imports?
> 
> The actual validation just happens on line 149 in `sources.py`, and some pipfile changes to import in the jsonschema library. All the rest is just test file change.

Do you mean `parse_vulnerability_from_dict` ( https://cs.github.com/google/osv.dev/blob/29ac697409101d1af65f367ec427302ecca16c79/osv/sources.py#L105)

So PR is just making that parsing stricter and more likely to raise an error in the event of invalidity? Wow, cool.

andrewpollock:
> > > This seems to be only changing tests and not actual code that imports?
> > 
> > 
> > The actual validation just happens on line 149 in `sources.py`, and some pipfile changes to import in the jsonschema library. All the rest is just test file change.
> 
> Do you mean `parse_vulnerability_from_dict` ( https://cs.github.com/google/osv.dev/blob/29ac697409101d1af65f367ec427302ecca16c79/osv/sources.py#L105)
> 
> So PR is just making that parsing stricter and more likely to raise an error in the event of invalidity? Wow, cool.

Oh, disregard, I found the change hiding at the bottom of `sources.py` in this PR.

andrewpollock:
> The test is failing because pipenv lock on python3.10 and on python3.8 actually results in different pipfile.lock files, because the transient deps that's needed changes based on python version.

Ugh. This sounds like time to sort out this dev vs prod Python version problem once and for all.

osv/sources.py
```diff
@@ -27,13 +29,16 @@
...
 from . import vulnerability_pb2
+from . import cache
 
 AUTHOR_EMAIL = 'infra@osv.dev'
 PUSH_RETRIES = 2
 PUSH_RETRY_SLEEP_SECONDS = 10
 
 YAML_EXTENSIONS = ('.yaml', '.yml')
-JSON_EXTENSIONS = ('.json')
+JSON_EXTENSIONS = '.json'
```
oliverchang, line 39: I think the extension was for this to be a tuple (which it previously was not because of a missing ,).

Can you make this `('.json.',)` instead?

docker/worker/testdata/BLAH-127.yaml
```diff
@@ -10,9 +10,12 @@ affected:
 - package:
     name: blah.com/package
     ecosystem: golang
+  versions:
```
oliverchang, line 13: There's a bit of a disconnect here in that we technically accept vulnerabilities with only git ranges and no versions arrays, because our infra will populate that if needed. 

We'll need to relax this rule here before we merge this: https://github.com/ossf/osv-schema/blob/92001ed827205e50509c8a286e298286674cc96d/validation/schema.json#L244

oliverchang (approved):
nice, thanks!! some minor comments.

docker/worker/testdata/BLAH-127.yaml
another-rex: Created a PR here: https://github.com/ossf/osv-schema/pull/93

another-rex:
Ran the validation against every repository and bucket we are currently importing. (apart from oss-fuzz, which is not being validated in this code anyway).

`android-osv` and `cve-osv` buckets have null values for some of the array entries. This is being loosened here https://github.com/ossf/osv-schema/pull/95 to allow those to pass. 

`android-osv` also have some entries that have `withdrawn`, but no `modified`.

Other than those entries, there's only 1 other invalid entry, `2021/44xxx/GSD-2021-44142.json`, which has `VENDOR` being a reference type. Everything else passes.

andrewpollock:
Can we put the actual behaviour behind a flag so we can toggle the behaviour without requiring a code change + push (just a flag change)?

I think this would require a new flag on the importer (which would default to non-rejection, so we don't need to make config changes) and an extra parameter to the `parse_vulnerability_from_dict` function that plumbs the intent of that flag through to it.

osv/schema.json
```diff
@@ -0,0 +1 @@
+../osv-schema/validation/schema.json
```
oliverchang, line 1: Can we just include osv-schema in the "osv" directory instead to avoid the "../" ?

osv/sources.py
```diff
@@ -27,13 +29,16 @@
 # pylint: disable=relative-beyond-top-level
 from . import repos
 from . import vulnerability_pb2
+from . import cache
```
oliverchang, line 32: nit: put this before line 30 to preserve lexicographic ordering.

osv/sources.py
```diff
@@ -105,6 +146,11 @@ def _get_nested_vulnerability(data, key_path=None):
 def parse_vulnerability_from_dict(data, key_path=None):
   """Parse vulnerability from dict."""
   data = _get_nested_vulnerability(data, key_path)
+  try:
+    jsonschema.validate(data, load_schema())
+  except jsonschema.exceptions.ValidationError as e:
+    logging.warning("Failed to validate loaded OSV entry: %s", e.message)
```
oliverchang, line 152: nit: use single quotes.

docker/worker/Dockerfile
```diff
@@ -24,6 +24,10 @@ VOLUME /var/lib/docker
 COPY docker/worker/Pipfile* /env/docker/worker/
 COPY setup.py Pipfile* README.md /env/
 COPY osv /env/osv
+# Remove the symbolic link
+RUN rm /env/osv/schema.json
```
oliverchang, line 28: Can we avoid this if we just check out the submodule into the "osv" directory directly?

oliverchang (approved):

docker/worker/Dockerfile
another-rex: We can, but osv-schema has a lot of unrelated files inside it which I wanted to avoid putting into worker. (i..e the docs project, the converting tooling (ghsa and debian)...etc).

docker/importer/importer.py
```diff
@@ -291,7 +291,12 @@ def _process_updates_git(self, source_repo: osv.SourceRepository):
       try:
         _ = osv.parse_vulnerability(path, key_path=source_repo.key_path)
       except Exception as e:
-        logging.error('Failed to parse %s: %s', changed_entry, str(e))
+        if isinstance(e, KeyError) and source_repo.key_path in e.args:
+          # Key path doesn't exist in the vulnerability.
```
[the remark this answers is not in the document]
another-rex: Added a new error :+1:

docker/worker/Dockerfile
oliverchang: That seems OK, it's less than 1MB in total (not including the .git repo, which seems to be 2.6MB on my machine which is also not too bad), and we can move out the tooling stuff later if needed.

another-rex: Moved the entire folder into docker.

another-rex:
@andrewpollock It is now gated by a flag, with it being logging warnings by default.

 We'll need to make a change to the docker image to enable strict mode.

andrewpollock (approved):

PR #789, opened 2022-10-26
Some minor frontend changes

Annoraaq:
Changing the following:

## Home page
- Make the decorative hero background on the home page stay on top and not scrolling
- Remove the padding from the ecosystems sections
- Add a decorative image to the data sources section
- Add footer decoration
- Change the spacing of the sections

## About page
- Avoid "jumping" of the FAQs when opening an answer
- Make cursor a pointer when hovering a question
- Change background color when hovering a question
- Change FAQ title


gcp/appengine/frontend3/src/templates/home.html
```diff
@@ -2,9 +2,8 @@
 {% set active_section = 'home' %}
 
 {% block content %}
-<div class="mdc-layout-grid">
-  <div class="mdc-layout-grid__inner home-page">
-    <img class="background-1" src="/static/img/background-1.png">
+<div class="mdc-layout-grid home-page">
+  <div class="mdc-layout-grid__inner">
     <img class="background-2" src="/static/img/background-2.png">
```
another-rex, line 7: I believe this line is also suppose to be removed (since the backgorund-2 css classes has been removed as well). Right now it's just a tiny image on the top left corner of the page.

another-rex (commented):
Thank you! LGTM with one comment.

gcp/appengine/frontend3/src/templates/home.html
Annoraaq: Absolutely. Good catch.

Done.

another-rex (approved):

PR #793, opened 2022-10-31
Use python3.11 docker for worker image

### LINKED ISSUE #623: Upgrade python to the latest version (3.10 or 3.11)
Upgrade python to the latest version (3.10 or 3.11)

another-rex:
Custom build for python3.11, overriding the system python3.8 that comes installed with `software-properties-common`

The code is mostly from clusterfuzz base image here: https://github.com/google/clusterfuzz/blob/master/docker/base/Dockerfile

The only change from the clusterfuzz example is to change `make altinstall` to `make install`, which makes it override python3 to point to `python3.11` instead of `python3.8`

Fixes #623


docker/worker/Dockerfile
```diff
@@ -18,13 +18,21 @@ ENV DEBIAN_FRONTEND noninteractive
...
         gnupg-agent \
-        python3-pip \
+        zlib1g-dev \
+        libffi-dev \
+        libssl-dev \
         software-properties-common
 
+RUN curl -sS https://www.python.org/ftp/python/3.11.0/Python-3.11.0.tgz | tar -C /tmp -xzv && \
+    cd /tmp/Python-3.11.0 && \
+    ./configure --enable-optimizations --with-lto --enable-loadable-sqlite-extensions && make && make install && \
```
andrewpollock, line 33: Could you separate the `make`/`make install` to be on separate lines so they're more clear to a reader that this isn't just a `./configure`.

andrewpollock:
I guess as a result of this, a weekly deployment will be longer because it's going to always involve sucking down and rebuilding Python...?

another-rex:
Yeah, we could improve that by creating a new base image for worker (and probably the other images) to be based on.

docker/worker/Dockerfile
oliverchang: One consideration with using `make install` is that it may potentially cause incompatibilities with the system/apt-packaged python (if it's installed indirectly through some other package?). `altinstall` is safer because it leaves the apt python intact.

another-rex: Python3.8 is installed through the `software-properties-common` package. 

It was using `altinstall` which skips overriding the default python binary, which causes problem with pipenv (I think it was getting confused on which python version to use). Switching to `install` overrides the default python3 executable in `/usr/bin`, which seems to fix the compile issues.

oliverchang: OK, hopefully there shouldn't be any backwards incompatibility issues from this.

andrewpollock: Honestly, this feels almost as messy as upgrading the base Docker image at this point. I'm assuming that `make altinstall` installs into `/usr/local/bin`? Can we just ensure that all invocations of `pipenv` are (for now) using `/usr/local/bin` (or better yet a central variable we can flip down the track when we _do_ upgrade the base image to Ubuntu 22.04 and can revert back to the system Python? To this end, are we going to need to adjust any shebang lines to use a different Python if we do go down the path of `make altinstall`?

I feel like we're still going to have to "test in Production" this Python change, which if we _don't_ do a `make altinstall` may result in unknown breakage of other parts of the base image.

oliverchang (approved):

PR #794, opened 2022-11-01
Fixes appengine releases by ignoring testdata

another-rex:
When running tests it will pull down alpine repository into testdata folder, massively inflating file count beyond the 10000 file limit. This is not needed for appengine, so should be ignored.


andrewpollock (approved):

oliverchang (approved):
thanks!! 

as discussed, it may also be a good idea to at a git clean step prior to our staging deployment.

oliverchang (approved):

PR #807, opened 2022-11-07
Fix alpine enumeration

another-rex:
Since version history is tracked by git, it is really hard to correct for errors in a package's history, so if a version is not valid, it will likely still keep showing up when enumerating history for a package. Because of this, we will ignore all invalid versions when enumerating versions.


oliverchang (approved):
thanks!!

Issue #243, opened 2021-10-15
vulnfeeds: Extract commit ranges from other git provider links


westonsteimel:
At least the following from the pypa advisory set:

- gitlab.com
- bitbucket.org

oliverchang:
This is the relevant code that currently handles parsing of github links to extract the commit hash: https://github.com/google/osv.dev/blob/85e7db46f2bea3fbfa4ea939e0f123ff5ec1a36a/vulnfeeds/cves/versions.go#L55

PR #808, opened 2022-11-07
Add more git repo hosts

andrewpollock:
Fixes: #243


vulnfeeds/cves/versions.go
```diff
@@ -52,35 +53,115 @@ type CPE struct {
 	Other      string
 }
 
-func extractGitHubCommit(link string) *FixCommit {
-	// Example: https://github.com/google/osv/commit/cd4e934d0527e5010e373e7fed54ef5daefba2f5
-	u, err := url.Parse(link)
+// Returns the base repository URL for supported repository hosts.
+func repo(u string) (string, bool) {
```
oliverchang, line 57: Can we make this public and merge the code you have in https://github.com/google/osv.dev/blob/4724b9229883b104ffea7fd71e49b16e3a33819b/vulnfeeds/cpp/main.go#L42, and just call this there?

vulnfeeds/cves/versions.go
```diff
@@ -52,35 +53,115 @@ type CPE struct {
...
+	}
+
+	// GitWeb URLs are structured another way, e.g.
+	// https://git.dpkg.org/cgit/dpkg/dpkg.git/commit/?id=faa4c92debe45412bfcf8a44f26e827800bb24be
+	// https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/commit/?id=817b8b9c5396d2b2d92311b46719aad5d3339dbe
+	if strings.HasPrefix(parsedURL.Path, "/cgit") && strings.HasSuffix(parsedURL.Path, "commit/") && strings.HasPrefix(parsedURL.RawQuery, "id=") {
+		return strings.Split(parsedURL.RawQuery, "=")[1], true
+	}
+
+	// GitHub and GitLab commit URLs are structured one way, e.g.
```
oliverchang, line 129: Maybe add a TODO to support extracting a commit hash from a github pull request URL. I've seen some cases of these in CVEs.

vulnfeeds/cves/versions.go
```diff
@@ -52,35 +53,115 @@ type CPE struct {
 	Other      string
 }
 
-func extractGitHubCommit(link string) *FixCommit {
-	// Example: https://github.com/google/osv/commit/cd4e934d0527e5010e373e7fed54ef5daefba2f5
-	u, err := url.Parse(link)
+// Returns the base repository URL for supported repository hosts.
+func repo(u string) (string, bool) {
+	parsedURL, err := url.Parse(u)
```
oliverchang, line 58: Would it make more sense to have a single function for parsing a repo url, and return a struct containing both the repo and commit in separate fields?

another-rex:
The if lines are getting quite long, can you put each && or || part on a separate line to reduce the width of the code making it easier to read?

vulnfeeds/cves/versions.go
andrewpollock: Yes, that was my intention once this has landed. I'll make them public from the get go.

vulnfeeds/cves/versions.go
andrewpollock: I debated this, but felt that for smaller units of testability to pursue the approach I took here.

vulnfeeds/cves/versions.go
andrewpollock: Ack, I've seen them. TODO added.

I suspect this will require using the GitHub API, which expands the complexity of this function from mere string manipulation.

vulnfeeds/cves/versions.go
oliverchang: You could have both: 

you could have a underlying function that does both, which could simplify the implementation. Then have 2 convenience/public functions that only do 1.

andrewpollock: I think we need to align on requirements and callers for these functions. 

Right now, you can feed a URL, any URL to the `Repo()` function for the purposes of getting a known Git repo from it. That URL may or may not also align with being a Git commit (think GitHub Issue, GitHub Pull Request before we're able to resolve them to commits). There's a larger set of URLs than just Git commits, basically. There are going to be URLs where `Repo()` works and `Commit()` doesn't.

Should potential callers of `Repo()` be penalised because a Git commit wasn't derivable, but a base Git repository could?

Right now it's `extractGitCommit()`'s job to return the struct and originally `repo()` and `commit()` were individually testable implementation details for how that struct got constructed.

oliverchang: The returned value of this combined function can simplify set `nil` on the fields it's not able to derive, which aligns with accounting for cases where `Repo()` works but `Commit()` doesn't. 

That said, there may be benefits in terms of error handling if we separate too. I don't feel strongly enough about this, so feel free to go either way for this.

vulnfeeds/cves/versions.go
```diff
@@ -52,35 +53,141 @@ type CPE struct {
...
 
-func extractGitHubCommit(link string) *FixCommit {
-	// Example: https://github.com/google/osv/commit/cd4e934d0527e5010e373e7fed54ef5daefba2f5
-	u, err := url.Parse(link)
+// Returns the base repository URL for supported repository hosts.
+func Repo(u string) (string, bool) {
+	parsedURL, err := url.Parse(u)
 	if err != nil {
-		return nil
+		log.Printf("%+v", err)
```
oliverchang, line 60: We should just return the error instead of logging it here, and have callers handle it. Same for `Commit()`.

vulnfeeds/cves/versions.go
andrewpollock: > The returned value of this combined function can simplify set `nil` on the fields it's not able to derive, which aligns with accounting for cases where `Repo()` works but `Commit()` doesn't.

Right now, you know if `Commit()` or `Repo()` works because you get your boolean back. Alternatively, the caller has to start digging around in structs to know if it got what it wanted, which seems more complicated to me?

vulnfeeds/cves/versions.go
andrewpollock: > We should just return the error instead of logging it here, and have callers handle it. Same for `Commit()`.

If we do that, then the simplicity of having a boolean return value for things working correctly is lost.

The current implementation was inspired from [go/gotip/episodes/30](https://goto.google.com/gotip/episodes/30)

If you don't feel that this is appropriate here, I can refactor accordingly.

oliverchang: That example lists a motivating factor as there being only a single possible sentinel error value, which isn't very useful. In this case, we have have at least two: one where the url.Parse failed, and the other is if it's unsupported. 

I also generally like returning errors, because they can propagate useful context up the stack, and leave it up to the caller to decide when the most appropriate time to log is, as opposed to here where we just log regardless. 

Anyway, not going to split any hairs over this minor thing and happy to just leave this as is  :) This is something we can easily refactor in the future if we need to.

oliverchang (approved):

PR #815, opened 2022-11-11
feat: inline `g-rath/osv-detector/pkg/lockfile` package

G-Rath:
This is hopefully the first step towards folding `osv-detector` into `osv-scanner`, so that we grow a single strong tool*

*: now that I've used the scanner I suspect I might end up keeping the detector around for a bit because of how difference the CLI interface is, and I suspect folks here are keen to stick with how stuff is currently being handled by the scanner, but that shouldn't stop our collaboration and development here 🙂


oliverchang:
Thank you so much Gareth!! we'll take a look.

Re the CLI differences, mind summarising where the missing gaps are? feel free to create new issues for these.

G-Rath:
> Re the CLI differences, mind summarising where the missing gaps are?

* it seems like you're recursing dwn folders i.e. if I do `osv-scanner .` in one of our client apps, it goes looking into `node_modules`
  * my experience has been its better to default to just the top-level because of how deep directories can go (due to things like `node_modules`) and its typically not that hard to handle it in userland (i.e. with globs)
  * (that's not to say you can't/shouldn't have something like a `--recursive` flag)
* the config implementation being implemented in #778 looks reasonably different to the interface used by `osv-detector`
  * I've [put a comment](https://github.com/google/osv.dev/pull/778#issuecomment-1310842275) about this
  * this extends to some of the flags like `--config`, `--config-ignore`, etc which could totally be landed in another PR so you might already be planning to add those, and don't want to jump the gun :)
* the general output feels more noisy and less clear to me (see comparison screenshots below)

image

image

another-rex:
Thanks for making the PR!

> * it seems like you're recursing dwn folders i.e. if I do `osv-scanner .` in one of our client apps, it goes looking into `node_modules`
>   
>   * my experience has been its better to default to just the top-level because of how deep directories can go (due to things like `node_modules`) and its typically not that hard to handle it in userland (i.e. with globs)
>   * (that's not to say you can't/shouldn't have something like a `--recursive` flag)

I've been looking at adding an option to respect `.gitignore` files and ignore files that have also been ignored by git, which could mitigate deep directories like `node_modules` somewhat. Having a flag to specify whether to run recursively or not is definitely a good idea.

> this extends to some of the flags like --config, --config-ignore, etc which could totally be landed in another PR so you might already be planning to add those, and don't want to jump the gun :)

The additional flags for ignoring configs are also certainly something we want to add.

> * the general output feels more noisy and less clear to me (see comparison screenshots below)
> 
> ...snip...

Yeah, the table output for osv-scanner did not turn out as well as I would like, it's something I would like to improve hopefully soon.

G-Rath:
> Having a flag to specify whether to run recursively or not is definitely a good idea.

That sounds like something that'd be straightforward for me to pick up, if you like - though how would you feel if I made it default to not be recursive though? imo that's better because even if you have support for ignoring that's something you need to know to setup and it can be common on projects to commit stuff you might not want to do security checks on (i.e. GitHub Actions encourages you to commit `node_modules`, at work our WordPress applications typically have their plugins committed which sometimes include stray `package.json` and `composer.json` files, etc).

----

btw those are minor things - I think you're done a great job with the scanner so far 😄

another-rex (approved):
LGTM, thanks!

PR #816, opened 2022-11-11
feat: implement `--recursive` flag

G-Rath:
As mentioned [here](https://github.com/google/osv.dev/pull/815#issuecomment-1311245059), I've made the default to be checking only the files in the target directories, as I think that is the better UX: generally easier to manage inclusion in userland (i.e. loops, globs, `tree`, etc) than exclusion (case and point, right now there is no way to exclude anything from the scanner).

I suspect it is probably going to be best to replace `filepath.Walk` with a recursive loop, at least if `recursive` is not passed because otherwise I believe you load the whole file tree into memory every time.

I have started writing some tests for this, but will do a seperate PR for that since it requires more unrelated work.

### DIFF
--- a/tools/osv-scanner/cmd/osv-scanner/main.go
+++ b/tools/osv-scanner/cmd/osv-scanner/main.go
@@ -206,9 +206,10 @@ func main() {
 				Value: false,
 			},
 			&cli.BoolFlag{
-				Name:  "recursive",
-				Usage: "check subdirectories",
-				Value: false,
+				Name:    "recursive",
+				Aliases: []string{"r"},
+				Usage:   "check subdirectories",
+				Value:   false,
 			},
 		},
 		ArgsUsage: "[directory1 directory2...]",
--- a/tools/osv-scanner/cmd/osv-scanner/main.go
+++ b/tools/osv-scanner/cmd/osv-scanner/main.go
@@ -19,8 +19,9 @@ import (
 )
 
 // scanDir walks through the given directory to try to find any relevant files
-func scanDir(query *osv.BatchedQuery, dir string, skipGit bool) error {
+func scanDir(query *osv.BatchedQuery, dir string, skipGit bool, recursive bool) error {
 	log.Printf("Scanning dir %s\n", dir)
+	root := true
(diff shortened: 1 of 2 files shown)


tools/osv-scanner/cmd/osv-scanner/main.go
```diff
@@ -199,6 +205,11 @@ func main() {
 				Usage: "skip scanning git repositories",
 				Value: false,
 			},
+			&cli.BoolFlag{
+				Name:  "recursive",
+				Usage: "check subdirectories",
```
another-rex, line 210: Can you add an alias with `-r` here as well?

tools/osv-scanner/cmd/osv-scanner/main.go
```diff
@@ -19,14 +19,20 @@ import (
...
+func scanDir(query *osv.BatchedQuery, dir string, skipGit bool, recursive bool) error {
 	log.Printf("Scanning dir %s\n", dir)
+	root := true
 	return filepath.Walk(dir, func(path string, info os.FileInfo, err error) error {
 		if err != nil {
 			log.Printf("Failed to walk %s: %v", path, err)
 			return err
 		}
 
+		if !root && !recursive && info.IsDir() {
```
another-rex, line 31: This will skip the .git scanning on the next if block. Can you move this to the bottom of the function instead?

another-rex (commented):
Thank you! Just 2 comments then LGTM.

> I suspect it is probably going to be best to replace filepath.Walk with a recursive loop, at least if recursive is not passed because otherwise I believe you load the whole file tree into memory every time.

I think walk/walkDir is also implemented recursively, and when you call skipdir it doesn't load the next level down: https://cs.opensource.google/go/go/+/refs/tags/go1.19.3:src/path/filepath/path.go;l=437;drc=122a22e0e9eba7fe712030d429fc4bcf6f447f5e

oliverchang (approved):

G-Rath:
> I think walk/walkDir is also implemented recursively,

ah yeah you're right - re-reading the function comment it says

> requires Walk to read *an entire directory* into memory before proceeding to walk that directory.

I misread that as "reads all of the directories it'll walk into memory" 😅

PR #817, opened 2022-11-11
feat: adjust logging and general restructure to enable writing cli tests

G-Rath:
This is an initial start on being able to test the cli component, based on approach I've used in `osv-detector`.

I'm happy to continue with this but want to check on how folks feel with the approach of having a `Reporter` and potentially changing up the output at the same time.

(this is based off https://github.com/google/osv.dev/pull/816 just to give me something easy to test)


oliverchang:
Can you rebase? Or we can defer this to when we set up our new standalone repo :)

G-Rath:
@oliverchang I've rebased :)

tools/osv-scanner/cmd/osv-scanner/main_test.go
```diff
@@ -0,0 +1,186 @@
...
+
+type cliTestCase struct {
+	name         string
+	args         []string
+	wantExitCode int
+	wantStdout   string
+	wantStderr   string
+}
+
+func testCli(t *testing.T, tc cliTestCase) {
```
oliverchang, line 74: nit: this should be testCLI per https://github.com/golang/go/wiki/CodeReviewComments#initialisms

tools/osv-scanner/internal/output/reporter.go
```diff
@@ -0,0 +1,41 @@
...
+
+import (
+	"fmt"
+	"io"
+)
+
+type Reporter struct {
+	stdout       io.Writer
+	stderr       io.Writer
+	outputAsJSON bool
```
oliverchang, line 11: Mind walking through the use case for this? 

When would someone want JSON output for general logging? Is the caller expected to format the JSON themselves? 

As is, this flag is more like a `outputToStderr` right now.

oliverchang (commented):
thanks! just some questions

tools/osv-scanner/internal/output/reporter.go
G-Rath: This is meant to be used alongside a `PrintJSONResults` function which gets called [if we're outputting JSON](https://github.com/G-Rath/osv-detector/blob/main/main.go#L508-L511) and that outputs the final results in JSON format - that way most parts of tool don't have to worry about where they log to, they can just call `PrintText` and it'll end up in the right place without creating invalid JSON.

I didn't want to sink too much time into this without confirming if you were happy with the approach, so I axed `PrintJSONResult` but kept this as a pull-over since without it `PrintText` becomes effectively redundant and it all just sort of falls over - it sounds like you happy with it though, so I can put some more time into it and then this should start making more sense :)

oliverchang: I see -- so PrintText is intended for human consumption of log output, while PrintJSONResults is intended for the machine readable scan results, which ends up being a no-op if JSON output is not enabled? 

I think my confusion was with the PR as is, where the meaning of `outputAsJSON` is rather unclear and just controls where the text output goes. Maybe we can remove this flag for now to keep things clearer, and once we move things over to the new repo we can expand this to what you mentioned.

G-Rath: Correct - the reason I have this in `osv-detector` is to allow us to output without effectively blocking the CLI output (which could take a few minutes), because we loop over each lockfile in order outputting as we go by passing the `results` of each check to the `Reporter` - that's fine when outputting human text, but for JSON this is really ugly (its technically possible by outputting JSON parts before and after the loop, but that's super brittle); so instead when the `Reporter` is set to output JSON it stores each result its given in an array while still outputting the human stuff (e.g. "checking file xyz", "loaded db abc", etc) to stderr to keep the user informed on what it's doing, and then when its finished before the CLI exits it calls `r.PrintJSONResults` which has the reporter output the `results` array to stdout.

Like I said, I wanted to make sure you were ok with me exploring this before I got too deep into it since you might have some plans in this area internally that crossed over that'd make that effort moot; since it sounds like you don't, I'll expand this further and then it should start making more sense.

oliverchang: Got it, in that case we'll get the repo moved first and try this again there once you've expanded this? 

Thanks!

oliverchang:
fyi we've moved the code to https://github.com/google/osv-scanner

oliverchang:
Closing, please re-open in the new repo :)

PR #820, opened 2022-11-14
Filter unsupported ecosystems

### LINKED ISSUE #753: Filter unsupported ecosystems.
e.g. "GSD", "DWF", "Javascript" on https://osv.dev/list. Supported ecosystems are part of the OSV schema at https://ossf.github.io/osv-schema/#affectedpackage-field

These don't have well defined rules for package naming/versioning, so should be excluded from our DB completely.

michaelkedar:
Resolves #753 
Adds a filter in the worker to prevent vulnerabilities from unsupported ecosystems from being ingested.
Use `osv/ecosystems.py` as source of truth for supported ecosystems, and added placeholder helpers for unimplemented supported ecosystems.


docker/worker/worker_test.py
```diff
@@ -609,6 +609,12 @@ def setUp(self):
 
     osv.ecosystems.work_dir = 'testdata/tmp/'
 
+    # add fake ecosystems used in tests to supported ecosystems
+    osv.ecosystems._ecosystems.update({
+        'ecosystem': osv.ecosystems.OrderingUnsupportedEcosystem(),
+        'golang': osv.ecosystems.OrderingUnsupportedEcosystem(),
```
oliverchang, line 615: Can you update the tests instead of adding this `golang` one here? The correct ecosystem is "Go".

docker/worker/worker.py
```diff
@@ -316,6 +316,21 @@ def maybe_normalize_package_names(vulnerability):
...
+def filter_unsupported_ecosystems(vulnerability):
+  """Remove unsupported ecosystems from vulnerabiltiy
+
+  Returns number of remaining affected packages after filtering
+  """
+  filtered = []
+  for affected in vulnerability.affected:
+    if osv.ecosystems.get(affected.package.ecosystem):
+      filtered.append(affected)
+  del vulnerability.affected[:]
```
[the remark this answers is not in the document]
michaelkedar: No - `vulnerability.affected` is a protobuf repeated field here, which doesn't have a `clear` method.

This way is what is shown in [the documentation](https://developers.google.com/protocol-buffers/docs/reference/python-generated?csw=1#repeated-fields), and is used elsewhere in this file.

docker/worker/worker.py
```diff
@@ -476,6 +491,12 @@ def _do_update(self, source_repo, repo, vulnerability, relative_path,
       logging.warning('%s has an encoding error, skipping.', vulnerability.id)
       return
 
+    # If the vulnerability has affected packages, but none of them are supported
+    # then discard it
+    if len(vulnerability.affected) > 0:
+      if filter_unsupported_ecosystems(vulnerability) == 0:
```
another-rex, line 497: I think we still want to update existing bugs if affected packages changes.

e.g. if a package has 2 affected packages, one for an ecosystem we support, one we don't. If they then remove the one that affects the ecosystem we support, we would want to update that in the database. 

I don't think we need to filter out the vulnerability at all, just import them and set status to invalid if there are no affected packages. This way importer won't try to keep importing filtered out vulns for unsupported ecosystems.

another-rex (commented):
Thanks! Just a few comments.

docker/worker/worker.py
```diff
@@ -316,6 +316,21 @@ def maybe_normalize_package_names(vulnerability):
...
+  """Remove unsupported ecosystems from vulnerabiltiy
+
+  Returns number of remaining affected packages after filtering
+  """
+  filtered = []
+  for affected in vulnerability.affected:
+    if osv.ecosystems.get(affected.package.ecosystem):
+      filtered.append(affected)
+  del vulnerability.affected[:]
+  vulnerability.affected.extend(filtered)
```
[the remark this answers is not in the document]
another-rex: Oh I see the reply to Oliver's question

osv/ecosystems.py
```diff
@@ -604,6 +627,15 @@ def enumerate_versions(self,
...
     'PyPI': PyPI(),
     'RubyGems': RubyGems(),
+    # Ecosystems missing implementations
+    'Android': OrderingUnsupportedEcosystem(),
+    'Github Actions': OrderingUnsupportedEcosystem(),
+    'Linux': OrderingUnsupportedEcosystem(),
+    'OSS-Fuzz': OrderingUnsupportedEcosystem(),
+    'Pub': OrderingUnsupportedEcosystem(),
+    # Ecosystems handled specially
+    'Alpine': OrderingUnsupportedEcosystem(),
```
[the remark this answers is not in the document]
michaelkedar: Honestly, I'm not sure.

If these aren't included here, the ecosystems of just "Debian" and "Alpine" (without the release numbers) would be considered invalid and thus filtered, even though [the schema says the release number is optional](https://ossf.github.io/osv-schema/#affectedpackage-field).

I included them in the dictionary as OrderingUnsupported to preserve the current behaviour in that case, rather than filtering them out.

docker/worker/worker.py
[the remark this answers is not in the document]
michaelkedar: `vulnerability.affected` is a protobuf repeated field, which [doesn't support assignment](https://developers.google.com/protocol-buffers/docs/reference/python-generated?csw=1#repeated-fields)

docker/worker/worker.py
```diff
@@ -316,6 +316,16 @@ def maybe_normalize_package_names(vulnerability):
   return vulnerability
 
 
+def filter_unsupported_ecosystems(vulnerability):
+  """Remove unsupported ecosystems from vulnerability"""
```
andrewpollock, line 320: Nit: use periods at the end of sentences. https://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings

osv/ecosystems.py
```diff
@@ -144,6 +144,29 @@ def _get_affected_versions(self, versions, introduced, fixed, last_affected,
...
+  @property
+  def supports_ordering(self):
+    return True
+
+
+class OrderingUnsupportedEcosystem(Ecosystem):
+  """Placeholder ecosystem helper for unimplemented ecosystems"""
+
+  def sort_key(self, version):
+    raise ValueError("Ecosystem helper does not support sorting")
```
andrewpollock, line 156: Would `NotImplementedError` seems like it would make more sense in this class?

docker/worker/worker.py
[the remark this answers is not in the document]
oliverchang: Got it, thanks!

osv/ecosystems.py
oliverchang: +1

osv/ecosystems.py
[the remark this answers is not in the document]
oliverchang: Got it, thanks! 

Please add a comment explaining this above these two.

Something like:
```
# Alpine and Debian require a release version for enumeration, which is
# handled separately in get().
'Alpine': ...,
'Debian': ...,
```

docker/worker/worker.py
oliverchang: +1 good point. There's a case here of an existing entry getting updated to be invalid later with no valid ecosystenms. 

We can just import everything as Rex suggested, and simply mark them as invalid in these cases (similar to https://github.com/google/osv.dev/blob/7ccaa3a3bba95364039bac82c8dd4bb48d952d1b/docker/worker/worker.py#L419).

oliverchang (commented):
nice! just some remaining comments to address. otherwise this is looking pretty good :)

osv/ecosystems.py
```diff
@@ -144,6 +144,29 @@ def _get_affected_versions(self, versions, introduced, fixed, last_affected,
...
+  @property
+  def supports_ordering(self):
+    return True
+
+
+class OrderingUnsupportedEcosystem(Ecosystem):
+  """Placeholder ecosystem helper for unimplemented ecosystems."""
+
+  def sort_key(self, version):
+    raise NotImplementedError("Ecosystem helper does not support sorting")
```
oliverchang, line 156: nit: please use single quotes for all strings (except for docstrings).

osv/ecosystems.py
```diff
@@ -144,6 +144,29 @@ def _get_affected_versions(self, versions, introduced, fixed, last_affected,
...
+
+  def enumerate_versions(self,
+                         package,
+                         introduced,
+                         fixed=None,
+                         last_affected=None,
+                         limits=None):
+    raise NotImplementedError("Ecosystem helper does not support enumeration")
```
oliverchang, line 164: nit: please use single quotes for all strings (except for docstrings).

andrewpollock (approved):

osv/ecosystems.py
```diff
@@ -604,6 +627,16 @@ def enumerate_versions(self,
     'Packagist': Packagist(),
     'PyPI': PyPI(),
     'RubyGems': RubyGems(),
+    # Ecosystems missing implementations:
+    'Android': OrderingUnsupportedEcosystem(),
+    'Github Actions': OrderingUnsupportedEcosystem(),
```
oliverchang, line 632: this should be "GitHub"

another-rex (approved):

oliverchang (approved):
thanks!! Please work with @another-rex next monday to get this released.

PR #822, opened 2022-11-14
Remove all references to service account

andrewpollock:
This should result in configuration and code that uses the default service account when a test for a PR containing this commit gets run, so we can see how it fails.


oliverchang:
The "fix" here is likely to use like: https://github.com/google/clusterfuzz/blob/master/local/emulators/metadata.go

- For local users: To emulate a GCP metadata server locally, backed by the local credentials. 
- For Cloud Build CI: we can likely get away with not requiring that, and just using the GCP metadata server that's part of the environment.

andrewpollock:
> The "fix" here is likely to use like: https://github.com/google/clusterfuzz/blob/master/local/emulators/metadata.go
> 
> * For local users: To emulate a GCP metadata server locally, backed by the local credentials.
> * For Cloud Build CI: we can likely get away with not requiring that, and just using the GCP metadata server that's part of the environment.

Drat, just saw this now, I've been deep in understanding how this all hangs together...

So far, is "working" in that everything seems to start up, but I'm seeing connectivity issues: https://gist.github.com/andrewpollock/0d65b05bf1dd9472d7985517700903e2

Any thoughts?

oliverchang:
> > The "fix" here is likely to use like: https://github.com/google/clusterfuzz/blob/master/local/emulators/metadata.go
> > 
> > * For local users: To emulate a GCP metadata server locally, backed by the local credentials.
> > * For Cloud Build CI: we can likely get away with not requiring that, and just using the GCP metadata server that's part of the environment.
> 
> Drat, just saw this now, I've been deep in understanding how this all hangs together...
> 
> So far, is "working" in that everything seems to start up, but I'm seeing connectivity issues: https://gist.github.com/andrewpollock/0d65b05bf1dd9472d7985517700903e2
> 
> Any thoughts?

Not sure..  We need to see the contents of backend.log and esp.log.

 Assuming the default cloud build service account is being propagated correctly, then maybe the service account itself doesn't have the right permissions to access the necessary service. See the roles on the "esp-test@oss-vdb.iam.gserviceaccount.com" account and compare that to the cloudbuild one.

andrewpollock:
> Assuming the default cloud build service account is being propagated correctly, then maybe the service account itself doesn't have the right permissions to access the necessary service. See the roles on the "[esp-test@oss-vdb.iam.gserviceaccount.com](mailto:esp-test@oss-vdb.iam.gserviceaccount.com)" account and compare that to the cloudbuild one.

Hmm, yeah. If you look at [this attempt](https://pantheon.corp.google.com/cloud-build/builds/b293e08e-6701-4e29-961a-62d649ffd421;step=6?project=oss-vdb) there's lots of juicy Envoy logs in the output, including lots of refused connections (which to me says more of a network connectivity issue, but hey, I know nothing about this)

Could you please grant the following roles to `service-651737493649@gcp-sa-cloudbuild.iam.gserviceaccount.com`
* `roles/cloudtrace.agent`
* `roles/datastore.viewer`
* `roles/servicemanagement.serviceController`

and we can see if that has any bearing on things?

`fail to initialize config manager: http call to GET https://servicemanagement.googleapis.com/v1/services/api-test.osv.dev/rollouts?filter=status=SUCCESS returns not 200 OK: 403 Forbidden` is one standout item that perhaps `roles/servicemanagement.serviceController` will rectify...

andrewpollock:
Current status:
* working as desired (using the default service account for Cloud Build) when running under Cloud Build
* thoroughly busted<sup>1</sup> when run locally

<sup>1</sup>: https://github.com/GoogleCloudPlatform/esp-v2/blob/5c4538a5a46f34e30fca009cd190934fd366ceee/src/go/configmanager/config_manager.go#L137 looks like the culprit (tries to contact the metadata server to fetch an access token, and can't

```
F1115 01:05:08.847584       7 server.go:54] fail to initialize config manager: fail to get access token: Get "http://169.254.169.254/computeMetadata/v1/instance/service-accounts/default/token": dial tcp 169.254.169.254:80: connect: connection timed out
```

If having working local tests is a hard requirement to proceed, I'll have to park this for the time being.

oliverchang:
> Current status:
> 
> * working as desired (using the default service account for Cloud Build) when running under Cloud Build
> * thoroughly busted1 when run locally
> 
> 1: https://github.com/GoogleCloudPlatform/esp-v2/blob/5c4538a5a46f34e30fca009cd190934fd366ceee/src/go/configmanager/config_manager.go#L137 looks like the culprit (tries to contact the metadata server to fetch an access token, and can't
> 
> ```
> F1115 01:05:08.847584       7 server.go:54] fail to initialize config manager: fail to get access token: Get "http://169.254.169.254/computeMetadata/v1/instance/service-accounts/default/token": dial tcp 169.254.169.254:80: connect: connection timed out
> ```
> 
> If having working local tests is a hard requirement to proceed, I'll have to park this for the time being.

Local tests are indeed a pretty important part of our workflow when we are changing the API implementation. Without any other good alternatives, we'd need to run something like https://github.com/google/clusterfuzz/blob/master/local/emulators/metadata.go and pipe that through to the ESPv2 container to get this working.

andrewpollock:
> Local tests are indeed a pretty important part of our workflow when we are changing the API implementation. Without any other good alternatives, we'd need to run something like https://github.com/google/clusterfuzz/blob/master/local/emulators/metadata.go and pipe that through to the ESPv2 container to get this working.

Okay I'll park this for (at least) the rest of this week to see what sort of response we get on https://github.com/GoogleCloudPlatform/esp-v2/issues/752 and then determine next steps.

andrewpollock:
I think this has been obsoleted by @rjerrems excellent work

PR #823, opened 2022-11-14
Fix API dependency issues.

oliverchang:
Revert package updates to avoid protobuf version mismatches.

The root cause was grpcio-tools, but this is a dev dependency so move it there instead to avoid issues with our production image build.
So, we move this package into the dev-packages section.


another-rex (approved):
LGTM!

PR #824, opened 2022-11-15
Use library code

andrewpollock:
https://github.com/google/osv.dev/pull/808 moved some code into an existing library, so remove it from here


andrewpollock:
Got myself into a right royal mess with this branch, running away to try again.

PR #826, opened 2022-11-15
Improve CPE parsing

andrewpollock:
Make the CPE string parsing more robust by using go-cpe

Return errors rather than bools and adjust all callsites.


vulnfeeds/go.mod
```diff
@@ -18,6 +18,8 @@ require (
 	github.com/google/go-cmp v0.5.8 // indirect
 	github.com/googleapis/enterprise-certificate-proxy v0.0.0-20220520183353-fd19c99a87aa // indirect
 	github.com/googleapis/gax-go/v2 v2.4.0 // indirect
+	github.com/knqyf263/go-cpe v0.0.0-20201213041631-54f6ab28673f // indirect
```
another-rex, line 21: Move this up to the direct dependencies, since it's being used in the code.

another-rex (approved):
LGTM, just 1 comment.

vulnfeeds/go.mod
andrewpollock: Good catch, thanks. Done in https://github.com/google/osv.dev/pull/826/commits/aebf77d323a99a1b53a70d89883064e91c200208

PR #829, opened 2022-11-16
Don't mark deleted git entries as invalid.

oliverchang:
The supported way to handle making entries as invalid is to mark them as "withdrawn" and
leave the entry intact.

This also guards against availability issues from source feeds, such as when a bug in their
system causes their entries to be deleted (e.g. https://github.com/github/advisory-database/commit/06071dd8e490ed354657f8ee158880433cf04997)

### DIFF
--- a/docker/importer/importer.py
+++ b/docker/importer/importer.py
@@ -265,12 +265,10 @@ class Importer:
 
     # Get list of changed files since last sync.
     changed_entries = set()
-    deleted_entries = set()
 
     if source_repo.last_synced_hash:
       # Syncing from a previous commit.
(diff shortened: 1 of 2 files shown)


andrewpollock:
Could you please add a bit more of the "why" to the description?

oliverchang:
> Could you please add a bit more of the "why" to the description?

done.

andrewpollock:
> The supported way to handle these is to mark entries as "withdrawn" and leave the entry intact.

Is it correct that this behaviour is not currently implemented?

andrewpollock (approved):

PR #830, opened 2022-11-16
Set the package name in the OSV record based off the CPE product name

andrewpollock:


vulnfeeds/cpp/main.go
```diff
@@ -186,8 +186,11 @@ func CVEToOSV(CVE cves.CVEItem, repo, directory string) {
 	affected.AttachExtractedVersionInfo(versions)
 	v.Affected = append(v.Affected, affected)
 	if len(v.Affected[0].Ranges) == 0 {
-		Logger.Infof("No affected versions detected.")
+		Logger.Infof("%s: No affected versions detected.", CVE.CVE.CVEDataMeta.ID)
 	}
+	// divine a package name from the CPE
+	v.Affected[0].Package.Ecosystem = "GIT"
```
oliverchang, line 192: "GIT" isn't a valid ecosystem, and it's also not clear that the Name we extract here would be correct or meaningful at all. 

Could i understand a bit more the intention behind this?

andrewpollock: Oh I thought this was the plan of record we'd previously discussed for the ecosystem value in this situation.

Do you have alternative suggestions for how to handle this value in this situation?

oliverchang: There's not hard requirement to have a "package" here for now. Instead, we could just list the git repo in the "affected" section as a GIT range.

andrewpollock: Oh, my bad. I thought it was a necessary part of the record. I was trying to get a cleaner looking record than

```
"affected": [
    {
      "package": {},
      "ranges": [
        {
          "type": "ECOSYSTEM",
          "events": [
            {
              "introduced": "0"
            },
            {
              "fixed": "1.34"
            }
          ]
        }
      ]
    }
  ],
```

Putting `omitempty` on the struct json struct tag, but it doesn't seem to make a difference...

PR #831, opened 2022-11-16
Move osv-scanner

another-rex:
Removing osv-scanner and different wording in text to point to the new repository.


oliverchang (approved):
nice! can you update the osv.dev home page as well?

oliverchang (approved):

PR #835, opened 2022-11-16
Improve handling of quoted strings in CPEs

andrewpollock:
It turns out CPE strings have some funky quoting rules, which can translate to weird looking directory names.


vulnfeeds/cves/versions.go
```diff
@@ -379,6 +379,26 @@ func CPEs(cve CVEItem) []string {
 	return cpes
 }
 
+// There are some weird and wonderful rules about quoting with strings in CPEs
+// this uses a variation of how they wind up in there from go-cpe's
+// processQuotedChars()
+func RemoveQuoting(s string) (result string) {
```
oliverchang, line 385: Can you add some more comments about what this function is actually doing? 

Is this something that we can't just use say, https://pkg.go.dev/strings#Replace for this?

andrewpollock: > Can you add some more comments about what this function is actually doing?

Done

> Is this something that we can't just use say, https://pkg.go.dev/strings#Replace for this?

Very good point, looks like I can. Thanks!

oliverchang (approved):

PR #836, opened 2022-11-17
Fix bug with SemVer range evaluation in API.

### LINKED ISSUE #834: Missing result from query
```
curl -X POST -d '{"package": {"ecosystem": "npm", "name": "ws"}, "version": "7.1.1"}' https://api.osv.dev/v1/query
```

Should return at least https://osv.dev/vulnerability/GHSA-6fc8-4gx4-v693

oliverchang:
This didn't work properly if there are multiple packages/ranges in the relevant vulnerability, and the matching range
is not the last one.

This implementation now matches the OSV schema pseudocode at https://ossf.github.io/osv-schema/#evaluation.

Fixes #834.

### DIFF
--- a/gcp/api/integration_tests.py
+++ b/gcp/api/integration_tests.py
@@ -320,6 +320,22 @@ class IntegrationTests(unittest.TestCase):
     self.assertCountEqual(['GO-2021-0061', 'GO-2020-0036'],
                           [vuln['id'] for vuln in response_json['vulns']])
 
+    response = requests.post(
+        _api() + '/v1/query',
+        data=json.dumps({
+            'version': '7.1.1',
+            'package': {
+                'name': 'ws',
+                'ecosystem': 'npm',
+            }
+        }),
+        timeout=_TIMEOUT)
+
+    response_json = response.json()
+    self.assertEqual(1, len(response_json['vulns']))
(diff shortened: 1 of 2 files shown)


andrewpollock (approved):

andrewpollock:
So if I'm understanding correctly, these API integration tests query against the live data?
Based on that, is the existing set of queries we're performing sufficiently representative (it seems like we were blind to this bug due to the selection of queries being performed)?

oliverchang:
> So if I'm understanding correctly, these API integration tests query against the live data? Based on that, is the existing set of queries we're performing sufficiently representative (it seems like we were blind to this bug due to the selection of queries being performed)?

Yes, our queries did not encounter this edge case unfortunately.

andrewpollock:
> Yes, our queries did not encounter this edge case unfortunately.

So with the addition of the extra OSV ID in this PR, do you feel we now have all the possible permutations and combinations of record scenarios, or do we need to add more?

oliverchang:
> > Yes, our queries did not encounter this edge case unfortunately.
> 
> So with the addition of the extra OSV ID in this PR, do you feel we now have all the possible permutations and combinations of record scenarios, or do we need to add more?

That's hard to quantify, but I'm fairly confident our coverage is pretty good with the addition of this.

PR #837, opened 2022-11-18
Fix link on front page to point to new scanner

another-rex:


oliverchang (approved):

gcp/appengine/frontend3/src/templates/home.html
```diff
@@ -172,7 +172,7 @@ <h3 class="code-card-title">Scan an SBOM or Lockfile (preview)</h3>
       </div>
       <div class="mdc-layout-grid__cell--span-12">
         <div class="cta">
-          <a class="cta-primary link-button" href="https://github.com/google/osv/tree/master/tools/osv-scanner">More details</a>
+          <a class="cta-primary link-button" href="https://github.com/google/osv-scanner/">More details</a>
```
oliverchang, line 175: actually is that trailing / in the link needed?

another-rex: Not really, I can remove it

oliverchang (approved):

PR #845, opened 2022-11-21
Only add an ecosystem version range if there's package information

andrewpollock:
ecosystem-specific version ranges are only meaningful in conjuction with package information, so only add them to the resulting vulnerability record if package information is present.

Break out the Package to its own struct so it doesn't get emitted when there is no package information.

Adjust AddPkgInfo accordingly

### DIFF
--- a/vulnfeeds/vulns/vulns.go
+++ b/vulnfeeds/vulns/vulns.go
@@ -369,38 +369,40 @@ func (affected *Affected) AttachExtractedVersionInfo(version cves.VersionInfo) {
 	}
 
 	// Adding an ECOSYSTEM version range only makes sense if we have package information.
-	if affected.Package != nil {
-		versionRange := AffectedRange{
-			Type: "ECOSYSTEM",
-		}
-		seenIntroduced := map[string]bool{}
-		seenFixed := map[string]bool{}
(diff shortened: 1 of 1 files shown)


vulnfeeds/vulns/vulns.go
```diff
@@ -364,36 +368,39 @@ func (affected *Affected) AttachExtractedVersionInfo(version cves.VersionInfo) {
...
-	seenFixed := map[string]bool{}
-
-	for _, v := range version.AffectedVersions {
-		var introduced string
-		if v.Introduced == "" {
-			introduced = "0"
-		} else {
-			introduced = v.Introduced
+	// Adding an ECOSYSTEM version range only makes sense if we have package information.
+	if affected.Package != nil {
```
oliverchang, line 372: nit: can we invert this condition and continue instead? This would avoid a lot of indentation and make the code a bit easier to read.

andrewpollock: Done in [ec3cf43](https://github.com/google/osv.dev/pull/845/commits/ec3cf43d42d8af6018a528f1440999b467b3f388)

oliverchang (approved):

PR #847, opened 2022-11-22
Add read-all default permissions

another-rex:
Fixes 4/5 high issues in scorecard

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -18,6 +18,8 @@ on:
     # The branches below must be a subset of the branches above
     branches: [ master ]
 
+permissions: read-all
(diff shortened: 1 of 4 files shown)


andrewpollock:
It's infuriatingly hard to understand from https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions what the default permissions are to understand whether this is actually changing anything or not :-(

oliverchang (approved):

another-rex:
> If you specify the access for any of these scopes, all of those that are not specified are set to none.

I think the default permissions are just access to everything, only once you set a specific permission, do the others get disabled.

andrewpollock (approved):

PR #848, opened 2022-11-22
Only emit valid records

andrewpollock:


vulnfeeds/cpp/main.go
```diff
@@ -185,27 +185,28 @@ func CVEToOSV(CVE cves.CVEItem, repo, directory string) {
 	affected := vulns.Affected{}
 	affected.AttachExtractedVersionInfo(versions)
 	v.Affected = append(v.Affected, affected)
+
 	if len(v.Affected[0].Ranges) == 0 {
-		Logger.Infof("No affected versions detected.")
-	}
+		Logger.Infof("No affected versions detected for %s for %q", CVE.CVE.CVEDataMeta.ID, CPE.Product)
```
oliverchang, line 190: nit: similar comment here from last PR. We can just early return here.

andrewpollock: Done in [4699967](https://github.com/google/osv.dev/pull/848/commits/46999670fbc51ecfcbf13d9fc56f7d113fe467ce)

oliverchang (approved):

PR #849, opened 2022-11-23
Utility to do some analysis of the CPE dictionary

andrewpollock:
Currently, this parses the CPE dictionary, selectively looks for repo-shaped URLs and creates a map of products to repositories.

The utility of the CPE dictionary seems pretty limited so far: this approach is only yielding about 500 repositories from 32,488 application CPEs (~1.5%)

Correction: with the improvements to `Repo()` in #851, this is now yielding close to 6,000 repositories for application CPEs, more like 18%.

What remains to be determined is if these ~6,000 repositories are not already found in the NVD CVE data itself...


vulnfeeds/cmd/cperepos/main.go
```diff
@@ -0,0 +1,184 @@
...
+				Logger.Infof("Disregarding %q for %q (%s) because %v", r.URL, CPE.Product, r.Description, err)
+				continue
+			}
+			// Disregard the repos we know we don't like.
+			if slices.Contains(InvalidRepos, repo) {
+				Logger.Infof("Disliking %q for %q (%s)", repo, CPE.Product, r.Description)
+				continue
+			}
+			// Don't consider URL descriptions not explicitly allowlisted to reduce invalidity.
+			if !slices.Contains(ValidDescriptions, r.Description) {
```
oliverchang, line 149: What kind of noise gets introduced because of this? 

Since we are already vetting the URL value itself through the `Repo()` call, I wonder if this could inadvertently cause us to miss valid URLs.

vulnfeeds/cmd/cperepos/main.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+	defer client.Close()
+	Logger.GCloudLogger = client.Logger("cperepos")
+	flag.Parse()
+
+	CPEDictionary, err := LoadCPEDictionary(*CPEDictionaryFile)
+	if err != nil {
+		Logger.Fatalf("Failed to load %s: %v", *CPEDictionaryFile, err)
+	}
+	ProductToRepo := make(map[string]*string)
```
oliverchang, line 113: Should we output the ProductToRepo map somewhere as e.g. a JSON file?

vulnfeeds/cmd/cperepos/main.go
```diff
@@ -0,0 +1,184 @@
...
+
+	byteValue, _ := ioutil.ReadAll(xmlFile)
+
+	var c CPEDict
+	xml.Unmarshal(byteValue, &c)
+
+	return c, nil
+}
+
+func main() {
```
oliverchang, line 100: Maybe add a line or two of comments explaining what the output of this is. It wasn't clear that the immediate output of this is the description frequency analysis, as opposed to the CPE -> repo mappings themselves.

vulnfeeds/cmd/cperepos/main.go
```diff
@@ -0,0 +1,184 @@
...
+		CPE, err := cves.ParseCPE(c.CPE23.Name)
+		if err != nil {
+			Logger.Infof("Failed to parse %q", c.CPE23.Name)
+			continue
+		}
+		if CPE.Part != "a" {
+			// Not interested in hardware or operating systems.
+			continue
+		}
+		if _, exists := ProductToRepo[CPE.Product]; !exists {
```
oliverchang, line 125: Do we not need to qualify product with the vendor?

oliverchang (approved):
nice! some questions + minor suggestions

vulnfeeds/cmd/cperepos/main.go
andrewpollock: I wrote this before I added the lines immediately before it, because I'd found `Advisory` references would bring in  https://github.com/CVEProject/cvelist for everything. I subsequently added the denylist, because it was still creeping in in a few places anyway.

Trialling the removal of the allowlist brought in another 1,030 product/repo combos, so I'll pull out the allowlist in favour of the denylist.

vulnfeeds/cmd/cperepos/main.go
andrewpollock: I will ditch this entirely in favour of emitting a CSV file in addition to the JSON file in the followup PR I'll make.

vulnfeeds/cmd/cperepos/main.go
andrewpollock: Removed in c6b70ea

PR #851, opened 2022-11-23
Make Repo() correctly handle being called with a bare repo URL

andrewpollock:
cperepos may call it with an exact bare repo URL, which then gets incorrectly rejected.


andrewpollock:
Obsoleted

PR #854, opened 2022-11-25
Improve invalid repo exclusion based on eyeballing the results

andrewpollock:
This is mostly compensating for the removal of the allowlist of reference descriptions ("Advisory" in particular)

### DIFF
--- a/vulnfeeds/cmd/cperepos/main.go
+++ b/vulnfeeds/cmd/cperepos/main.go
@@ -36,6 +36,7 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"regexp"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


vulnfeeds/cmd/cperepos/main.go
```diff
@@ -77,10 +78,43 @@ const (
 var (
 	Logger utility.LoggerWrapper
 	// These repos should never be considered authoritative for a product.
+	// TODO(apollock): read this from an external file
 	InvalidRepos = []string{
+		"https://github.com/abhiunix/goo-blog-App-CVE",
```
oliverchang, line 83: Do we need all of these, if you already have the Regex to look for CVEs in the repo name?

oliverchang (approved):

vulnfeeds/cmd/cperepos/main.go
andrewpollock: My regex is moderately conservative to avoid potential false positives

PR #856, opened 2022-11-25
Handle the special case of cgit.freedesktop.org

andrewpollock:
References to cgit.freedesktop.org have been discovered in the wild, and this is doesn't quite walk and talk like other cGit webservers. Furthermore, it's a mirror of gitlab.freedesktop.org, so canonicalise to this in case callers want to clone the repo returned

### DIFF
--- a/vulnfeeds/cves/versions.go
+++ b/vulnfeeds/cves/versions.go
@@ -95,13 +95,17 @@ func Repo(u string) (string, error) {
 		if strings.HasSuffix(parsedURL.Path, "commit/") &&
 			strings.HasPrefix(parsedURL.RawQuery, "id=") {
 			repo := strings.TrimSuffix(parsedURL.Path, "/commit/")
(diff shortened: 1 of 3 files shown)


vulnfeeds/cves/versions.go
```diff
@@ -89,6 +89,26 @@ func Repo(u string) (string, error) {
 			parsedURL.Hostname(), repo), nil
 	}
 
+	// cgit.freedesktop.org is a special snowflake with enough repos to warrant special handling
+	// it is a mirror of gitlab.freedesktop.org
```
oliverchang, line 93: Can you add some examples in comments to demonstrate the variations of cgit.freedesktop.org URLs you're handling here?

oliverchang (approved):

vulnfeeds/cves/versions.go
andrewpollock: Done in 646884ca7bb9b91ff1ca1fdce1780c189e080f0d

PR #867, opened 2022-11-29
Pipfile.lock was out of date, breaking things

andrewpollock:
https://pantheon.corp.google.com/cloud-build/builds/10a63863-bf1c-41c7-a424-7b0c57cc655a?project=oss-vdb


docker/worker/Pipfile.lock
```diff
@@ -894,7 +810,7 @@
                 "sha256:ee6acae74a2b91865910eef5e7de37dc6895ad96fa23603d1d27ea69df545015",
                 "sha256:ef3f72c9666bba2bab70d2a8b79f2c6d2c1a42a7f7e2b0ec83bb2f9e383950af"
             ],
-            "markers": "python_version >= '2.7' and python_version not in '3.0, 3.1, 3.2, 3.3, 3.4'",
+            "markers": "python_version < '3.11'",
```
another-rex, line 813: Ooo... this might cause some issues, since worker is running python 3.11?

another-rex: Why did this requirement suddenly change though...

andrewpollock: I'm assuming it's because of `osv = {editable = true, path = "./../.."}` and https://github.com/google/osv.dev/blob/e21c5dd7056983159901811770055312cf64b0f4/Pipfile.lock#L703 has this in it?

another-rex (approved):
Seems like it's on a dependency that's only for devdependencies, so should be fine for release.

PR #869, opened 2022-11-30
Add some basic metrics capture

andrewpollock:
Also pave the way for refactoring out some of the repository determination over to the CPE Dictionary parser.

### DIFF
--- a/vulnfeeds/cpp/main.go
+++ b/vulnfeeds/cpp/main.go
@@ -181,7 +181,7 @@ func InScopeRepo(repoURL string) bool {
 }
 
 // Takes an NVD CVE record and outputs an OSV file in the specified directory.
-func CVEToOSV(CVE cves.CVEItem, repo, directory string) error {
+func CVEToOSV(CVE cves.CVEItem, directory string) error {
 	CPEs := cves.CPEs(CVE)
 	CPE, err := cves.ParseCPE(CPEs[0])
 	if err != nil {
@@ -199,7 +199,7 @@ func CVEToOSV(CVE cves.CVEItem, repo, directory string) error {
 		return fmt.Errorf("No affected versions detected for %s for %q", CVE.CVE.CVEDataMeta.ID, CPE.Product)
 	}
 
-	vulnDir := filepath.Join(directory, CPE.Product)
+	vulnDir := filepath.Join(directory, CPE.Vendor, CPE.Product)
 	err = os.MkdirAll(vulnDir, 0755)
 	if err != nil {
 		Logger.Warnf("Failed to create dir: %v", err)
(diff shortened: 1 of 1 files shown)


vulnfeeds/cpp/main.go
```diff
@@ -174,11 +181,12 @@ func InScopeRepo(repoURL string) bool {
 }
 
 // Takes an NVD CVE record and outputs an OSV file in the specified directory.
-func CVEToOSV(CVE cves.CVEItem, repo, directory string) {
+func CVEToOSV(CVE cves.CVEItem, directory string) error {
 	CPEs := cves.CPEs(CVE)
 	CPE, err := cves.ParseCPE(CPEs[0])
 	if err != nil {
-		Logger.Fatalf("Can't generate an OSV record for %s without valid CPE data", CVE.CVE.CVEDataMeta.ID)
+		Logger.Warnf("Can't generate an OSV record for %s without valid CPE data", CVE.CVE.CVEDataMeta.ID)
```
oliverchang, line 188: This is pretty similar to the error that's returned. Should we just have the caller log this?

vulnfeeds/cpp/main.go
```diff
@@ -187,26 +195,30 @@ func CVEToOSV(CVE cves.CVEItem, repo, directory string) {
 	v.Affected = append(v.Affected, affected)
 
 	if len(v.Affected[0].Ranges) == 0 {
-		Logger.Infof("No affected versions detected for %s for %q", CVE.CVE.CVEDataMeta.ID, CPE.Product)
-		return
+		Logger.Warnf("No affected versions detected for %s for %q", CVE.CVE.CVEDataMeta.ID, CPE.Product)
```
oliverchang, line 198: Same comment as above. We can just have a single log call from the caller instead of all this duplicaiton.

vulnfeeds/cpp/main.go
```diff
@@ -235,16 +247,18 @@ func main() {
 		Logger.Fatalf("Failed to parse NVD CVE JSON: %v", err)
 	}
 
+	// TODO(apollock): preload with CPE Dictionary preprocessed JSON
+	VPRepoCache := make(map[string]map[string]string)
```
oliverchang, line 251: What does "VP" mean in this case?

andrewpollock: "Vendor/Product"

vulnfeeds/cpp/main.go
andrewpollock: My thinking had been to log as close to the source of the error, not the call site... But I've done as you've requested.

vulnfeeds/cpp/main.go
andrewpollock: Done.

vulnfeeds/cpp/main.go
oliverchang: In general, the way to pinpoint where things are going wrong is by adding additional context by wrapping (see  
 https://go.dev/blog/go1.13-errors).

oliverchang (approved):

PR #870, opened 2022-11-30
Some more instructions, and add a ID to allow links to specific section of page

another-rex:
Some more instructions, and add a ID to allow links to specific section of page. Also adds the term OSV-Scanner to the main page.

### DIFF
--- a/gcp/appengine/frontend3/src/templates/home.html
+++ b/gcp/appengine/frontend3/src/templates/home.html
@@ -157,14 +157,15 @@
         </div>
       </div>
     </div>
-    <div id="use-the-api" class="mdc-layout-grid__cell--span-12 usage-examples">
-      <h2 class="heading">Command line tools</h2>
+    <div id="use-the-cli" class="mdc-layout-grid__cell--span-12 usage-examples">
+      <h2 class="heading">Command Line Tools (OSV-Scanner)</h2>
       <div class="cmdline code-card mdc-layout-grid__cell--span-12 mdc-layout-grid__cell--span-12-tablet">
         <h3 class="code-card-title">Scan an SBOM or Lockfile (preview)</h3>
         <pre class="code-card-content" id="example-sbom-scan">
 go install github.com/google/osv-scanner/cmd/osv-scanner@latest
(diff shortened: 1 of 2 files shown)


andrewpollock (approved):

gcp/appengine/frontend3/src/templates/home.html
```diff
@@ -157,14 +157,15 @@ <h3 class="code-card-title">Query by version number</h3>
...
-      <h2 class="heading">Command line tools</h2>
+    <div id="use-the-cli" class="mdc-layout-grid__cell--span-12 usage-examples">
+      <h2 class="heading">Command Line Tools (OSV-Scanner)</h2>
       <div class="cmdline code-card mdc-layout-grid__cell--span-12 mdc-layout-grid__cell--span-12-tablet">
         <h3 class="code-card-title">Scan an SBOM or Lockfile (preview)</h3>
         <pre class="code-card-content" id="example-sbom-scan">
 go install github.com/google/osv-scanner/cmd/osv-scanner@latest
 $GOPATH/bin/osv-scanner --sbom=cycloned-or-spdx-sbom.json
 $GOPATH/bin/osv-scanner --lockfile=package-lock.json
+$GOPATH/bin/osv-scanner -r path/to/your/project
```
oliverchang, line 168: Would it make sense to add a "# Recursively scan entire local directory" explanation for this?

another-rex:
I went ahead and did some more changes to improve the command line section a bit, it now looks like this:

Fixes:
- Use non breaking hyphen for osv-scanner
- Slightly increases the size of the install command box to remove the horizontal scrollbar when at full size
- Separated the commands into 3 boxes,1 for install, and 2 boxes for different scanning commands. 
- Added a small margin under the subheading so that the command line part is not so close to the heading. 

image

More vertical viewing now looks like this:
image

gcp/appengine/frontend3/src/templates/home.html
another-rex: Separated it into a different box, which might look a bit better.

oliverchang (approved):

PR #875, opened 2022-12-01
Add filter for ecosystem count

another-rex:
This removes the UVI and GSD ecosystem entries on the vulnerabilities list. They still show up in the ecosystem list because there are entries that are marked as invalid.

We would require adding an additional index to have the `osv_get_ecosystems` function return only ecosystems with non invalid bugs.


oliverchang (approved):

andrewpollock:
Can you add a bit of "why" to the PR description?

andrewpollock (approved):

PR #877, opened 2022-12-01
Only consult Debian as a last resort

andrewpollock:
Also add a few more known-bad repos to ignore

### DIFF
--- a/vulnfeeds/cmd/cperepos/main.go
+++ b/vulnfeeds/cmd/cperepos/main.go
@@ -43,7 +43,6 @@ import (
 	"os"
 	"path"
 	"path/filepath"
-	"reflect"
 	"regexp"
 	"sort"
 	"strconv"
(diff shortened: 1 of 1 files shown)


vulnfeeds/cmd/cperepos/main.go
```diff
@@ -318,6 +332,7 @@ func MaybeGetSourceRepoFromDebian(mdir string, pkg string) string {
 func analyzeCPEDictionary(d CPEDict) (ProductToRepo VendorProductToRepoMap, DescriptionFrequency map[string]int) {
 	ProductToRepo = make(VendorProductToRepoMap)
 	DescriptionFrequency = make(map[string]int)
+	MaybeTryDebian := make(map[string]map[string]bool)
```
oliverchang, line 335: Rather than do this complicated nesting, can we just create a struct to include both the Vendor and Product? We should be able to use that as a map key as well.

andrewpollock: Done in cb8158aec6e877cc7d19f6b8ce0d76466e9f2de3

oliverchang (approved):

PR #878, opened 2022-12-02
Add beginnings of Terraform configuration for setting up GCP Project

michaelkedar:
Beginnings of Terraform setup for managing oss-vdb and oss-vdb-test. Currently only sets up what is needed for the workers on GKE.

Formatting and structure of terraform files will probably need refactoring.
I've temporarily copied and changed a few yaml files from osv-config to help deploy onto the oss-vdb-test project.

Mostly still a WIP.


deployment/README.md
```diff
@@ -0,0 +1,92 @@
+# GCP Project Setup (WIP)
+
+Ensure the `GOOGLE_CLOUD_PROJECT` environment variable is exported for the project:
+```bash
+export GOOGLE_CLOUD_PROJECT=<PROJECT_ID>
+```
+The official project id of OSV is `oss-vdb`
```
oliverchang, line 7: nit: end with period.

deployment/cloudbuild/cloudbuild.yaml
```diff
@@ -0,0 +1,51 @@
+steps:
```
oliverchang, line 1: Can you put all of this under an "deployment/oss-vdb-test/..." folder?

deployment/main.tf
```diff
@@ -0,0 +1,182 @@
...
+    disk_type       = "pd-ssd"
+    disk_size_gb    = 64
+    local_ssd_count = 1
+
+    oauth_scopes = ["https://www.googleapis.com/auth/cloud-platform"]
+
+  }
+}
+
+resource "google_container_node_pool" "highend2" {
```
oliverchang, line 92: nit: can we just call this "highend" instead of "highend2" ?

deployment/main.tf
```diff
@@ -0,0 +1,182 @@
...
+  region                   = "us-central1"
+}
+
+resource "google_compute_router" "router" {
+  name    = "router"
+  network = "default"
+  region  = "us-central1"
+}
+
+resource "google_compute_router_nat" "nat-config" {
```
oliverchang, line 31: very nice! did you test that internet access works?

oliverchang (commented):
some initial comments! very nice work!!!

deployment/main.tf
michaelkedar: Yep! The worker was able to pull the oss-fuzz repo from github

deployment/README.md
```diff
@@ -0,0 +1,92 @@
...
+```
+
+Then, inside `docker/worker/`
+```bash
+./build.sh <TAG>
+```
+
+## Deploy builds to cluster
+
+First, edit the image path in `deployment/oss-vdb-test/cloudbuild/gke/workers/workers.yaml` to match the project ID:
```
oliverchang, line 62: Can you add a TODO to use https://kubernetes.io/docs/tasks/manage-kubernetes-objects/kustomization/ ?

deployment/README.md
```diff
@@ -0,0 +1,92 @@
...
+
+The following instructions are from [here](https://cloud.google.com/kubernetes-engine/docs/tutorials/external-metrics-autoscaling#step1).
+
+```bash
+kubectl create clusterrolebinding cluster-admin-binding \
+    --clusterrole cluster-admin --user "$(gcloud config get-value account)"
+```
+
+```bash
+kubectl apply -f https://raw.githubusercontent.com/GoogleCloudPlatform/k8s-stackdriver/master/custom-metrics-stackdriver-adapter/deploy/production/adapter_new_resource_model.yaml
```
oliverchang, line 41: This is different from the one we actually use in our production project:

```
https://raw.githubusercontent.com/GoogleCloudPlatform/k8s-stackdriver/master/custom-metrics-stackdriver-adapter/deploy/production/adapter.yaml
```

Is there a meaningful difference between the two?

deployment/README.md
```diff
@@ -0,0 +1,92 @@
...
+In this directory (`/deployment`):
+
+Initialise terraform:
+```bash
+terraform init
+```
+
+Plan and apply required project infastructure:
+```bash
+terraform plan
```
oliverchang, line 20: Can you add some more detail on what `plan` does and what we should be doing with the output of that? Is this just to verify that the things it will do look sane, before actually applying them?

andrewpollock: If there's any good outlinks you can include at these steps to help an unfamiliar reader understand the "why" of these that would be helpful too.

michaelkedar: > Is this just to verify that the things it will do look sane, before actually applying them?

Yeah. `plan` is not actually necessary to run, but it shows you what resources would be added / changed / deleted with the configuration.
The output of plan is also shown when you run `apply`, which prompts you to accept the proposed changes.

I'll convey this in the README.

> If there's any good outlinks you can include at these steps to help an unfamiliar reader understand the "why" of these that would be helpful too.

I don't know if I've seen any good links, but will add more "why" information.

deployment/README.md
michaelkedar: I didn't notice that. The [linked steps](https://cloud.google.com/kubernetes-engine/docs/tutorials/autoscaling-metrics#pubsub_1) point to using the new resource model for pubsub, rather than that legacy one.

From the [GitHub page](https://github.com/GoogleCloudPlatform/k8s-stackdriver/tree/master/custom-metrics-stackdriver-adapter#configure-cluster):
> Stackdriver supports two models of Kubernetes resources: **the legacy model**
using monitored resource `gke_container` and **the new model** using different
Kubernetes monitored resources, including for example `k8s_pod`, `k8s_node`. See
[monitored resources documentation](https://cloud.google.com/monitoring/api/resources) for more details.

I don't know what actual difference this makes, but I can change this to the legacy one if that would be better.

deployment/oss-vdb-test/cloudbuild/gke/workers/workers.yaml
```diff
@@ -0,0 +1,72 @@
...
+        # env:
+        #   - name: DEPS_DEV_API_KEY
+        #     valueFrom:
+        #        secretKeyRef:
+        #         name: secrets
+        #         key: deps.dev
+        # args:
+        #   - "--ssh_key_public=/secrets/ssh.pub"
+        #   - "--ssh_key_private=/secrets/ssh"
+        #   - "--redis_host=10.102.25.213"
```
oliverchang, line 44: This (setting up the Cloud Memorystore Redis instance) should be another TODO :)

deployment/README.md
oliverchang: Did you verify if the autoscaling works? I think @another-rex tried the non-legacy one a while back on the production project and it didn't work. 

Fine to keep this as is for now, just flagging that this may need a bit more investigation.

deployment/oss-vdb-test/cloudbuild/gke/workers/workers.yaml
michaelkedar: TODO: everything

deployment/README.md
michaelkedar: Noted.
It did scale down from 4 nodes to 1 when I activated it, and I don't think it's spat out any errors. I've not tried publishing anything to see if it scales up, though.

oliverchang (approved):
Thanks! this looks pretty good. just some final typos to fix.

PR #890, opened 2022-12-07
Fix alpine enumeration with release-versions

another-rex:
Now alpine will enumerate release versions, should allow vulnerabilities to specify versions in the format `v1.2.3-r4` to be successfully parsed.


osv/third_party/univers/alpine.py
```diff
@@ -5,6 +5,7 @@
 # Visit https://aboutcode.org and
 # https://github.com/nexB/univers for support and download.
 import attr
+import re
```
oliverchang, line 8: Can you add a comment here describing the changes you made? i.e. supporting revisions?

osv/ecosystems.py
```diff
@@ -445,39 +449,54 @@ def sort_key(self, version):
...
     current_ver = None
+    current_rel = '0'
+
+    def clean_versions(ver: str) -> str:
+      ver = ver.split(' #')[0]  # Remove comment lines
+      ver = ver.strip(' "\'')  # Remove (occasional) quotes
+      ver = ver.removeprefix('r')
+      return ver
 
     for x in lines:
```
oliverchang, line 473: missed this earlier, but can we rename `x` to `line` ?

osv/ecosystems.py
```diff
@@ -445,39 +449,54 @@ def sort_key(self, version):
...
       if len(x) == 0:
```
oliverchang, line 474: Can we do

```python
if not line:
``` 

instead per https://google.github.io/styleguide/pyguide.html#214-truefalse-evaluations ?

osv/ecosystems.py
```diff
@@ -445,39 +449,54 @@ def sort_key(self, version):
...
-        if current_ver is None:
+        if current_ver is None or current_rel is None:
```
oliverchang, line 475: Can we do implicit false value comparison here?

```python
if not current_ver or not current_rel:
```

Also can current_rel ever be `None` in practice? It starts of as '0', and all other assignments seem to only be assigning strings.

oliverchang (approved):
LGTM with some comments.

osv/ecosystems.py
```diff
@@ -76,7 +76,7 @@ def _before_limits(self, version, limits):
   def next_version(self, package, version):
     """Get the next version after the given version."""
     versions = self.enumerate_versions(package, version, fixed=None)
-    if versions and versions[0] != version:
+    if versions and self.sort_key(versions[0]) != self.sort_key(version):
```
oliverchang, line 79: Can you explain a bit more what this was running into before? Good to add a comment here to add more context for future readers.

PR #891, opened 2022-12-07
Make Python building use multiple cores

michaelkedar:
For worker-base

### DIFF
--- a/docker/worker-base/build.sh
+++ b/docker/worker-base/build.sh
@@ -18,6 +18,6 @@ cd ../../
 
 PROJECT=${GOOGLE_CLOUD_PROJECT:-oss-vdb}
 
-docker build -t gcr.io/$PROJECT/worker-base:$1 -t gcr.io/$PROJECT/worker-base:latest -f docker/worker-base/Dockerfile . && \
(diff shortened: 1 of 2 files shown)


docker/worker-base/build.sh
```diff
@@ -18,6 +18,6 @@ cd ../../
 
 PROJECT=${GOOGLE_CLOUD_PROJECT:-oss-vdb}
 
-docker build -t gcr.io/$PROJECT/worker-base:$1 -t gcr.io/$PROJECT/worker-base:latest -f docker/worker-base/Dockerfile . && \
+docker build -t gcr.io/$PROJECT/worker-base:$1 -t gcr.io/$PROJECT/worker-base:latest -f docker/worker-base/Dockerfile --pull . && \
```
oliverchang, line 21: What does the `--pull` do here?

oliverchang (approved):
thanks!!

docker/worker-base/build.sh
michaelkedar: It tells it to always grab the latest version of the base image (ubuntu:20.04), rather than using the cached one.

PR #907, opened 2022-12-13
Update install instructions for osv-scanner

another-rex:


gcp/appengine/frontend3/src/templates/home.html
```diff
@@ -165,7 +165,7 @@ <h2 class="heading">Command Line Tools (OSV&#8209;Scanner)</h2>
         <div class="code-card mdc-layout-grid__cell--span-6 mdc-layout-grid__cell--span-12-tablet">
           <h3 class="code-card-title">Install OSV&#8209;Scanner</h3>
           <pre class="code-card-content" id="example-scanner-install">
-go install github.com/google/osv-scanner/cmd/osv-scanner@latest
+go install github.com/google/osv-scanner/cmd/osv-scanner@v1.0.0
```
oliverchang, line 168: can we just do "@v1" ? so that users get the latest 1.x ?

oliverchang (approved):

gcp/appengine/frontend3/src/templates/home.html
another-rex: Done

PR #914, opened 2022-12-14
Add exporter support for oss-vdb-test

michaelkedar:


deployment/terraform/environments/oss-vdb-test/main.tf
```diff
@@ -3,7 +3,8 @@ module "osv_test" {
 
   project_id = "oss-vdb-test"
 
-  public_import_logs_bucket = "osv-test-public-import-logs"
+  public_import_logs_bucket  = "osv-test-public-import-logs"
+  osv_vulnerabilities_bucket = "osv-test-vulnerabilities"
```
oliverchang, line 7: nit: let's rename this to "vulnerabilities_export_bucket" to be clearer what this is.

oliverchang (approved):

deployment/oss-vdb-test/cloudbuild/gke/exporter/exporter.yaml
```diff
@@ -0,0 +1,33 @@
+apiVersion: batch/v1
```
oliverchang, line 1: Actually (can be for a future PR), can we move these GKE configs into `oss-vdb-test/k8s` instead living in this `cloudbuild` directory which is inaccurate?

michaelkedar: I'll just do this in this PR quickly

oliverchang (approved):

PR #917, opened 2022-12-14
Make App Engine deployment more re-usable.

oliverchang:
- Move App Engine YAML files to deployment/<project>/gae.
- Modify the App Engine deployment script to symlink in the YAML files
  from the source config directory prior to deployment. This is
  necessary because `gcloud app deploy` requires the yaml files to be
  in the same directory as the application root.
- Also add GAE configs for "oss-vdb-test".


andrewpollock:
> * gcloud auth deploy

`gcloud app deploy`?

deployment/staging.yaml
```diff
@@ -31,8 +31,7 @@ steps:
   entrypoint: 'bash'
   args: [ '-c', "rm /workspace/service_account.json" ]
 - name: 'gcr.io/oss-vdb/deployment'
-  args: ['bash', '-ex', 'deploy.sh', 'oss-vdb', 'app.yaml', '--no-promote']
-  dir: gcp/appengine
+  args: ['bash', '-ex', 'gcp/appengine/deploy.sh', 'oss-vdb', 'deployment/oss-vdb/gae', 'app.yaml', '--no-promote']
```
andrewpollock, line 34: I think the project ID here should be `oss-vdb-staging` based on the filename?

andrewpollock:
> `gcloud app deploy`?

Pretty confident the answer is "yes", so I'll update the description accordingly.

michaelkedar (approved):

deployment/staging.yaml
oliverchang: This is the staging instance within the prod GCP project, so this is intentional.

oliverchang:
> > `gcloud app deploy`?
> 
> Pretty confident the answer is "yes", so I'll update the description accordingly.

Yep thanks !!!

PR #919, opened 2022-12-16
Add pubsub service account permissions to terraform config

michaelkedar:
Fixes issues with dead lettering not working


oliverchang (approved):
thanks!

PR #921, opened 2022-12-19
Add support for no underscore p alpine version variants

another-rex:
From the reply here: https://gitlab.alpinelinux.org/alpine/abuild/-/issues/10088, it seems the _ is not required, this PR relaxes the requirement by inserting an underscore if it is not found.


osv/third_party/univers/alpine.py
```diff
@@ -146,19 +146,26 @@ def __ge__(self, other):
...
+  # Some suffixes are not separated with an underscore. E.g. 1.9.5p2
+  # This should find them for inserting an underscore (replace with r'\1_\2')
+  patch_finder = re.compile(r'(\d+)(p\d+)')
+
   @classmethod
   def build_value(cls, string: str):
     search = cls.version_extractor.search(string)
-    return (search.group(1), int(search.group(2) or 0))
+    main_group = search.group(1)
+    main_group_patched = re.sub(cls.patch_finder, r'\1_\2', main_group, 1)
```
oliverchang, line 161: nit: instead of `patch_finder`,  should we just make a classmethod (`add_underscores`) that can be called in both places here, instead of duplicating the re.sub call?

oliverchang (approved):
LGTM with one suggestion

PR #924, opened 2022-12-19
Add API to oss-vdb-test

michaelkedar:
The `oss-vdb-test` API is currently being served on `https://osv-grpc-v1-ivqje5d2hq-uc.a.run.app/`, and requires authorization to access.

For `curl`, you can authorization with `-H "Authorization: Bearer $(gcloud auth print-identity-token)"`


gcp/api/v1/README.md
```diff
@@ -25,16 +25,26 @@ To deploy the service proxy,
...
+../deploy_service_proxy oss-vdb api.osv.dev osv-grpc-v1 /path/to/api_config.yaml
 ```
+
+### oss-vdb-test Project
+
+```
+../deploy_service_proxy oss-vdb-test osv-grpc-v1-ivqje5d2hq-uc.a.run.app osv-grpc-v1 /path/to/api_config.yaml
+```
+
+Configuring the Cloud Endpoint requires the Cloud Run service to already exist before running, as it needs the URL.
```
oliverchang, line 48: Is Cloud Run something that can be managed with terraform as well? Does it make sense to?

oliverchang (approved):
awesome! 
LGTM with a question.

gcp/api/v1/README.md
michaelkedar: [There is a terraform resoure for it](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/cloud_run_service).
I'll take a look into it - I'm not sure how updating the container image interacts with terraform.

michaelkedar: The terraform resource for the cloud run service requires the container image for the service, and changing outside of terraform (as `deploy_service_proxy` does) causes problems.

I suppose we could separate the building and deploying of the container image, and use terraform solely to deploy image versions, but that would be a significant workflow change that can be done in another PR.

PR #925, opened 2022-12-19
Only filter ecosystems

andrewpollock:
OSV records converted from NVD CVE entries do not have a package,
therefore they do not have an ecosystem, and are getting filtered out.

This adds a test for this type of vulnerability, and modifies the
ecosystem filtering behaviour to accommodate these vulnerabilities.

I also discovered that the existing
test_update_partly_bad_ecosystem_delete  was not indented correctly and
not being run, and when it is, fails unexpectedly.


docker/worker/worker_test.py
```diff
@@ -1250,6 +1250,36 @@ def test_update_partly_bad_ecosystem_delete(self):
...
+    self.source_repo.editable = False
+    self.source_repo.put()
+
+    task_runner = worker.TaskRunner(ndb_client, None, self.tmp_dir.name, None,
+                                    None)
+
+    message = mock.Mock()
+    message.attributes = {
+        'source': 'source',
+        'path': 'a/b/CVE-2022-0128.json',
```
oliverchang, line 1266: Does this file need to be checked in?

andrewpollock: It's sitting in the test GCS bucket, which seemed how the other similar tests operated (I'm thinking the bucket needs to be mocked out in future so these tests are hermetic).

oliverchang: Got it, missed that earlier thanks.

docker/worker/worker.py
```diff
@@ -322,6 +322,9 @@ def filter_unsupported_ecosystems(vulnerability):
   for affected in vulnerability.affected:
     if osv.ecosystems.get(affected.package.ecosystem):
       filtered.append(affected)
+    # CVE-converted OSV records have no package information.
+    if affected.package.ecosystem == '':
```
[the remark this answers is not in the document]
oliverchang: `affected.HasField('package')` communicates the intention of what you're trying to do much better. i.e. allow in cases where there is just no `package` specified to begin with. 

`affected.HasField('package')` works differently than your check: it returns false only if no fields in the nested structure are set. In the example I gave, it would return true because one of the fields is set, which fails the check.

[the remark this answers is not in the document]
andrewpollock: Got it, thanks for clarifying. Addressed in https://github.com/google/osv.dev/pull/925/commits/2ecf0d4443aa6fd0670d1258bb44e0502cc25be4

oliverchang (approved):

PR #939, opened 2023-01-03
Fix semver integration test

andrewpollock:
The current state of https://github.com/advisories/GHSA-hrm3-3xm6-x33h breaks the test, so use a
different package altogether. Adjust for test scenarios accordingly.


oliverchang:
Hmm, still seems like it's failing on https://pantheon.corp.google.com/cloud-build/builds/ec3a5df2-d29f-4205-bc76-43967efba578;step=6?project=oss-vdb ?

andrewpollock:
Yes, the problem is that 

```
{
            'version': '0.0.0-20200131131040-063a3fb69896',
            'package': {
                'name': 'github.com/nanobox-io/golang-nanoauth',
                'ecosystem': 'Go',
            }
}
```

is now returning `GHSA-hrm3-3xm6-x33h`, when it previously returned nothing. I think the problem is actually with how GHSA-hrm3-3xm6-x33h is expressing itself?

oliverchang (approved):

PR #945, opened 2023-01-05
Use the precomputed CPE-to-repo map as a cache

andrewpollock:
Look up the repo mapping first, before doing CVE-specific reference
examination.

For the 2022 CVE data, and the current state of the CPE Dictionary,
there's still an additional 1,629 potential repositories picked up from
CVE references, and an additional 210 OSV records generated, so it's
premature to exclusively rely on the CPE Dictionary.


vulnfeeds/cpp/main.go
```diff
@@ -97,8 +112,18 @@ func CVEToOSV(CVE cves.CVEItem, directory string) error {
 	return nil
 }
 
+func loadCPEDictionary(ProductToRepo *VendorProductToRepoMap, f string) error {
+	data, err := ioutil.ReadFile(f)
+	if err != nil {
+		return err
+	}
+	err = json.Unmarshal(data, &ProductToRepo)
```
oliverchang, line 120: nit: just return the result of `json.Unmarshal` directly.

oliverchang (approved):
LGTM with a nit.

PR #949, opened 2023-01-05
Don't run the integration tests on every PR

andrewpollock:
The integration tests can randomly fail because they're operating on live data that can change unexpectedly. This results in tests failing on a PR through no fault of the code in the PR. We already run the integration tests on every *commit*, so it's just a matter of surfacing those failures to the people who can do something about them.


oliverchang:
Thinking more, would it make sense to still make this run (but have passing be optional?). 

This can be useful for making sure API changes don't break things before they land.

andrewpollock:
> This can be useful for making sure API changes don't break things before they land.

Is there a way we can confine it to only run when there *are* API changes?

oliverchang:
> > This can be useful for making sure API changes don't break things before they land.
> 
> Is there a way we can confine it to only run when there _are_ API changes?

Not sure if there's an easy way, but if it's non blocking it should be a minor issue even if it does run.

PR #952, opened 2023-01-06
Create short redirect link format for vulnerabilities.

### LINKED ISSUE #930: Create shorter redirect link format for vulnerabilities
e.g. `https://osv.dev/v/PYSEC-foo` should redirect to the full `https://osv.dev/vulnerability/PYSEC-foo`.

This can help with osv-scanner's human readable output, where horizontal space is a premium.

oliverchang:
https://osv.dev/VULN-ID would redirect to
https://osv.dev/vulnerability/VULN-ID, if VULN-ID exists.

Fixes #930.

### DIFF
--- a/gcp/appengine/frontend_handlers.py
+++ b/gcp/appengine/frontend_handlers.py
@@ -171,6 +171,17 @@ def vulnerability(vuln_id):
   return render_template('vulnerability.html', vulnerability=vuln)
 
 
+@blueprint.route('/<potential_vuln_id>')
+def vulnerability_redirector(potential_vuln_id):
+  """Convenience redirector for /VULN-ID to /vulnerability/VULN-ID."""
+  vuln = osv_get_by_id(potential_vuln_id)
+  if vuln:
+    return redirect(f'/vulnerability/{potential_vuln_id}')
+
+  abort(404)
+  return None
+
+
 def bug_to_response(bug, detailed=True):
   """Convert a Bug entity to a response object."""
   response = osv.vulnerability_to_dict(bug.to_vulnerability())
--- a/gcp/appengine/frontend_handlers.py
+++ b/gcp/appengine/frontend_handlers.py
@@ -40,7 +40,9 @@ blueprint = Blueprint('frontend_handlers', __name__)
 _PAGE_SIZE = 16
 _PAGE_LOOKAHEAD = 4
 _REQUESTS_PER_MIN = 30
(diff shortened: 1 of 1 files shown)


gcp/appengine/frontend_handlers.py
```diff
@@ -171,6 +171,17 @@
   return render_template('vulnerability.html', vulnerability=vuln)
 
 
+@blueprint.route('/<potential_vuln_id>')
+def vulnerability_redirector(potential_vuln_id):
+  """Convenience redirector for /VULN-ID to /vulnerability/VULN-ID."""
+  vuln = osv_get_by_id(potential_vuln_id)
+  if vuln:
+    return redirect(f'/vulnerability/{potential_vuln_id}')
```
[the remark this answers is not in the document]
oliverchang: This isn't valid, because we already check it's a valid ID before we redirect.

That said, I'll add a regex match to be extra safe, and that can also save a Datastore read when the path is completely bogus.

gcp/appengine/frontend_handlers.py
```diff
@@ -40,7 +40,9 @@
 _PAGE_SIZE = 16
 _PAGE_LOOKAHEAD = 4
 _REQUESTS_PER_MIN = 30
-_VALID_BLOG_NAME = re.compile(r'^[\w-]+$')
+_WORD_CHARACTERS_OR_DASH = re.compile(r'^[\w-]+$')
+_VALID_BLOG_NAME = _WORD_CHARACTERS_OR_DASH
+_VALID_VULN_ID = _WORD_CHARACTERS_OR_DASH
```
andrewpollock, line 45: You could tighten this regex up a bit more to say `^\w+-[\w-]+$` and be less likely to match on things that aren't actually intended to be vulnerability IDs now, or in the future...

(aside: I can't find any signs of LBSEC IDs, even though they get a mention in the schema docs)

andrewpollock (approved):

another-rex (approved):

gcp/appengine/frontend_handlers.py
oliverchang: Yeah, but it does add a tad bit more complexity (adding a similar but different regex) for little benefit (since we ultimately validate the ID to exist in our DB), and this is a rare edge case that would save us a single Datastore entity read. 

I'll keep this as is in the interests of simplicity.

PR #961, opened 2023-01-09
Preliminary tag mapping

andrewpollock:
This is dreadfully inefficient in the long run, but great for proof-of-concept purposes.

I'm not sure why `go.sum` has gone bonkers?


vulnfeeds/cpp/main.go
```diff
@@ -73,21 +75,65 @@ func InScopeGitRepo(repoURL string) bool {
...
+		Logger.Infof("Cloning %s", repo)
+		r, err := git.Clone(memory.NewStorage(), nil, &git.CloneOptions{
+			URL: repo,
+		})
+		if err != nil {
+			Logger.Warnf("Failed clone %s: %v", repo, err)
+			continue
+		}
+		for _, av := range versions.AffectedVersions {
+			if av.Fixed == "" {
```
oliverchang, line 93: Can you add a comment on what this case means and why we skip?

vulnfeeds/cpp/main.go
```diff
@@ -73,21 +75,65 @@ func InScopeGitRepo(repoURL string) bool {
...
+				continue
+			}
+			// Need to come up with a way to decide when to prefix
+			ref, err := r.Tag("v" + av.Fixed)
```
oliverchang, line 97: Let's add a TODO to support some more sophisticated/fuzzy tag -> version matching.

vulnfeeds/cpp/main.go
```diff
@@ -73,21 +75,65 @@ func InScopeGitRepo(repoURL string) bool {
...
+			if err != nil {
+				Logger.Warnf("Failed to find a tag for %q in %s: %v", av.Fixed, repo, err)
```
oliverchang, line 99: I wonder if these need to be surfaced as "notes" for potential human triagers.

oliverchang (commented):
nice! 

IIUC `go.sum` is not meant to be a lockfile -- it contains additional state necessary for https://research.swtch.com/vgo-mvs.

oliverchang (approved):
LGTM with minor nits.

vulnfeeds/cpp/main.go
andrewpollock: Yes. At the moment I'm not generating per-record notes at all, but I see the value in doing this, and will implement this in the not too distant future.

vulnfeeds/cpp/main.go
andrewpollock: Addressed in 7ed688a7d9e25ab364426d06556083b9ffa0229f

vulnfeeds/cpp/main.go
andrewpollock: Addressed in 7ed688a7d9e25ab364426d06556083b9ffa0229f

PR #965, opened 2023-01-11
Update alpine mirror

another-rex:
Changed the alpine git url, and also update existing repo url instead of throwing an exception. 

Was there a specific reason for the original code to throw an exception instead of updating if the url does not match?

### DIFF
--- a/osv/repos.py
+++ b/osv/repos.py
@@ -131,7 +131,9 @@ def _use_existing_checkout(git_url,
   repo = pygit2.Repository(checkout_dir)
   repo.cache = {}
   if repo.remotes['origin'].url != _git_mirror(git_url):
-    raise RuntimeError('Repo URL changed.')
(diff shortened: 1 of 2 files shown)


osv/repos.py
```diff
@@ -131,7 +131,9 @@ def _use_existing_checkout(git_url,
   repo = pygit2.Repository(checkout_dir)
   repo.cache = {}
   if repo.remotes['origin'].url != _git_mirror(git_url):
-    raise RuntimeError('Repo URL changed.')
+    logging.warn('origin url updated:\nOld: %s\nNew: %s',
```
andrewpollock, line 134: Nit: uppercase URL

oliverchang (approved):

andrewpollock (approved):

osv/repos.py
andrewpollock: Please also update the description to capture why you're making this URL change.

PR #970, opened 2023-01-12
Make the schema version match the current version

andrewpollock:


oliverchang (approved):

osv/models.py
```diff
@@ -30,7 +30,7 @@
 from . import sources
 from . import vulnerability_pb2
 
-SCHEMA_VERSION = '1.3.0'
+SCHEMA_VERSION = '1.3.1'
```
another-rex, line 33:
```suggestion
SCHEMA_VERSION = '1.4.0'
```

another-rex: Updated again to 1.4.0

oliverchang: We don't actually support 1.4.0 properly yet. I created https://github.com/google/osv.dev/issues/1093 to track the needed changes.

michaelkedar:
This was done in #1115

PR #971, opened 2023-01-12
Migrate unit tests from Cloud Build to GitHub Workflows

andrewpollock:
This way they get run automatically for PRs from external contributors and do not require manual triggering by repository owners.

### DIFF
--- a/.github/workflows/unittest.yaml
+++ b/.github/workflows/unittest.yaml
@@ -36,6 +36,9 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: '3.8'
+      - name: Install pipenv
+        run: |
+          python -m install --upgrade pipenv
       - name: Run top-level unittests
         run: |
           bash -ex run_tests.sh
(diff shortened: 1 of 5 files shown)


.github/workflows/unittest.yaml
```diff
@@ -0,0 +1,106 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: "OSV Unittests"
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
```
another-rex, line 20:
```suggestion
  pull_request:
  merge_group:
```

andrewpollock:
Note to self: this is likely to require some credential plumbling like that seen at https://github.com/google/clusterfuzz/blob/ca679d958a209598d30ad4bbe565ff7dd320ea29/src/clusterfuzz/_internal/google_cloud_utils/storage.py#L578

PR #972, opened 2023-01-13
Fix alpine import by filtering out invalid data

another-rex:
There are only be 7 unique versions that are invalid currently out of 2221 unique versions.


vulnfeeds/cmd/alpine/fixtures/valid_versions.txt
```diff
@@ -0,0 +1,187 @@
...
+0.105-r8
+17.2_p1-r2
+1.0.2m-r0
+11.0.2r-r0
+13.3-r0
+16.8_p1-r8
+2.7.51.0
+110.73-r0
+1
+
```
andrewpollock, line 184: Do you mean to have this blank line here?

vulnfeeds/cmd/alpine/fixtures/invalid_versions.txt
```diff
@@ -0,0 +1,7 @@
+1.18.-r2
```
andrewpollock, line 1: I've just skimmed https://github.com/alpinelinux/abuild/blob/master/APKBUILD.5.scd to try and better understand validity and invalidity.

From my skim, it's a bit light on details on how _pkgver_ and _pkgrel_ get glommed together, am I right?

It seems like _pkgrel_ itself is an ordinal number, and the "r" gets introduced as part of the concatenation with the _pkgver_ and the hyphen.

Based on that, any other letter than "r" would be, by this convention if not by the letter of what's documented, invalid. So maybe add a non-"r" version and also something without a hyphen (or underscore by the looks of it?) for completeness.

Oh wait, I see `2.19.10.4_p8` in the valid versions file.

another-rex: Added some more invalid versions examples.

`2.19.10.4_p8` specifically works because `p` is one of the allowed suffixes specified in the APKBUILD when paired with connected with an `_`. Another letter there would not work.

vulnfeeds/cmd/alpine/fixtures/valid_versions.txt
another-rex: Yep, the entries below that blank line are more borderline cases, where there is a . separating the revision part and the version, which breaks the convention set by all the other entries. Added a comment there.

andrewpollock (approved):

vulnfeeds/cmd/alpine/verify.go
```diff
@@ -0,0 +1,25 @@
+package main
+
+import "regexp"
+
+func valid_version(ver string) bool {
+	// Following the version requirement specified here: https://github.com/alpinelinux/abuild/blob/master/APKBUILD.5.scd
+	checker, err := regexp.Compile(
```
oliverchang, line 7: nit: Just use regexp.MustCompile so we don't have to add the error handling ourselves.

vulnfeeds/cmd/alpine/verify.go
```diff
@@ -0,0 +1,25 @@
+package main
+
+import "regexp"
+
+func valid_version(ver string) bool {
```
oliverchang, line 5: This should follow camelCase. i.e. `validVersion`

PR #981, opened 2023-01-16
Add cve conversion and alpine k8s tasks to osv-vdb-test

another-rex:
Adds the cve conversion release configs and process into the cloudbuild for the test environment, also sets up the cloud bucket in terraform.

### DIFF
--- a/deployment/oss-vdb-test/k8s/gke/alpine-cve-convert/alpine-cve-convert.yaml
+++ b/deployment/oss-vdb-test/k8s/gke/alpine-cve-convert/alpine-cve-convert.yaml
@@ -12,7 +12,7 @@ spec:
         spec:
           containers:
           - name: alpine-cve-convert
-            image: gcr.io/oss-vdb/alpine-cve-convert
+            image: gcr.io/oss-vdb-test/alpine-cve-convert:latest
             imagePullPolicy: Always
             securityContext:
               privileged: true
--- a/deployment/oss-vdb-test/k8s/gke/alpine-cve-convert/alpine-cve-convert.yaml
+++ b/deployment/oss-vdb-test/k8s/gke/alpine-cve-convert/alpine-cve-convert.yaml
@@ -1,0 +1,30 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: alpine-cve-convert
+spec:
+  schedule: "0 */1 * * *"
+  concurrencyPolicy: Forbid
+  jobTemplate:
+    spec:
+      activeDeadlineSeconds: 3600
+      template:
+        spec:
+          containers:
+          - name: alpine-cve-convert
+            image: gcr.io/oss-vdb/alpine-cve-convert
+            imagePullPolicy: Always
+            securityContext:
+              privileged: true
+            resources:
+              requests:
+                cpu: 1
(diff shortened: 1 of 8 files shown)


deployment/oss-vdb-test/k8s/gke/alpine-cve-convert/alpine-cve-convert.yaml
```diff
@@ -0,0 +1,30 @@
...
+  schedule: "0 */1 * * *"
+  concurrencyPolicy: Forbid
+  jobTemplate:
+    spec:
+      activeDeadlineSeconds: 3600
+      template:
+        spec:
+          containers:
+          - name: alpine-cve-convert
+            image: gcr.io/oss-vdb/alpine-cve-convert
```
michaelkedar, line 15: I believe this has to be `gcr.io/oss-vdb-test/alpine-cve-convert` for `gke-deploy` to correctly substitute the image.

deployment/oss-vdb-test/k8s/gke/combine-to-osv/combine-to-osv.yaml
```diff
@@ -0,0 +1,30 @@
...
+  schedule: "30 */1 * * *"
+  concurrencyPolicy: Forbid
+  jobTemplate:
+    spec:
+      activeDeadlineSeconds: 3600
+      template:
+        spec:
+          containers:
+          - name: combine-to-osv
+            image: gcr.io/oss-vdb/combine-to-osv
```
michaelkedar, line 15: As above - `gcr.io/oss-vdb-test/combine-to-osv`

deployment/terraform/environments/oss-vdb-test/main.tf
```diff
@@ -5,6 +5,7 @@ module "osv_test" {
 
   public_import_logs_bucket     = "osv-test-public-import-logs"
   vulnerabilities_export_bucket = "osv-test-vulnerabilities"
+  cve_osv_conversion_bucket = "cve-osv-conversion"
```
michaelkedar, line 8: GCP bucket names need to be globally unique, and this is the same name in `oss-vdb`. Terraform won't be able to create it and the tasks in `oss-vdb-test` shouldn't be modifying the bucket in `oss-vdb`.

You're going to need to change this name and parameterize anything that uses the bucket.

also nit: Equals signs should be aligned. Run `terraform fmt` in this directory (and in `modules/osv/`)

michaelkedar (changes requested):

deployment/terraform/environments/oss-vdb-test/main.tf
another-rex: I think it should be fixed now? I'll have to test it to make sure I didn't miss anything.

andrewpollock: @michaelkedar sounds like there's an opportunity to have our lint check pick up on Terraform format issues?

michaelkedar: Looks like I've been beaten to it :slightly_smiling_face: 

 The changes look good to me.

michaelkedar (approved):

PR #982, opened 2023-01-16
Make the Maven tests use deps.dev

andrewpollock:
This matches production behaviour.

Avoid leaking the API key for deps.dev in test output.

This is an interim measure to address Maven flakiness causing intermittent test failures and won't work when we move tests over to GitHub Workflows. The ultimate solution is to make the tests hermetic and mock out the network dependencies.

### DIFF
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -1,6 +1,9 @@
 #!/bin/bash -ex
 
 export PIPENV_IGNORE_VIRTUALENVS=1
+set +x  # Keep the API key out of execution logs
+export DEPSDEV_API_KEY=$(kubectl get secret secrets -o jsonpath='{.data.deps\.dev}' | base64 --decode)
+set -x
 python3 -m pipenv sync
 python3 -m pipenv run python -m unittest osv.bug_test
 python3 -m pipenv run python -m unittest osv.ecosystems_test
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -2,12 +2,12 @@
 
 export PIPENV_IGNORE_VIRTUALENVS=1
 set +x  # Keep the API key out of execution logs
-export DEPSDEV_API_KEY="$(gcloud --project oss-vdb secrets versions access latest --secret depsdotdev-key)"
-test -z "$DEPSDEV_API_KEY" && echo "No API key for deps.dev" && exit 1
+export DEPSDEV_API_KEY="$(gcloud --project oss-vdb secrets versions access latest --secret depsdotdev-key || true)"
+test -z "$DEPSDEV_API_KEY" && echo "FYI no API key for deps.dev, tests needing it will be skipped"
 set -x
 python3 -m pipenv sync
 python3 -m pipenv run python -m unittest osv.bug_test
-python3 -m pipenv run python -m unittest -v osv.ecosystems_test
+python3 -m pipenv run python -m unittest osv.ecosystems_test
 python3 -m pipenv run python -m unittest osv.maven.version_test
 python3 -m pipenv run python -m unittest osv.nuget_test
 python3 -m pipenv run python -m unittest osv.purl_helpers_test
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -2,13 +2,12 @@
 
 export PIPENV_IGNORE_VIRTUALENVS=1
 set +x  # Keep the API key out of execution logs
-# Temporarily confirm that we're getting expected output
-kubectl get secret secrets
-export DEPSDEV_API_KEY=$(kubectl get secret secrets -o jsonpath='{.data.deps\.dev}' | base64 --decode)
+export DEPSDEV_API_KEY="$(gcloud --project oss-vdb secrets versions access latest --secret depsdotdev-key)"
+test -z "$DEPSDEV_API_KEY" && echo "No API key for deps.dev" && exit 1
 set -x
 python3 -m pipenv sync
 python3 -m pipenv run python -m unittest osv.bug_test
-python3 -m pipenv run python -m unittest osv.ecosystems_test
+python3 -m pipenv run python -m unittest -v osv.ecosystems_test
 python3 -m pipenv run python -m unittest osv.maven.version_test
 python3 -m pipenv run python -m unittest osv.nuget_test
 python3 -m pipenv run python -m unittest osv.purl_helpers_test
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -2,6 +2,8 @@
 
 export PIPENV_IGNORE_VIRTUALENVS=1
 set +x  # Keep the API key out of execution logs
+# Temporarily confirm that we're getting expected output
+kubectl get secret secrets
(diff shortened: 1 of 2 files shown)


run_tests.sh
```diff
@@ -1,6 +1,9 @@
 #!/bin/bash -ex
 
 export PIPENV_IGNORE_VIRTUALENVS=1
+set +x  # Keep the API key out of execution logs
+export DEPSDEV_API_KEY=$(kubectl get secret secrets -o jsonpath='{.data.deps\.dev}' | base64 --decode)
```
oliverchang, line 5: Does this require a `gcloud container clusters get-credentials ...` call before? The secret is tied to a GKE cluster right.

andrewpollock: > Does this require a `gcloud container clusters get-credentials ...` call before? The secret is tied to a GKE cluster right.

All I can say definitively is it would appear to be working correctly in Cloud Build (I imagine because it's running on GCP, there's ambient access to the necessary credentials)

oliverchang: Interesting, I was curious because there can be multiple GKE clusters in a single GCP project, so it's not clear how this is able to determine the right one..

oliverchang: I wonder if this is just failing (i.e. the key is getting assigned an empty string) and  tests pass because you added `unittest.skipIf`. 

Doing a local  test and adding this to `run_tests.sh`:

```bash
export FOO=$(false)
```

Things seem to proceed despite the `-e`. `FOO` just gets set as empty.

andrewpollock: > I wonder if this is just failing (i.e. the key is getting assigned an empty string) and tests pass because you added `unittest.skipIf`.

I wouldn't have expected `test_maven_deps_dev` to then pass if `DEPSDEV_API_KEY` was set to an unusable value?

oliverchang: A failed kubectl get secret command seems to give me an empty string to stdout (with error logs in stderr), so it's possible this is is being passed through as an empty string?

andrewpollock: > A failed kubectl get secret command seems to give me an empty string to stdout (with error logs in stderr), so it's possible this is is being passed through as an empty string?

If that is the case, why does `test_maven_deps_dev` pass?

oliverchang: Is there a way we can list out the secrets (without revealing them) on GCB to verify this is definitely the right one?

andrewpollock: Switching over to Cloud Secrets in b06550f6db9bdb5200125117f794f130b098853a

run_tests.sh
```diff
@@ -1,9 +1,13 @@
...
 
 export PIPENV_IGNORE_VIRTUALENVS=1
+set +x  # Keep the API key out of execution logs
+export DEPSDEV_API_KEY="$(gcloud --project oss-vdb secrets versions access latest --secret depsdotdev-key)"
+test -z "$DEPSDEV_API_KEY" && echo "No API key for deps.dev" && exit 1
+set -x
 python3 -m pipenv sync
 python3 -m pipenv run python -m unittest osv.bug_test
-python3 -m pipenv run python -m unittest osv.ecosystems_test
+python3 -m pipenv run python -m unittest -v osv.ecosystems_test
```
oliverchang, line 10: Is this -v part still needed? 

IF it's useful, is this something we want to apply to all?

run_tests.sh
```diff
@@ -1,9 +1,13 @@
 #!/bin/bash -ex
 
 export PIPENV_IGNORE_VIRTUALENVS=1
+set +x  # Keep the API key out of execution logs
+export DEPSDEV_API_KEY="$(gcloud --project oss-vdb secrets versions access latest --secret depsdotdev-key)"
+test -z "$DEPSDEV_API_KEY" && echo "No API key for deps.dev" && exit 1
```
oliverchang, line 6: Can we make this optional and runnable by external/non-maintainer people?

Otherwise we'd also have to fix up https://github.com/google/osv.dev/blob/master/CONTRIBUTING.md#running-tests to say that it's only runnable for maintainers.

run_tests.sh
andrewpollock: It's (in my opinion) useful in this instance so we know which tests are being skipped, given we have two sets of tests with two different criteria for being skipped. Happy to put it on all of them for consistency, I don't mind the verbose output myself...

```
+ python3 -m pipenv run python -m unittest -v osv.ecosystems_test
test_maven_deps_dev (osv.ecosystems_test.EnumerateTest.test_maven_deps_dev)
Test Maven using deps.dev. ... ok
test_alpine (osv.ecosystems_test.GetNextVersionTest.test_alpine)
Cloning into '/workspace/osv/testdata/version_enum/aports'...
Updating files:  89% (16158/18117)
Test Alpine ecosystem enumeration and caching behaviour ... ok
test_cache (osv.ecosystems_test.GetNextVersionTest.test_cache) ... ok
test_debian (osv.ecosystems_test.GetNextVersionTest.test_debian)
Test Debian ecosystem enumeration and caching behaviour ... WARNING:root:Package nginx has invalid fixed version: <end-of-life>. In debian release 9
ok
test_gems (osv.ecosystems_test.GetNextVersionTest.test_gems)
Test RubyGems. ... ok
test_maven (osv.ecosystems_test.GetNextVersionTest.test_maven)
Test Maven. ... skipped 'Unnecessary if using deps.dev'
test_maven_deps_dev (osv.ecosystems_test.GetNextVersionTest.test_maven_deps_dev)
Test Maven using deps.dev. ... ok
test_maven_with_cache (osv.ecosystems_test.GetNextVersionTest.test_maven_with_cache)
Test Maven. ... skipped 'Unnecessary if using deps.dev'
test_nuget (osv.ecosystems_test.GetNextVersionTest.test_nuget)
Test NuGet. ... ok
test_packagist (osv.ecosystems_test.GetNextVersionTest.test_packagist)
Test Packagist. ... ok
test_pypi (osv.ecosystems_test.GetNextVersionTest.test_pypi)
Test PyPI. ... ok
test_semver (osv.ecosystems_test.GetNextVersionTest.test_semver)
Test SemVer. ... ok

Ran 12 tests in 80.321s

OK (skipped=2)
```

I'm satisfied the right tests are being skipped for the right reasons on Cloud Build.

Now, to your [point](https://github.com/google/osv.dev/pull/982#discussion_r1070848873) about external contributors, there's not a lot of value in them running a different set of tests, so should I just remove the non-deps.dev ones and have the inability to retrieve the secrets fail silently, meaning external contributors won't be able to test this?

oliverchang: yeah, let's just have it fail (with a harmless log message) for external folks.

run_tests.sh
andrewpollock: Done in a40a406de8e0cd31f3c5eee473ddf603f341c2f1

oliverchang (approved):

run_tests.sh
andrewpollock: It should do this in a40a406de8e0cd31f3c5eee473ddf603f341c2f1

PR #986, opened 2023-01-17
Update alpine purl

### LINKED ISSUE #985: The purl of Alpine ecosystem is inconsistent with purl-spec
According to [**purl-type `apk`**](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#apk) ([PR](https://github.com/package-url/purl-spec/pull/171)), it seems that the purl type of `Alpine` should be `apk` and the purl namespace can be `alpine`. In this project, however, the purl type of `Alpine` is `alpine`.

As a result, the purl of a package is **different** when apply different standards. For example, the purl of the `curl` package is `pkg:apk/alpine/curl` according to purl-spec while it is `pkg:alpine/curl` in this project.

another-rex:
Fixes #985 

Also updates purl_helper.py to support alpine, even though it's not being called at the moment since vulnfeeds supplies the purl.


andrewpollock (commented):
Could you please add a test for Alpine while you're here?

another-rex:
Added a test for purl_helper

oliverchang (approved):

PR #991, opened 2023-01-23
Fix semver_index_test not being run

michaelkedar:
`semver_index_test` was not in `run_tests.sh`.
A change from a while ago also seems to have broken the test, which hadn't been picked up.

### DIFF
--- a/osv/semver_index_test.py
+++ b/osv/semver_index_test.py
@@ -17,7 +17,7 @@ import unittest
 
 import semver
 
-import semver_index
+from . import semver_index
 
 
 class SemverIndexTests(unittest.TestCase):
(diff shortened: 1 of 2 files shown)


michaelkedar:
Wondering if we should automatically detect all `*_test.py` to run tests on.

oliverchang (approved):

andrewpollock:
Good catch! Are there any others missing?

oliverchang:
> Wondering if we should automatically detect all `*_test.py` to run tests on.

+1 in general, but we do have e.g. worker_test.py that have other dependencies, so we have to be careful about scoping the discovery here.

michaelkedar:
> Good catch! Are there any others missing?

Looks like this was the only one missing.

PR #1004, opened 2023-01-30
Return more information from determineVersion

susperius:
Restructured the determineVersion endpoint and proto message to return the available information. 
Instead of just the version it now returns the name of the package, upstream repo information and the version string.


gcp/api/v1/osv_service_v1.proto
```diff
@@ -96,20 +96,41 @@ message VersionMatchList {
 
 // Match information for the provided VersionQuery.
 message VersionMatch {
+  // Identified package name.
+  string name = 1;
+  // Identified version number.
+  string version = 2;
```
oliverchang, line 102: What are "name" and "version" exactly here? 

Is version the git tag? Is it some normalized version number? Previously we made this clear to users what format to expect based on the type. 



Can I suggest the following structure instead: 

```json
{
  "confidence": 0.5,
  "repo": {
    "type": "GIT",
    "url": "http://...",
    "commit": "..."
    "exact_tag": "..."  // If the commit is at that exact tag. 
    "closest_tag": "..."  // The latest tag that comes before this commit. This could be the same as "exact_tag". 
  } ,
  "osv": {
     "ecosystem": "PyPI",
     "name": "flask"
     "version": "1.0.0"
   },
  "cpe": {
    "2.3": "cpe:/...."
   }
}
```

Where only one of repo/osv/cpe are specified in the structure depending on the type. I think these are all different enough that it might make things more confusing to try to all encode them in the one structure.

oliverchang (commented):
thanks! I just have a suggestion on simplifying this entire structure a little and making it more consistent with the rest of the OSV API's conventions.

gcp/api/v1/osv_service_v1.proto
susperius: Name is the potential package name and version is the parsed git tag (we don't store the raw git tag).

The idea was to send what we know about a package in one reply instead of breaking it up into multiple ones. Before this GIT and VERSION were also different and only VERSION was returned.

This structure reflects the data collection a bit more (at least for now). Every entry has a version set, and also repo information. In the future the extra information could be populated from other sources in OSV.

Perhaps we need to change the indexer to also store the raw tags.

oliverchang: Indeed we may need this. 

"Name" still seems a bit ambiguous to me -- what namespace is this name under? Same with "version" -- the namespace for versions can also be different depending on the context e.g. git tag vs a published version name may differ **a lot**. AFAIK only Go enforces a specific format for git tags that matches the actual published version.

oliverchang: Hmm, but what would "name" be exactly? Is it the basename from the repo URL? Is it a name that we choose (and so it's unpredictable to others?) IMO the repo URL would be the primary key for this (and it would be consistent and sufficient).

susperius: Yes, the name is just the repository name without ".git". It just takes away a step users might want to do. I don't mind removing it.

oliverchang: Let's avoid this redundancy and potential then :) It's a very simple step users can do if this is something they need.

oliverchang (approved):

PR #1006, opened 2023-02-01
Move debian converters to osv.dev

another-rex:
Move debian converters to osv.dev

Add release to the cloudbuild.yaml (We might want to move to separate this cloudbuild out since it's not released as often?)


deployment/oss-vdb-test/k8s/cloudbuild.yaml
```diff
@@ -55,6 +56,17 @@ steps:
...
+- name: gcr.io/cloud-builders/docker
+  args: ['build', '-t', 'gcr.io/$PROJECT_ID/debian-first-version', '-t', 'gcr.io/$PROJECT_ID/debian-first-version:$COMMIT_SHA', '--build-arg', 'PROJECT=$PROJECT_ID', '.']
+  dir: 'osv.dev/vulnfeeds/tools/debian'
+- name: gcr.io/cloud-builders/docker
+  args: ['build', '-t', 'gcr.io/$PROJECT_ID/debian-convert', '-t', 'gcr.io/$PROJECT_ID/debian-convert:$COMMIT_SHA', '--build-arg', 'PROJECT=$PROJECT_ID', '.']
+  dir: 'osv.dev/vulnfeeds/tools/debian'
+- name: gcr.io/cloud-builders/docker
+  args: ['push', 'gcr.io/$PROJECT_ID/debian-first-version']
+- name: gcr.io/cloud-builders/docker
+  args: ['push', 'gcr.io/$PROJECT_ID/debian-convert']
```
michaelkedar, line 69: Do you know if you also need to run `docker push gcr.io/$PROJECT_ID/[image]:$COMMIT_SHA`, or is the tag applied automatically?

vulnfeeds/tools/debian/run_convert_debian.sh
```diff
@@ -0,0 +1,32 @@
...
+echo "Converting DLAs"
+pipenv run python3 convert_debian.py --adv_type=DLA -o $OSV_DLA_OUT $WEBWML_PATH $DSA_PATH
+echo "Converting DTSAs"
+pipenv run python3 convert_debian.py --adv_type=DTSA -o $OSV_DTSA_OUT $WEBWML_PATH $DSA_PATH
+popd
+
+echo "Begin Syncing with cloud"
+gsutil -q -m rsync -d $OSV_DSA_OUT gs://debian-osv/dsa-osv
+gsutil -q -m rsync -d $OSV_DLA_OUT gs://debian-osv/dla-osv
+gsutil -q -m rsync -d $OSV_DTSA_OUT gs://debian-osv/dtsa-osv
```
michaelkedar, line 31: Actually, are these the buckets in `oss-vdb`? Does it need to parameterized for `oss-vdb-test`?

vulnfeeds/tools/debian/run_first_package_finder.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash -e
+
+# Use the first_package_finder script to generate a first version cache.
+pushd /src/debian_converter
+echo "Finding first packages"
+pipenv run python3 first_package_finder.py
+
+echo "Syncing with cloud first_package_output"
+gsutil -q -m rsync -d 'first_package_output' gs://debian-osv/first_package_output
```
michaelkedar, line 9: Also, is this the `oss-vdb` bucket? Does there need to be one for `oss-vdb-test`?

deployment/oss-vdb-test/k8s/cloudbuild.yaml
another-rex: Ah I missed it, added now. I don't think it's tagged automatically

deployment/oss-vdb-test/k8s/cloudbuild.yaml
```diff
@@ -55,6 +56,17 @@ steps:
   args: ['push', 'gcr.io/$PROJECT_ID/combine-to-osv:$COMMIT_SHA']
 - name: gcr.io/cloud-builders/docker
   args: ['push', 'gcr.io/$PROJECT_ID/combine-to-osv']
+# Build/push debian converter images
+- name: gcr.io/cloud-builders/docker
+  args: ['build', '-t', 'gcr.io/$PROJECT_ID/debian-first-version', '-t', 'gcr.io/$PROJECT_ID/debian-first-version:$COMMIT_SHA', '--build-arg', 'PROJECT=$PROJECT_ID', '.']
+  dir: 'osv.dev/vulnfeeds/tools/debian'
+- name: gcr.io/cloud-builders/docker
+  args: ['build', '-t', 'gcr.io/$PROJECT_ID/debian-convert', '-t', 'gcr.io/$PROJECT_ID/debian-convert:$COMMIT_SHA', '--build-arg', 'PROJECT=$PROJECT_ID', '.']
+  dir: 'osv.dev/vulnfeeds/tools/debian'
```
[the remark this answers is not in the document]
another-rex: Good point, they are the same, changed to just make one.

osv-scanner.toml
```diff
@@ -17,3 +17,23 @@ reason = "No ssh servers"
 id = "GO-2022-0969"
 # ignoreUntil = 2022-11-09 # Optional exception expiry date
 reason = "No external http servers in Go lang."
+
+[CallAnalysis]
```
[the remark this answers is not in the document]
another-rex: Ah my bad, accidentally got added in with the other changes, removed now.

vulnfeeds/tools/debian/run_first_package_finder.sh
another-rex: Ah yep, added the terraform config and pointed to the env variable, however, the code that points to that is located here: https://github.com/google/osv.dev/blob/bce1acb9a71ff10bbd5ba0eef26684043a557b86/osv/ecosystems/debian.py#L27-L29

Is the way to configure this other than just hard coding it in?

vulnfeeds/tools/debian/debian_converter/convert_debian.py
```diff
@@ -0,0 +1,398 @@
+# Copyright 2022 OSV Schema Authors
```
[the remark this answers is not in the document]
another-rex: Is it a bit weird that this is the only conversion code that is in osv-schema, while the rest is in vulnfeeds? That just feels inconsistent.

vulnfeeds/tools/debian/run_first_package_finder.sh
michaelkedar: I believe I dealt with this before by adding a command line argument for the bucket, which was coded into the kubernetes manifest:
https://github.com/google/osv.dev/blob/bce1acb9a71ff10bbd5ba0eef26684043a557b86/deployment/oss-vdb-test/k8s/gke/importer/importer.yaml#L19-L21
(though, It's not in that url format)

I'm still not really sure the best way to handle these situations - there might be a way to generate a config file with [terraform outputs variables](https://developer.hashicorp.com/terraform/language/values/outputs)

vulnfeeds/tools/debian/debian_converter/convert_debian.py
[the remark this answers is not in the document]
oliverchang: Good point, happy to leave it here then.

deployment/oss-vdb-test/k8s/gke/debian/debian-first-version.yaml
```diff
@@ -0,0 +1,38 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: debian-first-version
+spec:
+  schedule: "0 1 * * *"
```
andrewpollock, line 6: Just wanted to confirm that this timespec is correct, as it differs from the `"0 */1 * * *"` of deployment/oss-vdb-test/k8s/gke/debian/debian-convert.yaml

vulnfeeds/tools/debian/debian_converter/first_package_finder.py
```diff
@@ -0,0 +1,154 @@
+"""Functions to help find the first package version for a specific release"""
```
andrewpollock, line 1: This should probably have the standard copyright boilerplate added to the top?

vulnfeeds/tools/debian/run_convert_debian.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash -e
```
andrewpollock, line 1: This should probably have the standard copyright boilerplate added?

vulnfeeds/tools/debian/run_first_package_finder.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash -e
```
andrewpollock, line 1: This should probably have the standard copyright boilerplate added?

andrewpollock (approved):
I think this LGTM, assuming @michaelkedar is cool with the Terraform bits? Made left some comments about copyright boilerplate on a few files.

vulnfeeds/tools/debian/run_first_package_finder.sh
another-rex: That makes sense, in this case I don't think adding more options to worker.py is worth it for this change (first_package_output is very unlikely to have any changes in the short term), I added a TODO and we can come back to fix this when we have a better way to set cloud bucket urls.

oliverchang (approved):

PR #1011, opened 2023-02-08
Remove API backend image from Terraform

michaelkedar:
In preparation for moving to Cloud Deploy.

### DIFF
--- a/deployment/terraform/modules/osv/osv_api.tf
+++ b/deployment/terraform/modules/osv/osv_api.tf
@@ -1,12 +1,6 @@
 # The OSV API on Cloud Run
 # Adapted from https://github.com/hashicorp/terraform-provider-google/issues/5528#issuecomment-1136040976
 
-data "google_container_registry_image" "api_backend" {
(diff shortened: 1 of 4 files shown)


deployment/terraform/modules/osv/osv_api.tf
```diff
@@ -15,11 +9,18 @@ resource "google_cloud_run_service" "api_backend" {
   template {
     spec {
       containers {
-        image = data.google_container_registry_image.api_backend.image_url
+        image = "us-docker.pkg.dev/cloudrun/container/hello:latest" # Placeholder image.
```
andrewpollock, line 12: Given this is just a template, does it even need to be a valid image?

andrewpollock (approved):

deployment/terraform/modules/osv/osv_api.tf
michaelkedar: Terraform spits out an error if it's not valid when it's first created.

PR #1014, opened 2023-02-10
Trigger workflow on merge group

another-rex:
Support running workflows when merge queue is switched on. Cloud build doesn't seem to support merge groups yet though, so have to merge #971 before enabling merge queues again.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -15,6 +15,7 @@ on:
   push:
     branches: [ master ]
   pull_request:
+  merge_group:
     # The branches below must be a subset of the branches above
     branches: [ master ]
 
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -15,9 +15,11 @@ on:
   push:
     branches: [ master ]
   pull_request:
-  merge_group:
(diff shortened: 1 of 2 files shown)


andrewpollock:
Ugh, I think #971 isn't likely to land any time soon, there's a web of authentication-related fun that I don't have the time to untangle.

.github/workflows/codeql-analysis.yml
```diff
@@ -15,6 +15,7 @@ on:
   push:
     branches: [ master ]
   pull_request:
+  merge_group:
```
oliverchang, line 18: Does adding this change any semantics since the indented `branches` is now under `merge_group` and not `pull_request` ?

another-rex: Not actually sure, I originally thought it means `branches` would apply to both `pull_request` and `merge_group` because of this example: https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue#triggering-merge-group-checks-with-github-actions. 

But after testing it out in a yaml parser that's not how it works. I think github just takes null as meaning it will apply to all pull requests. 

I'll update this.

oliverchang (approved):

PR #1015, opened 2023-02-10
Regenerate protobufs

michaelkedar:
I don't think the protobufs were regenerated after the last commit of #1004 

Just happened to notice this when working on something else.


oliverchang (approved):
thanks!

PR #1017, opened 2023-02-10
Automate Terraform for oss-vdb-test

michaelkedar:
Adds a number of things needed to start automatically running terraform changes on the staging environment.
~~(I know Andrew wants to discuss potentially using KCC, but I'm using this to start setting up our staging pipeline)~~

-  Adds Terraform Docker image, usable by Cloud Build to run Terraform for our project.
- Adds a cloud build trigger on `oss-vdb` to run `terraform plan` on `oss-vdb-test` when terraform files are changed
  - Output is only viewable in Cloud Build at the moment
  - ~~Also added a Github workflow to comment on PRs when something changes.~~
- Adds a cloud build trigger on `oss-vdb` to run `terraform apply` on `oss-vdb-test` on pushes to the master branch.
  - The cloud build yaml will eventually be used for building Docker images & running Cloud Deploy

The Cloud Builds use the existing `deployment` service account from `oss-vdb`. IAM permissions have been granted to it on `oss-vdb-test` to allow it to modify the project as needed.


michaelkedar:
Well, the Github commenting didn't work :frowning_face:

deployment/terraform/cloudbuild-tf-plan.yaml
```diff
@@ -0,0 +1,29 @@
...
+  dir:  deployment/terraform/environments/oss-vdb-test
+- name: gcr.io/oss-vdb/terraform
+  args: ['plan', '-no-color']
+  dir:  deployment/terraform/environments/oss-vdb-test
+# TODO(michaelkedar): plan for oss-vdb
+
+# Set in Cloud Build:
+# serviceAccount: 'projects/oss-vdb/serviceAccounts/deployment@oss-vdb.iam.gserviceaccount.com'
+options:
+  logging: CLOUD_LOGGING_ONLY
```
michaelkedar, line 29: Personally, I find Cloud Logging output hard to read.
We could make a logging bucket for this to make it [look like this](https://pantheon.corp.google.com/cloud-build/builds;region=global/950ef9cc-1e27-4eff-aa65-558f3fc86545?project=oss-vdb).

I'm not really sure how much you'd have to manage it / delete old logs.

oliverchang: That seems a bit better than having to read Cloud Logging logs. Re deleting old logs, we can just set a lifecycle policy: https://cloud.google.com/storage/docs/lifecycle

docker/terraform/cloudbuild.yaml
```diff
@@ -0,0 +1,29 @@
+# In this directory, run the following command to build this builder.
+# $ gcloud builds submit . --config=cloudbuild.yaml
+substitutions:
+  _TERRAFORM_VERSION: 1.3.7
+  _TERRAFORM_VERSION_SHA256SUM: b8cf184dee15dfa89713fe56085313ab23db22e17284a9a27c0999c67ce3021e
+steps:
+- name: 'gcr.io/cloud-builders/wget'
+  args: ["https://releases.hashicorp.com/terraform/${_TERRAFORM_VERSION}/terraform_${_TERRAFORM_VERSION}_linux_amd64.zip"]
```
oliverchang, line 8: Can we instead build from source using `go install` ? 

This reduces a bit of supply chain risk (given that these builds are not signed/verifiable), by reducing it to the GitHub repo rather than including `releases.hashicorp.com` as well and every step that led to the binary getting uploaded there.

michaelkedar:
Turns out GitHub Actions coming from public forked repositories [can't have more than read permissions](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token), so I can't use an action to make a comment.

I'm removing the action, and we'll just have to remember to check the plans.

docker/terraform/cloudbuild.yaml
michaelkedar: (Disclaimer: I didn't write this file)
Aren't these builds signed? [Hashicorp provides a signature for the checksum files they provide](https://www.hashicorp.com/security)

I can move to building with go anyway, if that's your preference.

oliverchang: Missed that there are some signatures provided here for the checksums (but we're not verifying them anywhere anyway?)

That said, it's not as good as [SLSA](https://slsa.dev/) in that it doesn't verify the chain of trust all the way back to the source repo itself and reduce the risk of compromise to just the builder. The key management, security of hashicorp.com etc are all still included in our risk by relying on this current approach.

deployment/terraform/cloudbuild-tf-plan.yaml
michaelkedar: Moved logs to the terraform bucket & made a lifecycle policy to delete after 60 days

docker/terraform/cloudbuild.yaml
michaelkedar: I've rejigged this to now build with `go install`

deployment/build-and-stage.yaml
```diff
@@ -0,0 +1,43 @@
+# Copyright 2023 Google LLC
```
michaelkedar, line 1: Note to self: enable this trigger in cloud build before merging

deployment/terraform/cloudbuild-tf-plan.yaml
```diff
@@ -0,0 +1,28 @@
...
+- name: gcr.io/oss-vdb/terraform
+  args: ['init', '-no-color']
+  dir:  deployment/terraform/environments/oss-vdb-test
+- name: gcr.io/oss-vdb/terraform
+  args: ['plan', '-no-color']
+  dir:  deployment/terraform/environments/oss-vdb-test
+# TODO(michaelkedar): plan for oss-vdb
+
+# Set in Cloud Build:
+# serviceAccount: 'projects/oss-vdb/serviceAccounts/deployment@oss-vdb.iam.gserviceaccount.com'
```
michaelkedar, line 27: Just calling this out:
This service account has a lot of permissions, and Terraform lets you execute arbitrary scripts in `plan` (and `apply`).

You do need to do the /gcbrun comment for external contributors, though maybe the planner should have its own service account with the minimum required permissions to be safe.

deployment/build-and-stage.yaml
```diff
@@ -0,0 +1,43 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# TODO(michaelkedar): This is intentionally separate from staging.yaml
+# to not break existing deployment process as new one is implemented
+# eventually, they can probably be combined
+
+steps:
+# TODO: Docker image creation & tagging
```
oliverchang, line 20: nit: for consistency add TODO(michaelkedar).

Also, this seems pretty easy to add here in this PR. Let's just do that?

deployment/terraform/cloudbuild-tf-plan.yaml
oliverchang: Do you know what the minimum required permissions are?

docker/terraform/Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+
+ARG TERRAFORM_VERSION
+WORKDIR /build/
+RUN GOBIN=$(pwd) go install github.com/hashicorp/terraform@v${TERRAFORM_VERSION}
+
+FROM gcr.io/cloud-builders/gcloud
+
+COPY --from=build /build/terraform /usr/bin/terraform
+COPY entrypoint.bash /builder/entrypoint.bash
+ADD https://github.com/stedolan/jq/releases/download/jq-1.6/jq-linux64 /usr/bin/jq
```
oliverchang, line 11: Can we `apt-get` install this instead?

docker/terraform/README.md
```diff
@@ -0,0 +1,27 @@
+# Terraform Cloud Builder Image
+Used for running Terraform in Cloud Build. Contains the required `terraform`, `gcloud`, and `jq` executables.
```
oliverchang, line 2: What is `jq` used for?

oliverchang (commented):
thanks! some initial comments.

deployment/build-and-stage.yaml
michaelkedar: I was going to add it when I properly consolidated the staging/production docker registries, since that will affect some other staging deployment scripts as well.

docker/terraform/README.md
michaelkedar: It's used by a the [`esp_full_version`](https://github.com/michaelkedar/osv.dev/blob/terraform-cloudbuild/deployment/terraform/modules/osv/scripts/esp_full_version) script to interface with [Terraform inputs/outputs](https://github.com/michaelkedar/osv.dev/blob/6ead7f770340b70144c69e82af85c1acd0115dbc/deployment/terraform/modules/osv/osv_api.tf#L55-L58) (which are in json)

Actually, the output is simple enough that I can probably avoid using it. I might remove it.

deployment/terraform/cloudbuild-tf-plan.yaml
michaelkedar: From some playing around, this seems to work for the planner:

Roles on the project:
- Viewer
- Pub/Sub Admin

On the Terraform bucket
- Storage Legacy Bucket Owner
- Storage Legacy Object Reader

It might be possible to make a more minimum set with a custom role, but I'd need a deeper look into it.

docker/terraform/Dockerfile
michaelkedar: removed the `jq` dependency

deployment/terraform/cloudbuild-tf-plan.yaml
oliverchang: That seems like a pretty limited set to begin with already. Let's create a new one for this. 

Is the Pub/Sub one definitely needed?

deployment/build-and-stage.yaml
```diff
@@ -0,0 +1,43 @@
...
+  dir:  deployment/terraform/environments/oss-vdb-test
+
+# TODO(michaelkedar): App Engine for oss-vdb-test
+# TODO(michaelkedar): Cloud Deploy
+
+timeout: 7200s
+# Set in Cloud Build:
+# serviceAccount: 'projects/oss-vdb/serviceAccounts/deployment@oss-vdb.iam.gserviceaccount.com'
+
+# TODO(michaelkedar): the terraform bucket is not the best place for this
```
oliverchang, line 38: This bucket seems reasonable to me actually. Only comment would be to rename the "directory" "apply-logs"

deployment/terraform/cloudbuild-tf-plan.yaml
```diff
@@ -0,0 +1,28 @@
...
+logsBucket: gs://oss-vdb-tf/cloudbuild-logs
```
oliverchang, line 28: Let's rename this one "plan-logs"

oliverchang (approved):
LGTM with some minor comments remaining.

@andrewpollock can you please review and sign off on this too?

deployment/build-and-stage.yaml
michaelkedar: It's just that eventually this will also be building the container images and deploying to App Engine / Cloud Deploy, which I didn't really think belonged in the terraform bucket.

deployment/terraform/cloudbuild-tf-plan.yaml
andrewpollock: Is deployment@oss-vdb.iam.gserviceaccount.com a custom-created service account for this? https://cloud.google.com/deploy/docs/cloud-deploy-service-account#execution_service_account is saying to me that the default compute account is normally used. We should definitely follow the principle of least privilege here.

michaelkedar: Yeah, Pub/Sub Admin is a strange one, but Pub/Sub Editor and Pub/Sub Viewer don't seem to give permissions to look at publishers and subscribers for some reason.

michaelkedar: > Is [deployment@oss-vdb.iam.gserviceaccount.com](mailto:deployment@oss-vdb.iam.gserviceaccount.com) a custom-created service account for this? https://cloud.google.com/deploy/docs/cloud-deploy-service-account#execution_service_account is saying to me that the default compute account is normally used. We should definitely follow the principle of least privilege here.

The deployment service account already existed (for App Engine iirc) I was just reusing it. Will make a dedicated "terraform plan" service account for this one

docker/terraform/entrypoint.bash
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
michaelkedar: I didn't actually write this or edit file - it came from https://github.com/GoogleCloudPlatform/cloud-builders-community/tree/master/terraform 

I mentioned it in the README but I might add a comment in the file (and in Dockerfile and cloud build file) to note this as well

deployment/terraform/cloudbuild-tf-plan.yaml
michaelkedar: Created a terraform-plan@oss-vdb.iam.gserviceaccount.com for the planner.

deployment/terraform/modules/osv/scripts/esp_full_version
```diff
@@ -40,4 +37,4 @@ for tag in "${TAGS_ARRAY[@]}"; do
 done
 
 # Produce JSON object containing esp_full_version
-jq -n --arg esp_full_version "$ESP_FULL_VERSION" '{"esp_full_version":$esp_full_version}'
\ No newline at end of file
+echo "{\"esp_full_version\":\"${ESP_FULL_VERSION}\"}"
```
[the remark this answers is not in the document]
michaelkedar: Changed to here document

deployment/build-and-stage.yaml
andrewpollock: I think it's okay to have reasonably well-affiliated things in the one bucket, rather than having a proliferation of buckets

docker/terraform/entrypoint.bash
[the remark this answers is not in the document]
andrewpollock: Oh I'd glossed over that in the README. Treat all of these as nits then... I filed https://github.com/GoogleCloudPlatform/cloud-builders-community/issues/611 against the source.

andrewpollock (approved):

docker/terraform/entrypoint.bash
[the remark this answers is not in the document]
michaelkedar: I'm just going to leave it as-is, so I don't inadvertently break anything.

Issue #1018, opened 2023-02-13
missing GHSA advisories


josieang:
There are some advisories that are in GitHub but not in OSV, @oliverchang mentioned that there might be some encoding issues but possible bugs as well.
```
GHSA-x4rg-4545-4w7w
GHSA-v62j-cxhh-fq22
GHSA-2w7w-2j92-44hx
GHSA-g2qw-6vrr-v6pq
GHSA-fj6c-prgj-gr3r
GHSA-5pg8-f89x-wjcx
GHSA-789v-h9hw-38pg
GHSA-wh77-3x4m-4q9g
```

oliverchang:
Following up on this:

Thanks to @another-rex and @michaelkedar, 4 of these have been fixed (GHSA-v62j-cxhh-fq22, GHSA-g2qw-6vrr-v6pq, GHSA-5pg8-f89x-wjcx, GHSA-789v-h9hw-38pg)

The rest appear to be due to data quality issues from GitHub that we'll follow up on.

oliverchang:
All affected GHSAs are now in the OSV.dev database.

PR #1021, opened 2023-02-13
Fix Maven handling of version 0

### LINKED ISSUE #1020: Maven 'zero' versions causes errors when comparing
Maven versions that are 'zero' (e.g. `0.0.0` and `0.0.0-X.Y.Z`) cause assertion errors when comparing with non-zero versions, because the comparisons assume there's always an initial token without a `.` or `-` 'prefix'.

This seems to be causing some of the missing vulnerabilities in #1018

michaelkedar:
Fixes #1020 

This should also fix some of the missing vulnerabilities in #1018 on reimport:
- GHSA-v62j-cxhh-fq22
- GHSA-g2qw-6vrr-v6pq
- GHSA-789v-h9hw-38pg

### DIFF
--- a/osv/ecosystems/maven.py
+++ b/osv/ecosystems/maven.py
@@ -221,7 +221,7 @@ class Version:
     # Then, starting from the end of the version, the trailing "null" values
     # (0, "", "final", "ga") are trimmed.
     i = len(version.tokens) - 1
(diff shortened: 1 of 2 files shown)


oliverchang (approved):
thanks!

PR #1022, opened 2023-02-13
Reproduce the existing normalization functionality from osv/bugs.py

andrewpollock:
This is slightly horrible due to RE2's lack of support for lookarounds (negative lookbehind assertions in this instance). The most pragmatic approach I could come up with is to sanity check the results. Thank goodness for existing comprehensive test cases.

### DIFF
--- a/vulnfeeds/cves/versions_test.go
+++ b/vulnfeeds/cves/versions_test.go
@@ -374,3 +374,112 @@ func TestExtractGitCommit(t *testing.T) {
 		}
 	}
 }
+
+func TestNormalize(t *testing.T) {
+	tests := []struct {
+		description               string
+		inputVersion              string
+		expectedNormalizedVersion string
+		expectedOk                bool
+	}{
+		{
+			description:               "Empty string",
+			inputVersion:              "",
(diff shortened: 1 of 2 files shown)


vulnfeeds/cves/versions_test.go
```diff
@@ -374,3 +374,112 @@ func TestExtractGitCommit(t *testing.T) {
...
+		{
+			description:               "Valid supported version #3",
+			inputVersion:              "1.2.3.4.5-rc1",
+			expectedNormalizedVersion: "1-2-3-4-5-rc1",
+			expectedOk:                true,
+		},
+		{
+			description:               "Valid supported version #4",
+			inputVersion:              ".1",
+			expectedNormalizedVersion: "1",
```
oliverchang, line 418: I wonder if this special case should be 0-1 instead. (the bug.py implementation may be wrong here). 

This is also more consistent with https://maven.apache.org/pom.html#version-order-specification where ".1" actually means "0.1"

oliverchang (approved):
LGTM with a comment about a potential pre-existing bug in bug.py

Issue #1019, opened 2023-02-13
missing OSS-Fuzz fixes 


oliverchang:
The OSS-Fuzz bisection infra failed to identify the fixes for:

https://osv.dev/vulnerability/OSV-2022-837 , https://osv.dev/vulnerability/OSV-2022-1173 , https://osv.dev/vulnerability/OSV-2023-49 , https://osv.dev/vulnerability/OSV-2022-1087 ,

oliverchang:
For https://osv.dev/vulnerability/OSV-2022-1173, the bisector failed with:

```
Traceback (most recent call last):
  File "/usr/local/bin/oss_fuzz.py", line 129, in do_bisect
    result = bisector.bisect(bisect_type, old_commit, new_commit, f.name,
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/work/oss-fuzz/infra/bisector.py", line 307, in bisect
    return _bisect(bisect_type, old_commit, new_commit, testcase_path,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/work/oss-fuzz/infra/bisector.py", line 199, in _bisect
    bisect_repo_manager.fetch_all_remotes()
  File "/work/oss-fuzz/infra/repo_manager.py", line 134, in fetch_all_remotes
    self.git([
  File "/work/oss-fuzz/infra/repo_manager.py", line 59, in git
    return utils.execute(['git'] + cmd,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/work/oss-fuzz/infra/utils.py", line 73, in execute
    process = subprocess.Popen(command,
              ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/subprocess.py", line 1022, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/local/lib/python3.11/subprocess.py", line 1899, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: '/work/tmp/tmpadyhr3au/src/opensips'

Traceback (most recent call last):
  File "/usr/local/bin/oss_fuzz.py", line 129, in do_bisect
[...]

PR #1023, opened 2023-02-13
Downgrade Docker to older 20.x version.

oliverchang:
This is the previous version at time of writing.

This is part of addressing #1019.


andrewpollock (approved):
Have we surfaced the bad interaction to the gVizor folks?

oliverchang:
> Have we surfaced the bad interaction to the gVizor folks?

Not yet, I'll create a bug. Thanks for reminding!

PR #1025, opened 2023-02-13
Preliminary remote repository tag querying code

andrewpollock:
Queries tags and supports optional cacheing

The tests need to be fully mocked out so they're hermetic

### DIFF
--- a/vulnfeeds/git/repository.go
+++ b/vulnfeeds/git/repository.go
@@ -53,6 +53,7 @@ func RepoTags(repoURL string, repoTagsCache *RepoTagsMap) (versions Versions, e 
 		v := Version{Tag: ref.Name().Short(), Commit: ref.Hash().String()}
 		versions = append(versions, v)
 	}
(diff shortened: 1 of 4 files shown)


vulnfeeds/git/repository.go
```diff
@@ -0,0 +1,62 @@
...
+		return versions, err
+	}
+	for _, ref := range refs {
+		if !ref.Name().IsTag() {
+			continue
+		}
+		v := Version{Tag: ref.Name().Short(), Commit: ref.Hash().String()}
+		versions = append(versions, v)
+	}
+	sort.Sort(versions)
```
oliverchang, line 56: Is sorting purely for consistency/testing sake? Let's add a comment here on why this is needed.

oliverchang (approved):

vulnfeeds/git/repository.go
andrewpollock: Yes, added a comment in d1e812d

PR #1039, opened 2023-02-15
Output notes

andrewpollock:
We were tossing away the notes that were handed back from ExtractVersionInfo, stop doing that in case there's useful insights available

### DIFF
--- a/vulnfeeds/cpp/main.go
+++ b/vulnfeeds/cpp/main.go
@@ -140,8 +140,9 @@ func CVEToOSV(CVE cves.CVEItem, repos []string, repobasedir string, directory st
 	if err != nil {
 		return fmt.Errorf("Can't generate an OSV record for %s without valid CPE data", CVE.CVE.CVEDataMeta.ID)
 	}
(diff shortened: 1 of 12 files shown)


vulnfeeds/cpp/main.go
```diff
@@ -181,6 +183,18 @@ func CVEToOSV(CVE cves.CVEItem, repos []string, repobasedir string, directory st
 		return fmt.Errorf("Failed to write %s: %v", outputFile, err)
 	}
 	Logger.Infof("Generated OSV record from %s for %q", CVE.CVE.CVEDataMeta.ID, CPE.Product)
+	if len(notes) > 0 {
```
oliverchang, line 186: Can we simplify this by just using https://pkg.go.dev/os#WriteFile ?

oliverchang (approved):

vulnfeeds/cpp/main.go
andrewpollock: Great suggestion, thanks! Done in https://github.com/google/osv.dev/pull/1039/commits/f609b90eaca02da4e3d95f4406113c66f16ec01f

PR #1042, opened 2023-02-15
Datastore: Flatten issue

susperius:
Fixing the datastore flatten issue.

* Manually flatten the result struct.
* Adding the tag to the datastore entry


oliverchang (approved):
thanks! did you check if the python ndb model is able to read this without any changes as well?

PR #1044, opened 2023-02-16
Only skip exporting bugs that have an UNPROCESSED status

### LINKED ISSUE #1007: "withdrawn" entries not getting exported correctly
Identified by @andrewpollock : 

https://github.com/google/osv.dev/blob/26050deb42785bc5a4dc7d802eac8e7f95135509/docker/exporter/exporter.py#L94

withdrawn entries are marked as status = INVALID in our DB, so they're not included.

They should be included when we export.

andrewpollock:
This fixes #1007 by also exporting "withdrawn" bugs, i.e. ones with status INVALID.


oliverchang (approved):
thanks!

Issue #951, opened 2023-01-06
Batch AffectedCommit entities


oliverchang:
In hindsight, one entity per commit (as the current [AffectedCommit](https://github.com/google/osv.dev/blob/edacbd3c32b1d632bad8c5b506a14d0850b2e537/osv/models.py#L82) model uses) was a mistake -- this often results in large bursty Datastore writes which could potentially degrade our read latencies. 

We should look at migrating to a batched index instead which stores multiple commits per entity.

PR #1050, opened 2023-02-20
Don't write legacy AffectedCommit entries for Linux.

oliverchang:
This is needed for our backfill, as we don't want to be writing the legacy format while re-importing and re-generating indexes in the new one.

Part of #951.


docker/worker/testdata/UpdateTest_update_bucket_cve.txt
```diff
@@ -1,13 +1,17 @@
 {   'affected': [],
```
oliverchang, line 1: These are all automated formatting updates from `TESTS_GENERATE=1`.

andrewpollock (approved):

PR #1051, opened 2023-02-20
Fix request size limit exceptions with AffectedCommits.

oliverchang:
There is a limit on the total size of a Datastore put() request (10 MiB) that we were hitting when batch writing large AffectedCommits entities.

### DIFF
--- a/osv/impact.py
+++ b/osv/impact.py
@@ -39,7 +39,13 @@ AnalyzeResult = collections.namedtuple('AnalyzeResult', 'has_changes commits')
 
 TagsInfo = collections.namedtuple('TagsInfo', 'tags latest_tag')
 
+# Limit for writing small entities.
 _DATASTORE_BATCH_SIZE = 5000
+# Limit for writing large entities that are close to the 1MiB max entity size.
+# There is additionally a request size limit of 10MiB for batched put()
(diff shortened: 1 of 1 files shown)


osv/impact.py
```diff
@@ -322,17 +328,17 @@ def _batcher(entries, batch_size):
     yield entries[i:i + batch_size], i + batch_size >= len(entries)
 
 
-def _throttled_put(to_put):
+def _throttled_put(to_put, batch_size=_DATASTORE_BATCH_SIZE):
```
another-rex, line 331: nit: Prefer removing the defaults and explicitly setting the batch_size variable. `_throttled_put` and `_throttled_delete` are only used once each with the default value, so only two more line changes. 

 This avoids potential problem in the future where someone uses throttled_put assuming it will be throttling to `_DATASTORE_LARGE_BATCH_SIZE = 8`, but instead is throttling a lot less than that.

another-rex (approved):
LGTM with one comment

another-rex:
Oops, didn't realize auto merge was on.

osv/impact.py
oliverchang: Hmm good point, I'm going to delete the other two uses completely very soon after the migration is done. I'll clean this up then!

PR #1066, opened 2023-02-24
Migrate API serving based on new AffectedCommits entities.

oliverchang:
Also add a missed case of updating the new entries in the App Engine cron handling making OSS-Fuzz vulns public. We will need to do some manual queries to detect mismatching cases and fix them up if necessary.

Part of #951.


oliverchang:
Note: Do not merge for this week's prod release.

gcp/api/server.py
```diff
@@ -301,13 +308,16 @@ def _clean_purl(purl):
 @ndb.tasklet
 def query_by_commit(commit, to_response=bug_to_response):
   """Query by commit."""
-  query = osv.AffectedCommit.query(osv.AffectedCommit.commit == commit,
-                                   osv.AffectedCommit.public == True)  # pylint: disable=singleton-comparison
+  query = osv.AffectedCommits.query(osv.AffectedCommits.commits == commit)
```
michaelkedar, line 311: Is it worth automatically verifying with or falling back to the legacy `AffectedCommit` here, or will that cause too much latency?

gcp/appengine/handlers.py
```diff
@@ -59,6 +59,7 @@ def make_affected_commits_public(bug):
   """Make related AffectedCommit entities public."""
```
michaelkedar, line 59: Did a quick search and found this: https://github.com/google/osv.dev/blob/0e31a369fb9055dab2ddf97e0ec3ed7ef5978823/docker/worker/worker.py#L189-L200

Does this also need some handling for the new `AffectedCommits`?

michaelkedar (approved):
LGTM with a few questions

gcp/appengine/handlers.py
oliverchang: Good catch. I've fixed this up as well.

gcp/api/server.py
oliverchang: Good point. It adds a bit of complexity and I'm a bit wary of latency here. I'd like to validate that these indexes are complete a different way.

another-rex (approved):
LGTM

andrewpollock (approved):

PR #1082, opened 2023-03-02
Removing -X POST

### LINKED ISSUE #1081: Don't suggest using -X POST
On https://osv.dev/#use-the-api is the instruction
```
Query by commit hash

curl -X POST -d \
  '{"commit": "6879efc2c1596d11a6a6ad296f80063b558d5e0f"}' \
  "https://api.osv.dev/v1/query"
```

Using `-X POST` here is unnecessary, redundant and potentially dangerous as people cut and paste this into more places. curl will actually tell you this if you add `-v` to this command:

`Note: Unnecessary use of -X or --request, POST is already inferred.`

See also https://daniel.haxx.se/blog/2015/09/11/unnecessary-use-of-curl-x/

hayleycd:
Resolves #1081.


andrewpollock (approved):
Nice work finding it in even more places than initially sprang to mind!

hayleycd:
I'm not sure what is going on with the linting. 

I'm also not sure if we want to check that the docs/website looks ok on stage.

andrewpollock:
> I'm not sure what is going on with the linting.

Can you view https://github.com/google/osv.dev/actions/runs/4309942099/jobs/7517876146?

```
-      'source':
-          ("""cat <<EOF | curl -d @- "https://api.osv.dev/v1/querybatch"
+      'source': ("""cat <<EOF | curl -d @- "https://api.osv.dev/v1/querybatch"
```

It looks like it's of the opinion this can be on one line?

> I'm also not sure if we want to check that the docs/website looks ok on stage.

IMO this change is easy enough to reason about that previewing it isn't necessary.

See also https://github.com/google/osv.dev/issues/1081#issuecomment-1451219234 re some potential overthinking of fitness-for-purpose of using `curl`...

andrewpollock:
> See also [#1081 (comment)](https://github.com/google/osv.dev/issues/1081#issuecomment-1451219234) re some potential overthinking of fitness-for-purpose of using `curl`...

You can disregard that and the ensuing stream of consciousness from me

andrewpollock (approved):

PR #1090, opened 2023-03-06
Blogposts that are in the future are not generated

another-rex:


oliverchang (approved):

gcp/appengine/blog/content/posts/automating-and-scaling-vex-generation.md
```diff
@@ -1,6 +1,6 @@
 ---
 title: "Automating and Scaling Vex Generation"
-date: 2023-03-06T09:00:00Z
+date: 2023-03-01T09:00:00Z
```
oliverchang, line 3: Actually can we make this more like March 5th instead of 1st? That's closer to the actual date.

PR #1095, opened 2023-03-06
Modified frontend to safely output static images under blog

another-rex:
Fixes image output for the blog. Might still be too strict if we want to allow blog to have images in directories.

Also updates blog to use safe_join to add another layer to preventing escapes.

### DIFF
--- a/gcp/appengine/frontend_handlers.py
+++ b/gcp/appengine/frontend_handlers.py
@@ -136,7 +136,7 @@ def blog_post(blog_name):
 
 @blueprint.route('/blog/posts/<blog_name>/<file_name>', strict_slashes=False)
 def blog_post_static_files(blog_name: str, file_name: str):
-  print(file_name)
+  """Return static files under blog post directories"""
   if not _VALID_BLOG_NAME.match(blog_name):
     abort(404)
 
--- a/gcp/appengine/frontend_handlers.py
+++ b/gcp/appengine/frontend_handlers.py
@@ -43,8 +43,7 @@ blueprint = Blueprint('frontend_handlers', __name__)
 _PAGE_SIZE = 16
 _PAGE_LOOKAHEAD = 4
 _REQUESTS_PER_MIN = 30
-_WORD_CHARACTERS_OR_DASH = re.compile(r'^[\w\-]+$')
-_VALID_FILE_NAME = re.compile(r'^[\w\-_]+(\.\w+)*$')
+_WORD_CHARACTERS_OR_DASH = re.compile(r'^[\w-]+$')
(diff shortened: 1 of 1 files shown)


gcp/appengine/frontend_handlers.py
```diff
@@ -40,7 +43,8 @@
 _PAGE_SIZE = 16
 _PAGE_LOOKAHEAD = 4
 _REQUESTS_PER_MIN = 30
-_WORD_CHARACTERS_OR_DASH = re.compile(r'^[\w-]+$')
+_WORD_CHARACTERS_OR_DASH = re.compile(r'^[\w\-]+$')
+_VALID_FILE_NAME = re.compile(r'^[\w\-_]+(\.\w+)*$')
```
oliverchang, line 47: Is _ needed in the first `['\w\-_]` ? `\w` already includes underscores right? 

Also, can you add a comment on what you're trying to ensure here? The regex is starting to get a bit  complicated to understand the intention :)

oliverchang (approved):
LGTM with some nits

gcp/appengine/frontend_handlers.py
```diff
@@ -40,7 +43,8 @@
 _PAGE_SIZE = 16
 _PAGE_LOOKAHEAD = 4
 _REQUESTS_PER_MIN = 30
-_WORD_CHARACTERS_OR_DASH = re.compile(r'^[\w-]+$')
+_WORD_CHARACTERS_OR_DASH = re.compile(r'^[\w\-]+$')
```
andrewpollock, line 46: Based on the explanation regex101.com gives me, I'm not seeing the need for escaping the hyphen, are you doing this for a particular reason? Same question for the next line.

andrewpollock (approved):

gcp/appengine/frontend_handlers.py
andrewpollock: +1 to @oliverchang's comment about comments :-)

gcp/appengine/frontend_handlers.py
another-rex: Ah you're right, _ is not needed, in fact the regex is not actually needed, but just another layer to stop any invalid names. 

It seems like flask blueprint routes already strip out any / (even url encoded), and we are using `send_from_directory`, so this is actually overkill, so I removed the regex now to simplify things.

gcp/appengine/frontend_handlers.py
another-rex: You're right it's not needed to be escaped, but if you want to add one more character it becomes a range (e.g. `[A-Z]`), so I just added `\` to be consistent with the other regex. The other one's removed now so I reverted this.

PR #1100, opened 2023-03-06
Adding a new line

hayleycd:
On the preview, it looked like another new line was needed for the formatting to look good.


another-rex:
Just tested this, and it doesn't seem to do anything sadly, still squashed. I increased the margins a bit in the css to try and fix it, do you think this looks good?

### New:
image

### Old:
image

hayleycd:
Yes! That looks a lot better. Thanks @another-rex

PR #1103, opened 2023-03-07
Use sydney time

another-rex:
Currently OSS-Fuzz vulnerabilities are still being imported at around midnight UTC, which causes a lot of worker ramp up at around noon AEST. This causes trouble when deploying as GKE need to wait till works finished before updating the worker instances (not 100% sure how this interaction works with the new deployment process).

Use AEST instead of UTC, so the imports happen at AEST midnight instead.


andrewpollock (approved):
This will mean there's DST oscillations, as Australia/Sydney moves (relative to UTC). Not sure if that's going to be an issue or not?

another-rex:
That should not be an issue, since the requirement is just to move it close to midnight sydney time.

Interestingly when testing it doesn't seem to account for DST, even though the documentation says it should. I changed it to AEST time just to be sure DST is not a factor.

another-rex:
Turns out, it didn't work and I was reading the time wrong.  :facepalm: 

Updated it again to be correctly offsetting, and compatible with ndb.

docker/importer/importer.py
```diff
@@ -55,6 +56,17 @@ def _is_vulnerability_file(source_repo, file_path):
...
 
 
+def aestnow() -> datetime:
+  """Get the current AEST time"""
+  # - First make datetime timezone aware
+  # - Then set timezone to AEST without DST (+10)
+  # - Then make datetime timezone unaware again to keep
+  #   compatibility with ndb
+  return utcnow().replace(tzinfo=tz.UTC).astimezone(
+      tz.tzoffset('AEST_NO_DST',
```
oliverchang, line 66: What does dateutil.tz give us if we still need to specify the offsets ourself here?

Can we just use https://docs.python.org/3/library/datetime.html#timezone-objects instead?

another-rex: Good point, removed dateutil now.

docker/importer/importer.py
```diff
@@ -55,6 +55,16 @@ def _is_vulnerability_file(source_repo, file_path):
...
 
 
+def aestnow() -> datetime:
+  """Get the current AEST time"""
+  # - First make datetime timezone aware
+  # - Then set timezone to AEST without DST (+10)
+  # - Then make datetime timezone unaware again to keep
+  #   compatibility with ndb
+  return utcnow().replace(tzinfo=datetime.timezone.utc).astimezone(
+      datetime.timezone(datetime.timedelta(hours=11))).replace(tzinfo=None)
```
oliverchang, line 65: Can you explain in the comment why this is 11 when the comment says +10 ??

oliverchang (approved):

docker/importer/importer.py
another-rex: Oops, that should be 10, thought I fixed it earlier, but didn't commit it properly,

PR #1104, opened 2023-03-07
Better surface the OSV FAQ

andrewpollock:
This better surfaces the FAQ, which will only grow over time, and is our front line against being overwhelmed by frequently asked questions.

I've manually tested this locally:

- /faq URI works (and is what is advertised by default)
- /about URI works
- underlining works as intended/expected


gcp/appengine/frontend_handlers.py
```diff
@@ -148,8 +148,9 @@ def blog_post_static_files(blog_name: str, file_name: str):
 
 
 @blueprint.route('/about')
```
another-rex, line 150: Might be better to just have `/about` be a redirect rather than serving the same page. It makes it slightly confusing to the user, and also is unclear what url `url_for('frontend_handlers.faq')` will return (though in this case I think it will always return the bottom route (faq))

another-rex (commented):
LGTM, just one comment

gcp/appengine/frontend_handlers.py
oliverchang: +1. redirecting makes it clearer that there is a new canonical link for this.

andrewpollock: Ack, I believe I have done this correctly, PTAL.

another-rex (approved):
LGTM

PR #1105, opened 2023-03-07
Preliminary tag mapping

andrewpollock:
Very basic version to commit mapping via normalized versions and Git tags


vulnfeeds/cpp/main.go
```diff
@@ -75,84 +76,85 @@ func InScopeGitRepo(repoURL string) bool {
...
-	if err != nil {
-		return nil, err
-	}
-	return r, nil
-}
-
-// Examines repos and tries to convert version tags to commits
-func GitVersionToCommit(versions cves.VersionInfo, repos []string, repobasedir string) (v cves.VersionInfo, e error) {
+// Examines repos and tries to convert versions to commits by treating them as Git tags.
+// Takes a CVE string (for logging), cves.VersionInfo with AffectedVersions and
```
oliverchang, line 80: What exactly is this string? a CVE ID?

vulnfeeds/cpp/main.go
```diff
@@ -75,84 +76,85 @@ func InScopeGitRepo(repoURL string) bool {
...
+					}
+					Logger.Infof("[%s]: Successfully derived %+v for fixed version %q", CVE, gc, av.Fixed)
+					v.FixCommits = append(v.FixCommits, gc)
+				}
 			}
-			fc := cves.GitCommit{
-				Repo:   repo,
-				Commit: ref.Hash().String(),
+			if av.LastAffected != "" {
+				normalizedLastAffected, err := cves.NormalizeVersion(av.LastAffected)
```
oliverchang, line 118: These two branches for "Fixed" and "LastAffected" look **very)** similar. Can we create a helper function instead for these, and parameterize the bits that change?

vulnfeeds/cpp/main.go
andrewpollock: Yes, a CVE ID. Clarified in https://github.com/google/osv.dev/pull/1105/commits/db519e9ea51359703d144ecd26c1f241c8e8d16c

vulnfeeds/cpp/main.go
andrewpollock: Had a stab at addressing in [020dfb1](https://github.com/google/osv.dev/pull/1105/commits/020dfb1ca92fdc38f5aa34f4c9b847e17682a833)

oliverchang (approved):

vulnfeeds/cpp/main.go
```diff
@@ -75,84 +76,82 @@ func InScopeGitRepo(repoURL string) bool {
...
+		normalizedTags, err := git.NormalizeRepoTags(repo, cache)
 		if err != nil {
-			Logger.Warnf("Failed to get %s: %v", repo, err)
+			Logger.Warnf("[%s]: Failed to normalize tags for %s: %v", CVE, repo, err)
 			continue
 		}
 		for _, av := range versions.AffectedVersions {
 			// We can't map a non-existent version to a tag and commit, so skip.
-			if av.Fixed == "" {
+			if av.Fixed == "" && av.LastAffected == "" {
```
oliverchang, line 112: nit: we can probably just skip this condition completely since it's redundant.

andrewpollock: Removed in [993ea85](https://github.com/google/osv.dev/pull/1105/commits/993ea8584d7e429f00b8b358c25d33815fdf4d2b)

PR #1107, opened 2023-03-08
Performance improvements

another-rex:
Some performance improvements to the search page. 

- Use cached count for searches on just the ecosystem, rather than counting everything again. (The base `/list` page load time goes from (~1.3 seconds to ~270ms), and with smaller improvements to other ecosystem only searches (e.g. `/list?ecosystem=Debian`)
- Asynchronously query for search totals/counts while querying for actual search results. (should be faster, but small enough that it has improved a measurable amount)

With perf improvements: https://fc850a0288fe2bd9d9c7c0fbeaade561be64b1e2-dot-oss-vdb-test.wl.r.appspot.com/list

Current master branch: https://oss-vdb-test.wl.r.appspot.com/list


andrewpollock (approved):
Awesome performance improvement!

gcp/appengine/frontend_handlers.py
```diff
@@ -373,20 +374,37 @@ def osv_query(search_string, page, affected_only, ecosystem):
...
+
+  results = {
+      'total': total_future.get_result(),
+      'items': result_items,
+  }
 
   return results
 
 
+def get_ecosystem_count_for_eco(ecosystem: str) -> int:
```
oliverchang, line 400: nit: rename this function to `get_vuln_count_for_ecosystem`?

oliverchang (approved):

another-rex:
Upgraded to python3.11 as well since it's now officially supported by app engine.

PR #1108, opened 2023-03-08
Fix Terraform Redis Issues

michaelkedar:
Make a new Redis resource for the frontend because Terraform can't reproduce the one on production.
To not break the existing website & allow for rollbacks, the old one will stay around for about a month.

I've already run `terraform apply` on production and staging to generate the IPs, so no changes will appear in the plan check.

### DIFF
--- a/deployment/terraform/modules/osv/main.tf
+++ b/deployment/terraform/modules/osv/main.tf
@@ -12,7 +12,14 @@ resource "google_app_engine_application" "app" {
 }
 
 # MemoryStore
+# TODO(michaelkedar): The way this was initially created on production is not (easily) reproducable in Terraform.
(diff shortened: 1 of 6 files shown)


deployment/terraform/modules/osv/main.tf
```diff
@@ -12,7 +12,14 @@ resource "google_app_engine_application" "app" {
 }
 
 # MemoryStore
+# TODO(michaelkedar): The way this was initially created on production is not (easily) reproducable in Terraform.
```
andrewpollock, line 15:
```suggestion
# TODO(michaelkedar): The way this was initially created on production is not (easily) reproducible in Terraform.
```

oliverchang (approved):

PR #1109, opened 2023-03-08
Expand the VersionInfo struct to support Limit and LastAffected

andrewpollock:
Also expand AttachExtractedVersionInfo to support these fields.

This way we can generate OSV records containing Limit or LastAffected if Fixed cannot be determined.


vulnfeeds/vulns/vulns.go
```diff
@@ -369,6 +379,30 @@ func (affected *Affected) AttachExtractedVersionInfo(version cves.VersionInfo) {
 		affected.Ranges = append(affected.Ranges, gitRange)
 	}
 
+	for repo, commits := range repoToLimitCommits {
```
oliverchang, line 382: Can we consolidate everything (all  fixed, limit, last_affected events) in the one events array?

andrewpollock: I think I've done this in e45b64c29992a1047f2ad866f94911a2c9e439fc, PTAL?

oliverchang (approved):

Issue #1053, opened 2023-02-20
Make all errors reported through error reporting dashboard.


oliverchang:
Currently, it looks like a lot of our error logs are not properly propagated to the GCP error logging dashboard.

jayvdb:
https://github.com/google/osv.dev/pull/1111 was merged. Is there more to do, or can this be closed?

andrewpollock:
I'm not sure if #1111 captured "all" potential errors?

michaelkedar:
#1111 only covered the Python worker, importer and exporter.
Off the top of my head, what needs to be looked at:
- Converters - alpine, cve, debian
- The API backend
- The frontend osv.dev website

andrewpollock:
> * Converters - alpine, cve, debian

I *think* that all of this, being Go code and using https://github.com/google/osv.dev/blob/master/vulnfeeds/utility/logginghelpers.go should Do The Right Thing today. @another-rex, am I right?

another-rex:
Yep, that's correct as far as I know. 

Apart from Debian converters which is in python, and is just using normal print statements at the moment.

michaelkedar:
The Go logging helper LGTM.
Though if we were to add a `LoggerWrapper.Errorf` without a stack trace (that we want to show up on the Error Reporting dashboard) we'd need to do something like we have for the workers: https://github.com/google/osv.dev/blob/a2d45dae06d0de9acb9579d319da4e6d01eaf8f1/osv/logs.py#L38-L52

per https://cloud.google.com/error-reporting/docs/formatting-error-messages#log-text

PR #1111, opened 2023-03-10
Propagate Python workers' error logs to Error Reporting

michaelkedar:
Part of #1053


another-rex (commented):
Maybe it'll be worth it to pull the `_setup_gcp_logging` out into it's own library? Avoids duplicate code, and appengine / API / other google cloud pieces can use it as well.

michaelkedar:
> Maybe it'll be worth it to pull the `_setup_gcp_logging` out into it's own library? Avoids duplicate code, and appengine / API / other google cloud pieces can use it as well.

Good point. I've created `osv.logging_helper` and put it there.

another-rex (approved):
LGTM

osv/logging_helper.py
```diff
@@ -0,0 +1,60 @@
...
+                  'functionName': record.funcName,
+              }
+          },
+      })
+
+    return record
+
+  logging.setLogRecordFactory(record_factory)
+  logging.getLogger().setLevel(logging.INFO)
+  logging.getLogger('google.api_core.bidi').setLevel(logging.ERROR)
```
oliverchang, line 58: As a bit of documentation improvement while we're here, can you please add comments here to explain what these are doing? (i.e. suppressing noisy logs in our dependencies?)

osv/logging_helper.py
```diff
@@ -0,0 +1,60 @@
+# Copyright 2023 Google LLC
```
oliverchang, line 1: nit: can we just name this module "logging" ? "helper" is generally not a very descriptive word for naming.

oliverchang (approved):
thanks!! just some nits.

osv/logging_helper.py
michaelkedar: I could do that, but I'm a bit wary of some ambiguity - osv's `__init__.py` imports everything from a few files, including the python `logging` module. So you end up with something like
```python
>>> import osv
>>> osv.logging
"<module 'logging' from '/usr/lib/python3.11/logging/__init__.py'>"
>>> import osv.logging
>>> osv.logging
"<module 'osv.logging' from '/path/to/osv.dev/osv/logging.py'>"
```

(I actually forgot to import `osv.logging_helper` anyway - will fix that)

oliverchang: Ah, very good point here. 

In that case, can we name this something like "logs" instead?

osv/logging_helper.py
michaelkedar: Added that comment.

osv/logging_helper.py
michaelkedar: Changed to `osv.logs`

oliverchang (approved):

PR #1129, opened 2023-03-15
Add `--to-target` flag to Cloud Deploy promotion

michaelkedar:
Cloud Deploy doesn't automatically know where to promote a release if it's not currently deployed on staging (i.e. if it's not the most recent commit).
Explicitly telling which target to deploy to should fix this.


oliverchang (approved):
Thanks! This LGTM, but could you please add some more context/explanations for this change in your PR description?

PR #1130, opened 2023-03-15
Completely remove old AffectedCommit code.

oliverchang:
Part of #951.

### DIFF
--- a/docker/worker/worker.py
+++ b/docker/worker/worker.py
@@ -306,12 +306,6 @@ def filter_unsupported_ecosystems(vulnerability):
   vulnerability.affected.extend(filtered)
 
 
-def _is_linux_ecosystem(vulnerability):
-  """Return whether or not the vulnerability is for the Linux ecosystem."""
-  return any(affected.package.ecosystem == 'Linux'
-             for affected in vulnerability.affected)
-
-
 class TaskRunner:
   """Task runner."""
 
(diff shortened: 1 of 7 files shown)


docker/worker/worker.py
```diff
@@ -520,12 +515,7 @@ def _do_update(self, source_repo, repo, vulnerability, relative_path,
 
     bug.put()
 
-    # Temporary migration exclusion: Don't write legacy Linux AffectedComit
-    # entries as they cause Datastore scaling issues.
-    write_legacy = not _is_linux_ecosystem(vulnerability)
```
michaelkedar, line 525: Should `_is_linux_ecosystem()` also be removed?

docker/worker/worker_test.py
```diff
@@ -145,15 +145,6 @@ def setUp(self):
...
 
-    should_be_deleted = osv.AffectedCommit(
-        id='OSV-2020-1337-abcd',
-        bug_id='OSV-2020-1337',
-        commit='abcd',
-        project='project',
-        ecosystem='ecosystem',
-        public=False)
-    should_be_deleted.put()
-
```
michaelkedar, line 156: Is there no equivalent new version of this?
(I don't really know what this was testing...)

michaelkedar (approved):
LGTM

docker/worker/worker_test.py
oliverchang: Good catch, this was just checking if the new analysis overwrites previous results. I'll add the new equivalent here and a comment to explain.

PR #1143, opened 2023-03-21
Importer improvements

another-rex:
Couple of changes to improve performance of importer:

- Created and assigned importer to a specific node pool with exactly 1 node. This will hopefully keep importer's git clones on the same ssd and not erase it.

Seems to be working so far (changes were made at 2pm:
image

- Made debian and CVE scripts only upload changed entries by checking the checksum. This improves the rsync speed, and maintains the created time metadata. 
- Update importer script to only download files that have been created after the last update date.
- Updated the importer to interpret last_update_date as a datetime, rather than just a date. Inside datastore these map to the same datetime type, so no real datastore changes. 
- Added a missing -d flag for alpine convert.

Haven't been able to test this end to end (need to merge to do a staging release to test), but at a guess from the existing import times this should reduce import times to be < 2 mins

### Import times requiring first clone

Source | Time
-- | --
android | 0:00:23
cve-osv | 0:02:11
debian | 0:03:31
ghsa | 0:14:13
go | 0:00:09
nvd-poc | 0:01:28
python | 0:00:01
rust | 0:00:03
test-oss-fuzz | 0:00:02
uvi | 0:02:30
**Total** | **0:24:30**

### All imports that don't require fresh clone
Should be all imports after the importer pool change

Source | Time
-- | --
android | 0:00:23
cve-osv | 0:02:11
debian | 0:03:31
ghsa | 0:00:32
go | 0:00:09
nvd-poc | 0:01:28
python | 0:00:01
rust | 0:00:01
test-oss-fuzz | 0:00:01
uvi | 0:00:05
**Total** | **0:08:21**

### Predicted import times after changes in this PR

Source | Time
-- | --
android | 0:00:23
cve-osv | 0:00:11
debian | 0:00:11
ghsa | 0:00:32
go | 0:00:09
nvd-poc | 0:00:11
python | 0:00:01
rust | 0:00:01
test-oss-fuzz | 0:00:01
uvi | 0:00:05
**Total** | **0:01:44**


docker/importer/importer.py
```diff
@@ -201,12 +201,14 @@ def import_new_oss_fuzz_entries(self, repo, oss_fuzz_source):
                                self._git_callbacks(oss_fuzz_source)):
       ndb.put_multi(exported)
 
-  def schedule_regular_updates(self, repo, source_repo):
+  def schedule_regular_updates(self, repo, source_repo: osv.SourceRepository):
     """Schedule regular updates."""
+    aest_time_now = aestnow()
+
     if (source_repo.last_update_date and
         # OSV devs are mostly in located in australia,
```
andrewpollock, line 209:
```suggestion
        # OSV devs are mostly located in australia,
```

docker/importer/importer.py
```diff
@@ -217,7 +219,7 @@ def schedule_regular_updates(self, repo, source_repo):
 
     # Re-compute existing Bugs for a period of time, as upstream changes may
```
andrewpollock, line 220:
```suggestion
    # Re-compute existing Bugs after a period of time, as upstream changes may
```

andrewpollock: I think this makes more sense to readers?

docker/importer/importer.py
```diff
@@ -337,62 +339,72 @@ def _process_updates_git(self, source_repo: osv.SourceRepository):
...
     # First retrieve a list of files to parallel download
     storage_client = storage.Client()
-    listed_blob_names = [
-        blob.name for blob in storage_client.list_blobs(source_repo.bucket)
-    ]
+    utc_last_update_date = source_repo.last_update_date.replace(
+        tzinfo=datetime.timezone.utc)
+
+    # Convert to list to retrieve all information into memory
+    # This makes it's use in the concurrent map later faster
```
oliverchang, line 365:
```suggestion
    # This makes its use in the concurrent map later faster
```

docker/importer/importer.py
```diff
@@ -337,62 +339,72 @@ def _process_updates_git(self, source_repo: osv.SourceRepository):
...
+    listed_blob_names = list(storage_client.list_blobs(source_repo.bucket))
```
oliverchang, line 366: nit: these aren't names anymore right? but rather just `listed_blobs`.

oliverchang (approved):

michaelkedar (approved):

docker/importer/importer.py
andrewpollock: I guess what's ambiguous here is what's happening after a period of time? The re-computing, or the selection of bugs being recomputed?

another-rex: Had a quick chat with Oliver and expanded the comment a bit, can you see if it's clearer now?

andrewpollock (approved):

PR #1151, opened 2023-03-23
Refactor and test git commit mapping, add initial fuzzy version mapping

andrewpollock:


vulnfeeds/git/versions.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+	for i, t := range candidateTags {
+		// Handle the case where we were given say "12.0", but what we have is "12.0.0"
+		if strings.TrimPrefix(t, normalizedVersion) == "-0" {
+			return cves.GitCommit{
+				Repo:   repo,
+				Commit: normalizedTags[candidateTags[i]].Commit,
+			}, true
+		}
+	}
```
oliverchang, line 52: Should we log the candidates that we searched through but failed to match to aid in debugging later?

vulnfeeds/git/versions.go
```diff
@@ -0,0 +1,76 @@
...
+	if len(candidateTags) == 1 {
+		return cves.GitCommit{
+			Repo:   repo,
+			Commit: normalizedTags[candidateTags[0]].Commit,
+		}, true
+	}
+	for i, t := range candidateTags {
+		// Handle the case where we were given say "12.0", but what we have is "12.0.0"
+		if strings.TrimPrefix(t, normalizedVersion) == "-0" {
```
oliverchang, line 46: Would it be worth generalising this a bit more to an arbitrary number of -0s? 

e.g. 12.0.0.0 (which is a valid NuGet version), matching against 12.0.

vulnfeeds/git/versions_test.go
```diff
@@ -0,0 +1,53 @@
...
+			expectedResult: "ecf535f6cb70c97beb7c44dceadc14423086ca0d",
+			expectedOk:     true,
+		},
+		{
+			description:    "A fuzzy version match",
+			inputRepoURL:   "https://gitlab.com/gitlab-org/gitlab",
+			cache:          cache,
+			inputVersion:   "12.0",
+			expectedResult: "1d73db751369085d05d939c798e7da086646696c",
+			expectedOk:     true,
```
oliverchang, line 33: Can we add a false case as well?

oliverchang (approved):

vulnfeeds/git/versions.go
andrewpollock: Yes it absolutely is. 

I'll do that in a future PR, as until I've faked out remote Git repositories, I have to find a test case in the wild to use.

vulnfeeds/git/versions_test.go
andrewpollock: Done in [4d32497](https://github.com/google/osv.dev/pull/1151/commits/4d3249741636a1311a19d70f2af571f9f69e2e70)

vulnfeeds/git/versions.go
andrewpollock: I was trying to keep the (Cloud) logging code out of these functions to reduce complexity of running them under test.

For now, I'll see how things go without it, I think the error returned from `VersionToCommit` will add sufficient logging to enable band inspection of a failing match and the addition of more test cases to further assist with debugging.

PR #1157, opened 2023-03-24
Add another late-breaking discovered invalid repo

andrewpollock:


oliverchang (approved):

vulnfeeds/cmd/cperepos/main.go
```diff
@@ -162,6 +162,7 @@ var (
 		"https://github.com/Orange-Cyberdefense/CVE-repository",
 		"https://github.com/passtheticket/vulnerability-research",
 		"https://github.com/post-cyberlabs/CVE-Advisory",
+		"https://github.com/rapid7/metasploit-framework",
```
oliverchang, line 165: Hmm actually, are there any legitimate vulnerabilities in this repo? I can imagine there being CVEs for this.

andrewpollock: Good call, there are as it happens... 

I'll make a note to circle back to this particular one, because it's being invalidly attributed to a bunch of other CPEs other than just itself.

PR #1164, opened 2023-03-28
Rework processing of records

andrewpollock:
This is a bit of a grab bag of changes:

* Prevent duplicate repo URLs being derived for CPEs
* Reorder to be CPE-centric, rather than reference-centric
* Stop excluding references with particular tags, it was causing a reduction in successful record generation
  * This has the net effect of causing a spike in hypothetically convertible CVEs, because a whole bunch of repo-shaped URLs get derived for CPEs, with an overall drop in the successful conversion ratio

The next step will be to abstract out the known-irrelevant repo URLs from https://github.com/google/osv.dev/blob/f26a381ed5b6cca84936266adc49ca72c0ea776a/vulnfeeds/cmd/cperepos/main.go#L110 into a shared config file that both utilities can reference, and that'll bring the ratio back into better territory.


vulnfeeds/cpp/main.go
```diff
@@ -131,6 +131,41 @@ func GitVersionsToCommit(CVE string, versions cves.VersionInfo, repos []string,
 	return v, nil
 }
 
+func refAcceptable(ref cves.CVEReferenceData, tagDenyList []string) (acceptable bool) {
```
oliverchang, line 134: nit:  the return variable `acceptable` seems unused

vulnfeeds/cpp/main.go
```diff
@@ -131,6 +131,41 @@ func GitVersionsToCommit(CVE string, versions cves.VersionInfo, repos []string,
...
+		}
+	}
+	return true
+}
+
+// Examines the CVE references for a CVE's CPE and derives repos for it.
+func ReposForCPE(CVE string, cache VendorProductToRepoMap, vp VendorProduct, refs []cves.CVEReferenceData, tagDenyList []string) (repos []string) {
+	// This currently only gets called for cache misses, but make it not rely on that assumption.
+	cachedRepos, ok := cache[vp]
+	if ok {
```
oliverchang, line 147: nit: prefer

```golang

if cachedRepos, ok := cache[vp]; ok {
  return cachedRepos
}
```

It's shorter and scopes `cachedRepos` better

oliverchang (approved):

vulnfeeds/cpp/main.go
```diff
@@ -53,8 +53,8 @@ var Metrics struct {
 // References with these tags have been found to contain completely unrelated
 // repositories and can be misleading as to the software's true repository,
 var RefTagDenyList = []string{
-	"Exploit",
-	"Third Party Advisory",
+	// "Exploit",
```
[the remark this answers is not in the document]
andrewpollock: Fixed in [b2298f2](https://github.com/google/osv.dev/pull/1164/commits/b2298f296852fd5a931bb67374faaa0c733701fa)

vulnfeeds/cpp/main.go
andrewpollock: Fixed in [b2298f2](https://github.com/google/osv.dev/pull/1164/commits/b2298f296852fd5a931bb67374faaa0c733701fa)

vulnfeeds/cpp/main.go
andrewpollock: Fixed in [b2298f2](https://github.com/google/osv.dev/pull/1164/commits/b2298f296852fd5a931bb67374faaa0c733701fa)

PR #1165, opened 2023-03-28
Productionizing Debian copyright metadata mirroring

andrewpollock:
This is my first attempt at setting up all the additional plumbing for the Debian copyright metadata mirroring.

The objective is to get this up and going in the staging environment, prior to a release getting it going in the production environment.

### DIFF
--- a/deployment/clouddeploy/gke-workers/environments/oss-vdb-test/debian-copyright-mirror.yaml
+++ b/deployment/clouddeploy/gke-workers/environments/oss-vdb-test/debian-copyright-mirror.yaml
@@ -12,7 +12,5 @@ spec:
             env:
             - name: GCS_PATH
               value: gs://osv-test-cve-osv-conversion/debian_copyright
-            - name: WORK_DIR
-              value: /scratch
             - name: BE_VERBOSE
               value:
--- a/deployment/clouddeploy/gke-workers/environments/oss-vdb-test/debian-copyright-mirror.yaml
+++ b/deployment/clouddeploy/gke-workers/environments/oss-vdb-test/debian-copyright-mirror.yaml
@@ -1,0 +1,18 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: debian-copyright-mirror
+spec:
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          containers:
+          - name: debian-copyright-mirror
+            env:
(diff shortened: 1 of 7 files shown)


deployment/clouddeploy/gke-workers/environments/oss-vdb-test/debian-copyright-mirror.yaml
```diff
@@ -0,0 +1,18 @@
...
+    spec:
+      template:
+        spec:
+          containers:
+          - name: debian-copyright-mirror
+            env:
+            - name: GCS_PATH
+              value: gs://osv-test-cve-osv-conversion/debian_copyright
+            - name: WORK_DIR
+              value: /scratch
```
michaelkedar, line 16: nit: this `WORK_DIR` env can be moved to the base yaml, since it's the same in both environments.

michaelkedar (approved):
LGTM, with a tiny nit.

michaelkedar:
Also note: the mapping for the `debian-copyright-mirror` 'image' to the actual image address is done in `build-and-stage.yaml`:
https://github.com/google/osv.dev/blob/a2d45dae06d0de9acb9579d319da4e6d01eaf8f1/deployment/build-and-stage.yaml#L177-L188

Which was already added in #1146 - though in the future we'd probably want to do that change with these ones.

deployment/clouddeploy/gke-workers/environments/oss-vdb-test/debian-copyright-mirror.yaml
andrewpollock: Thanks, fixed in f47b8a5

Out of interest, with all this inheritance, there's a non-trivial cognitive burden when looking at the files to determine what the resultant spec actually looks like. Is there something you can run to see the merged YAML?

andrewpollock:
> though in the future we'd probably want to do that change with these ones

Got it. Chalk this up to me not fully understanding what I was cargo culting.

deployment/clouddeploy/gke-workers/environments/oss-vdb-test/debian-copyright-mirror.yaml
michaelkedar:
```
kubectl kustomize environments/oss-vdb-test > out.yaml
```
should output the kustomize'd yaml for that environment.

PR #1169, opened 2023-03-29
Add missing comma

andrewpollock:
I left out a comma and that seems to be making something not quite right...


deployment/build-and-stage.yaml
```diff
@@ -183,7 +183,7 @@ steps:
      debian-convert=gcr.io/oss-vdb/debian-convert:$COMMIT_SHA,\
      combine-to-osv=gcr.io/oss-vdb/combine-to-osv:$COMMIT_SHA,\
      alpine-cve-convert=gcr.io/oss-vdb/alpine-cve-convert:$COMMIT_SHA, \
-     debian-copyright-mirror=gcr.io/oss-vdb/debian-copyright-mirror:$COMMIT_SHA
+     debian-copyright-mirror=gcr.io/oss-vdb/debian-copyright-mirror:$COMMIT_SHA, \
```
michaelkedar, line 186: Not sure about having a space after the commas here, but I think it should be fine.

michaelkedar (approved):

PR #1171, opened 2023-03-29
Initial commit of productionization of gen-cperepos-map

andrewpollock:
Set up all the additional plumbing for the CPE repo map generation.

The objective is to get this up and going in the staging environment, prior to a release getting it going in the production environment.

### DIFF
--- a/deployment/clouddeploy/gke-workers/base/gen-cperepos-map.yaml
+++ b/deployment/clouddeploy/gke-workers/base/gen-cperepos-map.yaml
@@ -1,0 +1,33 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: gen-cperepos-map
+spec:
+  schedule: "0 6 */1 * *"
+  concurrencyPolicy: Forbid
+  jobTemplate:
+    spec:
+      activeDeadlineSeconds: 86400
+      template:
+        spec:
+          containers:
+          - name: gen-cperepos-map
+            image: gen-cperepos-map
+            imagePullPolicy: Always
+            securityContext:
+              privileged: true
+            resources:
+              requests:
+                cpu: 1
+                memory: "1G"
+              limits:
+                cpu: 1
+                memory: "2G"
+            env:
+              - name: WORK_DIR
+                value: /scratch
+          restartPolicy: OnFailure
+          volumes:
+            - name: "ssd"
+              hostPath:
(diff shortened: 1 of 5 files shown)


deployment/clouddeploy/gke-workers/base/gen-cperepos-map.yaml
```diff
@@ -0,0 +1,33 @@
...
+  schedule: "0 6 */1 * *"
+  concurrencyPolicy: Forbid
+  jobTemplate:
+    spec:
+      activeDeadlineSeconds: 86400
+      template:
+        spec:
+          containers:
+          - name: gen-cperepos-map
+            image: gen-cperepos-map
```
michaelkedar, line 15: This image needs to match `build-and-stage.yaml`, where it's called `cpe-repo-gen`
https://github.com/google/osv.dev/blob/cd244c5cd936674ce7dac0baec6eeb80f65394f8/deployment/build-and-stage.yaml#L187

deployment/clouddeploy/gke-workers/base/gen-cperepos-map.yaml
```diff
@@ -0,0 +1,33 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: gen-cperepos-map
+spec:
+  schedule: "0 6 */1 * *"
```
michaelkedar, line 6: Does this use what debian_copyright_mirror outputs? Would it be better to schedule this for a bit after it?

deployment/clouddeploy/gke-workers/base/gen-cperepos-map.yaml
andrewpollock: Thank you for catching that. Fixed in [69122a9](https://github.com/google/osv.dev/pull/1171/commits/69122a9771b00b4fc6ef73f3d195214859c1d304)

deployment/clouddeploy/gke-workers/base/gen-cperepos-map.yaml
andrewpollock: It does, but I'm not too worried about this being tightly coupled. 

I'll potentially dial down the frequency of the Debian copyright mirroring once it's proved itself, because I don't expect it to significantly change (although for the same reason, continuing to run it daily won't be a big deal).

michaelkedar (approved):

PR #1188, opened 2023-04-03
Update highend node type to much better fit indexer behavior

another-rex:
Each indexer pod needs a lot more memory than it needs CPUs. The current standard-32 leaves about 22 cores completely unused because there's not enough memory available to spin up more pods on each node.


oliverchang (approved):
good catch! @susperius FYI

PR #1190, opened 2023-04-04
Remove local_ssd_count, and convert to ssd

another-rex:
Works when running apply locally.

### DIFF
--- a/deployment/terraform/modules/osv/workers_gke.tf
+++ b/deployment/terraform/modules/osv/workers_gke.tf
@@ -92,9 +92,8 @@ resource "google_container_node_pool" "highend" {
 
   node_config {
     machine_type    = "n2-highmem-32"
-    disk_type       = "pd-standard"
+    disk_type       = "pd-ssd"
     disk_size_gb    = 100
-    local_ssd_count = 1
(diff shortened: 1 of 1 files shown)


michaelkedar (approved):
Looking at the indexer kubernetes files in `osv-config`, I can't tell if the indexer was even using the local SSDs

another-rex:
I think the ssd is mounted to /tmp, which is used to clone/copy the repositories that are being analysed.

oliverchang:
Note that local SSDs are a fair bit faster than pd-ssd: https://cloud.google.com/compute/docs/disks#localssds

Does this have any performance implications for us?

another-rex:
Ohh, I got mixed up persistent disks and local ssds mixed up. 

@michaelkedar I see what you mean now, I don't see the configs using the local disk either. Maybe it automatically uses local ssds when they are available? 

> Does this have any performance implications for us?

Yes and no? This should be faster than what is currently being used (just standard persistent hard disks I believe, since no requests specifically for the ephemeral ssds), but we would probably get a speed boost by switching to using local ssds. I'll look into configuring it.

PR #1191, opened 2023-04-04
Automate cleaning up excessive App Engine versions

andrewpollock:
Add a build stage that executes a new script that cleans up all but the last 5 App Engine versions.

### DIFF
--- a/deployment/build-and-stage.yaml
+++ b/deployment/build-and-stage.yaml
@@ -172,6 +172,10 @@ steps:
 - name: 'gcr.io/oss-vdb/deployment'
   args: ['bash', '-ex', 'gcp/appengine/deploy.sh', 'oss-vdb-test', 'deployment/gae/oss-vdb-test', 'app.yaml']
 
+# Clean up all but the last 5 App Engine versions to avoid hitting the ceiling on versions
+- name: 'gcr.io/cloud-builders/gcloud'
+  args: ['bash', '-ex', 'gcp/appengine/cleanup_versions.sh', 'oss-vdb-test']
+
 # Cloud Deploy
 # GKE Workers
 - name: 'gcr.io/cloud-builders/gcloud'
--- a/deployment/build-and-stage.yaml
+++ b/deployment/build-and-stage.yaml
@@ -172,7 +172,7 @@ steps:
 - name: 'gcr.io/oss-vdb/deployment'
   args: ['bash', '-ex', 'gcp/appengine/deploy.sh', 'oss-vdb-test', 'deployment/gae/oss-vdb-test', 'app.yaml']
 
-# Clean up all but the last 5 App Engine versions to avoid hitting the ceiling on versions
+# Clean up older, non-serving App Engine versions to avoid hitting the ceiling on versions
 - name: 'gcr.io/cloud-builders/gcloud'
   args: ['bash', '-ex', 'gcp/appengine/cleanup_versions.sh', 'oss-vdb-test']
 
--- a/gcp/appengine/cleanup_versions.sh
+++ b/gcp/appengine/cleanup_versions.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+# Helper to clean up all but the most recent 5 versions of the default service
+# in the given project
+
+# Inspiration: https://stackoverflow.com/questions/34499875/how-to-automatically-delete-old-google-app-engine-version-instances
+
+if [[ $# -lt 1 ]]; then
+  echo "Usage: $(basename $0) project_id"
+  exit 1
+fi
+
+PROJECT_ID="$1"
+
+VERSIONS="$(gcloud \
+  --project=$PROJECT_ID \
+  app versions list --service=default \
+  --format="value(version.id)" --sort-by="~version.createTime" \
+  | tail -n +6)"
+
+gcloud --project=$PROJECT_ID app versions delete --service=default --quiet $VERSIONS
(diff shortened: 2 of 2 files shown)


michaelkedar (commented):
We also should be running the script on the production branch. It's probably easiest to just run it manually (**after** deployment - production still stages every commit to master). Otherwise, it should go in `deploy-prod.yaml` but that currently won't have access to any files in the repo.

Also deleting all these versions will get rid the testing versions @another-rex has been working on. Is this alright / annoying?

LGTM otherwise.

andrewpollock:
> Also deleting all these versions will get rid the testing versions @another-rex has been working on. Is this alright / annoying?

I'll let @another-rex be the judge of that :-)

andrewpollock:
> We also should be running the script on the production branch.

Yeah I figured we'd prove this in the staging project first and then figure out how to move it over to production.

another-rex:
Will this delete master / production appengine instances? Don't we currently do a app engine release every PR merge, so if we clean up the last 5 we will "cleanup" the current production instance. 

Even moving this to deploy-prod.yaml might still cause problems, since we will want to have the previous production appengine instance to roll back to if we discover something wrong with the newly deployed frontend, and if there has been more than 5 commits / PR merged between releases, we will lose the old app engine instance.

andrewpollock:
> Will this delete master / production appengine instances? Don't we currently do a app engine release every PR merge, so if we clean up the last 5 we will "cleanup" the current production instance.

It keeps all but the most recent 5, not the last 5.

another-rex:
> > Will this delete master / production appengine instances? Don't we currently do a app engine release every PR merge, so if we clean up the last 5 we will "cleanup" the current production instance.
> 
> It deletes all but the most recent 5, not the last 5.

Ah yes, that was a typo, edited the message, basically each PR that is merged creates a new app engine instance that's newer than the production instance, so production will quickly become older than the 5 most recent instances, and get removed.

oliverchang:
Yeah only keeping the most recent 5 seems a bit too little. We should keep a lot more here. 

The limit is 210. Maybe let's keep half of that (105), which should be plenty of room.

andrewpollock:
> Ah yes, that was a typo, edited the message, basically each PR that is merged creates a new app engine instance that's newer than the production instance, so production will quickly become older than the 5 most recent instances, and get removed.

Oh, got it. `gcloud app versions delete --help` says "You cannot delete a version of a service that is currently receiving traffic." so the "live" version won't get unceremoniously deleted, the question is what is the failure mode...

Let's experiment! (I <3 having a test sandbox)

```
$ gcloud --project=oss-vdb-test app versions list --service=default --sort-by="~version.createTime" | head
SERVICE  VERSION.ID                                TRAFFIC_SPLIT  LAST_DEPLOYED              SERVING_STATUS
default  fc61316310651bda9f81f8e1005d71f50f0d09b3  1.00           2023-04-04T12:03:46+10:00  SERVING
default  fc22abcc4f63e25241dbb35613055b1e0ec5abfc  0.00           2023-04-04T11:56:44+10:00  SERVING
default  c32463f4d86d4acedfb82d26b4f67a07d46e15aa  0.00           2023-04-03T15:10:19+10:00  SERVING
default  72be03ba2f5a43af1d3d6caaae51f356f517420e  0.00           2023-04-03T12:18:00+10:00  SERVING
default  b53b4d3ef892c53d3afc34da14439a5ca9cd8c79  0.00           2023-04-03T12:10:52+10:00  SERVING
default  886ca915731f09a59d9bb1d7ea76d397ff510fb8  0.00           2023-04-03T12:03:55+10:00  SERVING
default  816adc3e99cfcc44abda8d01cd97ca72903c950c  0.00           2023-04-03T11:57:04+10:00  SERVING
default  3aecbeaaaf78e54894e1917839ce2ac4782641b0  0.00           2023-04-03T10:51:43+10:00  SERVING
default  b7cc034bbc83c7c579d4935f7f6a17a997ea96d5  0.00           2023-04-03T09:58:08+10:00  SERVING

$ gcloud --project=oss-vdb-test app versions delete fc61316310651bda9f81f8e1005d71f50f0d09b3
ERROR: (gcloud.app.versions.delete) Version [fc61316310651bda9f81f8e1005d71f50f0d09b3] is currently serving 100.00% of traffic for service [default].

Please move all traffic away via one of the following methods:
 - deploying a new version with the `--promote` argument
 - running `gcloud app services set-traffic`
 - running `gcloud app versions migrate`
 ```

So it's going to fail messily.

**Comment by andrewpollock** [comment]
Okay, I've upped the limit to 105, and explicitly filtered out versions serving traffic.

File: deployment/build-and-stage.yaml [hunk 1]
```diff
@@ -172,6 +172,10 @@ steps:
 - name: 'gcr.io/oss-vdb/deployment'
   args: ['bash', '-ex', 'gcp/appengine/deploy.sh', 'oss-vdb-test', 'deployment/gae/oss-vdb-test', 'app.yaml']
 
+# Clean up all but the last 5 App Engine versions to avoid hitting the ceiling on versions
```
another-rex, line 175: nit: Comment here still says 5

another-rex (approved):
LGTM with 1 nit. 

With 105 or so versions it will take a while for any testing app engine versions to get cleaned up, so that should be fine. I can always push another testing version if it does get cleaned up.

michaelkedar (approved):

deployment/build-and-stage.yaml
andrewpollock: Thanks fixed in [13bb8cc](https://github.com/google/osv.dev/pull/1191/commits/13bb8ccdadeba0cc0461e91d8babb767ab743b93)

PR #1197, opened 2023-04-05
Unpin python ndb

andrewpollock:
I need a newer version of the Datastore client to get the "IN" operator, it seems, and the original reasons for pinning it based on #916 don't appear to be applicable any more.


oliverchang (commented):
Hmm .IN should already exist in python-ndb <2.0.0. Or are you referring to the plain google cloud datastore client (which is blocked on this constraint) ?

andrewpollock:
> Hmm .IN should already exist in python-ndb <2.0.0. Or are you referring to the plain google cloud datastore client (which is blocked on this constraint) ?

Yes, the latter. I'd have thought that given [python-ndb is what has the dependency on the datastore client](https://deps.dev/pypi/google-cloud-ndb/2.1.1/dependencies), that it's still using it under the hood?

oliverchang:
LGTM, but please monitor this carefully. We had breakages with this in the past.

oliverchang (approved):
LGTM, but please monitor this carefully. We had breakages with this in the past.

PR #1205, opened 2023-04-10
Docs conversion PR #3: Delete Old Docs

hayleycd:
This is the third pull request in the docs conversion effort. 

Please merge #1203 and #1204 first. 

This PR deletes the existing docs. This will be ready once the first two PRs are merged and the resulting merge conflicts are resolved.


oliverchang:
Thanks Hayley! Note that we can't merge this as is, because we still need some of the tooling/scripts in `docs/` for the  swagger definition auto-generation. 

You don't have to deal with this though -- we can clean up the old docs such that these bits still exist.

PR #1206, opened 2023-04-10
Bringing homepage in line with docs

hayleycd:
When I was updating the docs, I saw another spot to adjust the API calls to bring them in line with the docs.


oliverchang (approved):
thanks!

PR #1208, opened 2023-04-11
Fix failing tests.

oliverchang:
Also replace github.com/go-test/deep with github.com/google/go-cmp which generates more readable diffs.


michaelkedar (approved):
LGTM

another-rex (approved):

PR #1214, opened 2023-04-12
Update determineversions API.

oliverchang:
- Make CPE field an explicit cpe23 string.
- Make the file_path comment clearer re its formatting and requirement.


another-rex (commented):
LGTM, but why is the api_descriptor removed though? Because it can be generated from the source?

oliverchang:
huh, weird. I think it's a bug in GitHub's UI, because that file definitely still exists in the branch for this PR: https://github.com/google/osv.dev/blob/update-api/gcp/api/v1/api_descriptor.pb

another-rex (approved):
Ah I think I now understand what the UI is saying, it's just 164 bytes smaller, and 100% of the binary file have changed, not that 164 bytes is 100% of the file.

PR #1216, opened 2023-04-12
Remove old docs from /docs

hayleycd:
I closed #1205 in favor of this. 

This pull request deletes the files from docs/ that are no longer needed with the new setup. 

You can check the sample docs [here](https://hayleycd.github.io/osv.dev) to see that the docs are still working despite removing these files. 

There are other files that should be deleted with this change, and I captured that in issue #1217


oliverchang:
We still need some of the bits here for teh openapi definition generation, so I opened https://github.com/google/osv.dev/pull/1216 instead for this!

PR #1218, opened 2023-04-14
Remove OpenAPI generation hacks.

oliverchang:
These were previously needed for our docs to render in the way we want. With our new docs, these are no longer necessary.


hayleycd:
I think something could be said for adding some information about the github page in the README alongside the info about OpenAPI. However, I can open a PR next week to add that. Otherwise this looks alright to me.

hayleycd (approved):

Issue #1211, opened 2023-04-11
Migrate to public deps.dev API 


oliverchang:
https://github.com/google/deps.dev

oliverchang:
(We may also no longer need an API key since there's no rate limits).

oliverchang:
This was done. Thanks a ton @sarnesjo!

PR #1219, opened 2023-04-14
Use deps.dev v3alpha HTTP API instead of v2alpha gRPC API

sarnesjo:
deps.dev recently released an updated API, which is available via HTTP as well as gRPC, and does not require authentication.

Update DepsDevMixin to use the new HTTP API when enumerating versions, and remove the fallback logic for Maven. Remove the old protos, along with the logic for handling API keys.

Addresses #1211.


osv/ecosystems/maven_test.py
```diff
@@ -275,8 +272,6 @@ def test_next_version(self):
     with self.assertRaises(ecosystems.EnumerateError):
       ecosystem.next_version('blah:doesnotexist123456', '1')
 
-  @unittest.skipIf(
-      os.getenv('DEPSDEV_API_KEY'), 'Unnecessary if using deps.dev')
   @mock.patch('requests.Session.get', side_effect=requests.get)
   def test_next_version_with_cache(self, mock_get):
     """Test next_version using cache."""
```
oliverchang, line 277: nit: I don't think this test is needed at all anymore, since we removed all the previous code that used the cache.

oliverchang (approved):
thanks a ton @sarnesjo !! 

just one minor nit

PR #1224, opened 2023-04-18
Add NoSchedule to see if it's fixed

another-rex:
Following the doc here: https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/


michaelkedar:
I'm not sure this'll do what you want it to:

A note in that doc says
> An empty `effect` matches all effects with key `key1`.

Also, the node pool doesn't have a `NoSchedule` taint.

another-rex:
Oh I see, I am confusing no execute with no schedule

PR #1229, opened 2023-04-19
Fix copyright consumption path

andrewpollock:
The GCS path wasn't being rsynced as I expected, so the wrong path was being passed to the tool.


andrewpollock:
Hmm, somehow I've managed to get bits of https://github.com/google/osv.dev/pull/1229 in here...

andrewpollock:
> Hmm, somehow I've managed to get bits of #1229 in here...

Actually no I didn't, I'd rebased or tried to rebase that out...

michaelkedar (approved):

PR #1230, opened 2023-04-19
Fix lint findings

andrewpollock:
It seems to feel differently about this than older versions did and now unrelated PRs are failing tests on GitHub.


michaelkedar (approved):
I really don't like this formatting, but I don't know if there's a specific lint rule that causes it.

andrewpollock:
> I really don't like this formatting, but I don't know if there's a specific lint rule that causes it.

Looking at https://github.com/google/yapf/commit/3684f3c6bab91581e80a40a16343fe8916cdb9ff and the changelog, it's a very deliberate "fix".

PR #1236, opened 2023-04-20
Bucketed Indexer

another-rex:
Indexer implementing bucketing to speed up and make more time consistent queries. 

Changes the API interface to add 2 extra fields in the response.

- minimum file matches
- estimated different files

Also added a number of bug fixes:

- Commit hash is now using the real commit hash rather than the reference hash of the tag. The datastore key is still using the reference hash for cases where multiple tags are on the same commit. An extra field `reference` in the datastore to save the reference hash
- Controller now correctly avoids pushing already existing items to the pub/sub
-


docker/indexer/stages/preparation/preparation.go
```diff
@@ -150,6 +151,13 @@ func (s *Stage) processGit(ctx context.Context, repoCfg *config.RepoConfig) erro
 	commitTracker := make(map[plumbing.Hash]bool)
 	// repoInfo is used as the iterator function to create RepositoryInformation structs.
 	repoInfo := func(ref *plumbing.Reference) error {
+		// Resolve the real commit hash
+		commitHash, err := repo.ResolveRevision(plumbing.Revision(ref.Name().String()))
```
oliverchang, line 155: Can you also add this fix info to the PR description? that we resolve the proper commit hash now as well?

docker/indexer/indexer.go
```diff
@@ -1,17 +1,17 @@
 /*
 Copyright 2022 Google LLC
 
- Licensed under the Apache License, Version 2.0 (the "License");
- you may not use this file except in compliance with the License.
- You may obtain a copy of the License at
+	 Licensed under the Apache License, Version 2.0 (the "License");
```
oliverchang, line 4: Is this change intentional?

tools/indexer-api-caller/caller.go
```diff
@@ -0,0 +1,99 @@
...
+		if i == len(fileResults)-1 {
+			fmt.Fprintf(&b, "{\"hash\": \"%s\"}", base64.StdEncoding.EncodeToString(fr.Hash))
+		} else {
+			fmt.Fprintf(&b, "{\"hash\": \"%s\"},", base64.StdEncoding.EncodeToString(fr.Hash))
+		}
+	}
+	b.WriteString("]}}")
+
+	cmd := exec.Command("bash")
+	cmd.Args = append(cmd.Args, "-c", `grpcurl -plaintext -d @ -protoset api_descriptor.pb 127.0.0.1:8000 osv.v1.OSV/DetermineVersion`)
```
oliverchang, line 82: nit: probably add a TODO to use proper gRPC library calls here.

gcp/api/testing_runner.py
```diff
@@ -0,0 +1,29 @@
+import subprocess
```
oliverchang, line 1: Can you add a license header? Also some comments around what this is doing?

oliverchang (approved):
very nice!! some initial cursory comments. will review in more detail after lunch.

docker/indexer/indexer.go
another-rex: Ah no, I think the auto formatter changed this, the indentation on the copyright notices are a bit weird though, so I made them all the same now.

tools/indexer-api-caller/caller.go
another-rex: This file is not meant to be added, I'll delete it. It was used to test the consistency of the random hash selection.

gcp/api/server.py
```diff
@@ -35,16 +39,16 @@
...
 _MAX_BATCH_QUERY = 1000
 _MAX_VULNERABILITIES_LISTED = 16
-_MAX_HASHES_TO_TRY = 50
-_MAX_COMMITS_TO_TRY = 10
+_MAX_MATCHES_TO_CARE = 100
+_MAX_RESULTS_TO_RETURN = 10
+_BUCKET_SIZE = 512
 
 _ndb_client = ndb.Client()
+profiler = cProfile.Profile()
```
oliverchang, line 51: Is this still neeed?

docker/indexer/stages/processing/processing.go
```diff
@@ -123,7 +144,54 @@ func (s *Stage) processGit(ctx context.Context, repoInfo *preparation.Result) er
...
+				}
+			}
+			return false
+		})
+
+		hasher := md5.New()
+		for _, v := range buckets[bucketIdx] {
+			_, err := hasher.Write(v.Hash)
+			if err != nil {
+				deflog.Panicf("Hasher error: %v", err)
```
oliverchang, line 186: Can we instead just use log.Fatalf from glog?

gcp/api/Pipfile
```diff
@@ -15,3 +15,4 @@ grpcio-reflection = "*"
 
 [dev-packages]
 grpcio-tools = "*"
+yapf = "*"
```
oliverchang, line 18: Is this needed when the root `Pipfile` also contains this? Introducing this here could also introduce conflicting versions and behaviour.

tools/indexer-api-caller/caller.go
oliverchang: Hmm, it's probably still useful to have a ready-to-use CLI tool to call the API.

gcp/api/v1/osv_service_v1.proto
```diff
@@ -97,12 +97,19 @@ message VersionMatchList {
...
   double score = 1;
+
   // Information about the upstream repository.
   VersionRepositoryInformation repo_info = 2;
   // The OSV identifier.
   osv.Package osv_identifier = 3;
   // CPE 2.3.
   string cpe23 = 5;
+
+  // Definite number of files that have matched
```
oliverchang, line 108: nit: add periods after these comments here to match the other comments in this file.

docker/indexer/stages/processing/processing.go
```diff
@@ -55,6 +69,9 @@ type Stage struct {
 	PubSubOutstandingMessages int
 }
 
+// bucketCount should be a divisor of 2^16
```
oliverchang, line 72: maybe also add a note that changing this will require some backfilling/reindexing?

docker/indexer/stages/processing/processing.go
```diff
@@ -123,7 +144,54 @@ func (s *Stage) processGit(ctx context.Context, repoInfo *preparation.Result) er
...
+		buckets[idx] = append(buckets[idx], fr)
+	}
+
+	results := make([]*BucketNode, bucketCount)
+
+	for bucketIdx := range buckets {
+		// Sort hashes to produce deterministic bucket hashes
+		sort.Slice(buckets[bucketIdx], func(i, j int) bool {
+			for k := 0; k < len(buckets[bucketIdx][i].Hash); k++ {
+				if buckets[bucketIdx][i].Hash[k] < buckets[bucketIdx][j].Hash[k] {
```
oliverchang, line 172: Can we just use https://pkg.go.dev/bytes#Compare here instead for lexicographic comparison of these hashes?

docker/indexer/stages/processing/processing.go
another-rex: I don't think Fatalf calls deferred functions while panics do, we need that to acknowledge the messages.

gcp/api/Pipfile
another-rex: I think it's needed? Root pipfile is not applied here since pipenv does not actually read other Pipfile's apart from the current one even if you import it.

oliverchang: Hmm, the way I typically do development (and as recommended in https://github.com/google/osv.dev/blob/master/CONTRIBUTING.md#prerequisites) 

Is to do a `pipenv shell`  at the root for all development. That gives us all the needed dev tools in a consistent way. This is how the github actions does things.
 
Then to run stuff/tests etc in e.g. gcp/api, we just do `pipenv run` which uses the Pipfile in those directories.

gcp/api/server.py
another-rex: Removed

docker/indexer/stages/processing/processing.go
another-rex: Yep, replaced it with bytes.Compare and added a test.

gcp/api/Pipfile
another-rex: That makes sense, removed the line.

docker/indexer/stages/processing/processing.go
oliverchang: In cases like these, should we actually ack the message? It seems like we should retry in these cases.

docker/indexer/stages/processing/processing.go
```diff
@@ -36,15 +41,25 @@ import (
...
 type FileResult struct {
 	Path string `datastore:"path,noindex"`
-	Hash []byte `datastore:"hash"`
+	Hash Hash   `datastore:"hash"`
+}
+
+// FileResult holds the per file hash and path information.
+type BucketNode struct {
+	NodeHash Hash `datastore:"node_hash"`
+	// ChildHashes    []Hash `datastore:"child_hashes,noindex"`
```
oliverchang, line 58: nit: delete these unused properties completely.

docker/indexer/stages/processing/processing.go
```diff
@@ -123,7 +147,46 @@ func (s *Stage) processGit(ctx context.Context, repoInfo *preparation.Result) er
 		}
 		return nil
 	}); err != nil {
-		return err
+		return fmt.Errorf("failed during file walk: %v", err)
+	}
+
+	log.Info("begin processing tree")
```
oliverchang, line 151: nit: s/tree/buckets/ ?

docker/indexer/storage/storage.go
```diff
@@ -30,12 +30,11 @@ import (
...
-	resultKind = "RepoIndexResult"
-	// Address-HashType-CommitHash
-	docKeyFmt = "%s-%s-%x"
-	// CommitHash-HashType-Page
-	resultKeyFmt = "%x-%s-%d"
-	pageSize     = 1000
+	bucketKind = "RepoIndexBucket"
+	// Address-HashType-ReferenceHash
+	docKeyFmt               = "%s-%s-%x"
+	treeKeyFmt              = "%x-%s-%d"
```
oliverchang, line 36: nit: add a comment for similar to line 34 here.

docker/indexer/storage/storage.go
```diff
@@ -125,7 +107,8 @@ func (s *Store) Exists(ctx context.Context, addr string, hashType string, hash p
 	if _, ok := s.cache.Load(fmt.Sprintf(docKeyFmt, addr, hashType, hash)); ok {
 		return true, nil
 	}
-	key := datastore.NameKey(docKind, fmt.Sprintf(docKeyFmt, addr, hashType, hash), nil)
+	// hash[:], the [:] is important, since the formatting uses %x, which will return a different result if not used
```
oliverchang, line 110: Can you explain this a bit more? Does this change the type which is why it's needed?

gcp/api/server.py
```diff
@@ -35,14 +38,13 @@
...
 
-from typing import List
-
 _SHUTDOWN_GRACE_DURATION = 5
 
 _MAX_BATCH_QUERY = 1000
 _MAX_VULNERABILITIES_LISTED = 16
-_MAX_HASHES_TO_TRY = 50
-_MAX_COMMITS_TO_TRY = 10
+_MAX_MATCHES_TO_CARE = 100
```
oliverchang, line 45: Can you add more comments here to describe these constants? 

Some of these (e.g. `_MAX_RESULTS_TO_RETURN` ) also sound a bit generic when they're only applicable ot the determine versions endpoint.

docker/indexer/stages/processing/processing.go
another-rex: The current process is that all messages are acked because it won't show up in RepoIndex, causing the controller to queue it up to be ran again the next day. 

(This actually makes me wonder if there's a specific reason the ack is done at the end rather than immediately when pulling the message.)

But looking at the implementation of md5.Write(), err is actually not set, it's only needed to fulfill the io.Writer interface, so I can actually just remove the check.

gcp/api/server.py
```diff
@@ -110,104 +112,132 @@ def QueryAffectedBatch(self, request, context):
...
+  """Build sorted determine version result from the input"""
+  idx_futures = ndb.get_multi_async(candidate_file_matches.keys())
+  output = []
+
+  for f in idx_futures:
+    idx: osv.RepoIndex = f.result()
+    version_match = osv_service_v1_pb2.VersionMatch(
+        score=candidate_file_matches[idx.key] / max_files,
+        minimum_file_matches=candidate_file_matches[idx.key],
+        estimated_diff_files=estimate_diff(
```
oliverchang, line 160: Not sure I fully understand the formula here just yet, but would it make sense to utilise this in the `score` calculation instead of the simple division?

oliverchang (commented):
very nice! This mostly looks good. just some more minor comments/questions

docker/indexer/storage/storage.go
another-rex: Added more comments here to explain it, basically plumbing.Hash already implements it's own String() method, so this is formating it into hex twice.

gcp/api/server.py
another-rex: That would make a lot of sense actually, changed to have the score be based off the estimated different files value.

oliverchang (approved):

PR #1237, opened 2023-04-20
Fix links to new documentation.

oliverchang:
The previous redirection had problems with Turbo. Instead just link directly to GitHub pages to avoid these issues.

Also fix the /docs/ redirection since the App Engine routing rules overrode things.


oliverchang:
Friendly ping!

PR #1238, opened 2023-04-20
Upgrade semver to >=3.0.0

michaelkedar:
semver 3.0.0 breakingly renamed `VersionInfo.isvalid` to `Version.is_valid`.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -6,7 +6,7 @@ name = "pypi"
 [packages]
 google-cloud-ndb = "*"
 google-cloud-logging = "*"
-semver = "*"
+semver = ">=3.0.0"
 pyyaml = "*"
 pygit2 = ">=1.9.2"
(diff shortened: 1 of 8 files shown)


michaelkedar:
The lockfiles in `docker/exporter`, `docker/worker`, and `gcp/api` still refer to semver v2.

andrewpollock:
> The lockfiles in `docker/exporter`, `docker/worker`, and `gcp/api` still refer to semver v2.

That would explain why the tests are failing for https://github.com/google/osv.dev/pull/1141

Is there a reason not to fix all the call sites?

michaelkedar:
> Is there a reason not to fix all the call sites?

I should fix them, relocking was just causing a bunch of other dependencies that might cause other issues like #1183

andrewpollock (approved):

PR #1250, opened 2023-04-26
Add last touches to the indexer API

another-rex:
and update Pipfile to include direct dependency of ndb.

Changes to API:
- Cutoff response after score is below 0.2
- Adjust formula to use the adjusted bucket size that ignores the empty buckets.
- Changed protobuf to make hash_type an enum
- Made osv folder in api a symbolic link instead of copying it every time tests are ran. This follows what is done in v1 folder and in appengine.
- Updated Pipfile dependencies to resolve conflicts. All latest versions of grpcio should work now. 
- Fully match against empty buckets via a empty bucket bitmap, this simplifies the estimation formula by a lot
- Account for file count differences by storing file count in RepoIndex
- Fixes bug where fetch is not pulling in new tags


oliverchang (approved):

oliverchang:
Can you also update the PR description to describe the main changes?

oliverchang:
Actually for that symbolic link change, that might break the deploy/Docker build process. Docker would see this directory has outside of the build context, and refuse to build.

another-rex:
The API docker image is built from the root directory, so it'll still be in context, and osv is already copied over, so the symbolic link will still be intact as well (but it won't be used anyway, since osv is already imported in as a library).

oliverchang:
> The API docker image is built from the root directory, so it'll still be in context, and osv is already copied over, so the symbolic link will still be intact as well (but it won't be used anyway, since osv is already imported in as a library).

Ah, I misremembered. Yeah this should work then.

oliverchang:
Is this ready to merge?

another-rex:
It should be ready to merge now, added two more fields to RepoIndex: empty_bucket_bitmap, and file_count.

gcp/api/server.py
```diff
@@ -165,10 +169,32 @@ def build_determine_version_result(
...
+
   for f in idx_futures:
     idx: osv.RepoIndex = f.result()
+
+    # Byte order little is how the bitmap is stored in the indexer originally
+    bitmap = int.from_bytes(idx.empty_bucket_bitmap, byteorder='little')
+
+    # We are looking to find cases where the bitmap generated by the user query
+    # gives a 0, but the bitmap of the repo is a 1. We do not want to count
+    # cases where bitmap of the repo is a 0, and the user query returns is 1,
```
oliverchang, line 184: nit: grammar is a little confusing from the second sentence onwards. 

Something like:

```
We do not want to count cases where the repo bitmap contains 0 but the user bitmap
contains 1, since these are already accounted for by not having these in the query
results in the first place.
```
 
seems clearer? (assuming I'm understanding this correctly).

oliverchang (approved):
very neat!

PR #1254, opened 2023-04-28
When all else fails, derive a last affected version from the CPE string

andrewpollock:
When there are no explicit versions in the CPE match record, use the version in the CPE string as the last affected version

This will allow an OSV record to be generated in these situations.


vulnfeeds/cves/versions.go
```diff
@@ -559,7 +559,24 @@ func ExtractVersionInfo(cve CVEItem, validVersions []string) (v VersionInfo, not
...
+				}
+				if CPE.Part != "a" {
+					// Skip operating system CPEs.
+					continue
+				}
+				if slices.Contains([]string{"NA", "ANY"}, lastaffected) {
+					// These are meaningless converting to commits.
+					continue
+				}
+				lastaffected = CPE.Version
```
oliverchang, line 576: Does a CPE version always mean last affected? Or is it supposed to refer to a single version?

andrewpollock: A single version with `vulnerable = true` means that specific version is vulnerable.

I can see there being a lot of CVEs where there's no fixed version in the CPE match info so all we'll be able to get is a last affected version.

oliverchang: I see, so we're just assuming that in such cases, everything that came prior is also vulnerable? Let's document that in comments at least to be clearer.

andrewpollock: Yes, no known introduced version. Added an additional comment to this effect in 
9ee3483

oliverchang: I also wonder if we should be adding all this context into a database_specific field somewhere, to aid future users in understanding the gaps of what we generated.

andrewpollock: > I also wonder if we should be adding all this context into a database_specific field somewhere, to aid future users in understanding the gaps of what we generated.

I don't think there's any useful context here to add (yet). I can see a scenario where we review all of the converted CVEs that are appearing as unfixed and determine fixed versions and contribute a *lot* of CPE Match updates back to the NVD...

oliverchang: I'm more referring to context for where our data came from. 

e.g. something like:

````
commit_source:
- last_affected: cpe
```

```
commit_source:
- fixed: reference_link
```

etc...

andrewpollock: > I'm more referring to context for where our data came from.

Ah, got it. That sounds like something that should be thought through in a little more detail and something I can look to add in a future PR.

oliverchang (approved):

PR #1256, opened 2023-05-02
Add more invalid repositories

andrewpollock:
Did a top to bottom eyeball of the CVEs that had repos but hadn't successfully found any ranges, and found plenty more invalid repositories.


oliverchang (approved):

oliverchang:
This list is starting to get quite big. I wonder if there's some other heuristic we can use here to determine if a repo is valid or not, based on either its contents or stats from GitHub (e.g. percentage of code in a real programmign language).

oliverchang:
(hit the wrong button by accident).

andrewpollock:
> I wonder if there's some other heuristic we can use here to determine if a repo is valid or not, based on either its contents or stats from GitHub (e.g. percentage of code in a real programmign language).

There's potential opportunity to expand the regex and chip away at the list a bit, but yeah, it'd be nice if there were some other (reliable) signal that a repo is for vulnerability research purposes. I was thinking it might be good to raise with the GitHub folks next time we're talking to them (tomorrow?)

PR #1267, opened 2023-05-04
Add AlmaLinux and Rocky Linux

hayleycd:
Adding the two new distros to the docs. 

Changes can be viewed [here](https://hayleycd.github.io/osv.dev/data/#covered-ecosystems)


oliverchang (approved):
thanks Hayley!

PR #1270, opened 2023-05-04
Downgrade packaging.

oliverchang:
https://github.com/pypa/packaging/issues/321 removed support for parsing legacy PyPI versions, which we needed to make our version enumeration/sorting work.

This upgrade happened in
https://github.com/google/osv.dev/commit/f96826094d1c2d766d4949c70cf2b9aae7ff45f3, but somehow we didn't notice this breakage.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -12,7 +12,7 @@ pygit2 = ">=1.9.2"
 attrs = "*"
 jsonschema = "*"
 grpcio = "*"
-packaging = "*"
+packaging = "<22.0"
 
 [dev-packages]
 yapf = "*"
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "685817003af27239474ea792fa8fba91379e3351444ab9bbe80415b32dad5201"
+            "sha256": "9a424a40988c0f77322d52b910443cddcc8b498d42664fbbee0abed0e759145a"
         },
         "pipfile-spec": 6,
         "requires": {},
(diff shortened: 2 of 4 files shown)


oliverchang:
Also CC @di if there are better workarounds for us here other than downgrading?

oliverchang:
I noticed this from https://github.com/pypa/advisory-database/actions/runs/4878347479/jobs/8703902390

Looking at our GCP logs, I don't see any instances of "packaging.version.InvalidVersion" in the past 30 days. It's possible we've just gotten lucky with not running into this in prod.

andrewpollock:
> but somehow we didn't notice this breakage

So how did we eventually notice the breakage and what needs to be done differently so we don't miss it? (i.e. what's the mini-postmortem learning here?)

andrewpollock (approved):

oliverchang:
> > but somehow we didn't notice this breakage
> 
> So how did we eventually notice the breakage and what needs to be done differently so we don't miss it? (i.e. what's the mini-postmortem learning here?)

I noticed separately via https://github.com/pypa/advisory-database/actions/runs/4878347479/jobs/8703902390

The error was on our error dashboard: https://pantheon.corp.google.com/errors/detail/CNGT29fs5Oj_DA;service=gke_instances?project=oss-vdb but we did not notice it :(

di:
> Also CC @di if there are better workarounds for us here other than downgrading?

You can use https://pypi.org/p/packaging-legacy as a drop-in replacement.

oliverchang:
> > Also CC @di if there are better workarounds for us here other than downgrading?
> 
> You can use https://pypi.org/p/packaging-legacy as a drop-in replacement.

Ah awesome, thanks Dustin!

Issue #1273, opened 2023-05-04
Reduce error spam with OSS-Fuzz bisection code


oliverchang:

oliverchang:
Once https://github.com/google/oss-fuzz/pull/10251 is merged we can just add to this:

https://github.com/google/osv.dev/blob/d14f916b46768749da4d92d76d322c91e81dd23b/osv/logs.py#L61

To set the log level to CRITICAL.

PR #1277, opened 2023-05-05
Suppress OSS-Fuzz helper logs.

oliverchang:
Part of #1273.


andrewpollock (approved):

another-rex (approved):

osv/logs.py
```diff
@@ -60,3 +60,7 @@ def record_factory(*args, **kwargs):
   logging.getLogger('google.api_core.bidi').setLevel(logging.ERROR)
   logging.getLogger('google.cloud.pubsub_v1.subscriber._protocol.'
                     'streaming_pull_manager').setLevel(logging.ERROR)
+
+  # Suppress OSS-Fuzz build error logs. These are expected as part of
+  # bisection.
+  logging.getLogger('helper').setLevel(logging.CRITICAL)
```
michaelkedar, line 66: Is `helper` the right path for this?

It might be better putting this here were the import path changes are happening.
https://github.com/google/osv.dev/blob/d14f916b46768749da4d92d76d322c91e81dd23b/docker/worker/worker.py#L653

oliverchang: Yes, "helper" should be the correct name for this, unless I'm misunderstanding how these logger names work. Within `helper.py`, this is set to `__name__`, which should be "helper" when we import this the way we are doing now. 

And you're right, worker.py would be the better logical place for this -- moved it. That said, I think we still have the problem you described, unless we make oss-fuzz into a proper python package. I'll file a bug.

PR #1278, opened 2023-05-05
Don't prematurely abort processing the entire AffectedVersions

andrewpollock:
Previous behaviour was resulting in premature skipping of potentially the whole triad of AffectedVersions fields if the first or second failed to resolve to a commit successfully, which is missing some potential candidates for partial record generation.

This seems to contribute an extra 2% of successful conversions.


another-rex (approved):
LGTM

PR #1286, opened 2023-05-15
Update version to return non-parsed tag

another-rex:
Update version that's returned in determine_version API to return the actual tag rather than the parsed version.


gcp/api/server.py
```diff
@@ -188,24 +191,24 @@ def build_determine_version_result(
...
+        score=(max_files - estimated_diff_files) / max_files,
         minimum_file_matches=file_matches_by_proj[idx.key],
-        estimated_diff_files=estimated_num_diff,
+        estimated_diff_files=estimated_diff_files,
         repo_info=osv_service_v1_pb2.VersionRepositoryInformation(
             type=osv_service_v1_pb2.VersionRepositoryInformation.GIT,
             address=idx.repo_addr,
             commit=idx.commit,
-            version=idx.version,
+            version=idx.tag.removeprefix(_TAG_PREFIX),
```
michaelkedar, line 211: Is it confusing to have this field called `version` when it refers to a git tag? Is `RepoIndex.version` used elsewhere?

another-rex: You're right it could be confusing, the main reason to keep it version here is to have the API not change for this update, though since it's still in experimental we probably should change it now rather than later.

another-rex: Changed to tag now, turns out there's already a tag entry in the protobuf.

michaelkedar (approved):

oliverchang (approved):

oliverchang (approved):

PR #1290, opened 2023-05-16
Add typing, add timeout, add kernel quick fail

another-rex:
Added the quick fail for Kernel queries, and tried matching the ndb query timeout with the GRPC time_remaining function. Added a log to check if that function does what I think it does and matches what we set in cloud run. 

Also added a bunch of type hints just to make development more efficient.

### DIFF
--- a/gcp/api/server.py
+++ b/gcp/api/server.py
@@ -91,7 +91,7 @@ class OSVServicer(osv_service_v1_pb2_grpc.OSVServicer):
 
     version.
     """
-    logging.info("Request time remaining: " + context.time_remaining())
+    logging.info(f"Request time remaining: {context.time_remaining()}")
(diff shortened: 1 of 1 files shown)


andrewpollock (approved):

gcp/api/server.py
```diff
@@ -86,11 +86,12 @@ def GetVulnById(self, request, context):
     return bug_to_response(bug)
 
   @ndb_context
-  def QueryAffected(self, request, context):
+  def QueryAffected(self, request, context: grpc.ServicerContext):
     """Query vulnerabilities for a particular project at a given commit or
 
     version.
     """
+    logging.info("Request time remaining: " + context.time_remaining())
```
andrewpollock, line 94: This has angered the linter

another-rex: Fixed

PR #1291, opened 2023-05-16
Make api backend timeout match frontend, reduce container concurrency

michaelkedar:
- Set `timeoutSeconds` to 60, instead of the default 300, for the backend api to match what is on the frontend
  - Also set the staging instance's frontend timeout to 60 to match prod
- Reduce the backend's `containerConcurrency` to 20 (from default 80). This is 2x the number of threads in the thread pool for the server. May reduce further in future.

### DIFF
--- a/deployment/clouddeploy/osv-api/run.yaml
+++ b/deployment/clouddeploy/osv-api/run.yaml
@@ -10,3 +10,5 @@ spec:
         resources:
           limits:
             memory: 4Gi
+      timeoutSeconds: 60
+      containerConcurrency: 20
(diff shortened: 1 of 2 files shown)


another-rex (commented):
Can you clarify why the production and staging deadline/timeout configs are in different types of files?

michaelkedar:
> Can you clarify why the production and staging deadline/timeout configs are in different types of files?

Both staging and production api backends use the `run.yaml` file.

The `api_config.tftpl` (Tensorflow template) file is for the api endpoints frontend. Production already has this set to 60:
https://github.com/google/osv.dev/blob/06bc562848d0f012ef262f8eb3773fb87bf64e0b/deployment/terraform/environments/oss-vdb/api/api_config.tftpl#L29-L33
It's managed by terraform mostly because it's not (yet?) supported by Cloud Deploy.

another-rex (approved):
Makes sense, thanks!

PR #1302, opened 2023-05-17
Only log the clone failure on the final attempt

andrewpollock:
Reasoning:

* it gets retried, so a first (or second) failure + successful reattempt is just noise

Reduces non-actionable instances of the likes of:

```
Clone failed: Command '['git', 'clone', 'https://github.com/github/advisory-database.git', '/work/sources/ghsa']' returned non-zero exit status 128.
```

in the error logs.

### DIFF
--- a/osv/repos.py
+++ b/osv/repos.py
@@ -115,7 +115,7 @@ def clone_with_retries(git_url, checkout_dir, git_callbacks=None, branch=None):
         _checkout_branch(repo, branch)
       return repo
     except (pygit2.GitError, subprocess.CalledProcessError) as e:
-      if attempt == range(CLONE_TRIES)[-1]:
+      if attempt == CLONE_TRIES - 1:
         logging.error('Clone failed on attempt %d of %d: %s', attempt + 1,
                       CLONE_TRIES, str(e))
       shutil.rmtree(checkout_dir, ignore_errors=True)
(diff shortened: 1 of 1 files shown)


osv/repos.py
```diff
@@ -107,15 +107,17 @@ def clone(git_url, checkout_dir, git_callbacks=None):
...
+  for attempt in range(CLONE_TRIES):
     try:
       repo = clone(git_url, checkout_dir, git_callbacks)
       repo.cache = {}
       if branch:
         _checkout_branch(repo, branch)
       return repo
     except (pygit2.GitError, subprocess.CalledProcessError) as e:
-      logging.error('Clone failed: %s', str(e))
+      if attempt == range(CLONE_TRIES)[-1]:
```
michaelkedar, line 118:
```suggestion
      if attempt == CLONE_TRIES - 1:
```

osv/repos.py
```diff
@@ -107,15 +107,17 @@ def clone(git_url, checkout_dir, git_callbacks=None):
...
+        logging.error('Clone failed on attempt %d of %d: %s', attempt + 1,
+                      CLONE_TRIES, str(e))
```
michaelkedar, line 120: `attempt + 1` will always equal `CLONE_TRIES` here.
```suggestion
        logging.error('Clone failed after %d attempts: %s', CLONE_TRIES,
                      str(e))
```
Maybe?

michaelkedar (approved):

another-rex (approved):

PR #1303, opened 2023-05-17
Fix timeout argument

another-rex:
Remove the deprecated default option and pass it in directly.

By filling out default options, every query now prints a warning log because of this line in the implementation code:

        if default_options is not None:
            _log.warning(
                "Deprecation warning: passing default_options to the Query"
                "constructor is deprecated. Please directly pass any "
                "arguments you want to use to the Query constructor or its "
                "methods."
            )
https://github.com/googleapis/python-ndb/blob/a3a181a427cc292882691d963b30bc78c05c6592/google/cloud/ndb/query.py#L1339

Though looking at the git blame has been "deprecated" since 4 years ago in version v0.0.1

### DIFF
--- a/gcp/api/server.py
+++ b/gcp/api/server.py
@@ -512,8 +512,9 @@ def _is_version_affected(affected_packages,
 
 
 @ndb.tasklet
-def _query_by_semver(query: ndb.Query, package_name, ecosystem,
-                     purl: PackageURL, version):
+def _query_by_semver(context: grpc.ServicerContext, query: ndb.Query,
+                     package_name: str, ecosystem: str, purl: PackageURL,
+                     version: str):
   """Query by semver."""
   if not semver_index.is_valid(version):
     return []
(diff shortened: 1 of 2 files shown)


michaelkedar (approved):

andrewpollock:
> Remove the deprecated default option and pass it in directly.

I'm not seeing anything at https://googleapis.dev/python/python-ndb/latest/query.html#google.cloud.ndb.query.Query that says the `default_options` attribute of the `Query` object is deprecated. Could you add some more context on why this change is necessary to the PR description?

andrewpollock (approved):
Change looks structurally sound, although I'm not understanding the necessity?

another-rex:
By filling out default options, every query now prints a warning log because of this line in the implementation code:
```
        if default_options is not None:
            _log.warning(
                "Deprecation warning: passing default_options to the Query"
                "constructor is deprecated. Please directly pass any "
                "arguments you want to use to the Query constructor or its "
                "methods."
            )
```
https://github.com/googleapis/python-ndb/blob/a3a181a427cc292882691d963b30bc78c05c6592/google/cloud/ndb/query.py#L1339

Though looking at the git blame has been "deprecated" the case since 4 years ago in version v0.0.1

PR #1305, opened 2023-05-18
Fix logging

another-rex:
Currently all logs default to the default severity since it's being gathered from stderr. This routes the logs properly by using the google cloud logging library.


gcp/api/server.py
```diff
@@ -690,9 +691,19 @@ def serve(port: int, local: bool):
     server.stop(_SHUTDOWN_GRACE_DURATION)
 
 
+def is_cloud_run() -> bool:
+  """Check if we are running in cloud run"""
```
andrewpollock, line 695:
```suggestion
  """Check if we are running in Cloud Run."""
```
Pedantic nit.

andrewpollock (approved):
Thanks for this. Based on https://cloud.google.com/profiler/docs/about-profiler we should rewrite the API in Go...

gcp/api/server.py
```diff
@@ -690,9 +691,19 @@ def serve(port: int, local: bool):
...
+  # https://cloud.google.com/run/docs/container-contract#env-vars
+  return os.getenv('K_SERVICE') is not None
+
+
 def main():
   """Entrypoint."""
-  logging.basicConfig(stream=sys.stderr)
+  if is_cloud_run():
+    logging_client = google.cloud.logging.Client()
+    logging_client.setup_logging()
```
michaelkedar, line 704: We could use https://github.com/google/osv.dev/blob/master/osv/logs.py which also surfaces error logs (without stack traces) to the error reporting dashboard.

another-rex: Good point! Imported that function to use instead.

andrewpollock (approved):

andrewpollock (approved):
Adjusted PR title to match new reality

gcp/api/run_tests.sh
```diff
@@ -18,12 +18,13 @@ if [ $# -lt 1 ]; then
...
 
-pipenv requirements > requirements.txt
-virtualenv ENV
-. ENV/bin/activate
-pip install -r requirements.txt
+# pipenv requirements > requirements.txt
+# virtualenv ENV
+# . ENV/bin/activate
+# pip install -r requirements.txt
+python3 -m pipenv sync
```
michaelkedar, line 25: nit: Want to delete the commented lines?
Any idea why it used virtualenv instead of pipenv in the first place?

michaelkedar (approved):

gcp/api/run_tests.sh
another-rex: Not sure why, I asked Oliver a few weeks ago and he can't remember exactly why either, I'll remove it.

PR #1308, opened 2023-05-23
Fix IN_SCOPE_YEARS in nvd-cve-osv

michaelkedar:
Looks like env vars need to be strings in kubernetes manifests.
Used parentheses instead of quotes since `IN_SCOPE_YEARS` is meant to be an array.


andrewpollock (approved):
Thank you!

Are you sure this is going to translate the right way? I wasn't sure if I should be double-quoting space-separated strings...

michaelkedar:
> Are you sure this is going to translate the right way? I wasn't sure if I should be double-quoting space-separated strings...

I'm about 85% sure :slightly_smiling_face:
`()` aren't special characters in yaml, and yaml treats each value as a string if it can't be parsed as a numeric type or list

PR #1309, opened 2023-05-23
Rework version extraction to handle CPE updates when present

andrewpollock:
While investigating some incomplete record generation, I found:

- when `ExtractVersionInfo` is able to get an `AffectedCommit` from the CVE's reference, this isn't accounted for (added an explicit test case here, and fixed in #1317)
- when the CPE string has an "update" part to it, this should be reflected in the generated `last_affected` value

Added additional test cases for both of these and also fixed test failure output (it was ordered the wrong way around)


andrewpollock:
> I don't see any code changes that include FixCommits.

Yeah, there's no code changes here, they ended up downstream in #1317

There's just an explicit test case for this behaviour added in this PR. I'll reword the PR description a bit for clarity...

another-rex (approved):

oliverchang (approved):

PR #1327, opened 2023-05-26
Add a convenience function for retrieving Cloud Secrets

andrewpollock:
GitHub API keys will be stored in Cloud Secrets, and a convenient way to retrieve them will be required


oliverchang (commented):
A potentially simpler alternative is to just pass this as an argument to the entrypoint program and propagate this as needed. 

We can then manage the secret through Kubernetes secrets, which just adds it as an argument to the container command. 

(This was how we managed the deps.dev API key previously).

andrewpollock:
> A potentially simpler alternative is to just pass this as an argument to the entrypoint program and propagate this as needed.
> 
> We can then manage the secret through Kubernetes secrets, which just adds it as an argument to the container command.

That complicates testing/execution outside of GKE though...

oliverchang:
> > A potentially simpler alternative is to just pass this as an argument to the entrypoint program and propagate this as needed.
> > We can then manage the secret through Kubernetes secrets, which just adds it as an argument to the container command.
> 
> That complicates testing/execution outside of GKE though...

We'd just be passing an additional argument to the entrypoint/CLI to test. Relying on Cloud Secrets has the additional burden of requiring your local gcloud creds to be refreshed etc..

andrewpollock:
> We'd just be passing an additional argument to the entrypoint/CLI to test. Relying on Cloud Secrets has the additional burden of requiring your local gcloud creds to be refreshed etc..

That's already a requirement for logging to work correctly.

oliverchang:
> > We'd just be passing an additional argument to the entrypoint/CLI to test. Relying on Cloud Secrets has the additional burden of requiring your local gcloud creds to be refreshed etc..
> 
> That's already a requirement for logging to work correctly.

Hmm, that shouldn't be the case if we're just testing locally :) logging can just go into stdout/stderr. 

Having this Cloud Secrets dependency still complicates things though. It makes development harder for anyone in the future who doesn't have access to the right secret, because now they need a GCP project with the right secret set up, rather than a string they can just provide at runtime. 

It also fragments the way we generally handle secrets: we are currently using GKE secrets for everything running there.

andrewpollock:
> Hmm, that shouldn't be the case if we're just testing locally :) logging can just go into stdout/stderr.

That's not how the logging wrapper currently behaves. It's Cloud Logging or bust, see https://github.com/google/osv.dev/blob/32cca687bb2ead7bcc7587388eb95e9c9a47a228/vulnfeeds/utility/logginghelpers.go#L15

> Having this Cloud Secrets dependency still complicates things though. It makes development harder for anyone in the future who doesn't have access to the right secret, because now they need a GCP project with the right secret set up, rather than a string they can just provide at runtime.

The intention is to have a flag to specify the Cloud Secret name and version, so yes, there will be a dependency on *a* Cloud Project with *an* appropriate secret in it, just as there's already a dependency for *a* Cloud Project to act as a log sink.

It also means that local testing requires a lot more direct handling of the secret itself, rather than the identity of it, which seems like an ongoing complexity for the humans doing it.

I'm not keen to be encouraging the passing of secrets on the command line in *any* environment, because making assumptions about the safety of the environment isn't a particularly good security practice. 

Yes, if the secret is visible on the commandline/environment in GKE, there shouldn't be any way for it to leak (unless it got logged), but we can't say the same for any local runtime environments during testing.

> It also fragments the way we generally handle secrets: we are currently using GKE secrets for everything running there.

We've had to store the deps.dev key in Cloud Secrets as well, for reasons I can't immediately recall.

oliverchang:
> > Hmm, that shouldn't be the case if we're just testing locally :) logging can just go into stdout/stderr.
> 
> That's not how the logging wrapper currently behaves. It's Cloud Logging or bust, see
> 
> https://github.com/google/osv.dev/blob/32cca687bb2ead7bcc7587388eb95e9c9a47a228/vulnfeeds/utility/logginghelpers.go#L15
> 

Yes, I'm suggesting that this should not be the case and we should support easier local development (with a stdout/stderr fallback) without this hard dependency :)

> > Having this Cloud Secrets dependency still complicates things though. It makes development harder for anyone in the future who doesn't have access to the right secret, because now they need a GCP project with the right secret set up, rather than a string they can just provide at runtime.
> 
> The intention is to have a flag to specify the Cloud Secret name and version, so yes, there will be a dependency on _a_ Cloud Project with _an_ appropriate secret in it, just as there's already a dependency for _a_ Cloud Project to act as a log sink.
> 
> It also means that local testing requires a lot more direct handling of the secret itself, rather than the identity of it, which seems like an ongoing complexity for the humans doing it.
> 
> I'm not keen to be encouraging the passing of secrets on the command line in _any_ environment, because making assumptions about the safety of the environment isn't a particularly good security practice.

If the concern is leaking via CLI, we could have this be a file or env var instead. Local gcloud credentials can similarly be leaked... if we assume the filesystem is compromised. 

> 
> Yes, if the secret is visible on the commandline/environment in GKE, there shouldn't be any way for it to leak (unless it got logged), but we can't say the same for any local runtime environments during testing.
> 
> > It also fragments the way we generally handle secrets: we are currently using GKE secrets for everything running there.
> 
> We've had to store the deps.dev key in Cloud Secrets as well, for reasons I can't immediately recall.

The deps.dev key was handled by GKE. It was removed since it's no longer needed. Our SSH secrets for our worker are also still handled by GKE: https://github.com/google/osv.dev/blob/32cca687bb2ead7bcc7587388eb95e9c9a47a228/deployment/clouddeploy/gke-workers/environments/oss-vdb/workers.yaml#L14

andrewpollock:
> Yes, I'm suggesting that this should not be the case and we should support easier local development (with a stdout/stderr fallback) without this hard dependency :)

We're digressing here, but that's not a workable solution for me. The volume of log output generated for a local run makes trying to do any sort of examination outside of Cloud Logging not a particularly tenable solution. I for one want Cloud Logging under all circumstances, regardless of where execution is occurring.

I had thought we had a workable middle ground previously where if a GCP project was supplied, it was used as a log sink, and if it wasn't the wrapper "failed open" to stdout, but I believe that all changed in #1241 @another-rex ?

another-rex:
I think we used to have hard coded GCP IDs, and changed it so that it's from the environment instead. 

The google cloud logger does have an option to redirect the output to stdout as JSON, which we can use when no GOOGLE_CLOUD_PROJECT is specified.

oliverchang (approved):
I'll approve this to unblock, but I do still think we should move away from too many GCP dependencies here.

I've been burned in the past on other projects where requiring GCP hindered external contributions and also made things much harder to test.

PR #1329, opened 2023-05-28
Remove apt-key usage in dockerfiles

### LINKED ISSUE #1074: Future-proof against deprecation of apt-key
We have [5](https://github.com/search?q=repo%3Agoogle%2Fosv.dev%20apt-key&type=code) Docker images that invoke `apt-key` as part of their builds, and this utility is deprecated and will go away entirely in a future release of Ubuntu.

The replacement is to directly manage key files in `/etc/apt/trusted.gpg.d/`

The change is pretty simple: just output the key file acquired via curl/wget et al directly into `/etc/apt/trusted.gpg.d/` and stop calling `apt-key`

hirotasoshu:
Resolves #1074


andrewpollock (approved):
Thank you so much for your contribution!

PR #1331, opened 2023-05-29
Make tooling use python 3.10

michaelkedar:
Use python 3.10 in lint, dependabot, and renovate-bot so updates in pipenv doesn't break our lockfiles for local development (using 3.10). Should hopefully fix the issue with #1311 breaking our lint action. We will still use python 3.11.0 in the worker docker image.

We really should be using 3.11 everywhere, but this is a short term fix to prevent dependabot breaking our dev environments.


michaelkedar:
Going to migrate everything to 3.11 instead.

PR #1336, opened 2023-05-30
Fallback log to stdout for vulnfeeds

another-rex:
When GOOGLE_CLOUD_PROJECT env variable is not passed in, we fall back to stdout logging rather than exiting.

### DIFF
--- a/vulnfeeds/utility/logginghelpers.go
+++ b/vulnfeeds/utility/logginghelpers.go
@@ -14,16 +14,19 @@ import (
 // and also returns a cleanup function to be deferred
 func CreateLoggerWrapper(name string) (LoggerWrapper, func()) {
 	projectId, ok := os.LookupEnv("GOOGLE_CLOUD_PROJECT")
+	var loggerOptions []logging.LoggerOption
 	if !ok {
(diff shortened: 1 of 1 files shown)


oliverchang (approved):

vulnfeeds/utility/logginghelpers.go
```diff
@@ -14,16 +14,19 @@ import (
 // and also returns a cleanup function to be deferred
 func CreateLoggerWrapper(name string) (LoggerWrapper, func()) {
 	projectId, ok := os.LookupEnv("GOOGLE_CLOUD_PROJECT")
+	var loggerOptions []logging.LoggerOption
 	if !ok {
-		log.Fatalf("GOOGLE_CLOUD_PROJECT not set")
+		log.Println("GOOGLE_CLOUD_PROJECT not set, routing logs to stdout")
+		loggerOptions = append(loggerOptions, logging.RedirectAsJSON(os.Stdout))
```
andrewpollock, line 20: I think this will be a line of JSON and not necessarily very locally consumer friendly.

Looking at https://pkg.go.dev/cloud.google.com/go/logging#hdr-The_Standard_Logger I'm wondering if it could return a standard https://pkg.go.dev/log#Logger instead?

PR #1337, opened 2023-05-30
Better logging redirection

another-rex:
LoggingHelper already had the functionality to redirect to stdout (when cloud logger is nil), so undo the JSON redirect and set the gcloud logger to nil instead.

### DIFF
--- a/vulnfeeds/utility/logginghelpers.go
+++ b/vulnfeeds/utility/logginghelpers.go
@@ -13,23 +13,21 @@ import (
 // CreateLoggerWrapper creates and initializes the LoggerWrapper,
 // and also returns a cleanup function to be deferred
 func CreateLoggerWrapper(name string) (LoggerWrapper, func()) {
-	projectId, ok := os.LookupEnv("GOOGLE_CLOUD_PROJECT")
-	var loggerOptions []logging.LoggerOption
-	if !ok {
-		log.Println("GOOGLE_CLOUD_PROJECT not set, routing logs to stdout")
-		loggerOptions = append(loggerOptions, logging.RedirectAsJSON(os.Stdout))
+	projectId, projectIdSet := os.LookupEnv("GOOGLE_CLOUD_PROJECT")
+	if !projectIdSet {
+		return LoggerWrapper{}, func() {}
 	}
-	client, err := logging.NewClient(context.Background(), projectId)
 
+	log.Println("Logging to project id: " + projectId)
+	client, err := logging.NewClient(context.Background(), projectId)
 	if err != nil {
 		log.Fatalf("Failed to create client: %v", err)
 	}
-
 	wrapper := LoggerWrapper{
-		GCloudLogger: client.Logger(name, loggerOptions...),
+		GCloudLogger: client.Logger(name),
 	}
-
 	return wrapper, func() { client.Close() }
+
 }
 
 // LoggerWrapper wraps the Logger provided by google cloud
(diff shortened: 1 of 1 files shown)


oliverchang (approved):

vulnfeeds/utility/logginghelpers.go
```diff
@@ -13,23 +13,21 @@ import (
 // CreateLoggerWrapper creates and initializes the LoggerWrapper,
 // and also returns a cleanup function to be deferred
 func CreateLoggerWrapper(name string) (LoggerWrapper, func()) {
```
andrewpollock, line 15: Nit: Be consistent with https://pkg.go.dev/cloud.google.com/go/logging#Client.Logger given this is being passed to that.

```suggestion
func CreateLoggerWrapper(logID string) (LoggerWrapper, func()) {
```

vulnfeeds/utility/logginghelpers.go
```diff
@@ -13,23 +13,21 @@ import (
...
 		log.Fatalf("Failed to create client: %v", err)
 	}
-
 	wrapper := LoggerWrapper{
-		GCloudLogger: client.Logger(name, loggerOptions...),
+		GCloudLogger: client.Logger(name),
 	}
-
 	return wrapper, func() { client.Close() }
+
```
another-rex, line 30:
```suggestion
```

vulnfeeds/utility/logginghelpers.go
```diff
@@ -13,23 +13,21 @@ import (
...
 // and also returns a cleanup function to be deferred
 func CreateLoggerWrapper(name string) (LoggerWrapper, func()) {
-	projectId, ok := os.LookupEnv("GOOGLE_CLOUD_PROJECT")
-	var loggerOptions []logging.LoggerOption
-	if !ok {
-		log.Println("GOOGLE_CLOUD_PROJECT not set, routing logs to stdout")
-		loggerOptions = append(loggerOptions, logging.RedirectAsJSON(os.Stdout))
+	projectId, projectIdSet := os.LookupEnv("GOOGLE_CLOUD_PROJECT")
+	if !projectIdSet {
+		return LoggerWrapper{}, func() {}
```
andrewpollock, line 18: I'm having a hard time understanding how this magically falls back to using the standard golang logging here?

vulnfeeds/utility/logginghelpers.go
```diff
@@ -13,23 +13,21 @@ import (
...
 
+	log.Println("Logging to project id: " + projectId)
+	client, err := logging.NewClient(context.Background(), projectId)
 	if err != nil {
 		log.Fatalf("Failed to create client: %v", err)
 	}
-
 	wrapper := LoggerWrapper{
-		GCloudLogger: client.Logger(name, loggerOptions...),
+		GCloudLogger: client.Logger(name),
```
andrewpollock, line 27: Per nit above.

```suggestion
		GCloudLogger: client.Logger(logID),
```

andrewpollock (commented):
I'm not quite understanding how this works :-(

vulnfeeds/utility/logginghelpers.go
andrewpollock: Oh nevermind, I looked at the wrapper functions and I see what they're doing.

andrewpollock (approved):

andrewpollock:
My only usability nit is I'd prefer to use a flag to set the project ID/logging behaviour rather than an environment variable, and there's no way to pass that into the wrapper.

PR #1338, opened 2023-05-30
Update Environment to Python 3.11.1

michaelkedar:
- Specify specific Python versions in Pipfiles & relock them (with minimal changes)
- Delete unused exporter Pipfile
- Update lint action to use Python 3.11.1

dependabot and renovate-bot should automatically use the correct python versions from the Pipfiles, so hopefully their dependency upgrades will be more reliable.

We will need to install Python 3.11 and manually re-create our local pipenv environments with the correct version:
```
pipenv --rm
pipenv sync --dev --python=python3.11
```


gcp/appengine/Pipfile.lock
```diff
@@ -14,6 +16,14 @@
...
     },
     "default": {
+        "async-timeout": {
+            "hashes": [
+                "sha256:2163e1640ddb52b7a8c80d0a67a08587e5d245cc9c553a74a847056bc2976b15",
+                "sha256:8ca1e4fcf50d07413d66d1a5e416e42cfdf5851c981d679a09851a6853383b3c"
+            ],
+            "markers": "python_full_version <= '3.11.2'",
+            "version": "==4.0.2"
+        },
```
michaelkedar, line 26: I don't really know what's going on here. Apparently `async-timeout` isn't needed in Python 3.11.3, or our App Engine app never actually needs to import it.

michaelkedar:
I haven't touched https://github.com/google/osv.dev/tree/bd7f1b5b1db67bd51cb13248f952b3e2a3f3e48d/gcp/functions/pypi

It doesn't get deployed automatically (I don't know when we last deployed it) and it's not on the staging environment.

It also uses the Python 3.9 runtime in the deploy script.

andrewpollock:
> * Delete unused exporter Pipfile

Why *is* it unused? What makes it unused but not the docker/worker one?

andrewpollock (approved):
LGTM to the best of my understanding of things.

michaelkedar:
> > * Delete unused exporter Pipfile
> 
> Why _is_ it unused? What makes it unused but not the docker/worker one?

[It's never actually used to install requirements when creating the exporter image](https://github.com/google/osv.dev/blob/5824aa6d8089e4134ddfc9b035da8634fdb032bf/docker/exporter/Dockerfile). It just uses the worker image which installs from the worker's pipenv: https://github.com/google/osv.dev/blob/5824aa6d8089e4134ddfc9b035da8634fdb032bf/docker/worker/Dockerfile#L27

The importer does the same as the exporter, and does not have its own Pipfile.

another-rex (approved):
LGTM!

PR #1339, opened 2023-05-30
Update `worker-base` image to Python 3.11.1

michaelkedar:
Need to do this before #1338 (which enforces Python 3.11.1) because the PR test runner image `gcr.io/oss-vdb/ci` uses the worker image.

`gcr.io/oss-vdb/ci` should probably be automatically rebuilt for PRs if the worker image changes...


michaelkedar:
Will merge this after this week's release in case something breaks

oliverchang (approved):

PR #1346, opened 2023-05-31
Fix indexer occasional error, change commit response to hex

another-rex:
Make the indexer API resilient to cases where buckets are created but RepoIndex entry is not (this might happen when something fails when writing to the datastore), these errors should only be temporary and be resolved on the next indexer run. 

Added a warning message there so we can observe how often this occurs. 

Also changed commit in the response from base64 to hex.


gcp/api/server.py
```diff
@@ -178,11 +178,17 @@ def build_determine_version_result(
...
 
+    if idx is None:
+      logging.warn(
+          'Bucket exists for project: %s, which does ' +
+          'not have a matching IndexRepo entry', bucket_match_items[i][0])
+      continue
+
     if idx.empty_bucket_bitmap is None:
-      logging.error('No empty bucket bitmap for: %s@%s', idx.name, idx.version)
+      logging.warn('No empty bucket bitmap for: %s@%s', idx.name, idx.version)
```
another-rex, line 191:
```suggestion
      logging.warning('No empty bucket bitmap for: %s@%s', idx.name, idx.version)
```

gcp/api/server.py
```diff
@@ -178,11 +178,17 @@ def build_determine_version_result(
   inverted_empty_bucket_bitmap = (1 << _BUCKET_SIZE) - 1 - empty_bucket_bitmap
   empty_bucket_count = inverted_empty_bucket_bitmap.bit_count()
 
-  for f in idx_futures:
+  for i, f in enumerate(idx_futures):
     idx: osv.RepoIndex = f.result()
 
+    if idx is None:
+      logging.warn(
```
another-rex, line 185:
```suggestion
      logging.warning(
```

gcp/api/server.py
```diff
@@ -178,11 +178,17 @@ def build_determine_version_result(
...
 
+    if idx is None:
+      logging.warning(
+          'Bucket exists for project: %s, which does ' +
+          'not have a matching IndexRepo entry', bucket_match_items[i][0])
+      continue
+
     if idx.empty_bucket_bitmap is None:
-      logging.error('No empty bucket bitmap for: %s@%s', idx.name, idx.version)
+      logging.warning('No empty bucket bitmap for: %s@%s', idx.name, idx.version)
```
another-rex, line 191:
```suggestion
      logging.warning('No empty bucket bitmap: %s@%s', idx.name, idx.version)
```

oliverchang (approved):

PR #1349, opened 2023-06-02
Use set version regex in indexer.

### LINKED ISSUE #1348: indexer: version matching regex not broad enough.
Noticed that we were missing some indexes for c-ares. and our indexer logs show:

```
failed to extract version for repo: https://github.com/c-ares/c-ares.git tag/branch: refs/tags/cares-1_5_0
```

We should should make the [indexer version extraction logic](https://github.com/google/osv.dev/blob/e756517994e3f250c5426e765a1287608b4e6174/docker/indexer/stages/preparation/preparation.go#L46) match our Python one here: https://github.com/google/osv.dev/blob/master/osv/bug.py

We should also just delete the config setting from most of our indexer yaml configs as well as just use this single generic one.

another-rex:
Remove version matching in Indexer and move logic into API. Closes #1348


docker/indexer/stages/preparation/preparation.go
```diff
@@ -177,16 +177,8 @@ func (s *Stage) processGit(ctx context.Context, repoCfg *config.RepoConfig) erro
...
-			if err != nil {
-				return err
-			}
-			version = versRE.FindString(commitTag)
-		}
-		if version == "" {
-			version = genericVersionRE.FindString(commitTag)
-		}
+		verComponents := genericVersionRE.FindAllString(commitTag, -1)
+		version = strings.Join(verComponents, "-")
```
oliverchang, line 181: Should we just use "." as the separator here since this version is what will be returned in our responses? "." probably makes a lot more sense to API users than "-".

docker/indexer/stages/preparation/preparation.go
```diff
@@ -43,7 +43,7 @@ import (
 const workers = 5
 
 var (
-	genericVersionRE = regexp.MustCompile(`(\d+\.\d+\.?\d?-?(rc|alpha|beta|preview)?\d?)`)
+	genericVersionRE = regexp.MustCompile(`(?i)(\d+|(?<![a-z])(?:rc|alpha|beta|preview)\d*)`)
```
oliverchang, line 46: Does ?<! work in Go's re syntax?

oliverchang (commented):
thanks Rex!

docker/indexer/stages/preparation/preparation.go
another-rex: It does not, just replaced it with an alternative, which should mostly have the same behavior. The main difference is:
```
1_5_0-6alpha4
```
will get `1-5-0-6-alpha4` in python, but `1-5-0-6-4` in golang.

Not sure how to resolve those without getting too complicated or using an alternative regex engine.

docker/indexer/stages/preparation/preparation_test.go
```diff
@@ -0,0 +1,66 @@
...
+				tag: "cares-1_5_0-alpha4",
+			},
+			want: "1.5.0.alpha4",
+		},
+		{
+			name: "",
+			args: args{
+				tag: "cares-1_5_0-6alpha4",
+			},
+			want: "1.5.0.6.4",
```
oliverchang, line 56: Should this be 1.5.0.6.alpha4 ?

another-rex: The python version will return `1.5.0.6.alpha4`, but I'm not sure how to replicate it in golang without lookbehinds, at least without changing behavior somewhere else, or complicating it by doing multiple passes with regexes.

oliverchang: We can probably just completely remove that negative lookbehind. I think that was just to capture a very rare edge case with things like "arc" getting interpreted as a version component. 

Also more broadly, we should just compute this at the API serving time rather than record it here. Then we can refine/fix this whenever without having to re-index. That can come later if it's complicated to do here.

oliverchang: (And doing this at the API serving layer also fixes the Python/Go discrepancy)

another-rex: Removed this entirely to favor parsing and filtering it at API serving time. PTAL

oliverchang (approved):
thanks!!

PR #1353, opened 2023-06-02
Expand coverage scope to all desired years, take 2

andrewpollock:
#1345 did not work as intended, and we can't set the environment variable to an array of values with YAML, so split the single string into an array at run time.

h/t @oliverchang


michaelkedar:
I don't think this is valid YAML.
Honestly, I'm not sure how you're supposed to get a bash array from a YAML string

andrewpollock:
> Honestly, I'm not sure how you're supposed to get a bash array from a YAML string

https://stackoverflow.com/questions/68528541/env-variable-array certainly suggests it's not possible :-(

oliverchang:
This is just an ENV var right? Sounds like you just need something like this in the bash script? https://stackoverflow.com/a/9294015

michaelkedar (approved):

PR #1354, opened 2023-06-02
API paging implementation

another-rex:
A draft implementation of API pagination. 

Removed the previous temporary mitigation around kernel and GSD queries. 

The new pagination system works by limiting the response with these 3 constants (open to changing these numbers, currently just a estimate):

```python
# Maximum number of responses to return before applying post exceeded limit
_MAX_VULN_RESP_THRESH = 3000
# Max responses after MAX_VULN_RESP_THRESH has been exceeded
_MAX_VULN_LISTED_POST_EXCEEDED = 5
# Max responses before MAX_VULN_RESP_THRESH been exceeded
_MAX_VULN_LISTED_PRE_EXCEEDED = 1000
```

Few things to do before merging:
- Assess how widely used the Version + Package name but no ecosystem query is used. This type of query is technically invalid according to our documentation, but was functional before. This is now disabled to simplify pagination.
  -  If this is widely used, we can reconsider removing this, figure out an implementation with pagination, and document official support for this
  - The implementation in this PR returns 1 page of responses for these queries, without including the page cursor so no way to retrieve the next page.
- Release OSV-Scanner and auto remediation versions that support pagination with the API
- Announce that we are doing this to API consumers (blog post?)

## Stats:

Running with 10 ish "queries of doom" from before in a querybatch will allow 6 queries to each return 500 successsful responses + a page cursor for the next query. The other 4 queries will return up to 5 successful responses + a page cursor for the next queries. Any other queries that has less than 5 responses will respond fully. This takes about 5 seconds when running locally.

Running with just under 1000   "queries of doom" (mix of tensorflow, kernel version, kernel commit) will respond in around 45 seconds, with ~8000 successful responses (994 * 5 + 500*6). However it uses up 13 GiBs of memory...

#


gcp/api/server.py
```diff
@@ -149,6 +171,48 @@ def Watch(self, request, context: grpc.ServicerContext):
...
+  if necessary
+  """
+  count: int
+
+  def change(self, amount):
+    # This is to prevent query `limit` parameter being smaller than
+    # the number that is checked later in the iter() loop for the last page
+    if amount < 0:
+      raise ValueError("change amount must be positive")
+    self.count += amount
```
michaelkedar, line 190: nit: maybe name this method something like `add` or `increase` instead

gcp/api/server.py
```diff
@@ -340,9 +404,9 @@ def do_query(query, context: grpc.ServicerContext, include_details=True):
     ecosystem = ''
     purl_str = ''
 
-  page_token = None
+  context.page_token = None
   if query.page_token:
-    page_token = ndb.Cursor(urlsafe=query.page_token)
+    context.page_token = ndb.Cursor(urlsafe=query.page_token)
```
michaelkedar, line 409: Does there need to be some validation here or around the queries to prevent an invalid page token causing problems?

gcp/api/server.py
```diff
@@ -149,6 +171,48 @@ def Watch(self, request, context: grpc.ServicerContext):
...
+    return _MAX_VULN_LISTED_PRE_EXCEEDED
+
+
+@dataclass
+class QueryContext:
+  service_context: grpc.ServicerContext
+  page_token: ndb.Cursor | None
+  request_start_time: datetime
+  # Use a list to copy by reference
+  total_responses: ResponsesCount
```
michaelkedar, line 213: nit:
```suggestion
  # Use a dataclass to copy by reference
  total_responses: ResponsesCount
```

michaelkedar (commented):
Looks good! A few things:

gcp/api/server.py
```diff
@@ -100,7 +107,13 @@ def QueryAffected(self, request, context: grpc.ServicerContext):
...
     version.
     """
-    results, next_page_token = do_query(request.query, context).result()
+    query_context = QueryContext(
+        service_context=context,
+        request_start_time=datetime.now(),
+        page_token=None,
+        total_responses=ResponsesCount(0))
+
+    results, next_page_token = do_query(request.query, query_context).result()
```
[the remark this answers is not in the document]
another-rex: Using yield here seems to give a "Failed to serialize response!" error from the esp server. Can't quite figure out why though.

gcp/api/server.py
another-rex: Not sure how I can validate it, as it's just a base64 encoded byte blob, I think it won't do anything if the bytes does not relate to the query.

gcp/api/integration_tests.py
```diff
@@ -479,8 +483,8 @@ def test_query_package(self):
         _api() + '/v1/query',
         data=json.dumps({
             'package': {
-                'ecosystem': 'Maven',
-                'name': 'org.apache.tomcat:tomcat',
+                'ecosystem': 'PyPI',
```
oliverchang, line 486: Are there any tests we could add that actually trigger the pagination? 

Also curious about the reason for these query changes too.

gcp/api/server.py
```diff
@@ -43,7 +47,12 @@
 _SHUTDOWN_GRACE_DURATION = 5
 
 _MAX_BATCH_QUERY = 1000
-_MAX_VULNERABILITIES_LISTED = 16
+# Maximum number of responses to return before applying post exceeded limit
+_MAX_VULN_RESP_THRESH = 3000
+# Max responses after MAX_VULN_RESP_THRESH has been exceeded
+_MAX_VULN_LISTED_POST_EXCEEDED = 5
+# Max responses before MAX_VULN_RESP_THRESH been exceeded
```
oliverchang, line 54: s/been/has been/

gcp/api/server.py
```diff
@@ -340,9 +421,14 @@ def do_query(query, context: grpc.ServicerContext, include_details=True):
     ecosystem = ''
     purl_str = ''
 
-  page_token = None
+  context.page_token = None
   if query.page_token:
-    page_token = ndb.Cursor(urlsafe=query.page_token)
+    try:
+      context.page_token = ndb.Cursor(urlsafe=query.page_token)
```
oliverchang, line 427: Can we do this earlier as part of constructing the QueryContext instead?

gcp/api/server.py
```diff
@@ -658,29 +762,54 @@ def query_by_version(context: grpc.ServicerContext,
...
-                                                 ecosystem, purl, version)))
 
-  return [to_response(bug) for bug in bugs]
+    # TODO: Remove after testing how many consumers are
+    # querying the API this way.
+    context.page_token = None
+    new_bugs, _ = yield _query_by_semver(context, query, package_name,
+                                         ecosystem, purl, version)
+    bugs.extend(new_bugs)
+    new_bugs, _ = (yield _query_by_generic_version(context, query, package_name,
```
oliverchang, line 797: Are the parentheses around the yield statement required here?

gcp/api/server.py
```diff
@@ -658,29 +762,54 @@ def query_by_version(context: grpc.ServicerContext,
   if ecosystem:
     query = query.filter(osv.Bug.ecosystem == ecosystem)
 
+  # Don't filter by ecosystem if it's a purl query
+  if purl and not ecosystem:
+    purl_ecosystem = purl_helpers.purl_to_ecosystem(purl.type)
+    if purl_ecosystem:
+      ecosystem = purl_ecosystem
+  else:
+    logging.warning('ecosystem specified in a purl query')
```
oliverchang, line 771: I'm not sure if this ever makes sense. We can probably just return an error here.

oliverchang (commented):
nice!! some minor comments.

gcp/api/server.py
michaelkedar, line 771: Shouldn't this be
```python
if purl:
  if not ecosystem:
    # ...
  else:
    logging.warning(...)
```

Otherwise this would be logged whenever purl is not set

gcp/api/server.py
another-rex: Not really, since the page token exists in each query within a batch_query, not just one page token for the entire batch_query

osv/purl_helpers.py
```diff
@@ -59,3 +59,9 @@ def package_to_purl(ecosystem, package_name):
     suffix = '?arch=source'
 
   return f'pkg:{purl_type}/{_url_encode(package_name)}{suffix}'
+
+
+def purl_to_ecosystem(purl_type: str) -> str | None:
+  """Convert purl to a specific ecosystem string"""
+  ecosystem_purl = {v: k for k, v in PURL_ECOSYSTEMS.items()}
+  return ecosystem_purl[purl_type]
```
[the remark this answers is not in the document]
another-rex: Good point, changed in latest commit

gcp/api/server.py
another-rex: Wow, good find! Updated it now to be correct.

gcp/api/integration_tests.py
another-rex: The previous paging implementation limited the pages for package queries to 16. Since this upped it to 500 the Maven query wasn't returning enough results to require showing another page.

michaelkedar (approved):

gcp/api/server.py
oliverchang: Discussed offline: looks like we can just initialise this earlier.

gcp/api/server.py
```diff
@@ -149,6 +187,48 @@ def Watch(self, request, context: grpc.ServicerContext):
...
+      return _MAX_VULN_LISTED_POST_EXCEEDED
+
+    return _MAX_VULN_LISTED_PRE_EXCEEDED
+
+
+@dataclass
+class QueryContext:
+  service_context: grpc.ServicerContext
+  page_token: ndb.Cursor | None
+  request_start_time: datetime
```
oliverchang, line 227: Is this used anywhere?

gcp/api/server.py
```diff
@@ -100,7 +108,20 @@ def QueryAffected(self, request, context: grpc.ServicerContext):
...
     """
-    results, next_page_token = do_query(request.query, context).result()
+    query_context = QueryContext(
+        service_context=context,
+        request_start_time=datetime.now(),
+        page_token=None,
+        total_responses=ResponsesCount(0))
+
+    try:
+      results, next_page_token = do_query(request.query, query_context).result()
```
oliverchang, line 118: Can we add a log as discussed to log the fact that we paginated?

oliverchang (approved):

gcp/api/server.py
oliverchang: (also need a similar log fot the batch case)

gcp/api/server.py
another-rex: Commented this out, might still be useful in the future to add a hard time limit to avoid continuing the query after the 60s timeout.

gcp/api/server.py
another-rex: Yep, done

oliverchang (approved):

PR #1359, opened 2023-06-06
Improve CPE version extraction

andrewpollock:
I'm not convinced this approach is sound, so I'm running some test conversions to see how it changes the results...

Problem: sometimes a CVE can have multiple unrelated CPEs on it, a
pathological example being [CVE-2022-27778](https://nvd.nist.gov/vuln/detail/CVE-2022-27778), which is for a curl
vulnerability, but also affects MySQL.

This changes the behaviour to effectively only focus on the first CPE in
the configurations of the CVE.

An alternative treatment of CVEs like [CVE-2022-27778](https://nvd.nist.gov/vuln/detail/CVE-2022-27778) would be to emit
multiple `affected[].ranges[]` entries.

This also refactors the `VendorProduct` struct into the `cves` package for centralised use.

### DIFF
--- a/vulnfeeds/cmd/cperepos/main.go
+++ b/vulnfeeds/cmd/cperepos/main.go
@@ -80,17 +80,17 @@ type CPE23Item struct {
 // product names, which cause undesired and incorrect repository attribution
 // when resolved via Debian copyright metadata.
 var DebianCopyrightDenylist = []cves.VendorProduct{
-	cves.VendorProduct{"apple", "pdfkit"},
-	cves.VendorProduct{"f-secure", "safe"},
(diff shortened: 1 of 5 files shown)


oliverchang (approved):
LGTM, but how does this change the existing conversions?

andrewpollock:
> LGTM, but how does this change the existing conversions?

For the record, I won't be proceeding with this approach, as based on some analysis of records excluded from the 2022 data set, it's now incorrectly omitting too many valid records that were previously not omitted. I'm going to pursue a different approach of fully walking the data structure and potentially generating per configuration node ranges.

PR #1360, opened 2023-06-06
Fix logging in tests

michaelkedar:
- Capture and check expected warning/error logs in tests
- Fix tests that were logging errors due to outdated test files

I've discovered that there were a few tests in the importer that didn't really check anything. I've explicitly skipped them for now.

Ideally, I think it'd be good if the tests started failing if they start logging something new. You can accomplish this by wrapping everthing with `with self.assertNoLogs(level='WARNING'):`, but it's annoying to add to every test.


andrewpollock (approved):
Great stuff!

PR #1369, opened 2023-06-08
Improve performance of affectedcommits 

another-rex:
- Improve performance of affected commits queries by making it keys_only.
- Adds logging to  package queries that don't have ecosystems.

### DIFF
--- a/gcp/api/server.py
+++ b/gcp/api/server.py
@@ -399,7 +399,7 @@ def _get_bugs(bug_ids, to_response=bug_to_response):
   """Get bugs from bug ids."""
   bugs = ndb.get_multi_async([ndb.Key(osv.Bug, bug_id) for bug_id in bug_ids])
 
-  responses = list()
+  responses = []
   for future_bug in bugs:
     bug: osv.Bug = future_bug.result()
     if bug and bug.status == osv.BugStatus.PROCESSED and bug.public:
--- a/gcp/api/server.py
+++ b/gcp/api/server.py
@@ -397,12 +397,15 @@ def bug_to_response(bug, include_details=True):
 
 def _get_bugs(bug_ids, to_response=bug_to_response):
   """Get bugs from bug ids."""
-  bugs = ndb.get_multi([ndb.Key(osv.Bug, bug_id) for bug_id in bug_ids])
-  return [
-      to_response(bug)
-      for bug in bugs
-      if bug and bug.status == osv.BugStatus.PROCESSED
-  ]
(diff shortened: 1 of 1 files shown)


oliverchang (approved):
LGTM

gcp/api/server.py
```diff
@@ -397,12 +397,15 @@ def bug_to_response(bug, include_details=True):
 
 def _get_bugs(bug_ids, to_response=bug_to_response):
   """Get bugs from bug ids."""
-  bugs = ndb.get_multi([ndb.Key(osv.Bug, bug_id) for bug_id in bug_ids])
-  return [
-      to_response(bug)
-      for bug in bugs
-      if bug and bug.status == osv.BugStatus.PROCESSED
-  ]
+  bugs = ndb.get_multi_async([ndb.Key(osv.Bug, bug_id) for bug_id in bug_ids])
```
michaelkedar, line 400: FYI: [This](https://cloud.google.com/appengine/docs/legacy/standard/python/ndb/queries#map) is the functional mapping I was thinking about yesterday (there's also an async version).

It only works on queries so some logic would have to move around, and you'd have to get each bug by id individually instead of using `get_multi_async`, so I'm not sure if it's really worth it. But if you were still looking into memory usage it might possibly maybe use less, though I'm not really convinced.

michaelkedar (approved):

gcp/api/server.py
```diff
@@ -429,25 +432,24 @@ def query_by_commit(context: grpc.ServicerContext,
...
   while (yield it.has_next_async()):
-    affected_commits = it.next()
-    # Avoid requiring a separate index to include this in the initial Datastore
-    # query. The number of these should be very little to just iterate through.
-    if not affected_commits.public:
-      continue
+    # Affect commits key follows this format:
+    # <BugID>-<PageNumber>
+    affected_commit_key: ndb.Key = it.next()
+    bug_id: str = affected_commit_key.id().rsplit("-", 1)[0]
```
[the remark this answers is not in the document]
another-rex: It's hard to get an estimate without just making the index,, but I'm guessing the index created will be huge because:
> Only indexed properties can be projected.
> This means that all properties used in a query (projected or filters) must exist in the same index. So, select tag from Task where priority = 1 requires a composite index on priority then tag.

Since we are filtering by commit_hash, we will have to hash that column, which will result in roughly another terabyte of indexes.

gcp/api/server.py
another-rex: Hmm it would require building a huge query of OR statements to make use of the map, not sure if that makes sense to do.
There is also this `done_callback` function which can be used for each of the futures, which might reduce memory usage by a bit more I can look into as well.

PR #1371, opened 2023-06-09
Fix Link

hayleycd:
I am working on integrating a link checker and it is bringing up some hits. This is the first one. 

Additionally, is this the best link for the call to action to submit a PR? I think so because it is the file that they would have to edit, but if something else makes more sense, I am happy to change where that link goes to. 

Edited to add preview: https://hayleycd.github.io/osv.dev/third-party/


docs/third-party.md
```diff
@@ -30,4 +30,4 @@ tools and unsupported by the core OSV maintainers.
     advisory databases](https://github.com/twu/skjold)
 -   [Trivy](https://github.com/aquasecurity/trivy)
 
-Feel free to [send a PR](https://github.com/google/osv.dev/docs/third-party.md) to add your project here.
\ No newline at end of file
+Feel free to [send a PR](https://github.com/google/osv.dev/blob/master/docs/third-party.md) to add your project here.
```
andrewpollock, line 33: This is acting as a self-link, which I don't think is the intention here?

andrewpollock (commented):
I note that the content here seems to be redundantly duplicated at https://github.com/google/osv.dev#third-party-tools-and-integrations

docs/third-party.md
andrewpollock: In teams of submitting a PR, using what the OSV Schema docs do as a precedent, I'd say link to https://github.com/google/osv.dev/compare

hayleycd: I added this so that the potential user would be linked to the file that they would need to change, so they didn't need to dig through our docs file structure. That seems helpful to me, but I was curious about a dev perspective on it. It looks funny when viewing the file on github, but I think it makes sense when going from the docs site to github. 

From my perspective, https://github.com/google/osv.dev/compare isn't particularly helpful, because for it requires you to have a branch with changes to compare. Where as from the file itself someone could just hit edit, commit to a new branch, and open a PR right on that page.

andrewpollock: > From my perspective, https://github.com/google/osv.dev/compare isn't particularly helpful, because for it requires you to have a branch with changes to compare. Where as from the file itself someone could just hit edit, commit to a new branch, and open a PR right on that page.

Ah, got it. Thanks for clarifying.

andrewpollock (approved):

PR #1375, opened 2023-06-12
Remove unavailable package

hayleycd:
The link checker caught another dead link. The previously linked go package was recently deprecated and is no longer available. 

If we wanted to, we could replace this link with a link to the [Go vuln database](https://go.dev/security/vuln/database) but I don't think that is the intention of this section. I am inclined to just delete the link.


andrewpollock (commented):
I wonder if this should be https://pkg.go.dev/github.com/julieqiu/vuln/osv instead?
@julieqiu do you know what the deal is here?

hayleycd:
> I wonder if this should be https://pkg.go.dev/github.com/julieqiu/vuln/osv instead? @julieqiu do you know what the deal is here?

I checked with Julie, and she told me that the package had been removed. She suggested linking to https://go.dev/security/vuln/database, but it isn't exactly the right fit here and is linked elsewhere.

andrewpollock (approved):

Issue #1380, opened 2023-06-13
Allow searching by the "related" field on the site


another-rex:
`search_indices` should also contain "related" field to allow for searching for related on the site.

Issue #888, opened 2022-12-05
Aliases, and how they are supposed to be used


nscuro:
Hey OSV team, thanks for your great work!

We're currently looking at how we can correlate vulnerabilities that describe the same thing.

As per specification, OSV has the [`aliases`](https://ossf.github.io/osv-schema/#aliases-field) field for this:

> The `aliases` field gives a list of **IDs of the same vulnerability** in other databases, in the form of the id field.

At least in my interpretation, aliasing is a bidirectional relationship that also applies transitively.  
If `X` aliases `Y` and `Z`, `Y` should also alias `X`, and `Y` should also alias `Z`. If they all describe the same thing, that should be a valid assumption.

However, in reality, we see that many vulnerability databases (ab-)use the OSV schema to publish *advisories*. In my understanding, a *vulnerability* would describe one defect, and that one defect only. Whereas an *advisory* can potentially refer to multiple vulnerabilities (as in "we patched all these vulnerabilities in version 1.2.3 of our package"). This appears to be a common thing for at least the Go, Rust, and (especially) Debian ecosystems in the OSV database. There are most likely more, but these have been the most obvious candidates to us.

For example, [GO-2022-0586](https://pkg.go.dev/vuln/GO-2022-0586) presumably aliases **four** CVEs and **four** GHSAs:
* [CVE-2022-26945](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-26945)
* [CVE-2022-30321](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-30321)
* [CVE-2022-30322](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-30322)
* [CVE-2022-30323](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-30323)
* [GHSA-28r2-q6m8-9hpx](https://github.com/advisories/GHSA-28r2-q6m8-9hpx)
* [GHSA-cjr4-fv6c-f3mv](https://github.com/advisories/GHSA-cjr4-fv6c-f3mv)
* [GHSA-fcgg-rvwg-jv58](https://github.com/advisories/GHSA-fcgg-rvwg-jv58)
* [GHSA-x24g-9w7v-vprh](https://github.com/advisories/GHSA-x24g-9w7v-vprh)

[...]

PR #1381, opened 2023-06-13
Change aliases for debian to related instead

another-rex:
Debian advisories' CVE lines lists multiple CVE's that's fixed by the new version. This is not what aliases field is used for, and is better suited to the related field. 

This will have the consequence of not being able to search by the CVE on the osv.dev website to find the DSAs, since only aliases are currently placed in the search_indices, not related. Created #1380 to track this.


andrewpollock (approved):

andrewpollock:
This partially addressed #888

PR #1382, opened 2023-06-13
Fix up a few links that the new link checker was picking up on

andrewpollock:
Also ran `mdformat` on these files (my vim does this on save).

### DIFF
--- a/docs/contributing/architecture.md
+++ b/docs/contributing/architecture.md
@@ -1,17 +1,18 @@
----
-layout: page
-title: Architecture
-permalink: /architecture/
-nav_order: 1
-parent: Contributing
----
+--------------------------------------------------------------------------------
+
+layout: page title: Architecture permalink: /architecture/ nav_order: 1
+
+## parent: Contributing
+
 # Architecture
+
 ![Architecture diagram](../images/architecture.png)
 
 ## Data source
 
-Our data is sourced from a variety of [sources](../data#current-data-sources), which we are looking to expand
-on over time.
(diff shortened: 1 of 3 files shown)


docs/contributing/architecture.md
```diff
@@ -1,17 +1,18 @@
----
-layout: page
-title: Architecture
-permalink: /architecture/
-nav_order: 1
-parent: Contributing
----
+--------------------------------------------------------------------------------
+
+layout: page title: Architecture permalink: /architecture/ nav_order: 1
```
another-rex, line 3: Probably caused by mdformat, but this looks incorrect, each option should be on a new line.

docs/index.md
```diff
@@ -1,17 +1,30 @@
----
+--------------------------------------------------------------------------------
+
 layout: home
-nav_order: 1
----
+
+## nav_order: 1
```
another-rex, line 5: This is also incorrect, probably caused by mdformat as well

docs/contributing/architecture.md
andrewpollock: Oh wow, thanks for flagging this, I'd missed it. I think I'll turn off `mdformat` in my `.vimrc` given it seems very determined to mutilate this.

Fixed in d9ee501

docs/index.md
andrewpollock: Also fixed in d9ee501

another-rex (approved):
LGTM!

PR #1388, opened 2023-06-14
Reflect GCS bucket access logging changes in Terraform

andrewpollock:
The other week I enabled GCS bucket access logging directly, without reflecting it in Terraform (I think I even did it incorrectly, the logging bucket specified was in my own project).

Also add the logs bucket itself

### DIFF
--- a/deployment/terraform/modules/osv/main.tf
+++ b/deployment/terraform/modules/osv/main.tf
@@ -96,6 +96,11 @@ resource "google_storage_bucket" "osv_vulnerabilities_export" {
   lifecycle {
     prevent_destroy = true
   }
+
+  logging {
+    log_bucket        = var.logs_bucket
+    log_object_prefix = "osv-vulnerabilities"
+  }
 }
 
 resource "google_storage_bucket" "cve_osv_conversion" {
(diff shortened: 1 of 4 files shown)


deployment/terraform/modules/osv/main.tf
```diff
@@ -120,6 +125,17 @@ resource "google_storage_bucket" "debian_osv_conversion_bucket" {
...
+  project                     = var.project_id
+  name                        = var.logs_bucket
+  location                    = "US"
+  uniform_bucket_level_access = true
+
+  lifecycle {
+    prevent_destroy = true
+  }
+}
+
```
michaelkedar, line 138: It looks like `osv-logs` and `osv-test-logs` both already exist (but are empty?) on their respective projects.
I don't think terraform is smart enough to automatically pick up that they already exist, the `terraform plan` output says it's trying to create these.

Since they're currently empty, could you delete it on `oss-vdb-test` before merging, and delete it on `oss-vdb` before running `terraform apply` locally?

michaelkedar (approved):
LGTM
One thing before you merge:

deployment/terraform/modules/osv/main.tf
andrewpollock: Oh, I had just created them (and manually redid the bucket configuration). My intention here was to make a future `terraform apply` a no-op...

michaelkedar: I think it'd generally be better to let terraform create them from scratch, but if you want a no-op you could `terraform import` them:
```sh
# in environments/oss-vdb-test
terraform import module.osv_test.google_storage_bucket.logs_bucket oss-vdb-test/osv-test-logs

# in environments/oss-vdb
terraform import module.osv.google_storage_bucket.logs_bucket oss-vdb/osv-logs
```

I'd re-run the `terraform-plan` check if you do that, to make sure nothing is changing.

andrewpollock: Okay I'll do as you suggested and let terraform (re)create the bucket by deleting it first.

PR #1395, opened 2023-06-16
Add a log message for a task's latency

andrewpollock:
Add a new timestamp attribute to pubsub messages emitted by the importer

Then this attribute is present, have the worker log how much time passed between when the messages were published and when the tasks were completed.

This will allow for a log based metric to measure how long things are taking, notably when it comes to importing, for SLO measurement.

### DIFF
--- a/docker/worker/worker.py
+++ b/docker/worker/worker.py
@@ -572,6 +572,20 @@ class TaskRunner:
     if task_type in ('fixed', 'regressed'):
       oss_fuzz.handle_timeout(task_type, source_id, self._oss_fuzz_dir, message)
 
+  def _log_task_latency(self, message):
+    """Determine how long ago the task was requested.
+
+    Log how long it took to be serviced."""
+    request_time = message.attributes.get('timestamp')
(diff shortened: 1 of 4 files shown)


docker/worker/worker.py
```diff
@@ -572,6 +572,20 @@ def handle_timeout(self, subscriber, subscription, ack_id, message):
...
       oss_fuzz.handle_timeout(task_type, source_id, self._oss_fuzz_dir, message)
 
+  def _log_task_latency(self, message):
+    """Determine how long ago the task was requested.
+
+    Log how long it took to be serviced."""
+    request_time = message.attributes.get('timestamp')
+    if request_time:
+      request_time = int(request_time)
+      latency = request_time = int(time.time())
```
oliverchang, line 582: should this be int(time.time()) - request_time ?

andrewpollock: Sigh. Yes. Yes, it should be. Fixed in 2af2b3c

oliverchang (approved):

PR #1397, opened 2023-06-19
Fix determine version api again, add diffing power to indexer-api-caller

another-rex:
Fix the API by using the greater file count for the score.


tools/indexer-api-caller/caller.go
```diff
@@ -16,10 +16,11 @@ import (
 
 var (
 	repoDir   = flag.String("repo", "", "repo directory")
+	repoDir2  = flag.String("repo2", "", "specify another directory to compare file hashes to the first")
```
oliverchang, line 19: What's this used for? Was this just for debugging this issue locally?

oliverchang (approved):

tools/indexer-api-caller/caller.go
another-rex: It helps debugging issues locally, also can be used to get an actual exact number of file matches once you identified what the estimate is with the API. (e.g. you download the repo@tag that determineversion API returns, and run it against what you initially are querying with)

oliverchang: SG. Let's merge and see if we can get this into this week's release?

PR #1402, opened 2023-06-20
Enable API Backend Liveness Probing

michaelkedar:
Looks like the grpc health check service needs to be defined as `grpc.health.v1.Health` for it to work with Cloud Run.
Removed `Check` and `Watch` from `osv.v1.OSV` and registered the `grpc.health.v1.Health` service defined in the `grpcio-health-checking` python package

Added `livenessProbe` into `run.yaml`, using [default parameters](https://cloud.google.com/run/docs/configuring/healthchecks#yaml_4):
- `initialDelaySeconds: 0`
- `timeoutSeconds: 1`
- `failureThreshold: 3`
- `periodSeconds: 10`


oliverchang (approved):
thanks!

PR #1409, opened 2023-06-22
Timestamp error fixes

another-rex:
Renames `timestamp` attribute in the pub/sub message used for logging to `req_timestamp`.

OSS-Fuzz also uses the timestamp attribute in the Pub/Sub message, with a different format and meaning. This clashes with the timestamp info added to help logging.


michaelkedar (approved):

oliverchang (approved):
LGTM, but please add some more context to the PR message!

PR #1410, opened 2023-06-22
Add support for JSON to the id allocator.

calebbrown:
- uses osv-scanner models instead of the local versions (they are more complete).
- currently the same input format is used for output.


oliverchang (approved):
thanks!

oliverchang (approved):

PR #1412, opened 2023-06-23
Minor determineversions docs changes.

oliverchang:
- Make the side panel heading consistent with other methods.
- Clarify that this identifies hashes
- Fix an incorrect link.
- Some more clear wording for the current set of libraries that we can identify.

### DIFF
--- a/docs/api/post-v1-determineversion.md
+++ b/docs/api/post-v1-determineversion.md
@@ -1,6 +1,6 @@
 ---
 layout: page
-title: POST /v1experimental/determineversion
+title: POST /v1experimental/ determineversion
 permalink: /post-v1-determineversion/
 parent: API
 nav_order: 5
--- a/docs/api/post-v1-determineversion.md
+++ b/docs/api/post-v1-determineversion.md
@@ -1,6 +1,6 @@
 ---
 layout: page
-title: Experimental determineversion
+title: POST /v1experimental/determineversion
 permalink: /post-v1-determineversion/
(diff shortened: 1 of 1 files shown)


docs/api/post-v1-determineversion.md
```diff
@@ -1,6 +1,6 @@
 ---
 layout: page
-title: Experimental determineversion
+title: POST /v1experimental/determineversion
```
another-rex, line 3: I think we specifically changed this to Experimental determineversion because of how badly this renders on the sidebar

oliverchang: OK, I found a hack. Adding a space after the second slash causes the `determineversion` to be rendered on the next line. I think it's important to be consistent here.

oliverchang: This is how it looks: image

another-rex: Ah, looks much better!

another-rex (approved):
LGTM!

docs/api/post-v1-determineversion.md
another-rex: the title (the text on the browser tab) might look weird maybe, but most likely it'll be too long to show up anyway.

oliverchang: It gets rendered with a newline very similarly to the side panel here for me.

hayleycd:
Thanks guys!

PR #1414, opened 2023-06-26
Make API backend startupProbe use grpc

michaelkedar:
The default setting uses TCP, which could potentially report the instance is available before the grpc server is actually ready.

### DIFF
--- a/deployment/clouddeploy/osv-api/run.yaml
+++ b/deployment/clouddeploy/osv-api/run.yaml
@@ -10,6 +10,9 @@ spec:
         resources:
           limits:
             memory: 4Gi
+        startupProbe:
+          grpc:
+            service: osv.v1.OSV
         livenessProbe:
           grpc:
             service: osv.v1.OSV


deployment/clouddeploy/osv-api/run.yaml
```diff
@@ -10,6 +10,9 @@ spec:
         resources:
           limits:
             memory: 4Gi
+        startupProbe:
```
rjerrems, line 13: Are you happy to accept the defaults for these fields?

```
          initialDelaySeconds: DELAY
          timeoutSeconds: TIMEOUT
          failureThreshold: THRESHOLD
          periodSeconds: PERIOD
```

michaelkedar: Personally I'm fine with leaving them as default, since that's what we have been using before anyway.

I'm not sure if there are better values to be setting these?

rjerrems: Yeah, it's something we can tune later if you don't have a strong intuition for what you would want to set these values now. We can check what they are once deployed in test and make sure it makes sense for this backend.

oliverchang (approved):

PR #1415, opened 2023-06-26
Include the remote repository URL in some of the impact assessment code

andrewpollock:
This is to help with troubleshooting why converted CVEs aren't successfully having their commits converted back to versions.

Also add type hints

### DIFF
--- a/osv/impact.py
+++ b/osv/impact.py
@@ -116,7 +116,7 @@ class RepoAnalyzer:
     tags = set()
     commits_to_tags = _get_commit_to_tag_mappings(repo)
     branch_to_limit = {}
-    repo_url = 'UNKNOWN_URL'
+    repo_url = None
     if 'origin' in repo.remotes.names():
       repo_url = repo.remotes['origin'].url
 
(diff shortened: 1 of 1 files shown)


osv/impact.py
```diff
@@ -105,17 +105,20 @@ def get_affected(self,
...
+                          fix_commits: list[str],
+                          limit_commits: list[str] = None):
     """Get affected range."""
     range_collector = RangeCollector()
     commits = set()
     seen_commits = set()
     tags = set()
     commits_to_tags = _get_commit_to_tag_mappings(repo)
     branch_to_limit = {}
+    repo_url = 'UNKNOWN_URL'
```
oliverchang, line 119: nit: typically `None` is better to denote no value rather than a sentinel value.

andrewpollock: > nit: typically `None` is better to denote no value rather than a sentinel value.

That's a lot harder to definitively find in the logs to determine if it's even happening, though...

oliverchang: The better practice would be to format this accordingly in the logs, rather than this potentially causing problems in the future because we have code that does `if repo_url: foo`. ,

andrewpollock: Fair enough, I think I've done as you advise in 1f2e404

oliverchang (approved):

PR #1416, opened 2023-06-27
The importer sets `source_id` selectively

andrewpollock:
Fall back to using `source` when it is not set

### DIFF
--- a/docker/worker/worker.py
+++ b/docker/worker/worker.py
@@ -581,7 +581,8 @@ class TaskRunner:
       request_time = int(request_time)
       latency = int(time.time()) - request_time
       task_type = message.attributes['type']
-      source_id = get_source_id(message)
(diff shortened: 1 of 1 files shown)


docker/worker/worker.py
```diff
@@ -581,7 +581,8 @@ def _log_task_latency(self, message):
       request_time = int(request_time)
       latency = int(time.time()) - request_time
       task_type = message.attributes['type']
-      source_id = get_source_id(message)
+      source_id = get_source_id(message) or message.attributes.get(
```
another-rex, line 584: Do you need to do this at 569 as well? it seems to be a similar log

oliverchang (approved):

docker/worker/worker.py
andrewpollock: Good catch, I think you're right, but this auto-merged before I could address, see #1420 1420

PR #1419, opened 2023-06-27
Plumb the bug ID through to more of the log messages

andrewpollock:
Because the worker is threaded the log entries don't readily correlate with the bug ID logged earlier in the process, which makes manually following the processing of a bug in the logs more difficult.

### DIFF
--- a/osv/impact.py
+++ b/osv/impact.py
@@ -93,6 +93,7 @@ class RepoAnalyzer:
     self.detect_cherrypicks = detect_cherrypicks
 
   def get_affected(self,
+                   bug_id: str,
                    repo: pygit2.Repository,
                    regress_commits: list[str],
                    fix_commits: list[str],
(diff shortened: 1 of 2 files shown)


osv/impact.py
```diff
@@ -264,13 +269,14 @@ def _get_commit_to_tag_mappings(repo: pygit2.Repository):
...
 
 
-def get_commit_and_tag_list(repo,
-                            start_commit,
-                            end_commit,
-                            commits_to_tags=None,
-                            include_start=False,
-                            include_end=True,
-                            limit_commit=None):
+def get_commit_and_tag_list(bug_id: str,
```
oliverchang, line 272: This is technically a public interface as part of the Python OSV library that we can't change.

osv/impact.py
```diff
@@ -264,13 +269,14 @@ def _get_commit_to_tag_mappings(repo: pygit2.Repository):
...
+                            repo: pygit2.Repository,
+                            start_commit: str,
+                            end_commit: str,
+                            commits_to_tags: dict = None,
+                            include_start: bool = False,
+                            include_end: bool = True,
+                            limit_commit: dict = None):
```
oliverchang, line 279: Is dict the correct type here? I think it should be str ?

oliverchang (commented):
Hmm, there's a fair bit of plumbing here and a breaking interface change. Could we instead plumb it through via https://github.com/google/osv.dev/blob/abde45effd50e16ceef8039d3fbb14dcfa837716/docker/worker/worker.py#L91 which keeps it consistent with how we plumb through metadata like these generally?

andrewpollock:
Abandoning this approach in favour of https://github.com/google/osv.dev/pull/1430

Issue #1321, opened 2023-05-24
Problems utilising Debian purls


andrewpollock:
I have also noticed an issue regarding debian purls, if you add distro to the purl then from what i've seen it is ignored, also arch=source is problematic as it automatically disqualifies querying by a specific architecture i.e. x86
example: 
    "package": {
        "purl": "pkg:deb/debian/imagemagick@8:6.9.10.23+dfsg-2.1+deb10u1?distro=debian-8"
    }
will ignore the distro query and return the vulnerabilities
and querying 
    "package": {
        "purl": "pkg:deb/debian/imagemagick@8:6.9.10.23+dfsg-2.1+deb10u1?arch=amd64"
    }
will result with empty object, querying with arch=any also yields empty result.
this is the purl returned by osv
![image](https://github.com/google/osv.dev/assets/38865738/595e7e7e-20be-4f0f-946b-2d32bdaacbc9)
purl spec for reference https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst

_Originally posted by @satanshiro in https://github.com/google/osv.dev/issues/1234#issuecomment-1559112256_

andrewpollock:
@satanshiro is it fair to say this is more of an issue when trying to search using a purl with the API?

Ideally you'd like the search for using

```
"package": {
  "purl": "pkg:deb/debian/imagemagick@8:6.9.10.23+dfsg-2.1+deb10u1?arch=amd64"
}
```

to match vulnerabilities that have `pkg:deb/debian/imagemagick?arch=source`?

and

```
"package": {
  "purl": "pkg:deb/debian/imagemagick@8:6.9.10.23+dfsg-2.1+deb10u1?distro=debian-8"
}
```

to only match vulnerabilities in what is essentially the `Debian:8` ecosystem?

satanshiro:
To be honest I'm not sure I can say that syft (https://github.com/anchore/syft) generates purls with distro and arch in the query,
in terms of vulnerabilities, I've seen on osv website, many vulnerabilities that are identified to be on specific debian versions. This implied to me that those vulnerabilities are not relevant to similar version and package names but different debian versions. is this the case?
and in regards to arch looking at the specification of purls([link](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst)): 
![image](https://github.com/google/osv.dev/assets/38865738/0e46c505-d2ae-4c88-b29b-84baf70f84e5)
[...]

PR #1421, opened 2023-06-27
Refactor and update PURL matching in the API

another-rex:
Main functional change in this PR is making `arch=source` to mean `arch=<anything>`, since a source vulnerability will affect all architectures. 

Fixes a problem discussed here: #1321 

Did a bit of refactoring to do this:

- Moved the PURL matching logic around so that PURLs are not cleaned when passed around, but only cleaned when matching specifically with the datastore. 

- Also added the same additional matching step of `_match_purl` to `query_package` that's being used in `query_version`. 

- Updated `_match_purl` to consider `arch=source` to mean `arch=<anything>`.

### DIFF
--- a/gcp/api/Pipfile.lock
+++ b/gcp/api/Pipfile.lock
@@ -762,11 +762,11 @@
         },
         "setuptools": {
             "hashes": [
-                "sha256:5df61bf30bb10c6f756eb19e7c9f3b473051f48db77fddbe06ff2ca307df9a6f",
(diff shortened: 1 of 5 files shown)


oliverchang (approved):
nice! are there any tests we can add ?

another-rex:
Updated the integration test to test for PURLs with arch=source when queries have arch=amd64, also added a unit test to server.py for the `_match_purl()` function

oliverchang (approved):
thanks for adding the tests!!!

PR #1432, opened 2023-06-28
Remove code path per TODO

andrewpollock:
Affected versions are then correctly calculated for CVE-2022-0074


oliverchang:
we had discussed this one offline: it's a little risky given that it may increase the number of false positives for commit based hashing. Let's park it for now and assess how important this is for the NVD ingestion.

PR #1434, opened 2023-06-28
Adding determineversion info to the API index page

hayleycd:
When reviewing the docs for pagination, I realized there was another place where we could add something about the determineversion API. 

[Preview available here](https://hayleycd.github.io/osv.dev/api/)


hayleycd:
I am open to adjusting wording.

oliverchang (approved):

PR #1439, opened 2023-07-03
Add missing reference types.

oliverchang:


osv/vulnerability.proto
```diff
@@ -152,13 +152,16 @@ message Credit {
 message Reference {
   enum Type {
     NONE = 0;
-    WEB = 1;
     ADVISORY = 2;
-    REPORT = 3;
-    FIX = 4;
-    PACKAGE = 5;
     ARTICLE = 6;
```
oliverchang, line 156: Also ordered these lexicographically

andrewpollock: Please add a comment with what the next ID is, given it's now much more difficult to determine this.

michaelkedar (approved):

andrewpollock (changes requested):
To confirm, the only actual change of significance is to the osv/vulnerability.proto file?

oliverchang:
> To confirm, the only actual change of significance is to the osv/vulnerability.proto file?

Yes, everything else is all auto-generated.

osv/vulnerability.proto
oliverchang: done!

andrewpollock (approved):

PR #1440, opened 2023-07-03
Fix test

andrewpollock:
Really need to come back and properly stub the remote Git repositories so this stops breaking...


oliverchang (approved):
Can we point this at an (immutable) fork instead as a stopgap?

PR #1442, opened 2023-07-04
Add version enumeration for last_affected

### LINKED ISSUE #700: support `last_affected`

andrewpollock:
I'd appreciate a thorough review of this:

a) I'm not sure how accurate my docstrings are, because I've basically tried to reverse engineer my understanding of the code to write them
b) the code changed has no test coverage, and I'd appreciate some assistance in changing that situation, but I'm having a hard time understanding how the current tests work...

Fixes #700.


osv/impact.py
```diff
@@ -512,41 +551,64 @@ def _analyze_git_ranges(repo_analyzer, checkout_path, affected_range,
...
             # package no longer exists).
             pass
         else:
           logging.warning('No ecosystem helpers implemented for %s',
-                          affected_package.package.ecosystem)
+                          affected.package.ecosystem)
 
       new_git_versions = set()
       new_introduced = set()
+      new_last_affected = set()
```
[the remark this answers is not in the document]
andrewpollock: > `new_last_affected` doesn't seem to be used anywhere, should there be another section after line 636 for `new_last_affected`?

Ah yes, good catch, thank you. This is what happens when I work on this across a weekend :-) Addressed in [cae6d61](https://github.com/google/osv.dev/pull/1442/commits/cae6d61d9c0ec6bff8051108b0fca54ca546c384)

osv/impact.py
```diff
@@ -33,9 +33,10 @@
...
 
 AffectedResult = collections.namedtuple('AffectedResult',
-                                        'tags commits affected_ranges')
+                                        ['tags', 'commits', 'affected_ranges'])
 
-AnalyzeResult = collections.namedtuple('AnalyzeResult', 'has_changes commits')
+AnalyzeResult = collections.namedtuple('AnalyzeResult',
+                                       ['has_changes', 'commits'])
 
 TagsInfo = collections.namedtuple('TagsInfo', 'tags latest_tag')
```
michaelkedar, line 41: nit: also change this last one to use the list notation

osv/impact.py
```diff
@@ -59,24 +60,25 @@ class RangeCollector:
...
+      if fixed_in is None and affected_in is None:
         # New range doesn't add anything new.
         return
 
       # Remove in-place as we need to preserve insertion order.
       ranges = self.grouped_ranges[introduced_in]
-      ranges.append((introduced_in, fixed_in))
+      ranges.append((introduced_in, fixed_in, affected_in))
       for value in ranges.copy():
         if value[1] is None:
```
michaelkedar, line 74: I don't fully understand what this function is doing, but I think this should now be
```python
if value[1] is None and value[2] is None:
```

osv/impact.py
```diff
@@ -174,11 +185,31 @@ def _get_affected_range(self,
...
+
+        range_collector.add(equivalent_regress_commit, equivalent_fix_commit,
+                            equivalent_last_affected_commit)
 
       if equivalent_fix_commit:
         end_commit = equivalent_fix_commit
         include_end = False
+      elif equivalent_last_affected_commit:
+        end_commit = equivalent_last_affected_commit
+        include_end = False
```
michaelkedar, line 212: `include_end` should be `True` for last_affected

michaelkedar (commented):
> b) the code changed has no test coverage, and I'd appreciate some assistance in changing that situation, but I'm having a hard time understanding how the current tests work...

Apparently this is somewhat tested in the worker's tests:
https://github.com/google/osv.dev/blob/efe2a73c48cfc8f358f304cf6d359d404b9806d8/docker/worker/worker_test.py#L127

~~I'm not sure what change has made them start failing, though...~~ [I think I found the problem](https://github.com/google/osv.dev/pull/1442#discussion_r1251558967)

osv/impact.py
```diff
@@ -174,11 +185,31 @@ def _get_affected_range(self,
...
+          equivalent_last_affected_commit = self._get_equivalent_commit(
+              repo,
+              ref,
+              last_affected_commit,
+              detect_cherrypicks=detect_cherrypicks)
+          if equivalent_last_affected_commit:
+            break
+
+        range_collector.add(equivalent_regress_commit, equivalent_fix_commit,
+                            equivalent_last_affected_commit)
```
michaelkedar, line 205: Ah, this is indented one level too deep (should be outside of `if last_affected_commits:`)

osv/impact.py
oliverchang: I'd probably just convert all of these to dataclasses instead, which are a lot nicer to work with.

osv/impact.py
```diff
@@ -459,9 +490,30 @@ def sort_key(event):
   return versions
 
 
-def _analyze_git_ranges(repo_analyzer, checkout_path, affected_range,
-                        new_versions, commits, new_introduced, new_fixed):
-  """Analyze git ranges."""
+def _analyze_git_ranges(repo_analyzer: RepoAnalyzer, checkout_path: str,
```
oliverchang, line 493: Please add tests to https://github.com/google/osv.dev/blob/efe2a73c48cfc8f358f304cf6d359d404b9806d8/docker/worker/worker_test.py#L127 as @michaelkedar mentioned. 

The tests are based on a fake repo from here: https://github.com/google/osv.dev/blob/efe2a73c48cfc8f358f304cf6d359d404b9806d8/.gitmodules#L6

osv/impact.py
```diff
@@ -459,9 +490,30 @@ def sort_key(event):
...
+
+  Args:
+    repo_analyzer: an instance of RepoAnalyzer to use.
+    checkout_path: If defined, used in lieu of cloning the repo.
+    affected_range: the GIT range from the vulnerability.
+    new_versions: a set that will be in-place modified to contain any new
+    versions detected by analysis.
+    commits: a set that will be in-place modified to contain any commits???
+    new_introduced: a set that will be in-place modified to contain additional
+    introduced commits determined.
```
oliverchang, line 508: nit: determined **by cherrypick detection**

osv/impact.py
andrewpollock: Thank you! Fixed in e8ac084

osv/impact.py
andrewpollock: Yes, you're right, thanks. Fixed in c38c637

osv/impact.py
andrewpollock: I've made it consistent for now in 65766cb and I'll look at how invasive a change dataclasses would be if I get a chance. I didn't want to go too much off on a tangent.

osv/impact.py
andrewpollock: I'm finding why a method named `add` is _removing_ anything to be... warranting more comments explaining what's going on here. I agreee with you on the addition, on further inspection. @oliverchang can you provide some more context so I can add more comments to the code while I'm here?

andrewpollock: Addressed in f4072fe and renamed some variables for better readability

oliverchang: This is to address a limitation with tuples being immutable. We remove them rather than setting them in place.

osv/impact.py
```diff
@@ -174,11 +190,31 @@ def _get_affected_range(self,
         if equivalent_fix_commit:
           break
 
-      range_collector.add(equivalent_regress_commit, equivalent_fix_commit)
+      # Get the latest equivalent commit in the last_affected range (if
+      # present).
+      equivalent_last_affected_commit = None
+      if last_affected_commits:
```
oliverchang, line 196: Do we need this if statement, or is it redundant given the for loop? The `fix_commits` case doesn't have it.

osv/impact.py
```diff
@@ -174,11 +190,31 @@ def _get_affected_range(self,
...
+          if equivalent_last_affected_commit:
+            break
+
+      range_collector.add(equivalent_regress_commit, equivalent_fix_commit,
+                          equivalent_last_affected_commit)
 
       if equivalent_fix_commit:
         end_commit = equivalent_fix_commit
         include_end = False
+      elif equivalent_last_affected_commit:
```
oliverchang, line 215: nit: add a comment here saying that it's technically invalid to have both fix and last_affected, but the way this is structured makes us prefer taking the fix.

oliverchang (commented):
nice! I think we just need to add some tests to verify this works.

oliverchang (commented):
I've pushed some fixes to this PR, and added a test. See https://github.com/google/osv.dev/pull/1442/commits/780ad77f78d3bcfebffef5d7625df1ea0d1f9113

oliverchang (approved):

osv/impact.py
andrewpollock: Converted to dataclasses in 03dac88

osv/impact.py
andrewpollock: > I don't fully understand what this function is doing, but I think this should now be
> 
> ```python
> if value[1] is None and value[2] is None:
> ```

I *think* @oliverchang did what needed to be done in 780ad77 so this has been addressed appropriately.

osv/impact.py
```diff
@@ -459,9 +495,30 @@ def sort_key(event):
...
+  Args:
+    repo_analyzer: an instance of RepoAnalyzer to use.
+    checkout_path: If defined, used in lieu of cloning the repo.
+    affected_range: the GIT range from the vulnerability.
+    new_versions: a set that will be in-place modified to contain any new
+    versions detected by analysis.
+    commits: a set that will be in-place modified to contain any commits???
+    new_introduced: a set that will be in-place modified to contain additional
+    introduced commits determined.
+    new_last_affected: set that will be in-place modified to contain additional
```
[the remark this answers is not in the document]
andrewpollock: It looks like you did this in 780ad77 so there's nothing further to be done here?

oliverchang (approved):

PR #1446, opened 2023-07-05
Update URL to get debian versions for debian_first_package

another-rex:
The previous URL points to a resource that is not being actively updated, this new endpoint points to the latest entry in the raw git repository. We only call this endpoint once per day so should not overload the git server.

- Update URL to git
- add additional logging to easily help identify errors 
- Properly use pandas to correctly set cell data (the old method will fail on newer versions of pandas)
- Update pipenv.lock to the latest version


andrewpollock (approved):
Please update the PR description to state why you're making this change.

oliverchang (approved):

PR #1447, opened 2023-07-05
frontend: implement better support for individual CVE record rendering

andrewpollock:
This is the changes for individual record rendering, as documented in go/osv-cve-ui-plan:

- when `introduced` is zero, don't render a bogus link
- add an explanatory tooltip for this scenario as well
- a new package info section links to the repo, when it is available
- the Affected Versions section is conditionally displayed only when there are versions available to display

Additionally:
- Remove the link decoration on the outlink to the GitHub repo in the top right
- Reduce spurious whitespace in generated HTML

### DIFF
--- a/gcp/appengine/frontend3/src/styles.scss
+++ b/gcp/appengine/frontend3/src/styles.scss
@@ -81,7 +81,7 @@ a {
   text-decoration: underline;
 }
 
-a:not([href^="https://osv.dev"]):not([href^="/"]):not([href^="#"]):not([href^="https://google.github.io/osv.dev"])::after {
+a:not([href^="https://osv.dev"]):not([href^="/"]):not([href^="#"]):not([href^="https://google.github.io/osv.dev"]):not([href^="https://github.com/google/osv.dev"])::after {
   display: inline-block;
   content: ' ';
   background-image: url('/static/img/external-link.svg');
@@ -112,6 +112,33 @@ pre {
 
 .link-button:hover {
   background: $osv-accent-color-hover;
+}
+
+.tooltip {
+  position: relative;
+  display: inline-block;
+}
+
+.tooltip .tooltiptext {
+  visibility: hidden;
+  width: 120px;
+  background-color: white;
+  color: #000;
+  text-align: center;
+  padding: 5px 0;
+  border-radius: 6px;
+
+  top: -5px;
+  left: 105%;
+  z-index: 1;
+
+  opacity: 0;
+  transition: opacity 1s;
+}
+
+.tooltip:hover .tooltiptext {
+  visibility: visible;
+  opacity: 1;
 }
 
 /** Top bar */
--- a/gcp/appengine/frontend3/src/styles.scss
+++ b/gcp/appengine/frontend3/src/styles.scss
@@ -122,11 +122,14 @@ pre {
 .tooltip .tooltiptext {
   visibility: hidden;
   width: 120px;
-  background-color: white;
-  color: #000;
+  background: #11111199;
+  color: #fff;
   text-align: center;
-  padding: 5px 0;
-  border-radius: 6px;
+  padding: 5px;
+  border-radius: 4px;
+
+  max-width: 250px;
+  line-height: 16px;
 
   top: -5px;
   left: 105%;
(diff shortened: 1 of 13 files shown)


gcp/appengine/frontend3/src/templates/vulnerability.html
```diff
@@ -60,17 +60,31 @@ <h2 class="title">Affected packages</h2>
   <spicy-sections
       class="vulnerability-packages{% if vulnerability.affected|should_collapse %} force-collapse{% endif %}">
     {% for affected in vulnerability.affected %}
+      {% if affected['package']['ecosystem'] %}
```
oliverchang, line 63: Would it make more sense to have a tab that just says "Source-only" or display the repo URL instead to keep things more consistent? 

This seems particularly important if there are other tabs for other ecosystems to render here in the same entry. I.e. if there is an Alpine tab.

gcp/appengine/frontend3/src/templates/vulnerability.html
```diff
@@ -60,17 +60,31 @@ <h2 class="title">Affected packages</h2>
...
+        <div class="vulnerability-package-subsection mdc-layout-grid__inner">
+          <h3 class="mdc-layout-grid__cell--span-3">
+            Package Info
+          </h3>
+          <div class="mdc-layout-grid__cell--span-9">
+            <dl>
+              <dt>Package Name</dt>
+              <dd>{{ affected['package']['name'] or 'Unknown' }}</dd>
+              <dt>Repo URL</dt>
+              <dd><a href="{{ vulnerability.repo }}">{{ vulnerability.repo }}</a></dd>
```
oliverchang, line 80: Thinking more on this, it's possible that there could be multiple repo URLs for the same tab/affected entry. We should handle such a case more gracefully, at least as a follow up item.

gcp/appengine/frontend3/src/styles.scss
```diff
@@ -114,6 +114,33 @@ pre {
   background: $osv-accent-color-hover;
 }
 
+.tooltip {
+  position: relative;
+  display: inline-block;
+}
+
+.tooltip .tooltiptext {
```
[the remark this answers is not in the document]
andrewpollock: Done in https://github.com/google/osv.dev/pull/1447/commits/026af26a3d232e858edd2e7c8dcbefccb3f2d571

gcp/appengine/frontend3/src/templates/vulnerability.html
andrewpollock: Addressed in [cd34ab9](https://github.com/google/osv.dev/pull/1447/commits/cd34ab994bd4fb1b57e5c01a49b9a503e32f58ea)

gcp/appengine/frontend3/src/templates/vulnerability.html
oliverchang: Let's add a TODO somewhere on this.

gcp/appengine/frontend3/src/templates/vulnerability.html
```diff
@@ -88,12 +100,17 @@ <h3 class="mdc-layout-grid__cell--span-3">
                         {% set link = event | event_link %}
                         {% if link %}
                           <a href="{{ link }}"}>
+                        {% elif event | event_type == 'Introduced' and event | event_value == '0' %}
+                          <div class="tooltip">
```
oliverchang, line 104: The <div> seems to overlap the <a>. Is that intentional?

oliverchang (approved):
Very nice! LGTM with some final nits/comments.

Can you also update the PR title to include "frontend:" or something similar? The PR description bit around the tab should also be updated.

gcp/appengine/frontend3/src/templates/vulnerability.html
andrewpollock: I'm not following? The opening and closing div tags are both in the equivalent `elif` branches as the anchor tags?

gcp/appengine/frontend3/src/templates/vulnerability.html
andrewpollock: Done in [2ea079e](https://github.com/google/osv.dev/pull/1447/commits/2ea079ebed15318614b5ea12bdae826de01d4e14)

gcp/appengine/frontend3/src/templates/vulnerability.html
oliverchang: Ah, disregard. I misinterpreted what was going on here. It seemed like this was 

```
<a>
<div>
</a>
</div>
```

But I now realise the <a> and <div> are mutually exclusive.

PR #1448, opened 2023-07-06
Fix various API errors

another-rex:
- Replace logging.error with logging.warning to keep our error dashboard cleaner
- Fix bug with aborting where we are using the wrong context object
- Catch BadValueErrors where some query parameters are not valid


andrewpollock (approved):
Woot!

michaelkedar (approved):

oliverchang (approved):
nice!!

PR #1460, opened 2023-07-12
Make API resolve Bugs from Commits properly asynchronous

michaelkedar:
[From what I can tell](https://groups.google.com/g/appengine-ndb-discuss/c/Krr68fAY-0M), using `bug = yield future_bug` allows other tasklets to run while waiting for a result, while using `bug = future_bug.result()` will force the entire function to finish before making ndb calls in other tasklets (but I still don't think I really understand it).

I've tested this locally, and this change seems to fix the strange error we've been getting, which seems to happens when there's lots (>100) of commit hashes in a batch query:
```
Exception in thread Thread-2 (_run):
  File "/usr/lib/python3.11/threading.py", line 1038, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.11/threading.py", line 975, in run
    self._target(*self._args, **self._kwargs)
  File "src/python/grpcio/grpc/_cython/_cygrpc/thread.pyx.pxi", line 53, in grpc._cython.cygrpc._run_with_context._run
  File "/usr/local/google/home/michaelkedar/.local/share/virtualenvs/api-CM6qCKPQ/lib/python3.11/site-packages/grpc/_channel.py", line 1392, in channel_spin
    event = state.channel.next_call_event()
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "src/python/grpcio/grpc/_cython/_cygrpc/channel.pyx.pxi", line 489, in grpc._cython.cygrpc.Channel.next_call_event
  File "src/python/grpcio/grpc/_cython/_cygrpc/channel.pyx.pxi", line 171, in grpc._cython.cygrpc._next_call_event
  File "src/python/grpcio/grpc/_cython/_cygrpc/channel.pyx.pxi", line 172, in grpc._cython.cygrpc._next_call_event
  File "src/python/grpcio/grpc/_cython/_cygrpc/channel.pyx.pxi", line 481, in grpc._cython.cygrpc.Channel.next_call_event.on_success
  File "src/python/grpcio/grpc/_cython/_cygrpc/channel.pyx.pxi", line 263, in grpc._cython.cygrpc._process_integrated_call_tag
KeyError: <grpc._cython.cygrpc._BatchOperationTag object at 0x7f90c04288b0>
```

I don't know *why* or *if* this actually fixes it (or if it just makes the error occur in a race condition).
My best guesses are that theirs either some limit on how many grpc requests can exist at once, or ndb/grpc sometimes accidentally signals to the wrong request when there a lots waiting, but these don't really line up with my understanding of ndb's `.result()` vs `yield` or why the change would fix it.


andrewpollock (approved):
Nice work. I presume given you've been able to reliably reproduce the failure mode (in staging?) you'll be able to validate that you can no longer reproduce it after this change goes live (in staging)?

oliverchang (approved):

michaelkedar:
> Nice work. I presume given you've been able to reliably reproduce the failure mode (in staging?) you'll be able to validate that you can no longer reproduce it after this change goes live (in staging)?

Yep - interestingly it currently seems to be ok with 86 (of the same) commits in a batch, but always errors with >=87 (both locally and on the test environment). After this change locally I haven't run into the exception even with our current maximum of 1000 commits.

another-rex (approved):

PR #1462, opened 2023-07-13
Bump worker-base to Python 3.11.4

michaelkedar:
Noticed my Cloud Top is now using Python 3.11.4.

I think there should be a way to get dependabot/rennovate to make these PRs.
Alternatively, we could change the base image so we don't need to build python from source.

note: The ci image that runs our tests needs to be manually deployed but is based on the worker image, so it can only really be rebuilt after this is merged...

### DIFF
--- a/docker/worker-base/Dockerfile
+++ b/docker/worker-base/Dockerfile
@@ -28,7 +28,7 @@ RUN apt-get update && apt-get upgrade -y && \
         libssl-dev \
         software-properties-common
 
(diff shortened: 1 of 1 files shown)


andrewpollock:
> Alternatively, we could change the base image so we don't need to build python from source.

Honestly, I'd love for us to be able to get to this point if at all possible...

andrewpollock (approved):

Issue #1418, opened 2023-06-27
add HSEC (Haskell) security advisories data source


frasertweedale:
The Haskell community is establishing a security advisory database: https://github.com/haskell/security-advisories .
We are currently bootstrapping the database with known vulnerabilities and developing our tooling and processes.

OSV export is nearly complete.  Of course, we would like osv.dev to ingest our data.

Could you please advise next steps?  What is the best way to publish our OSV data for osv.dev to ingest and process it?

I couldn't easily make sense of the OSV importer code or where the sources are configured.  But from what I can see,
it looks like a Git branch containing *just* the individual OSV json files for each of our advisories would be easy for you to
consume?  If you prefer something else please elaborate.

Thanks!

Fraser
on behalf of the Haskell Security Response Team

oliverchang:
Hey there! This is super exciting, thanks for filing this issue.  Sorry for the lack of detailed documentation on this. 

Indeed, a git repo with a branch containing OSV JSONs would work great for us. One additional piece here to make our API matching and indexing work is to add support for version ordering and enumeration in https://github.com/google/osv.dev/blob/1d395c94739bd50b919b21a1642f8fe4d697b539/osv/ecosystems/_ecosystems.py#L28. This doesn't block the initial ingestion, but it does make API queries work. 

Does Hackage/Haskell use its own versioning scheme, or does it follow something like SemVer 2.0 strictly? New ecosystems that follow SemVer can be added very easily with minimal work.

frasertweedale:
> Does Hackage/Haskell use its own versioning scheme, or does it follow something like SemVer 2.0 strictly? New ecosystems that follow SemVer can be added very easily with minimal work.

Haskell package versions are not SemVer.  They allows an arbitrary number (> 1) of version number components, and there are no optional labels.  Comparison logic is more or less what you'd expect, with the one quirk that e.g. `1.2.` < `1.2.0`.

oliverchang:
Thanks for clarifying. Is there a spec describing the Haskell package version scheme? Is https://pvp.haskell.org/ the one?

frasertweedale:
https://pvp.haskell.org/ has all the necessary info, but I would call it "informational".  The closest thing to a normative specification is the doc and source code of https://hackage.haskell.org/package/Cabal-syntax-3.10.1.0/docs/Distribution-Types-Version.html .

frasertweedale:
[...]

PR #1463, opened 2023-07-14
Hackage and GHC version enumeration (#1418)

frasertweedale:
Add version enumeration for the Haskell ecosystems: Hackage (the main package index) and GHC (the toolchain).

Fixes: https://github.com/google/osv.dev/issues/1418

### DIFF
--- a/osv/ecosystems/haskell.py
+++ b/osv/ecosystems/haskell.py
@@ -1,0 +1,153 @@
+# Copyright 2021 Google LLC
+# Copyright 2023 Fraser Tweedale
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+Haskell ecosystem helpers.
+
+Contact the Haskell Security Response Team <security-advisories@haskell.org>
+if something is broken and you need help to fix it.
+
+"""
+
+import requests
+import typing
+
+from . import config
+from .helper_base import Ecosystem, EnumerateError
+from .. import semver_index
+
+
+class Hackage(Ecosystem):
+  """Hackage (Haskell package index) ecosystem."""
+
+  _API_PACKAGE_URL = 'https://hackage.haskell.org/package/{package}.json'
+
+  def sort_key(self, version):
+    """Sort key.
+
+    The Haskell package version data type is defined at
+    https://hackage.haskell.org/package/Cabal-syntax/docs/Distribution-Types-Version.html
+
+    """
+    return [int(x) for x in version.split('.')]
+
+  def enumerate_versions(self,
+                         package,
+                         introduced,
+                         fixed=None,
+                         last_affected=None,
+                         limits=None):
+    """Enumerate versions."""
+    response = requests.get(
+        self._API_PACKAGE_URL.format(package=package), timeout=config.timeout)
+    if response.status_code == 404:
+      raise EnumerateError(f'Package {package} not found')
+    if response.status_code != 200:
+      raise RuntimeError(
+          f'Failed to get Hackage versions for {package} with: {response.text}')
+
+    response = response.json()
+    versions = list(response.keys())
+
+    self.sort_versions(versions)
+    return self._get_affected_versions(versions, introduced, fixed,
+                                       last_affected, limits)
+
+
+class GHC(Ecosystem):
+  """Glasgow Haskell Compiler (GHC) ecosystem."""
+
+  _API_PACKAGE_URL = ('https://gitlab.haskell.org'
+                      '/api/v4/projects/3561/repository/tags?per_page=-1')
+  """
+  Historical versions do not have tags in the Git repo, so we hardcode the
+  list.  See https://github.com/google/osv.dev/pull/1463 for discussion.
+  """
+  HISTORICAL_VERSIONS = [
+    '0.29',
+    '2.10',
+    '3.02', '3.03',
+    '4.02', '4.04', '4.06', '4.08', '4.08.1', '4.08.2',
+    '5.00', '5.00.1', '5.00.2', '5.02', '5.02.1', '5.02.2', '5.02.3',
+    '5.04', '5.04.1', '5.04.2', '5.04.3',
+    '6.0', '6.0.1',
+    '6.2', '6.2.1', '6.2.2',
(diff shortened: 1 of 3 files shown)


osv/ecosystems/hackage.py
```diff
@@ -0,0 +1,124 @@
...
+                                       last_affected, limits)
+
+
+class GHC(Ecosystem):
+  """Glasgow Haskell Compiler (GHC) ecosystem."""
+
+  _API_PACKAGE_URL = (
+    'https://gitlab.haskell.org'
+    '/api/v4/projects/3561/repository/tags?per_page=-1'
+  )
```
michaelkedar, line 73: It doesn't look like there's any tags for GHC versions before 7.2.1 in this repo. On the [GHC downloads page](https://www.haskell.org/ghc/download.html) I can see versions going all the way back to 0.29.

Is there a single source of truth for every GHC version ever released?

osv/ecosystems/hackage.py
```diff
@@ -0,0 +1,124 @@
+# Copyright 2021 Google LLC
```
michaelkedar, line 1: Could you please add some simple tests in a `hackage_test.py`, along the lines of whats in [pypi_test.py](https://github.com/google/osv.dev/blob/1838dfe41a1383261cd3402bdd962935414fa2b1/osv/ecosystems/pypi_test.py) or [rubygems_test.py](https://github.com/google/osv.dev/blob/1838dfe41a1383261cd3402bdd962935414fa2b1/osv/ecosystems/rubygems_test.py)?

michaelkedar (commented):
Thanks for doing this!

Looks mostly good to me, just a few small things

osv/ecosystems/hackage.py
frasertweedale: Not surprising for a 30 year old project that underwent several VCS migrations.  Even the download page omits several earlier versions.  I asked around and the general consensus was Git tags would be the most reliable data source on an ongoing basis.  I will ask whether we could add tags for older versions.

osv/ecosystems/hackage.py
frasertweedale: No problem.  I'll do it in the next day or so.

osv/ecosystems/hackage.py
frasertweedale: @michaelkedar adding tags for older versions is infeasible (because it is unknown/unknow*able* what commits particular releases were cut from).  What do you think about including a hardcoded list of the known historical public releases that don't have a corresponding Git tag?

michaelkedar: Yeah, a hardcoded list does seem like the only feasible way to have these historic versions. :+1:

osv/ecosystems/haskell.py
```diff
@@ -0,0 +1,145 @@
...
+
+    Different components of GHC are part of the same software release.
+    So we ignore the package (component) name.
+
+    Historical versions do not have tags in the Git repo, so we hardcode
+    the list.  See https://github.com/google/osv.dev/pull/1463 for
+    discussion.
+
+    """
+    historical_versions = [
```
michaelkedar, line 114: nit: maybe define the historical versions outside of `enumerate_versions()`, next to `_API_PACKAGE_URL`

michaelkedar (commented):
LGTM, just one typo

osv/ecosystems/haskell.py
```diff
@@ -0,0 +1,145 @@
...
+    'https://gitlab.haskell.org'
+    '/api/v4/projects/3561/repository/tags?per_page=-1'
+  )
+
+  def sort_key(self, version):
+    """Sort key."""
+    return semver_index.parse(version)
+
+  @staticmethod
+  def tag_to_version(tag)
```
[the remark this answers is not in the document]
frasertweedale: Thanks, will update it tomorrow.

michaelkedar:
I got this error when running the tests
```
ERROR: test_next_version (osv.ecosystems.haskell_test.GHCEcosystemTest.test_next_version)
Test next_version.
  File "/workspace/osv/ecosystems/haskell_test.py", line 41, in test_next_version
    self.assertEqual('0.29', ecosystem.next_version('GHC', '0'))
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspace/osv/ecosystems/helper_base.py", line 46, in next_version
    versions = self.enumerate_versions(package, version, fixed=None)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspace/osv/ecosystems/haskell.py", line 143, in enumerate_versions
    self.sort_versions(versions)
  File "/workspace/osv/ecosystems/helper_base.py", line 65, in sort_versions
    versions.sort(key=self.sort_key)
  File "/workspace/osv/ecosystems/haskell.py", line 96, in sort_key
    return semver_index.parse(version)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspace/osv/semver_index.py", line 104, in parse
    return semver.Version.parse(coerce(version))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/builder/home/.local/share/virtualenvs/workspace-dqq3IVyd/lib/python3.11/site-packages/semver/version.py", line 646, in parse
    raise ValueError(f"{version} is not valid SemVer string")
ValueError: 7.10.3a is not valid SemVer string

```

frasertweedale:
Thanks, looking into it.

osv/ecosystems/haskell.py
```diff
@@ -0,0 +1,155 @@
...
+    '6.4', '6.4.1', '6.4.2', '6.4.3',
+    '6.6', '6.6.1',
+    '6.8.1', '6.8.1', '6.8.3',
+    '6.10.1', '6.10.2-rc1', '6.10.2', '6.10.3', '6.10.4-rc1', '6.10.4',
+    '6.12.1-rc1', '6.12.1', '6.12.2-rc1', '6.12.2', '6.12.3-rc1', '6.12.3',
+    '7.0.1-rc1', '7.0.1-rc2', '7.0.1', '7.0.2-rc1', '7.0.2-rc2', '7.0.2',
+    '7.0.3', '7.0.4-rc1', '7.0.4',
+  ]  # yapf: disable
+
+  INT_PATTERN = re.compile(r'^[1-9][0-9]*$')
```
michaelkedar, line 96: This regex won't match `0`

Our `semver_index.parse()` will accept and trim numbers with leading 0's, so you can get away with just using the `isdigit()` method on the integer strings.

frasertweedale: of course, thank you for catching that.

michaelkedar (approved):
This all LGTM. Thanks again for working on this!

I'm going to merge this now and trigger a reimport on our [staging instance](https://oss-vdb-test.appspot.com/list?ecosystem=Hackage). We should start to see the affected versions being populated.

If everything's working, this will go into production when we do a release next week.

frasertweedale:
Thanks for your reviews!

PR #1466, opened 2023-07-18
Removed FAQ for pagination

hayleycd:
Only merge this when ready. 

Once pagination is up and running, I think we should remove the associated FAQ. 

I would wait to merge this until you are ready to remove [this](https://github.com/google/osv.dev/blob/1d7b86e2db7dc7747c6254bd522128020b6c11c6/gcp/api/server.py#L74) (_LINUX_ERROR)


hayleycd:
Seems like this was taken care of elsewhere.

PR #1475, opened 2023-07-19
Reenable Kernel queries

another-rex:
Kernel queries was disabled before paging was implemeneted to prevent timing out other queries. Now that paging is implemented, reenable kernel queries


michaelkedar (approved):
There's a typo in the PR name :)

Issue #1476, opened 2023-07-19
Add support for CRAN+Bioconductor


oliverchang:
https://github.com/ossf/osv-schema/pull/176

- [x] Initial import 
- [x] Ecosystem enumeration support (https://github.com/google/osv.dev/blob/1d395c94739bd50b919b21a1642f8fe4d697b539/osv/ecosystems/_ecosystems.py#L28) to enable OSV.dev API queries.

oliverchang:
@tylfin we may need some assistance with the second point above (similar to #1463).

michaelkedar:
CRAN & Bioconductor is now on production: https://osv.dev/list?ecosystem=CRAN&q= 
(API queries work too!)

PR #1477, opened 2023-07-19
Add support for CRAN and Bioconductor

tylfin:
This is still a work-in-progress but here we:

- [x] Add the R ecosystem to _ecosystems with accompanying R ecosystem class and unit tests
- [x] Use the Posit Public Package Manager as an API endpoint to query for package version information.
- [x] TODO: Add Bioconductor support

Refs: https://github.com/google/osv.dev/issues/1476


osv/ecosystems/_ecosystems.py
```diff
@@ -30,6 +31,7 @@
     'Packagist': Packagist(),
     'Pub': Pub(),
     'PyPI': PyPI(),
+    'R': R(),
```
michaelkedar, line 34: This needs to match the ecosystems defined in the schema, so it should be `CRAN` (and also `Bioconductor`) based on https://github.com/ossf/osv-schema/pull/176

michaelkedar (commented):
Hey, thanks for working on this!

Just wanted to point out one thing:

osv/ecosystems/cran.py
```diff
@@ -0,0 +1,69 @@
...
+class CRAN(Ecosystem):
+  """CRAN ecosystem helpers."""
+
+  # Use the Posit Public Package Manager API to pull both the current and archived versions for a specific package
+  # since CRAN doesn't natively support this functionality.
+  _API_PACKAGE_URL_POSIT_CRAN = 'https://packagemanager.posit.co/__api__/repos/2/packages/{package}'
+
+  def sort_key(self, version):
+    """Sort key."""
+    return packaging.version.parse(version)
```
michaelkedar, line 31: Is Python's versioning scheme suitable for CRAN?
Looking at [this page](https://r-pkgs.org/lifecycle.html#sec-lifecycle-version-number), R seems to treat both `.` and `-` as the same separator, which Python does not.

Though, I don't know if Posit will show versions using `-`'s?

Also, could you add a link to that R version number (or a better page?) in a comment for reference.

osv/ecosystems/bioconductor.py
```diff
@@ -0,0 +1,73 @@
...
+  """Bioconductor ecosystem helpers."""
+
+  # Use the Posit Public Package Manager API to pull both the current and archived versions for a specific package
+  # since CRAN doesn't natively support this functionality.
+  _API_PACKAGE_URL_POSIT_BIOCONDUCTOR = 'https://packagemanager.posit.co/__api__/repos/4/packages/{package}?bioc_version={bioc_version}'
+  _BIOC_VERSIONS = ['3.18', '3.17', '3.16', '3.15', '3.14', '3.13', '3.12', '3.11', '3.10', '3.9', '3.8', '3.7', '3.6', '3.5', '3.4', '3.3', '3.2', '3.1']
+
+  def sort_key(self, version):
+    """Sort key."""
+    return packaging.version.parse(version)
```
michaelkedar, line 32: Using Python's versioning scheme for this might be a bit misleading.

Looking at [this page](https://contributions.bioconductor.org/versionnum.html), it looks like Bioconductor versioning is mostly just semver. I think I'd prefer this to use [`semver_index.parse()`](https://github.com/google/osv.dev/blob/78296d3018200c3dcf8226310f58445341e29ee9/osv/semver_index.py#L102) to make it a bit clearer that this is not the same versioning scheme as Python.

Also, could you please add in a comment a link to that Bioconductor Version Numbering page for reference.

osv/ecosystems/bioconductor.py
```diff
@@ -0,0 +1,73 @@
...
+from . import config
+from .helper_base import Ecosystem, EnumerateError
+
+class Bioconductor(Ecosystem):
+  """Bioconductor ecosystem helpers."""
+
+  # Use the Posit Public Package Manager API to pull both the current and archived versions for a specific package
+  # since CRAN doesn't natively support this functionality.
+  _API_PACKAGE_URL_POSIT_BIOCONDUCTOR = 'https://packagemanager.posit.co/__api__/repos/4/packages/{package}?bioc_version={bioc_version}'
+  _BIOC_VERSIONS = ['3.18', '3.17', '3.16', '3.15', '3.14', '3.13', '3.12', '3.11', '3.10', '3.9', '3.8', '3.7', '3.6', '3.5', '3.4', '3.3', '3.2', '3.1']
```
michaelkedar, line 28: Do these `_BIOC_VERSIONS` need to be kept up to date? Is there somewhere we can pull this list from?

osv/ecosystems/bioconductor.py
```diff
@@ -0,0 +1,73 @@
...
+    return packaging.version.parse(version)
+
+  def _enumerate_versions(self, url, package, introduced, fixed=None, last_affected=None, limits=None):
+    """Helper method to enumerate versions from a specific URL."""
+
+    versions = []
+    for version in self._BIOC_VERSIONS:
+      response = requests.get(url.format(package=package, bioc_version=version), timeout=config.timeout)
+      if response.status_code == 404:
+        break
```
michaelkedar, line 41: Is it supposed to be `break` here vs `continue`?
Or is it not possible for a package that exists in earlier bioc_versions to not appear in later versions?

michaelkedar (commented):
Looks pretty good.
I just have a few questions:

osv/ecosystems/bioconductor.py
tylfin: > Or is it not possible for a package that exists in earlier bioc_versions to not appear in later versions

Yeah, this is the case, I'll add a comment for clarity

osv/ecosystems/bioconductor.py
tylfin: Yeah, I can pull this dynamically to keep it up to date

andrewpollock: With 5463e09853c4190de02f1ab00087ffefb3eea7a8, is this still required?

osv/ecosystems/bioconductor.py
```diff
@@ -0,0 +1,91 @@
...
+  def refresh_bioc_versions(self):
+    """
+    Pull latest Bioconductor versions
+    """
+    response = requests.get(self._API_BIOC_VERSIONS_URL)
+    data = response.json()
+
+    self._BIOC_VERSIONS = [bioc['bioc_version'] for bioc in data['bioc_versions']]
+
+  def sort_key(self, version):
```
andrewpollock, line 47: Could you not then subclass `SemverEcosystem` instead of `Ecosystem` making this not necessary?

osv/ecosystems/cran.py
```diff
@@ -0,0 +1,69 @@
...
+from .helper_base import Ecosystem, EnumerateError
+
+class CRAN(Ecosystem):
+  """CRAN ecosystem helpers."""
+
+  # Use the Posit Public Package Manager API to pull both the current and archived versions for a specific package
+  # since CRAN doesn't natively support this functionality.
+  _API_PACKAGE_URL_POSIT_CRAN = 'https://packagemanager.posit.co/__api__/repos/2/packages/{package}'
+
+  def sort_key(self, version):
```
andrewpollock, line 29: Could you not then subclass `SemverEcosystem` instead of `Ecosystem` making this not necessary?

osv/ecosystems/bioconductor.py
michaelkedar: I guess if Bioconductor is guaranteed to be semver-compliant you can forgo this whole file and just use the `SemverEcoystem` as others do: https://github.com/google/osv.dev/blob/434a7f55b59f186f3f32599a809842c6ce9d4369/osv/ecosystems/_ecosystems.py#L36-L39
https://github.com/google/osv.dev/blob/434a7f55b59f186f3f32599a809842c6ce9d4369/osv/ecosystems/_ecosystems.py#L58-L63

osv/ecosystems/cran.py
michaelkedar: `semver` is probably worse in this case, since it will error on `w.x.y.z`.

osv/ecosystems/cran.py
tylfin: Ah I should've seen this, thank you for the review, I'll switch it over.

osv/ecosystems/cran.py
tylfin: I added some reference links, and additional cases for the typical `X.Y-Z` I see in CRAN. Looks like the `packaging.version.parse(...)` method handles these well and should work moving forward.

michaelkedar: `packaging.version.parse(...)` has some odd behaviour surrounding `-` that would be wrong for R:

```python
from packaging.version import parse
# All these evaluate as True
parse('1.0') < parse('1.0-0')
parse('1.0.0') < parse('1.0-0') 
parse('1.0-0') > parse('1.0-0.0') 
parse('1.0-0') > parse('1.0-0.1')
parse('1.0-0') > parse('1.0-1.0')
```

I think parsing `version.replace('-', '.')` will give the correct comparisons, though.

osv/ecosystems/cran.py
michaelkedar: `SemverEcosystem` doesn't actually get the real package versions, `next_version` just returns the next semver version numerically.

Since Bioconductor doesn't truly use semver (it doesn't allow for prerelease `x.y.z-pre` versions) I think leaving it as is is best.

osv/ecosystems/bioconductor.py
```diff
@@ -0,0 +1,91 @@
...
+  _API_PACKAGE_URL_POSIT_BIOCONDUCTOR = 'https://packagemanager.posit.co/__api__/repos/4/packages/{package}?bioc_version={bioc_version}'
+  _API_BIOC_VERSIONS_URL = 'https://packagemanager.posit.co/__api__/status'
+  _BIOC_VERSIONS = ['3.18', '3.17', '3.16', '3.15', '3.14', '3.13', '3.12', '3.11', '3.10', '3.9', '3.8', '3.7', '3.6', '3.5', '3.4', '3.3', '3.2', '3.1']
+
+  def __init__(self, *args, **kwargs):
+    # Call the init of Ecosystem class
+    super().__init__(*args, **kwargs)
+
+    # Refresh the versions
+    self.refresh_bioc_versions()
```
michaelkedar, line 36: nit: I think it's better to just query these versions every time in `_enumerate_versions`

The ecosystem helper is created only once `_ecosystems.py` is loaded, and the worker process that uses these is long-running, so it may (on rare occasions) miss a Bioconductor version update.

osv/ecosystems/bioconductor.py
```diff
@@ -0,0 +1,86 @@
...
+  _BIOC_VERSIONS = ['3.18', '3.17', '3.16', '3.15', '3.14', '3.13', '3.12', '3.11', '3.10', '3.9', '3.8', '3.7', '3.6', '3.5', '3.4', '3.3', '3.2', '3.1']
+
+  def refresh_bioc_versions(self):
+    """
+    Pull latest Bioconductor versions
+    """
+    response = requests.get(self._API_BIOC_VERSIONS_URL)
+    data = response.json()
+
+    self._BIOC_VERSIONS = [bioc['bioc_version'] for bioc in data['bioc_versions']]
```
michaelkedar, line 38: nit: Since this is now pulled every time, can this method just return the list, and pass it into `_enumerate_versions` (or call it from inside).

Also, could you check the response status code and raise an explicit exception.

osv/ecosystems/cran_test.py
```diff
@@ -0,0 +1,36 @@
...
+    # Test typical semver X.Y.Z version
+    self.assertEqual('0.1.1', ecosystem.next_version('readxl', '0.1.0'))
+    self.assertEqual('1.0.0', ecosystem.next_version('readxl', '0.1.1'))
+
+    with self.assertRaises(ecosystems.EnumerateError):
+      ecosystem.next_version('doesnotexist123456', '1')
+
+    # Test versions with the X.Y-Z format
+    self.assertEqual('0.1-18', ecosystem.next_version('abd', '0.1-12'))
+    self.assertEqual('0.2-2', ecosystem.next_version('abd', '0.1-22'))
```
michaelkedar, line 36: Could you also add some tests for packages with atypical versions e.g. `0-1` or `1.2-3.4` (if you can find any)

michaelkedar (commented):
This LGTM - I'm to approve/merge this when the lints are fixed (as long as the tests pass).
Also, I have a few small comments:

Thanks again for your work!

osv/ecosystems/bioconductor.py
```diff
@@ -0,0 +1,100 @@
...
+  _API_PACKAGE_URL_POSIT_BIOCONDUCTOR = \
+    'https://packagemanager.posit.co/__api__/repos/4/packages/' + \
+    '{package}?bioc_version={bioc_version}'
+  _API_BIOC_VERSIONS_URL = 'https://packagemanager.posit.co/__api__/status'
+
+  def get_bioc_versions(self):
+    """
+    get latest Bioconductor versions
+    """
+    response = requests.get(self._API_BIOC_VERSIONS_UR, timeout=60)
```
michaelkedar, line 38: typo:
```suggestion
    response = requests.get(self._API_BIOC_VERSIONS_URL, timeout=60)
```

osv/ecosystems/bioconductor.py
```diff
@@ -0,0 +1,100 @@
...
+    data = response.json()
+    if response.status_code == 404:
+      raise RuntimeError('Failed to get Bioconducot versions')
```
michaelkedar, line 41: another typo I just noticed:
```suggestion
      raise RuntimeError('Failed to get Bioconductor versions')
```

michaelkedar (approved):
I think this is good to go now - thanks!

I'm going to go ahead and merge this so we can get importing on our [staging site](https://oss-vdb-test.appspot.com/list).

osv/ecosystems/bioconductor_test.py
```diff
@@ -0,0 +1,30 @@
...
+from .. import ecosystems
+
+
+class BioconductorEcosystemTest(unittest.TestCase):
+  """Bioconductor ecosystem helper tests."""
+
+  def test_next_version(self):
+    """Test next_version."""
+    ecosystem = ecosystems.get('Bioconductor')
+    self.assertEqual('1.18.0', ecosystem.next_version('a4', '1.16.0'))
```
andrewpollock, line 27: @tylfin this is starting to fail, but I'm not seeing why?

```
$ GET https://packagemanager.posit.co/__api__/repos/4/packages/a4?bioc_version=1.16.0
{"code":2,"error":"No sources found for Bioconductor 1.16.0","payload":null}
$ GET https://packagemanager.posit.co/__api__/repos/4/packages/a4?bioc_version=1.18.0
{"code":2,"error":"No sources found for Bioconductor 1.18.0","payload":null}
```

Yet https://bioconductor.org/packages/release/bioc/html/a4.html suggests the package still exists?

Also:

```
$ GET https://packagemanager.posit.co/__api__/repos/4/packages/a4?bioc_version=1.50.1
{"code":2,"error":"No sources found for Bioconductor 1.50.1","payload":null}
$ GET https://packagemanager.posit.co/__api__/repos/4/packages/a4?bioc_version=1.50.0
{"code":2,"error":"No sources found for Bioconductor 1.50.0","payload":null}
```

andrewpollock: Sorry, I was driving that wrong, here's what's happening:

```
$ GET https://packagemanager.posit.co/__api__/repos/4/packages/a4?bioc_version=3.19
{"code":4,"error":"Package 'a4' not found","payload":null}
```

It seems to work with all of the earlier versions than 3.19, based on a spot check with the bounds of `bioc_versions`

PR #1482, opened 2023-07-20
Fix renovate's config so language filtering works as intended

andrewpollock:
I cannot find any mention of `matchLanguages` at
https://docs.renovatebot.com/configuration-options and by my reading of https://docs.renovatebot.com/configuration-options/#matchcategories using `matchCategories` does what my understanding of the intention of the use of `matchLanguages` was trying to do.


another-rex:
This is the correct fix, it looks like it was changed 3 weeks ago: https://github.com/renovatebot/renovate/releases/tag/36.0.0

andrewpollock:
> This is the correct fix, it looks like it was changed 3 weeks ago: https://github.com/renovatebot/renovate/releases/tag/36.0.0

I'm surprised they didn't have more of a noisier transition period, and the breaking change is silent?

another-rex:
It is a major version change, which aligns with how they do breaking changes, but (with a very brief search) I haven't found any way to stop renovate-bot itself from auto updating? It might be something managed organisation (google) wide.

andrewpollock:
> It is a major version change, which aligns with how they do breaking changes, but (with a very brief search) I haven't found any way to stop renovate-bot itself from auto updating? It might be something managed organisation (google) wide.

#1489 may help surface future breakage, but potentially only when we touch the config...

another-rex (approved):

PR #1483, opened 2023-07-20
frontend: improve rendering of CVEs with partial package information

andrewpollock:
- Use the repo as the package name when the package name is unavailable
- Don't render an empty anchor when the repo URL is unavailable

h/t @michaelkedar for picking up on these defects


oliverchang (approved):

gcp/appengine/frontend3/src/templates/vulnerability.html
```diff
@@ -74,17 +74,17 @@ <h3 class="mdc-layout-grid__cell--span-3">
           <div class="mdc-layout-grid__cell--span-9">
             <dl>
               <dt>Package Name</dt>
-              <dd>{{ affected['package']['name'] or 'Unknown' }}</dd>
+              <dd>{{ affected['package']['name'] or vulnerability.repo | strip_scheme }}</dd>
```
oliverchang, line 77: Actually, thinking more on this, this doesn't seem very correct since this isn't actually a "Package Name", and it's also a bit repetitive.  IMO we should just strip this heading completely.

oliverchang (changes requested):

gcp/appengine/frontend3/src/templates/vulnerability.html
andrewpollock: Removed in https://github.com/google/osv.dev/pull/1483/commits/f6d276ead199d90463abb8edad478f2b59b82286

michaelkedar (approved):

gcp/appengine/frontend3/src/templates/vulnerability.html
```diff
@@ -3,169 +3,159 @@
...
+              <dd>
+                <div class="mdc-layout-grid__inner events">
+                  {% for event in range['events'] -%}
+                  <div class="mdc-layout-grid__cell--span-3">
+                    {{ event | event_type -}}
+                  </div>
+                  <div class="mdc-layout-grid__cell--span-9 version-value">
+                    {% set link = event | event_link -%}
+                    {% if link -%}
+                    <a href="{{ link }}" }>
```
another-rex, line 91: This `}` seems out of place...

andrewpollock: Well spotted, thanks! Fixed in https://github.com/google/osv.dev/pull/1483/commits/65884fcd8820292d955a1fb10880353808bc9195

gcp/appengine/frontend3/src/templates/vulnerability.html
```diff
@@ -3,169 +3,159 @@
...
       <div class="mdc-layout-grid">
-        <div class="vulnerability-package-subsection mdc-layout-grid__inner">
-          <h3 class="mdc-layout-grid__cell--span-3">
-            Package Info
-          </h3>
-          <div class="mdc-layout-grid__cell--span-9">
-            <dl>
-              <dt>Package Name</dt>
-              <dd>{{ affected['package']['name'] or 'Unknown' }}</dd>
-              <dt>Repo URL</dt>
```
oliverchang, line 78: Sorry my comment was very unclear. I was proposing to strip the individual "Package Name" and "Repo URL" bits only in the case when they do not exist, instead of writing "Unknown" or trying to force a repo name into a package name.

andrewpollock: How about in the interests of getting this merged so the release can proceed, I do that in a followup PR?

andrewpollock: Or feel free to commit exactly what you have in mind if that's sufficiently expedient...

oliverchang: SGTM, I'll approve to unblock.

oliverchang (approved):

PR #1484, opened 2023-07-20
CSS Changes to clean up blog

hayleycd:
Hi team!

I have added some CSS that is only specific to blog posts. I am hoping that it will clean things up a lot. 

I tried running this locally, but our current instructions are not great. I would appreciate being able to see a dev/stage version if that is possible. What would you guys recommend?

### DIFF
--- a/gcp/appengine/blog/hugo.toml
+++ b/gcp/appengine/blog/hugo.toml
@@ -1,3 +1,7 @@
 baseURL = 'http://osv.dev/blog/'
 languageCode = 'en-us'
 title = 'OSV Blog'
+[markup]
+[markup.highlight]
+style = 'native'
+tabWidth = 2
--- a/gcp/appengine/blog/hugo.toml
+++ b/gcp/appengine/blog/hugo.toml
@@ -1,3 +1,5 @@
 baseURL = 'http://osv.dev/blog/'
 languageCode = 'en-us'
 title = 'OSV Blog'
+[markup.goldmark.renderer]
+unsafe= true
--- a/gcp/appengine/blog/hugo.toml
+++ b/gcp/appengine/blog/hugo.toml
@@ -1,5 +1,3 @@
 baseURL = 'http://osv.dev/blog/'
 languageCode = 'en-us'
 title = 'OSV Blog'
-[markup.goldmark.renderer]
-unsafe= true
--- a/gcp/appengine/frontend3/package.json
+++ b/gcp/appengine/frontend3/package.json
@@ -33,7 +33,7 @@
     "sass-loader": "12.6.0",
     "style-loader": "3.3.3",
     "webpack": "5.87.0",
-    "webpack-cli": "4.10.0",
+    "webpack-cli": "^4.10.0",
     "webpack-dev-server": "4.15.1"
   }
 }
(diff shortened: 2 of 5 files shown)


gcp/appengine/blog/hugo.toml
```diff
@@ -1,3 +1,5 @@
 baseURL = 'http://osv.dev/blog/'
 languageCode = 'en-us'
 title = 'OSV Blog'
+[markup.goldmark.renderer]
+unsafe= true
```
hayleycd, line 5: This is needed for hugo to render raw html. It should be ok because we don't have any user-submitted html. https://discourse.gohugo.io/t/raw-html-getting-omitted-in-0-60-0/22032

hayleycd:
I renamed the config.toml file to hugo.toml based on the recommendation here: https://gohugo.io/getting-started/configuration/#hugotoml-vs-configtoml

hayleycd:
We may want to adjust the date depending on when you do the next release.

gcp/appengine/frontend3/src/styles.scss
```diff
@@ -1240,5 +1240,30 @@ dl.vulnerability-details,
...
+      color: black;
+    }
+
+    .API {
+      border-radius: 15px;
+      background: lightgrey;
+      color: black;
+      padding: 15px;
+      overflow-y: hidden;
+      overflow-x: auto;
```
hayleycd, line 1265: This allows us to scroll left-right to see all of the json.

gcp/appengine/frontend3/package.json
```diff
@@ -33,7 +33,7 @@
     "sass-loader": "12.6.0",
     "style-loader": "3.3.3",
     "webpack": "5.87.0",
-    "webpack-cli": "4.10.0",
+    "webpack-cli": "^4.10.0",
```
andrewpollock, line 36: I think technically this change shouldn't be necessary, and was a side-effect from you locally installing `webpack-cli`

I'm too green on how NPM works to know if this is a total no-op or whether it has any implications. Suggest reverting it unless someone with more NPM clue than me says it's benign.

andrewpollock (commented):
Per the out of band chat, I think there's an opportunity to reduce the embedded HTML in favour of pure MarkDown + Hugo-specific CSS styling tweaks (and this might then make the HTML rendering config change unnecessary)

another-rex:
I went a bit overboard and did a small rework of the entire blog post css, changes made:

- Fix backtick indentation
- Lighten inline code styling (similar to your change, but keeps white text instead of switching to black text)
  - Make all inline code styling the same
- reformat SCSS file
- Make a lot of pieces have the same border radius (4px)
- Removed html rendering as a css only solution seems to work
- Make the blog "list" page and each article share the same css to avoid inconsistencies
- Made the links on the blog list page the same blueish colour that we use inside the blog post. (Might be controversial, want some feedback on what people think about this)
- Use the "native" code rendering theme, seems to match pretty well with out site theme: https://xyproto.github.io/splash/docs/longer/native.html
- Indent code with tabs, set tab width to 2 spaces
- Fix mobile scaling/responsive design

Some screenshots and videos:

Screenshot from 2023-07-21 12-20-50
Screenshot from 2023-07-21 12-17-57
Screenshot from 2023-07-21 12-17-43
Screenshot from 2023-07-21 12-05-58

[Screencast from 2023-07-21 12-53-25.webm](https://github.com/google/osv.dev/assets/106129829/93004b8c-d469-4c56-90c6-3e3274ad624b)

oliverchang (approved):
Incredible work here Rex!

andrewpollock (approved):
LGTM. I'm not particularly opinionated about the link colour, it's incredibly subjective IMO.

Issue #892, opened 2022-12-08
Better validation for API requests.


oliverchang:
We can validate:

1. The provided version is valid for the given ecosystem. We can simply do this by attempting to parse the provided version, if an ecosystem is specified. 
2. The provided package and version number exists in the ecosystem (this will need https://deps.dev support).

andrewpollock:
Based on my read of b/291986718, also when an invalid/unsupported ecosystem is provided

PR #1485, opened 2023-07-21
Validate the ecosystem used in queries

andrewpollock:
Partially addresses #892

### DIFF
--- a/gcp/api/server.py
+++ b/gcp/api/server.py
@@ -439,7 +439,7 @@ def determine_version(version_query: osv_service_v1_pb2.VersionQuery,
 def valid_ecosystems():
   """Return the list of ecosystems considered valid."""
   query = osv.Bug.query(projection=[osv.Bug.ecosystem], distinct=True)
-  return [bug.ecosystem[0] for bug in query if bug.ecosystem]
+  yield [bug.ecosystem[0] for bug in query if bug.ecosystem]
 
 
 @ndb.tasklet
--- a/gcp/api/server.py
+++ b/gcp/api/server.py
@@ -454,7 +454,7 @@ def do_query(query, context: QueryContext, include_details=True):
     ecosystem = ''
     purl_str = ''
 
-  if ecosystem and ecosystem not in valid_ecosystems():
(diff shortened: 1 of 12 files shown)


another-rex (commented):
This feels like something that should be cached, even for just for 15 mins or so will save a lot of datastore queries.

gcp/api/server.py
```diff
@@ -447,6 +454,10 @@ def do_query(query, context: QueryContext, include_details=True):
     ecosystem = ''
     purl_str = ''
 
+  if ecosystem and ecosystem not in valid_ecosystems():
```
michaelkedar, line 457: I think we could check `ecosystems.get()` for this instead of querying
https://github.com/google/osv.dev/blob/310c28d0a708f28fcabd96a0d335a278d3fde549/osv/ecosystems/_ecosystems.py#L66

I'm pretty sure we only add things to datastore if we have helpers for them

andrewpollock: Yeah, that works for me and is far easier to implement, thanks for the suggestion. Implemented in https://github.com/google/osv.dev/pull/1485/commits/5c4385bcb5599476c4fa0bab0a2254464e787942 with test coverage in https://github.com/google/osv.dev/pull/1485/commits/36ae72269fb7f22d992736f0f0ad7cd2757a662b and https://github.com/google/osv.dev/pull/1485/commits/4c790de0e8ad8fd627ca9645b7b790be2fc771b5

gcp/api/server.py
```diff
@@ -435,6 +435,13 @@ def determine_version(version_query: osv_service_v1_pb2.VersionQuery,
                                         len(version_query.file_hashes))
 
 
+@ndb.tasklet
+def valid_ecosystems():
```
oliverchang, line 439: Should we delete this now?

oliverchang (approved):

gcp/api/server.py
andrewpollock: Yep! Removed in https://github.com/google/osv.dev/pull/1485/commits/10e894be6f4d9b506f5459ea296592c312f5c053

PR #1488, opened 2023-07-24
Update hugo and remove ^ from version

another-rex:
Attempt to update hugo and see if that helps it build.

### DIFF
--- a/docker/ci/Dockerfile
+++ b/docker/ci/Dockerfile
@@ -16,9 +16,9 @@ FROM gcr.io/oss-vdb/worker
 
 RUN apt-get update && \
     apt-get install -y \
-        wget \
-        google-cloud-sdk-datastore-emulator \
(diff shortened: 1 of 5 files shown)


michaelkedar (approved):
The ci and deployment images need to be manually built & pushed

another-rex:
Hmm can't even build it, I think we need a newer version of go, I'll test that out

andrewpollock:
> Hmm can't even build it, I think we need a newer version of go, I'll test that out

The path of least resistance might be to revert the config file rename for now?

another-rex:
Latest version of go fixed the build issue, but yeah, I think changing the go version might be too risky for a blog post change, and should be separate.

another-rex:
Found a nice middleground, updated hugo to 0.111.3, which is what I have running locally, so I know it works, and it compiles with our current version of go.

oliverchang (approved):

PR #1490, opened 2023-07-24
Add --pull to the new build scripts, and pin pipenv version to 2023.6.12

another-rex:
Add --pull to the new build scripts, and pin pipenv version to 2023.6.12 for the deployment image.

Also remove ^ from package-lock that  was missed in the previous PR.

> We've pinned pipenv because pipenv requirements was behaving differently on the most recent version, and we were matching on the output to remove the requirement from the App Engine deployment:
> 
> in 2023.6.12 the output for the editable osv package was
> 
> -e ./../..
> in 2023.7.23 it was
> 
> ../../

This does not seem to be an intentional change from pipenv, as it's not logged in the patch notes, but from bugs in their new implementation of the requirements: https://github.com/pypa/pipenv/pull/5757, so this might change back in the future versions.

### DIFF
--- a/docker/ci/build.sh
+++ b/docker/ci/build.sh
@@ -14,6 +14,6 @@
 # limitations under the License.
 
 
-docker build -t gcr.io/oss-vdb/ci:$1 -t gcr.io/oss-vdb/ci:latest . && \
-gcloud docker -- push gcr.io/oss-vdb/ci:$1 && \
-gcloud docker -- push gcr.io/oss-vdb/ci:latest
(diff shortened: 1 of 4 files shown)


michaelkedar (approved):
For future reference:
We've pinned `pipenv` because `pipenv requirements` was behaving differently on the most recent version, and we were matching on the output for App Engine:

in 2023.6.12 the output for the editable `osv` package was
```
-e ./../..
```

in 2023.7.23 it was
```
../../
```

PR #1491, opened 2023-07-24
Add retries to remote git repo tag listing

andrewpollock:
This will help for transient failures.

### DIFF
--- a/vulnfeeds/git/repository.go
+++ b/vulnfeeds/git/repository.go
@@ -16,13 +16,17 @@
 package git
 
 import (
+	"context"
 	"sort"
 	"strings"
+	"time"
 
 	"github.com/go-git/go-git/v5"
 	"github.com/go-git/go-git/v5/config"
+	"github.com/go-git/go-git/v5/plumbing"
 	"github.com/go-git/go-git/v5/plumbing/transport"
 	"github.com/go-git/go-git/v5/storage/memory"
+	"github.com/sethvargo/go-retry"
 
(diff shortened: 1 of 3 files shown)


vulnfeeds/git/repository.go
```diff
@@ -61,6 +65,36 @@ type RepoTagsMap struct {
...
+		},
+	}
+	repo := git.NewRemote(memory.NewStorage(), remoteConfig)
+
+	ctx := context.Background()
+
+	backoff := retry.NewExponential(1 * time.Second)
+	backoff = retry.WithMaxRetries(retries, backoff)
+
+	if err := retry.Do(ctx, backoff, func(ctx context.Context) error {
```
oliverchang, line 83: nit: this seems like it could all be simplified as 

```go
err := retry.Do(
   ....)
return refs, err
```

without needing the extra nil checks and repetition.

oliverchang (approved):

vulnfeeds/git/repository.go
andrewpollock: Hmm, I don't think so. There's an inline anonymous function that *only* returns an error, which is required by `retry.Do` and then there's the outer function that returns the refs and the error.

oliverchang: It's possible I'm misreading this, but could we just write this as:

```golang
err = retry.Do(ctx, backoff, func(ctx context.Context) error {
  ... anonymous inline function
});

return refs, err
```

?

PR #1492, opened 2023-07-25
pin pipenv version

another-rex:
Pin the pipenv version to prevent new updates introducing unknown bugs when building new images.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -37,7 +37,7 @@ jobs:
         with:
           python-version: '3.11'
       - name: Install pipenv
-        run: pip install pipenv
+        run: pip install pipenv==2023.6.12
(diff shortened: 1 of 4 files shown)


oliverchang (approved):

andrewpollock (approved):
Random maintainability musing, could we externalise and centralise this somewhere so that everything is guaranteed to stay in lockstep?

michaelkedar (approved):

another-rex:
> Random maintainability musing, could we externalise and centralise this somewhere so that everything is guaranteed to stay in lockstep?

I thought about this, but the work to do that seems tied to the renovate automatically updating thing as well, I think all of that can be bundled into our move away from `pipenv`. It's not urgent since we aren't using any new features in pipenv, so freezing it at one version for a long time should be fine.

Issue #973, opened 2023-01-13
Migrate App Engine to Cloud Run 


oliverchang:
Per https://cloud.google.com/appengine/docs/flexible/cloud-run-for-gae-customers :

> Cloud Run is the latest evolution of Google Cloud Serverless, building on the experience of running App Engine for more than a decade. Cloud Run runs on much of the same infrastructure as App Engine standard environment, so there are many similarities between these two platforms.
> 
> Cloud Run is designed to improve upon the App Engine experience, incorporating many of the best features of both App Engine standard environment and App Engine flexible environment. Cloud Run services can handle the same workloads as App Engine services, but Cloud Run offers customers much more flexibility in implementing these services. This flexibility, along with improved integrations with both Google Cloud and third-party services, also enables Cloud Run to handle workloads that cannot run on App Engine.

We should look at migrating the osv.dev frontend off App Engine, which will also give us the benefit of better homogenising our infra. This has benefits for our deployment and monitoring, as the relevant App Engine bits (versions, cron jobs etc) all live in a separate place from Cloud Run and Kubernetes. 

One consideration is Datastore latency -- we should ensure that there's no negative impact there.

PR #1506, opened 2023-07-26
Deploy App Engine from Docker Image

michaelkedar:
First step on our App Engine to Cloud Run migration #973 

- Made a Docker image for our App Engine website, and moved to App Engine flex to start using it in deployment.
- Experimenting with using [BuildKit](https://docs.docker.com/build/buildkit/) for faster Docker builds.
- I've kept the App Engine deployment workflow the same for now, just added the image building steps.
  - To preserve the current behaviour, images end up duplicated on both `gcr.io/oss-vdb` and `gcr.io/oss-vdb-test`
- I also haven't changed how we run the App Engine locally.

Unfortunately, the App Engine flex/custom deployment (without the docker builds) seems to take a few minutes longer than current, which will make our deployments take a bit longer.

The website performance doesn't seem to be changed: [current](https://oss-vdb-test.wl.r.appspot.com/) vs [docker](https://docker-dot-oss-vdb-test.wl.r.appspot.com/)

There are still some things using App Engine that I haven't looked into for Cloud Run migration:
- The `cron-service` (which doesn't actually get automatically deployed?)
- The static files in the `app.yaml`

### DIFF
--- a/deployment/gae/oss-vdb-test/app.yaml
+++ b/deployment/gae/oss-vdb-test/app.yaml
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-service: docker
+service: default
 runtime: custom
 env: flex
 
--- a/deployment/gae/oss-vdb-test/app.yaml
+++ b/deployment/gae/oss-vdb-test/app.yaml
@@ -12,8 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-service: default
-runtime: python311
+service: docker
+runtime: custom
+env: flex
 
 handlers:
   - url: /static
@@ -39,7 +40,7 @@ inbound_services:
   - warmup
 
 automatic_scaling:
-  min_instances: 1
+  min_num_instances: 1
 
 instance_class: F4
 
--- a/deployment/gae/oss-vdb-test/app.yaml
+++ b/deployment/gae/oss-vdb-test/app.yaml
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-service: default
+service: docker
 runtime: custom
 env: flex
 
@@ -42,7 +42,18 @@ inbound_services:
 automatic_scaling:
   min_num_instances: 1
 
-instance_class: F4
+resources:
+  cpu: 4
+  memory_gb: 3.6
+
+liveness_check:
+  path: "/health"
+
+readiness_check:
+  path: "/"
+  # App Engine Flex runs multiple redundant readiness checks
+  # The default value (5) was hitting our rate limiting and failing...
+  check_interval_sec: 15
 
 env_variables:
   REDISHOST: 10.189.34.180
--- a/gcp/appengine/Dockerfile
+++ b/gcp/appengine/Dockerfile
@@ -1,0 +1,52 @@
+# Build the Javascript frontend
+FROM node:latest AS FRONTEND3_BUILD
+WORKDIR /build/frontend3
+
+# Install dependencies first for better caching
+COPY gcp/appengine/frontend3/package.json gcp/appengine/frontend3/package-lock.json ./
+RUN npm ci
+
+COPY gcp/appengine/frontend3/webpack.prod.js ./
+COPY gcp/appengine/frontend3/img img
+COPY gcp/appengine/frontend3/src src
+
+RUN npm run build:prod
+
+# Build hugo blogs
+# The ci image is a bit heavy for this, but it keeps the hugo version we use in one place
+FROM gcr.io/oss-vdb/ci:latest AS HUGO_BUILD
+WORKDIR /build/blog
+COPY gcp/appengine/blog ./
+
+RUN hugo -d ../dist/static/blog
+
+# OSV.dev site image
+# Adapted from https://cloud.google.com/run/docs/quickstarts/build-and-deploy/deploy-python-service#writing
+FROM python:3.11-slim
+
+# Allow statements and log messages to immediately appear in the logs
+ENV PYTHONUNBUFFERED True
+WORKDIR /osv/gcp/appengine
+
+# Install Python dependencies
+COPY setup.py Pipfile* README.md /osv/
+COPY osv /osv/osv
+COPY gcp/appengine/Pipfile* ./
+RUN pip3 install -U pipenv==2023.6.12 && python3 -m pipenv install --deploy --system
+
+# Website Python code
+COPY gcp/appengine/*.py ./
+
+# JS/hugo builds
+COPY gcp/appengine/dist/public_keys dist/public_keys
+COPY gcp/appengine/docs docs
+COPY --from=FRONTEND3_BUILD /build/dist/ dist
+COPY --from=HUGO_BUILD /build/dist dist
+
+
+# Run the web service on container startup. Here we use the gunicorn
+# webserver, with one worker process and 8 threads.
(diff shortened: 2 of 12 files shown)


another-rex (approved):
Exciting! LGTM, though can you add a build.sh script in the blog page that just runs `hugo -d ../dist/static/blog` since it's no longer in the deploy script.

michaelkedar:
> can you add a build.sh script in the blog page that just runs `hugo -d ../dist/static/blog` since it's no longer in the deploy script.

It already gets run in the [Makefile](https://github.com/google/osv.dev/blob/3aa40c88c8367a23844c2a4ba9735707dec9f47e/Makefile#L33), so I don't know if it's necessary

gcp/appengine/Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+RUN npm ci
+
+COPY gcp/appengine/frontend3/webpack.prod.js ./
+COPY gcp/appengine/frontend3/img img
+COPY gcp/appengine/frontend3/src src
+
+RUN npm run build:prod
+
+# Build hugo blogs
+FROM klakegg/hugo:0.111.3 AS HUGO_BUILD
```
oliverchang, line 16: Are there any alternatives here to relying on this unofficial image?

Since Hugo is a Go program, can we just pull the binary from an official release instead in some minimal image?

deployment/gae/oss-vdb-test/app.yaml
```diff
@@ -13,7 +13,8 @@
 # limitations under the License.
 
 service: default
-runtime: python311
+runtime: custom
+env: flex
```
oliverchang, line 17: Are there any region / location changes as a result of this? 

Also, I assume our redis instances still work re networking ?

oliverchang (commented):
awesome! just some questions.

gcp/appengine/Dockerfile
oliverchang: or even build it from source if it's fast enough. that way we eliminate a few layers of potential compromise.

gcp/appengine/Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+COPY gcp/appengine/*.py ./
+
+# JS/hugo builds
+COPY gcp/appengine/dist/public_keys dist/public_keys
+COPY gcp/appengine/docs docs
+COPY --from=FRONTEND3_BUILD /build/dist/ dist/
+COPY --from=HUGO_BUILD /build/dist dist/
+
+
+# Run the web service on container startup. Here we use the gunicorn
```
andrewpollock, line 49: Could you speak (either here, or at next week's team meeting) to your choice of gunicorn? 

I'm mostly interested to understand how much due diligence has gone into the selection (I'm thinking through the lens of https://best.openssf.org/Concise-Guide-for-Evaluating-Open-Source-Software). 

While this is an interim measure to moving to Cloud Run, I imagine this particular aspect of things won't necessarily be temporary, so I'd like to ensure we're being mindful of the technical debt we're taking on.

gcp/appengine/deploy.sh
```diff
@@ -3,13 +3,16 @@
 dir=$(dirname "$0")
 
 if [ $# -lt 1 ]; then
```
andrewpollock, line 5:
```suggestion
if [[ $# -lt 2 ]]; then
```
I think we're going to be mandating two arguments now, right? Also adhere to https://google.github.io/styleguide/shellguide.html#test----and---

gcp/appengine/deploy.sh
```diff
@@ -21,17 +24,14 @@ for config in $configs_dir/*.yaml; do
 done
 
 cd "$dir"
+image="gcr.io/$project_id/osv-website"
+
+if [ -z "$CLOUDBUILD" ]; then
```
andrewpollock, line 29:
```suggestion
if [[ -z "$CLOUDBUILD" ]]; then
```
Adhere to https://google.github.io/styleguide/shellguide.html#test----and---

gcp/appengine/deploy.sh
```diff
@@ -21,17 +24,14 @@ for config in $configs_dir/*.yaml; do
...
+  pushd ../../
+  # Using BuildKit allows us to cache the multi-stage builds
+  DOCKER_BUILDKIT=1 docker build --build-arg=BUILDKIT_INLINE_CACHE=1 -t $image:$version -f gcp/appengine/Dockerfile --cache-from=$image .
```
andrewpollock, line 32:
```suggestion
  DOCKER_BUILDKIT=1 docker build --build-arg=BUILDKIT_INLINE_CACHE=1 -t "${image}:${version}" -f gcp/appengine/Dockerfile --cache-from="${image}" .
```
Comply with:
- https://google.github.io/styleguide/shellguide.html#variable-expansion
- https://google.github.io/styleguide/shellguide.html#quoting

gcp/appengine/deploy.sh
```diff
@@ -21,17 +24,14 @@ for config in $configs_dir/*.yaml; do
...
-
-pushd blog
-hugo -d ../dist/static/blog
-popd
-
-# Skip the '-e' editable library install as we copy in the "osv" library
-# directly instead for deployment.
-python3 -m pipenv requirements | grep -v '^-e '  > requirements.txt
-gcloud app deploy --quiet --project=$project_id --version=$(git rev-parse HEAD) "$@"
+gcloud app deploy --quiet --project=$project_id --version=$version --image-url=$image:$version "$@"
```
andrewpollock, line 37:
```suggestion
gcloud app deploy --quiet --project="${project_id}" --version="${version}" --image-url="${image}:${version}" "$@"
```
Comply with:
- https://google.github.io/styleguide/shellguide.html#variable-expansion
- https://google.github.io/styleguide/shellguide.html#quoting

andrewpollock:
> The website performance doesn't seem to be changed: [current](https://oss-vdb-test.wl.r.appspot.com/) vs [docker](https://docker-dot-oss-vdb-test.wl.r.appspot.com/)

Do you happen to know if any of the metrics we're using for SLO monitoring will require adjusting as part of this change?

gcp/appengine/Dockerfile
michaelkedar: I've changed it to build from source within the image, which takes some time but *should* be cached.
I could also just use the ci image, which builds Hugo (or make a bespoke Hugo image) if the caching doesn't work as intended.

gcp/appengine/Dockerfile
michaelkedar: I only chose gunicorn because App Engine standard uses it under the hood for our current website. (And also because it's used in [this quickstart](https://cloud.google.com/run/docs/quickstarts/build-and-deploy/deploy-python-service#writing))

I hadn't put much thought into it otherwise.

michaelkedar:
I've been running into some issues with App Engine Flex:
1. Deployment, promotion, and rollbacks all take a long time
2. There's downtime/server errors between promotions

Instead of trying to deal with these App Engine issues, it'd be better to just wholesale move to Cloud Run, so I'm going to close this PR for now. I'll probably use this branch to work on the Cloud Run migration, since the Dockerfile should be useful for that.

rjerrems:
I think this is the right call, App Engine Flex is a bit slower and cumbersome on a number of dimensions so if your going to the effort of containerising might as well go to Run.

> Instead of trying to deal with these App Engine issues, it'd be better to just wholesale move to Cloud Run, so I'm going to close this PR for now. I'll probably use this branch to work on the Cloud Run migration, since the Dockerfile should be useful for that.

PR #1514, opened 2023-07-31
CVE output for consolidation pipeline

andrewpollock:
This way it can integrate with the consolidation pipeline and merge with CVEs converted from Alpine.

Also address some go-staticcheck findings around deprecated libraries and error string case and make some other minor readability improvements.

Rearrange `vulns` package so struct methods appear correctly in the outline view of VS Code.


vulnfeeds/cpp/main.go
```diff
@@ -206,63 +208,133 @@ func ReposForCPE(CVE string, cache VendorProductToRepoMap, vp VendorProduct, ref
...
+
+	vulnDir := filepath.Join(directory, CPE.Vendor, CPE.Product)
+	err = os.MkdirAll(vulnDir, 0755)
+	if err != nil {
+		Logger.Warnf("Failed to create dir: %v", err)
+		return fmt.Errorf("failed to create dir: %v", err)
+	}
+
+	outputFile := filepath.Join(vulnDir, CVEID+".nvd"+extension)
+	notesFile := filepath.Join(vulnDir, CVEID+".nvd.notes")
```
oliverchang, line 309: Question: Are notes actually needed for this use case?

oliverchang (approved):

vulnfeeds/cpp/main.go
andrewpollock: > Question: Are notes actually needed for this use case?

Probably not in the long run, I'm keeping them around for now so they're available for manual human inspection (like when I'm running things locally) but I can see them getting removed in the near future.

PR #1530, opened 2023-08-03
fix: Update make command and point contributing docs at ADC.

rjerrems:
Update make command and docs so that you no longer need to supply a service account credential to run the API server locally.

Add mention of NodeJS and version.


Makefile
```diff
@@ -42,7 +42,10 @@ run-appengine-staging:
 
 run-api-server:
 	test $(SERVICE_ACCOUNT) || (echo "SERVICE_ACCOUNT variable not set"; exit 1)
-	cd gcp/api && pipenv sync && GOOGLE_CLOUD_PROJECT=oss-vdb pipenv run python test_server.py $(SERVICE_ACCOUNT)
+	# Required for ESP container to be able to read the credential, the configured alpine image doesn't respect the volume mapping args.
+	chmod 644 $(SERVICE_ACCOUNT)
+	# This will change permissions back to secure default if the program is interrupted or exits.
+	bash -c "trap 'chmod 600 $(SERVICE_ACCOUNT)' EXIT; cd gcp/api && pipenv sync && GOOGLE_CLOUD_PROJECT=oss-vdb pipenv run python test_server.py $(SERVICE_ACCOUNT)"
```
andrewpollock, line 48:
```suggestion
	bash -c "trap 'chmod 600 $(CREDENTIAL)' EXIT; cd gcp/api && pipenv sync && GOOGLE_CLOUD_PROJECT=oss-vdb pipenv run python test_server.py $(CREDENTIAL)"
```
Continuing to call it SERVICE_ACCOUNT is likely to cause confusion, let's not.

andrewpollock (changes requested):
Mostly LGTM, with some nits around readability and OpSec

Makefile
```diff
@@ -42,7 +42,10 @@ run-appengine-staging:
 
 run-api-server:
 	test $(SERVICE_ACCOUNT) || (echo "SERVICE_ACCOUNT variable not set"; exit 1)
-	cd gcp/api && pipenv sync && GOOGLE_CLOUD_PROJECT=oss-vdb pipenv run python test_server.py $(SERVICE_ACCOUNT)
+	# Required for ESP container to be able to read the credential, the configured alpine image doesn't respect the volume mapping args.
+	chmod 644 $(SERVICE_ACCOUNT)
```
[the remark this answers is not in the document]
rjerrems: Yeah this is a good point, I tried quite a few variations before I got here but this syntax:
`-v "$HOME/.config/gcloud:/gcp/config:ro"` does not seem to work with this particular container image. 

I was trying to avoid modifying the image, but that is the only alternative I can think of?  We would create our own image that adds the dependencies required to do that and/or adds the credentials themselves to the image, which would mean copies of those creds would live in docker images locally accessible to anybody in the docker group.

I'll see if I can amend the image without having to copy the creds in as that seems the least bad option

rjerrems:
I have fixed the issue around permissions by creating a temporary copy of the esp container with our own user and group that now allows us to map in files without changing the underlying permissions. I will do a PR on that repo as well and once that is approved/released we can remove our custom container.

gcp/api/Dockerfile.esp
```diff
@@ -0,0 +1,22 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+FROM gcr.io/endpoints-release/endpoints-runtime:2
+
+USER root
+
+RUN addgroup -g 1000 -S osv && \
```
andrewpollock, line 19: Please add a comment explaining why this is necessary.

andrewpollock (approved):

rjerrems:
I have added a comment explaining the configuration. I also changed the contributing guide so you don't have to supply the path since this make command does not seem to be referenced anywhere in our builds and seems to be for local use.

andrewpollock (approved):

PR #1531, opened 2023-08-03
Mark disputed CVEs as withdrawn

andrewpollock:
This consults the CVE 5.0 record in the CVE List, because the 4.0 schema in the NVD doesn't have a machine-readable way to determine dispute status.

### DIFF
--- a/vulnfeeds/vulns/vulns.go
+++ b/vulnfeeds/vulns/vulns.go
@@ -15,17 +15,35 @@
 package vulns
 
 import (
+	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
+	"net/http"
 	"net/url"
 	"strings"
 	"time"
 
+	"golang.org/x/exp/slices"
 	"gopkg.in/yaml.v2"
 
 	"github.com/google/osv/vulnfeeds/cves"
+	"github.com/sethvargo/go-retry"
 )
+
+const CVEListBaseURL = "https://raw.githubusercontent.com/CVEProject/cvelistV5/main/cves"
+
+var ErrVulnNotACVE = errors.New("not a CVE")
+
+type VulnsCVEListError struct {
+	URL string
+	Err error
+}
+
+func (e *VulnsCVEListError) Error() string {
+	return e.URL + ": " + e.Err.Error()
+}
 
 type Event struct {
 	Introduced   string `json:"introduced,omitempty" yaml:"introduced,omitempty"`
@@ -173,6 +191,7 @@ type Reference struct {
 
 type Vulnerability struct {
 	ID         string      `json:"id" yaml:"id"`
+	Withdrawn  string      `json:"withdrawn,omitempty" yaml:"modified,omitempty"`
 	Summary    string      `json:"summary,omitempty" yaml:"summary,omitempty"`
 	Severity   []Severity  `json:"severity,omitempty" yaml:"severity,omitempty"`
 	Details    string      `json:"details" yaml:"details"`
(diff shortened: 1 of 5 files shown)


vulnfeeds/vulns/vulns.go
```diff
@@ -275,6 +294,82 @@ func (v *Vulnerability) AddSeverity(CVEImpact cves.CVEImpact) {
 	v.Severity = append(v.Severity, severity)
 }
 
+// MarkDisputedCVEWithdrawn will set the Vulnerability "withdrawn" field based on the CVEList tag, if determinable.
+func (v *Vulnerability) MarkDisputedCVEWithdrawn() error {
```
oliverchang, line 298: This feels a bit weird to be a method on the actual vulnerability. It has a lot of external dependencies (i.e. CVE 5, github). The naming also isn't very clear -- it seems to imply it's marking it as withdrawn in all cases. 

Can we instead make this a separate function:

```go
func GetCVEWithdrawn(id string) (string, error) {
}
```

And then call that to explicitly set the withdrawn field based on hte results of that?

andrewpollock: Let's unpack this feedback before I proceed further:

> This feels a bit weird to be a method on the actual vulnerability

How so? It's a conditional vulnerability mutation method. Is it more idiomatic Go to have that as a standalone mutation function instead of a method of the object? 

> It has a lot of external dependencies (i.e. CVE 5, github).

This is why I've opted to make it as defensively coded as possible, and treat its success as "best effort" until we have an established track record. 

An alternative implementation could be to keep a local fresh clone and interrogate that, but that seemed like it required more ongoing supporting infrastructure, so I elected to go down this path instead. Doing that *would* also address @another-rex's need to be able to retrieve CVE descriptions.

> The naming also isn't very clear -- it seems to imply it's marking it as withdrawn in all cases.

That's a very easy fix. Perhaps `MarkWithdrawnIfCVEIsDisputed` instead?

> Can we instead make this a separate function

Yes. My thoughts are that this should return the timestamp to set the `withdrawn` field to if the CVE is disputed, and an error to check if the dispute state was successfully determined. Alternatively, a struct and an error:

```go
type GetCVEWithdrawnStatus struct {
  Disputed bool
  LastUpdated string // or a read-to-use for Withdrawn time.Time
}
``` 

I'd like to clarify how you'd like me to proceed before I do.

oliverchang: > How so? It's a conditional vulnerability mutation method. Is it more idiomatic Go to have that as a standalone mutation function instead of a method of the object?

This is just around a guiding design principle of making pieces more modular and scoped to doing a single thing well (i.e. unix philosophy). a `Vulnerability` should have low level methods on it that let others manipulate its values, but not necessarily encode complicated domain-specific knowledge such as fetching CVE metadata and making decisions based on withdrawn status. Tying all of these together will make all 

> An alternative implementation could be to keep a local fresh clone and interrogate that, but that seemed like it required more ongoing supporting infrastructure, so I elected to go down this path instead. 

The suggested diff is more minimal -- just break it out into a separate function rather than making this a method on a `Vulnerability` object. Your current approach seems fine to me, and is not part of my feedback. 

> Yes. My thoughts are that this should return the timestamp to set the withdrawn field to if the CVE is disputed, and an error to check if the dispute state was successfully determined.

This SGTM.

andrewpollock: Addressed in [24b1272](https://github.com/google/osv.dev/pull/1531/commits/24b127294fdba24538e1faca640ba55bdc7fb03e)

oliverchang (approved):

vulnfeeds/vulns/vulns.go
```diff
@@ -275,6 +294,82 @@ func (v *Vulnerability) AddSeverity(CVEImpact cves.CVEImpact) {
...
+			}
+			return &VulnsCVEListError{CVEListURL, err}
+		}
+		if res.StatusCode/100 == 5 {
+			return retry.RetryableError(fmt.Errorf("bad response: %v", res.StatusCode))
+		}
+		if res.Body != nil {
+			defer res.Body.Close()
+		}
+		CVERawJSON, err = io.ReadAll(res.Body)
```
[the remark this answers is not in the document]
andrewpollock: Addressed in [9b9deaa](https://github.com/google/osv.dev/pull/1531/commits/9b9deaa34869f83bad734ca27954283595dd8f1a)

andrewpollock:
This currently fails CI due to `url.JoinPath` being a Go 1.19-ism and the CI image having Go 1.18. I will rebuild the CI image on Monday with https://github.com/google/osv.dev/pull/1532 and then be able to merge this.

PR #1536, opened 2023-08-08
Make the CVE consolidation hermetic to staging

andrewpollock:
After much poking around, I realised that while combine-to-osv was writing to the staging GCS bucket, it was *reading* from the production one. This was made harder to track down by an incorrect variable being used in the logged output. Also use `set -u` to make incorrect variable usage more self-evident.

### DIFF
--- a/deployment/clouddeploy/gke-workers/environments/oss-vdb-test/combine-to-osv.yaml
+++ b/deployment/clouddeploy/gke-workers/environments/oss-vdb-test/combine-to-osv.yaml
@@ -12,5 +12,7 @@ spec:
             env:
             - name: GOOGLE_CLOUD_PROJECT
               value: oss-vdb-test
+            - name: INPUT_GCS_BUCKET
+              value: osv-test-cve-osv-conversion
             - name: OUTPUT_GCS_BUCKET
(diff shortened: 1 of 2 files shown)


deployment/clouddeploy/gke-workers/environments/oss-vdb-test/combine-to-osv.yaml
```diff
@@ -12,5 +12,7 @@ spec:
             env:
             - name: GOOGLE_CLOUD_PROJECT
               value: oss-vdb-test
+            - name: INPUT_GCS_BUCKET
```
michaelkedar, line 15: Probably want to set this in the production `oss-vdb/combine-to-osv.yaml` also

andrewpollock: It defaults when unset: https://github.com/google/osv.dev/blob/38a28cd7123e13f90206872cc2e9af57959cabfa/vulnfeeds/cmd/combine-to-osv/run_combine_to_osv_convert.sh#L11

michaelkedar (approved):

PR #1557, opened 2023-08-11
Throw even more CPU at the problem

andrewpollock:
Now that I (slightly) better understand Kubernetes, throw even more CPU at the problem.

(The GCS file transfer is no longer exceptionally slow, so I'm not sure how much this will help, but hey, let's find out)

### DIFF
--- a/deployment/clouddeploy/gke-workers/base/combine-to-osv.yaml
+++ b/deployment/clouddeploy/gke-workers/base/combine-to-osv.yaml
@@ -23,7 +23,7 @@ spec:
             resources:
               requests:
                 cpu: 30
-                memory: "1G"
+                memory: "4G"
               limits:
                 cpu: 30
                 memory: "8G"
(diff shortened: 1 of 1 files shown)


another-rex (dismissed):

deployment/clouddeploy/gke-workers/base/combine-to-osv.yaml
```diff
@@ -22,10 +22,10 @@ spec:
               privileged: true
             resources:
               requests:
-                cpu: 2
+                cpu: 30
                 memory: "1G"
               limits:
-                cpu: 2
+                cpu: 30
                 memory: "2G"
```
another-rex, line 29:
```suggestion
                memory: "8G"
```

Let's just bump memory up just in case we get bottlenecked on this, we are using a high mem node pool so this should not be a problem.

oliverchang (approved):

deployment/clouddeploy/gke-workers/base/combine-to-osv.yaml
```diff
@@ -22,10 +22,10 @@ spec:
               privileged: true
             resources:
               requests:
-                cpu: 2
+                cpu: 30
                 memory: "1G"
```
another-rex, line 26:
```suggestion
                memory: "4G"
```

PR #1561, opened 2023-08-15
Mirror OSV.dev website on Cloud Run in staging instance

michaelkedar:
Working on migration #973, partially repurposed from #1506

Start mirroring the website onto Cloud Run on the staging environment only.
Including:
- New website Docker image & building in `build-and-stage`
- Temporary Website Cloud Deploy pipeline on `oss-vdb-test`
- Temporary Terraform definitions for the website Cloud Run service

Still need to look into:
- AppEngine cron stuff
- Check if all AppEngine static files exist in Cloud Run
- Compression?
- Cloud Load Balancing for domain mapping
- rename/move the `gcp/appengine` folder


gcp/appengine/Dockerfile
```diff
@@ -0,0 +1,56 @@
...
+COPY --from=FRONTEND3_BUILD /build/dist/ dist/
+COPY --from=HUGO_BUILD /build/dist dist/
+
+
+# Run the web service on container startup. Here we use the gunicorn
+# webserver, with one worker process and 8 threads.
+# For environments with multiple CPU cores, increase the number of workers
+# to be equal to the cores available.
+# Timeout is set to 0 to disable the timeouts of the workers to allow Cloud Run to handle instance scaling.
+CMD exec gunicorn --bind :$PORT --workers 1 --threads 8 --timeout 0 main:app
```
andrewpollock, line 56: If I recall correctly, there hadn't been a whole lot of rigour around the suitability of the selection of gunicorn for the serving stack, right?

If https://www.fullstackpython.com/wsgi-servers.html is to be considered authoritative, uWSGI may be the most performant choice. I definitely think we should be running the final choice through https://best.openssf.org/Concise-Guide-for-Evaluating-Open-Source-Software

gcp/appengine/handlers.py
```diff
@@ -13,6 +13,9 @@
 # limitations under the License.
 """OSV cron handlers."""
 
+# TODO(michaelkedar): Cloud Run equivalents need to be made for these.
+# This cron logic should probably be removed from the website entirely.
```
andrewpollock, line 17:
```suggestion
# This cron logic should probably be removed from the website entirely.
# https://github.com/google/osv.dev/issues/1249
```
Add existing tracking issue.

andrewpollock (approved):
LGTM to the best of my understanding.

gcp/appengine/Dockerfile
rjerrems: It seems uWSGI is in maintenance mode FYI (at least according to their docs)

another-rex (approved):

gcp/appengine/Dockerfile
another-rex: I think gunicorn is the default appengine choice, so at least this maintains the same behavior as before. Agreed with going through some basic evaluation before moving this to production though.

rjerrems (approved):
LGTM - Happy to help with GCLB + NEG if needed

gcp/appengine/Dockerfile
andrewpollock: > It seems uWSGI is in maintenance mode FYI (at least according to their docs)

Is that a bad thing? Less changes == less bugs == better reliability (in theory)

rjerrems: > Is that a bad thing? Less changes == less bugs == better reliability (in theory)

That's certainly one way to look at it and can be true in some situations. There is some subjectivity about what maintenance mode means depending on the project & maintainers.

Rightly/wrongly, I would usually interpret it as less effort being put into the software in general and possibly delayed response for things that need to be worked on (bugs/security vulns etc)

I think it also means that shiny new protocols and features are less likely to be supported, which can be both good and bad depending on the situation.  Worth evaluating against the criteria you shared to see where it ends up.

PR #1510, opened 2023-07-27
Refactor cves.AffectedVersions

andrewpollock:
Store commits in a single array like versions are stored.

With PR https://github.com/google/osv.dev/pull/1508 making combine-to-osv use `cves.VersionInfo`, this makes the
struct more compact, as it's grown organically to be messier than it needs to be.

PR #1568, opened 2023-08-18
Fix logic error with preferencing predetermined fix commits

andrewpollock:
The refactoring from #1510 introduced an incorrect logic inversion, and records were failing to be converted successfully due to no commits being found. They also weren't being added to what was returned, when they should have been.

Output additional logging to assist with following the flow for troubleshooting purposes.


andrewpollock:
Before:

```
Metrics: {TotalCVEs:13287 CVEsForApplications:9632 CVEsForKnownRepos:4300 OSVRecordsGenerated:2734 Outcomes:map[]}
```

After:

```
Metrics: {TotalCVEs:13287 CVEsForApplications:9632 CVEsForKnownRepos:4300 OSVRecordsGenerated:3118 Outcomes:map[]} 
```

oliverchang (approved):

PR #1571, opened 2023-08-20
Tolerate bugs with mixed range types

andrewpollock:
Converted CVE records (without packages) are merged in with records that do have package fields, so the package field is not reliably existent for all items in a given record's `affected` array.

Also add support to outlink to a package in its ecosystem.

### DIFF
--- a/gcp/appengine/frontend3/src/templates/list.html
+++ b/gcp/appengine/frontend3/src/templates/list.html
@@ -97,20 +97,15 @@
             </span>
             <span role="cell" class="vuln-table-cell vuln-packages mdc-data-table__cell">
               <ul class="packages">
-                {%- if vulnerability.affected | map(attribute='package', default=[]) | list == [{}] -%}
-                {%- for repo in vulnerability.affected | git_repo | map('strip_scheme') -%}
-                {%- if loop.index < 8 -%} <li>{{ repo }}</li>
-                  {%- elif loop.index == 8 -%}
-                  <li>...</li>
-                  {%- endif -%}
-                  {%- else -%}
-                  <li>See details.</li>
-                  {%- endfor -%}
-                  {%- else -%}
-                  {%- for affected in vulnerability.affected -%}
-                  <li>{{ affected.package.ecosystem }}/{{ affected.package.name }}</li>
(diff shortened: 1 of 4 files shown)


gcp/appengine/frontend3/src/templates/vulnerability.html
```diff
@@ -60,12 +60,18 @@ <h2 class="title">Affected packages</h2>
     <spicy-sections
       class="vulnerability-packages{% if vulnerability.affected|should_collapse %} force-collapse{% endif %}">
       {% for affected in vulnerability.affected -%}
+      {% if 'package' in affected %}
```
another-rex, line 63: nit: Can this be indented? Or will the formatter always unindent it.

gcp/appengine/frontend3/src/templates/list.html
```diff
@@ -97,20 +97,15 @@ <h1 class="title">Vulnerability Library</h1>
...
-                  {%- else -%}
-                  <li>See details.</li>
-                  {%- endfor -%}
-                  {%- else -%}
-                  {%- for affected in vulnerability.affected -%}
-                  <li>{{ affected.package.ecosystem }}/{{ affected.package.name }}</li>
-                  {%- endfor -%}
-                  {%- endif -%}
+                {% for affected in vulnerability.affected %}
+                {%- for range in affected.ranges -%}
```
another-rex, line 101: Same with the indentation here.

another-rex (commented):
LGTM, just some comments about indentation.

gcp/appengine/frontend3/src/templates/vulnerability.html
andrewpollock: > nit: Can this be indented? Or will the formatter always unindent it.

The formatter very helpfully unindented it all for me after I'd made it way more readable at the time 🤬

gcp/appengine/frontend3/src/templates/list.html
andrewpollock: I would if I could...

another-rex (approved):

PR #1572, opened 2023-08-21
Fix: Give cloud build service account access so we can remove esp-test

rjerrems:
Give cloud build default service account access so we can remove esp-test.

### DIFF
--- a/deployment/terraform/modules/osv/gcp_apis.tf
+++ b/deployment/terraform/modules/osv/gcp_apis.tf
@@ -91,3 +91,9 @@ resource "google_project_service" "resource_manager" {
   service            = "cloudresourcemanager.googleapis.com"
   disable_on_destroy = false
 }
+
+resource "google_project_service" "cloud_build" {
+  project            = var.project_id
+  service            = "cloudbuild.googleapis.com"
+  disable_on_destroy = false
+}
--- a/deployment/terraform/modules/osv/project_iam.tf
+++ b/deployment/terraform/modules/osv/project_iam.tf
@@ -1,0 +1,12 @@
+# Used to dynamically retrieve the project number.
+data "google_project" "vdb" {
+  project_id = var.project_id
+}
+
+# Give the cloud build in-built service account access to retrieve cloud endpoints config.
+resource "google_project_iam_member" "project" {
+  project    = var.project_id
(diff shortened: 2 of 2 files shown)


andrewpollock (approved):

andrewpollock:
a) it already (manually) has this role (interestingly, one cannot interrogate the IAM policies for 
Google-managed service accounts with `gcloud` it seems, `gcloud --project=oss-vdb  iam service-accounts describe 651737493649@cloudbuild.gserviceaccount.com` fails miserably), because I'd been down this path previously
b) I'm a little bit torn about whether we should manage all the IAM stuff the Right Way (i.e. Latchkey) or not. We went down this path briefly with the staging project and then walked it back, but the reasoning wasn't captured on the related changelist. @oliverchang may be better able to remember specifically why...

rjerrems:
Yeah I suppose the other question is, do we want to support other people to be able to spin up OSV using this terraform? If the answer is yes, we will need to provide the code to setup the necessary permissions etc

michaelkedar (approved):
Does this permission need to exist for both projects (`oss-vdb-test` and `oss-vdb`)?
I think our CI stuff only runs from the `oss-vdb` project.

> Yeah I suppose the other question is, do we want to support other people to be able to spin up OSV using this terraform? If the answer is yes, we will need to provide the code to setup the necessary permissions etc

One thing I'm not sure on is how much should be captured here i.e. do we need to be able to reproduce our Cloud Build triggers & other CI/CD pipeline stuff? It's especially awkward to capture with it being cross-project between staging and production.

rjerrems:
> Does this permission need to exist for both projects (oss-vdb-test and oss-vdb)?
> I think our CI stuff only runs from the oss-vdb project.

This is a good question, my 2c is that we should make the test and prod projects identical and then move the build/deployment dependencies into another project/s, granting permissions to do the deployments etc in the other projects.

> One thing I'm not sure on is how much should be captured here i.e. do we need to be able to reproduce our Cloud Build > triggers & other CI/CD pipeline stuff? It's especially awkward to capture with it being cross-project between staging and > production.

I think this comes back to my earlier question, if we want to support others to be able to run there own OSV in GCP then we should invest the time, but if that is not a priority then maybe not. I do think separating build/deployment from the OSV stack would help solve part of this problem regardless.

PR #1582, opened 2023-08-23
Remove service account download from cloudbuild

rjerrems:
Once we have done this, we should be able to throw the service account onto the bonfire.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ vulnfeed-tests:
 	cd vulnfeeds && ./run_tests.sh
 
 api-server-tests:
-	test $(HOME)/.config/gcloud/application_default_credentials.json || (echo "GCP Application Default Credentials not set."; exit 1)
+	test $(HOME)/.config/gcloud/application_default_credentials.json || (echo "GCP Application Default Credentials not set, try 'gcloud auth login --update-adc'"; exit 1)
 	cd gcp/api && ./run_tests.sh $(HOME)/.config/gcloud/application_default_credentials.json
 
 lint:
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,10 @@ appengine-tests:
 vulnfeed-tests:
 	cd vulnfeeds && ./run_tests.sh
 
+api-server-tests:
+	test $(HOME)/.config/gcloud/application_default_credentials.json || (echo "GCP Application Default Credentials not set."; exit 1)
+	cd gcp/api && ./run_tests.sh $(HOME)/.config/gcloud/application_default_credentials.json
+
 lint:
 	tools/lint_and_format.sh
 
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ vulnfeed-tests:
 	cd vulnfeeds && ./run_tests.sh
 
 api-server-tests:
-	test $(HOME)/.config/gcloud/application_default_credentials.json || (echo "GCP Application Default Credentials not set, try 'gcloud auth login --update-adc'"; exit 1)
+	test -f $(HOME)/.config/gcloud/application_default_credentials.json || (echo "GCP Application Default Credentials not set, try 'gcloud auth login --update-adc'"; exit 1)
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -27,6 +27,10 @@ appengine-tests:
 vulnfeed-tests:
 	cd vulnfeeds && ./run_tests.sh
 
+api-server-tests:
+	test $(HOME)/.config/gcloud/application_default_credentials.json || (echo "GCP Application Default Credentials not set."; exit 1)
```
andrewpollock, line 31:
```suggestion
	test $(HOME)/.config/gcloud/application_default_credentials.json || (echo "GCP Application Default Credentials not set, try 'gcloud auth login --update-adc'"; exit 1)
```
Add a little more of a hint on how to resolve

gcp/api/run_tests.sh
```diff
@@ -14,14 +14,13 @@
 # limitations under the License.
 
 if [ $# -lt 1 ]; then
-  echo "Usage: $0 /path/to/service_account.json"
+  echo "Usage: $0 /path/to/credential.json"
```
andrewpollock, line 17: Is this any of this still necessary at all, given the use of $1 was removed?

cloudbuild.yaml
```diff
@@ -27,17 +27,12 @@ steps:
...
   dir: vulnfeeds
   args: ['bash', '-ex', 'run_tests.sh']
-- name: 'gcr.io/cloud-builders/gcloud'
-  entrypoint: 'bash'
-  args: [ '-c', "gcloud secrets versions access latest --secret=integration-test-account --format='get(payload.data)' | tr '_-' '/+' | base64 -d > /workspace/service_account.json" ]
 - name: 'gcr.io/oss-vdb/ci'
   dir: gcp/api
-  args: ['bash', '-ex', 'run_tests.sh', '/workspace/service_account.json']
+  #TODO: Update test scripts to support not supplying a credential.
+  args: ['bash', '-ex', 'run_tests.sh', '/workspace/dummy.json']
```
andrewpollock, line 33: Similar to my comment on run_tests, if you remove the requirement this TODO isn't necessary (or am I missing something)

andrewpollock (commented):
Please update the description to explain why you're making this change.

gcp/api/run_tests.sh
rjerrems: So this is still useful locally as it allows you to run the tests using ADC. Both the `test_server.py` and `integration_tests.py` expect the credential path as an argument when run locally to map the credentials into the container. We could potentially refactor this to remove the argument but I was avoiding doing it in this PR, but could have my arm twisted otherwise

cloudbuild.yaml
rjerrems: See comment above and let me know what you think

Makefile
```diff
@@ -27,6 +27,10 @@ appengine-tests:
 vulnfeed-tests:
 	cd vulnfeeds && ./run_tests.sh
 
+api-server-tests:
+	test $(HOME)/.config/gcloud/application_default_credentials.json || (echo "GCP Application Default Credentials not set, try 'gcloud auth login --update-adc'"; exit 1)
+	cd gcp/api && ./run_tests.sh $(HOME)/.config/gcloud/application_default_credentials.json
```
michaelkedar, line 32:
```suggestion
	test -f ${HOME}/.config/gcloud/application_default_credentials.json || (echo "GCP Application Default Credentials not set, try 'gcloud auth login --update-adc'"; exit 1)
	cd gcp/api && ./run_tests.sh ${HOME}/.config/gcloud/application_default_credentials.json
```

(I don't actually know bash too well, I was just trying to manually run the commands on my machine)

rjerrems: Good call that is more specific

oliverchang (approved):

michaelkedar (approved):

gcp/api/run_tests.sh
andrewpollock: I don't need to be twisting arms :-)

andrewpollock (approved):

PR #1599, opened 2023-08-30
Ease fuzzy tag matching by removing repo name

andrewpollock:
There's a particularly hairy fuzzy tag matching scenario that doesn't lend itself easily to the regex, where the repo name ends in a digit and the tags have this in their prefix. It leads to incorrect normalization of the tag. So try and identify such cases and preemptively remove the name-based prefix from the tags to be normalized.

### DIFF
--- a/vulnfeeds/git/repository.go
+++ b/vulnfeeds/git/repository.go
@@ -17,6 +17,8 @@ package git
 
 import (
 	"context"
+	"net/url"
+	"path"
 	"sort"
 	"strings"
 	"time"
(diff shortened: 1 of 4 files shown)


andrewpollock:
A trial conversion run with this code now successfully converts CVE-2023-2597, which was a failing scenario that surfaced this edge case

vulnfeeds/git/repository.go
```diff
@@ -151,13 +164,21 @@ func NormalizeRepoTags(repoURL string, repoTagsCache RepoTagsCache) (NormalizedT
...
 	tags, err := RepoTags(repoURL, repoTagsCache)
 	if err != nil {
 		return nil, err
 	}
 	NormalizedTags = make(map[string]NormalizedTag)
 	for _, t := range tags {
-		normalizedTag, err := cves.NormalizeVersion(t.Tag)
+		// Opportunistically remove parts determined to match the repo name,
+		// to ease particularly difficult to normalize cases like 'openj9-0.38.0'.
+		prenormalizedTag := strings.TrimPrefix(t.Tag, assumedReponame)
```
oliverchang, line 179: Can we do this in a case insensitive way?

oliverchang (approved):

oliverchang:
Also, can we create a bug to trrack implementing the same in Python?

andrewpollock:
> Also, can we create a bug to trrack implementing the same in Python?

Yes, I want to better understand how it's behaving with real-world data and I'll either fix it, or file a bug.

PR #1613, opened 2023-09-03
Add a repo that redirects

andrewpollock:
This is resulting in an unnecessary duplicate range entry for the same fix commit


andrewpollock:
Disregard, I'd like to handle synonymous repos in a better way...

PR #1614, opened 2023-09-04
Store an empty ecosystem as a null in Datastore

andrewpollock:
Datastore does not support searching for an empty array, so for cases where it is empty, set it to None, which will translate to a null.

This will allow for manual searching of the Bug Kind for ecosystem set to null (finding only the converted NVD CVEs being the primary use case).

### DIFF
--- a/osv/models.py
+++ b/osv/models.py
@@ -351,6 +351,12 @@ class Bug(ndb.Model):
     self.ecosystem = list(ecosystems_set)
     self.ecosystem.sort()
 
+    # Store an empty ecosystem list as null in Datastore to allow for searching
+    # by this value. (It is not possible to search Datastore for an empty array
+    # value).
+    if not self.ecosystem:
+      self.ecosystem = None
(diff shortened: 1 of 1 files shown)


michaelkedar (commented):
Have we tested that searching for `ecosystem == null` actually works as intended?
I have a suspicion that it might only match things with `null` listed within the ecosystems e.g. `ecosystems: [null, "something else"]`

andrewpollock:
> Have we tested that searching for `ecosystem == null` actually works as intended?
> I have a suspicion that it might only match things with `null` listed within the ecosystems e.g. `ecosystems: [null, "something else"]`

Sadly, I think you might be right... I just manually set a Bug entry's ecosystem to null, and it still doesn't seem to get found with a `select * from Bug where ecosystem = null` :-(

andrewpollock:
Abandoning, as this isn't going to work, either, unfortunately.

PR #1620, opened 2023-09-05
In-progress commit to make consulting the CVE List faster

andrewpollock:
This change makes consulting the CVE List to determine dispute status (for marking converted CVEs as withdrawn) conditional and is the quick interim step to making it consult a local Git clone rather than the CVE List remotely, in an attempt to speed up the run time.

This will temporarily prevent the calls to `vulns.CVEIsDisputed()`, which should speed up run time significantly.

The next PR will refactor `vulns.CVEIsDisputed` to consult a local clone rather than make numerous HTTP requests.


andrewpollock:
Superseded by #1621

Issue #1486, opened 2023-07-21
use natural sort order for affected versions


frasertweedale:
Affected versions are currently sorted lexicographically, meaning e.g. `0.10` comes before `0.4`.
See https://oss-vdb-test.appspot.com/vulnerability/HSEC-2023-0001 (`0.x` series) for example.

Versions should be sorted in the natural order, according to the Ecosystem's scheme.

PR #1631, opened 2023-09-15
Sort versions

hogo6002:
Related issue: https://github.com/google/osv.dev/issues/1486
Before:
image
After:
image


gcp/appengine/frontend_handlers.py
```diff
@@ -515,6 +517,16 @@ def group_versions(versions):
   return groups
 
 
+def sort_version(versions):
```
another-rex, line 520: Can you add a type annotation to this function e.g. `(versions: list[string]) -> list[string]`

gcp/appengine/frontend_handlers.py
```diff
@@ -515,6 +517,16 @@ def group_versions(versions):
...
+  """Sorts a list of version numbers in natural order."""
+  try:
+    return sorted(versions, key=Version)
```
oliverchang, line 523: This needs to be ecosystem specific, rather than always relying on `packging.version.Version`, which is Python's implementation.

Every ecosystem (npm, PyPI, NuGet etc) tend to have its own version ordering rules, which we implement here: https://github.com/google/osv.dev/blob/6551ae4c1877447e01896ed0fd32f797adf4d6fd/osv/ecosystems/_ecosystems.py#L88

Sample usage: https://github.com/google/osv.dev/blob/6551ae4c1877447e01896ed0fd32f797adf4d6fd/osv/models.py#L791

another-rex, line 523: This only supports python versions (and ecosystems that can be coerced into python versions) right?

another-rex: Can you have a look at `osv.ecosystems` and use the version sorting logic we have there depending on the ecosystem? And for the `OrderingUnsupportedEcosystem`s just do the lexicographical sorting.

another-rex (commented):
Thanks, added some comments

gcp/appengine/frontend_handlers.py
another-rex: Oh I'm a bit late haha.

hogo6002: Thanks!

hogo6002: Thanks!

gcp/appengine/frontend_handlers.py
```diff
@@ -515,6 +515,17 @@ def group_versions(versions):
   return groups
 
 
+def sort_versions(versions: list[str], ecosystem: str) -> list[str]:
+  """Sorts a list of version numbers in the given ecosystem's sorting order."""
+  try:
+    return sorted(
+        versions, key=lambda e: osv.ecosystems.get(ecosystem).sort_key(e))
```
another-rex, line 522: You can pass in the function `sort_key` directly rather than creating another function/lambda to wrap around it, this way we also don't need to run `ecosystem.get()` for every element (in this case it doesn't matter too much since it's just a dictionary lookup, but still good practice to avoid it.)

```suggestion
        versions, key=osv.ecosystems.get(ecosystem).sort_key
```

hogo6002: Thanks!

another-rex (approved):
LGTM!

Issue #1607, opened 2023-08-31
"ranges" is used both as a method name and as a variable name


UVvirus:
In Impact.py, the name **"ranges"** is used both as a method name([#L78](https://github.com/google/osv.dev/blob/c8d6717fd63ef85ac0c1e0cc551cd1cfcc76a9d9/osv/impact.py#L78)) and a variable name([#L80](https://github.com/google/osv.dev/blob/c8d6717fd63ef85ac0c1e0cc551cd1cfcc76a9d9/osv/impact.py#L80)) for a list. This is not recommended because it can lead to confusion and unexpected behavior.

I kindly request you to change the name.

karthikmurali60:
@andrewpollock can I contribute to this issue ?

andrewpollock:
By all means. Please review https://github.com/google/osv.dev/blob/master/CONTRIBUTING.md

another-rex:
Resolved in #1636

PR #1636, opened 2023-09-18
Change local variable name in ranges method

karthikmurali60:
### Description

This PR resolves the issue - https://github.com/google/osv.dev/issues/1607

In impact.py, the name "ranges" is used both as a method name and a variable name for a list. The PR changes the variable name to avoid confusion and unexpected behavior.


another-rex (approved):
LGTM

PR #1642, opened 2023-09-19
Update osv-scanner-pr.yml

hayleycd:
So I believe that this should indicate 'master' instead of 'main' for this action. I believe that is why it is not showing up on PRs. 

[Related](https://github.com/google/osv-scanner/pull/541#discussion_r1330519098)


another-rex (approved):
You're right, thanks for spotting it!

Issue #1645, opened 2023-09-20
slices.Delete


Deleplace:
Hello, while researching for [this Go proposal](https://github.com/golang/go/issues/62729) I noticed [this code](https://github.com/google/osv.dev/blob/master/vulnfeeds/cpp/main.go#L407):

```
slices.Delete(cacheEntry, i, i)
```

which probably doesn't behave as intended and should look like this instead:

```
cacheEntry = slices.Delete(cacheEntry, i, i+1)
```

(2 differences: using the returned value, and deleting `+1` element instead of zero).

I have not tried to execute it.

andrewpollock:
Thank you so much for bringing this to my attention!

Deleplace:
See my comment in #1646 : `i+1` as last arg would be more appropriate

PR #1646, opened 2023-09-20
Correctly use the return value of slices.Delete

andrewpollock:
Fixes: #1645


Deleplace:
⚠️ this line deletes `i`- `i` = 0 elements, this is probably not what we want

andrewpollock:
> ⚠️ this line deletes `i`- `i` = 0 elements, this is probably not what we want

Thank you for picking up on that @Deleplace, fixed in https://github.com/google/osv.dev/pull/1646/commits/13fc97e133ca204dbd13b91e82857c04c07f4c33

oliverchang (approved):

PR #1655, opened 2023-09-24
Make running the frontend locally more turnkey

andrewpollock:
I encountered some unfamiliar errors trying to run the frontend locally that were resolved by running `npm install`, so run this automatically as part of the relevant Makefile targets to make it easier for people not familiar with NPM to get things done.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,12 +36,12 @@ lint:
 	tools/lint_and_format.sh
 
 run-appengine:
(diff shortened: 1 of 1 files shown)


another-rex (approved):
Let's go with `npm install` for now and if we discover it changing the `package-lock.json` a lot we can change it to `npm ci` ?

michaelkedar (approved):

PR #1656, opened 2023-09-25
frontend: render multiple repos correctly

andrewpollock:
Rather than iterate over `affected[].packages`, iterate over each `affected`
element, this way elements with multiple `ranges[]` entries can be
rendered as separate package-like tabs.

This is to predominantly address CVEs that end up with multiple GIT
ranges on them.

### DIFF
--- a/gcp/appengine/frontend3/src/templates/vulnerability.html
+++ b/gcp/appengine/frontend3/src/templates/vulnerability.html
@@ -60,17 +60,17 @@
     <spicy-sections
       class="vulnerability-packages{% if vulnerability.affected|should_collapse %} force-collapse{% endif %}">
       {% for affected in vulnerability.affected -%}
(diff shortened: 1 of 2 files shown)


another-rex (approved):

oliverchang (approved):

oliverchang:
Revisiting this as part of debugging https://github.com/google/osv.dev/pull/1661.

I don't think this has the optimal behaviour.

This now renders https://osv.dev/vulnerability/PYSEC-2023-173 as 

image

as opposed to 

image

Which misrepresents the original OSV's entry's structure a bit.

Would it make more sense to instead just move the "Repository" field under "Type ECOSYSTEM" ? That more properly represents the OSV schema's structure.

For the tab name, we can just take the first repo URL we find within that affected[] entry.

PR #1658, opened 2023-09-25
Rework indexes

another-rex:
Rework indexes to remove the unneeded indexes, add comments on new indexes, and reorder indexes.

Added a comprehensive test which queries the API server with **every** possible query combination and check the response code to be correct (and never a 500).

This test does take an additional 45 seconds to run locally (compared  to the rest of the API  tests, which runs in 15 seconds in total) (not including server startup time), so I gated it behind a _LONG_TESTS boolean.

If the behavior needs to change, a clear diff output will be printed out thanks to the `difflib` library that's built into python.


gcp/api/integration_tests.py
```diff
@@ -26,6 +30,7 @@
 
 _PORT = 8080
 _TIMEOUT = 10  # Timeout for HTTP(S) requests
+_LONG_TESTS = False
```
oliverchang, line 33: Should this be an env var instead?

gcp/api/integration_tests.py
```diff
@@ -582,6 +587,90 @@ def test_query_package_purl(self):
...
+    combined_product = set()
+    for elem in product:
+      # Deep copy elem required since merge merges
+      # in-place to the first argument
+      elem = copy.deepcopy(elem)
+      functools.reduce(merge, elem)
+      combined_product.add(json.dumps(elem[0], sort_keys=True))
+
+    self.assertEqual(len(combined_product), 120)
+    with open('fixtures/api_query_response.txt') as h:
```
oliverchang, line 644: Can we use https://github.com/google/osv.dev/blob/master/osv/tests.py#L54 ? 

That also makes it very easy to regenerate fixtures.

gcp/api/integration_tests.py
another-rex: Yep, changed to a env var

gcp/api/integration_tests.py
another-rex: Yep, now using the test module

gcp/api/README.md
```diff
@@ -27,3 +27,14 @@ https://cloud.google.com/container-registry/docs/advanced-authentication#gcloud-
...
 Automatically built and deployed on pushes to master branch.
+
+# Contributing:
+
+## Writing tests
+To speed up running tests:
+
+1. `make run-api-server` at the root of this repository to start the test server.
+2. `pipenv shell` in this directory
+3. `python -m unittest integration_test.py` in this directory.
```
michaelkedar, line 38: How hard is it to do this automatically as part of the `api-server-tests` make command?

michaelkedar (approved):

oliverchang (approved):

gcp/api/README.md
another-rex: It is a decent amount of work, probably need a script that tests whether the server is running, if not build the docker image and run the server before running the test, otherwise run the test normally. Hmm... might be able to do a python script for it.

PR #1660, opened 2023-09-25
Variation of existing tool to delete all Bugs for a given source

andrewpollock:
For CVE unlaunching purposes.

Also update existing scripts to filter in non-warning inducing ways

### DIFF
--- a/tools/datafix/delete_bugs_with_source.py
+++ b/tools/datafix/delete_bugs_with_source.py
@@ -42,6 +42,7 @@ def main() -> None:
 
   query = client.query(kind="Bug")
   query.add_filter(filter=PropertyFilter("source", "=", args.source))
+  query.keys_only()
 
   print(f"Running query {query.filters} "
         f"on {query.kind} (in {query.project})...")
--- a/tools/datafix/delete_bugs_with_source.py
+++ b/tools/datafix/delete_bugs_with_source.py
@@ -42,7 +42,9 @@ def main() -> None:
 
   query = client.query(kind="Bug")
   query.add_filter(filter=PropertyFilter("source", "=", args.source))
-  query.keys_only()
+
+  if not args.verbose:
+    query.keys_only()
 
   print(f"Running query {query.filters} "
         f"on {query.kind} (in {query.project})...")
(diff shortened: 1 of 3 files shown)


tools/datafix/delete_bugs_with_source.py
```diff
@@ -0,0 +1,77 @@
...
+      "--project",
+      action="store",
+      dest="project",
+      default="oss-vdb-test",
+      help="GCP project to operate on")
+  args = parser.parse_args()
+
+  client = datastore.Client(project=args.project)
+
+  query = client.query(kind="Bug")
```
another-rex, line 43: This can run a lot faster as a keys_only query when verbose is set to false.

another-rex: But if it's already running at a decent speed no need to optimise it.

another-rex (approved):

tools/datafix/delete_bugs_with_source.py
andrewpollock: If I'm reading https://cloud.google.com/datastore/docs/samples/datastore-keys-only-query#datastore_keys_only_query-python correctly, it seems like a pretty straightforward optimisation? Thank you for bringing this to my attention.

andrewpollock: Added in [16860f6](https://github.com/google/osv.dev/pull/1660/commits/16860f6df6bedd3b02ddf4cccb17aa0ec125e195)

tools/datafix/delete_bugs_with_source.py
```diff
@@ -42,6 +42,7 @@ def main() -> None:
 
   query = client.query(kind="Bug")
   query.add_filter(filter=PropertyFilter("source", "=", args.source))
+  query.keys_only()
```
another-rex, line 45: Do you intend to print out the full bug when verbose is true? Or just the key, if you want to print out the full bug you probably want to only call this when verbose if false.

andrewpollock: Good point. Adjusted accordingly in 6631cd5

PR #1661, opened 2023-09-25
frontend: rework template for correct rendering of range events

andrewpollock:
For reasons not entirely clear, the subsequent range elements after the first one were not rendering correctly with the previous approach. This approach, while marginally less DRY does the right thing.

Example with CVE-2021-41272:

Before:

image

After:

image

### DIFF
--- a/gcp/appengine/frontend3/src/templates/vulnerability.html
+++ b/gcp/appengine/frontend3/src/templates/vulnerability.html
@@ -105,8 +105,6 @@
           <div class="mdc-layout-grid__cell--span-9">
             <dl>
               <dt>Type</dt>
-              <dd>{{ range.type -}}</dd>
-              <dt>Events</dt>
               <dd>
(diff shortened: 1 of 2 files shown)


another-rex:
Can you post an example link to see this problem?

andrewpollock:
> Can you post an example link to see this problem?

Updated PR description. Including screenshots is easier than I'd appreciated :-)

oliverchang:
Hmm, the "after" screenshot doesn't look quite correct to me. The "Type" row should show GIT/ECOSYSTEM etc.

another-rex:
Should this PR be closed now that #1656 is merged? Or is this a separate issue.

andrewpollock:
> Should this PR be closed now that #1656 is merged? Or is this a separate issue.

CVE-2021-41272 looks respectable in staging currently, so I think this PR is obsolete.

PR #1666, opened 2023-09-26
Remove temporary files when GKE pods start

michaelkedar:
GKE sends SIGTERM signals to pods when it needs to terminate them, and `tempfile.TemporaryDirectory()` does not clean up its files on the persistent SSD when that happens in its scope.
Manually clear the files when starting the pods.

### DIFF
--- a/docker/exporter/exporter.py
+++ b/docker/exporter/exporter.py
@@ -119,7 +119,7 @@ def main():
   args = parser.parse_args()
 
   tmp_dir = os.path.join(args.work_dir, 'tmp')
-  # Temp files are on the persistent SSD,
+  # Temp files are on the persistent local SSD,
   # and they do not get removed when GKE sends a SIGTERM to stop the pod.
   # Manually clear the tmp_dir folder of any leftover files
   # TODO(michaelkedar): use an ephemeral disk for temp storage.
--- a/docker/exporter/exporter.py
+++ b/docker/exporter/exporter.py
@@ -17,6 +17,7 @@ import argparse
 import concurrent.futures
 import logging
 import os
+import shutil
 import tempfile
 import zipfile
 from typing import List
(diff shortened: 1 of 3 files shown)


andrewpollock (approved):

docker/exporter/exporter.py
```diff
@@ -118,6 +118,12 @@ def main():
   args = parser.parse_args()
 
   tmp_dir = os.path.join(args.work_dir, 'tmp')
+  # Temp files are on the persistent SSD,
```
another-rex, line 121:
```suggestion
  # Temp files are on the persistent local SSD,
```

docker/importer/importer.py
```diff
@@ -524,6 +524,12 @@ def main():
   args = parser.parse_args()
 
   tmp_dir = os.path.join(args.work_dir, 'tmp')
+  # Temp files are on the persistent SSD,
```
another-rex, line 527:
```suggestion
  # Temp files are on the persistent local SSD,
```

docker/worker/worker.py
```diff
@@ -672,6 +672,12 @@ def main():
   oss_fuzz_dir = os.path.join(args.work_dir, 'oss-fuzz')
 
   tmp_dir = os.path.join(args.work_dir, 'tmp')
+  # Temp files are on the persistent SSD,
```
another-rex, line 675:
```suggestion
  # Temp files are on the persistent local SSD,
```

docker/exporter/exporter.py
```diff
@@ -118,6 +118,12 @@ def main():
...
+  # and they do not get removed when GKE sends a SIGTERM to stop the pod.
+  # Manually clear the tmp_dir folder of any leftover files
+  # TODO(michaelkedar): use an ephemeral disk for temp storage.
+  if os.path.exists(tmp_dir):
+    os.rmdir(tmp_dir)
```
michaelkedar, line 126: wait, this doesn't work if the directory's not empty...

another-rex (approved):
Just clarify that it's the local SSD

oliverchang (approved):

PR #1670, opened 2023-09-26
Fix vulnerability details page

oliverchang:
- Use alternate approach to rendering multiple repo ranges as proposed
  in https://github.com/google/osv.dev/pull/1656#issuecomment-1734978673.

- Fix various unbalanced tag issues and indent the template a bit better
  to make this more apparent. This seems to be the root cause of
  rendering issues in https://github.com/google/osv.dev/pull/1661.

image

### DIFF
--- a/gcp/appengine/frontend3/src/templates/vulnerability.html
+++ b/gcp/appengine/frontend3/src/templates/vulnerability.html
@@ -67,17 +67,17 @@
     <spicy-sections
       class="vulnerability-packages{% if vulnerability.affected|should_collapse %} force-collapse{% endif %}">
       {% for affected in vulnerability.affected -%}
-      {% for range in affected.ranges %}
       {% if 'package' in affected %}
       {% set ecosystem = affected.package.ecosystem %}
       {% set package = affected.package.name %}
       {% else %}
       {% set ecosystem = 'Git' %}
-      {% set package = range.repo | strip_scheme %}
+      {% set package = vulnerability.repo | strip_scheme %}
       {% endif %}
       <h2 class="package-header">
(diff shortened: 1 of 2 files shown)


oliverchang:
@andrewpollock do you have a CVE ID handy that has multiple repos? I'd like to verify that it renders reasonably with this PR.

andrewpollock:
Looking at the particular problem presented in https://github.com/google/osv.dev/pull/1656#issuecomment-1734978673 how about instead adjusting the multi-tab behaviour based on the presence of any `affected[].package`? That's the notable difference between PYSEC-2023-173 and CVE-2023-2597

That would mean if we have an Alpine CVE that *also* has multiple repos, it wouldn't do the right thing though...

oliverchang:
> > @andrewpollock do you have a CVE ID handy that has multiple repos? I'd like to verify that it renders reasonably with this PR.
> 
> [CVE-2023-2597](https://github.com/advisories/GHSA-4794-756c-cx7v) is the one I was testing against previously.

Thanks! I've updated the PR description with a screenshot. 

> Looking at the particular problem presented in https://github.com/google/osv.dev/pull/1656#issuecomment-1734978673 how about instead adjusting the multi-tab behaviour based on the presence of any affected[].package? That's the notable difference between PYSEC-2023-173 and https://github.com/advisories/GHSA-4794-756c-cx7v

Hmm, I'm a bit wary of custom behaviour like that. It's going to be harder to maintain/reason about and diverges from the source OSV entry.

andrewpollock (approved):
If this works for you, it works for me.

PR #1672, opened 2023-09-27
Reinstate gsutil rsync -c behaviour

andrewpollock:
It turns out that `-c` with `gcloud storage rsync` is the equivalent of `-C` with `gcloud storage rsync`, so we weren't getting the desired behaviour any more :-(


another-rex (approved):
Nice spotting! Though I think your PR description is incorrect? I think you're saying:

> It turns out that --checksum-only with gcloud storage rsync is the equivalent of -c with gsutil rsync, so we weren't getting the desired behaviour any more :-(

andrewpollock:
> I think you're saying

You are of course correct. Unfortunately this automerged as is :-(

andrewpollock:
> > I think you're saying
> 
> You are of course correct. Unfortunately this automerged as is :-(

Actually, no, I think what I originally wrote is correct. The problem was that `-c` with `gcloud storage rsync` is like `-C` with `gsutil rsync` (which is not `--checksum-only`-behaviour with `gcloud storage rsync`). The PR description was stating what the problem was, not what the solution was. (I try to aim for the "why" in the description, not the "what", as that's usually self-evident by the PR content itself).

another-rex:
Ah that makes more sense.

PR #1676, opened 2023-09-28
Log the creation of a worker task thread

andrewpollock:
Right now the completion of a worker task thread is logged, but the commencement of it (with any useful contextual information) is not, which makes it difficult to follow the journey of particular failing record imports.

### DIFF
--- a/docker/worker/worker.py
+++ b/docker/worker/worker.py
@@ -616,6 +616,7 @@ class TaskRunner:
           target=self._do_process_task,
           args=(subscriber, subscription, ack_id, message, done_event),
           daemon=True)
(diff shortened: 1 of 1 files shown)


andrewpollock:
Trawling through the logs, I think there's already this line that gets emitted, that serves as a good enough proxy for what this PR was adding: 

`{"logging.googleapis.com/diagnostic":{…}, "thread":1.3976104892128E14}`

andrewpollock:
No, I retract that, I'd like to see the PubSub message content for each _thread_ creation, not the overall startup of the worker...

hogo6002 (approved):

PR #1700, opened 2023-10-06
Omit unfixed from debian entries

### LINKED ISSUE #1698: Minor issues with Debian advisories revisited
This is directly related to the previous issue
https://github.com/google/osv.dev/issues/1525

This time `<unfixed>` versions seemed to be present, and I am unsure if this is expected. Perhaps just omitting the fixed version would be preferable?  
https://osv.dev/vulnerability/DSA-746-1
https://osv.dev/vulnerability/DSA-866-1
https://osv.dev/vulnerability/DSA-880-1

another-rex:
Fixes #1698 by removing "fixed" entries when the value is "`<unfixed>`". Currently there are only 3 entries with `<unfixed>`.

A similar case where the "fixed" entries is "`<end-of-life>`" is also now omitted (there are currently only 5 advisories with this).

Also bump the python version and update lockfile to make local development easier.

### DIFF
--- a/vulnfeeds/tools/debian/debian_converter/convert_debian.py
+++ b/vulnfeeds/tools/debian/debian_converter/convert_debian.py
@@ -106,7 +106,7 @@ class AffectedInfo:
         }],
     }
 
-    if self.fixed not in [UNFIXED_VERSION, END_OF_LIFE_VERSION]:
+    if self.fixed:
       result['ranges'][0]['events'].append({'fixed': self.fixed})
 
     return result
@@ -220,7 +220,14 @@ def parse_security_tracker_file(advisories: Advisories,
         release_name = version_match.group(1)
         package_name = version_match.group(2)
         fixed_ver = version_match.group(3)
+
+        # Only create advisory if the version is affected.
         if fixed_ver != NOT_AFFECTED_VERSION:
+          # If fixed version is one of the following special values
+          # fixed version essentially doesn't exist, so blank it
+          if fixed_ver in [UNFIXED_VERSION, END_OF_LIFE_VERSION]:
+            fixed_ver = ''
+
           advisories[current_advisory].affected.append(
               AffectedInfo(codename_to_version[release_name], package_name,
                            fixed_ver))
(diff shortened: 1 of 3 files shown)


oliverchang (approved):

vulnfeeds/tools/debian/debian_converter/convert_debian.py
```diff
@@ -99,12 +100,15 @@ def to_dict(self):
             'type': 'ECOSYSTEM',
             'events': [{
                 'introduced': '0'
-            }, {
-                'fixed': self.fixed
             }]
         }],
     }
 
+    if self.fixed != UNFIXED_VERSION:
```
andrewpollock, line 107: Could you please consolidate this and the existing filtering being done at https://github.com/google/osv.dev/blob/b686b04e3237cb7a50ea43b07e9896bb53d67159/vulnfeeds/tools/debian/debian_converter/convert_debian.py#L221 into the one place? I have no strong feelings about whether you have it here where you've added this, or at the other location where the other filtering is happening, but I just think for readability and cognitive load minimisation for future readers, it's better to have it all in the one place.

andrewpollock (changes requested):
IMO, consolidating all the filtering in one place is easier for future readers to reason about the flow of things...

vulnfeeds/tools/debian/debian_converter/convert_debian.py
another-rex: It's in different places because it performs different tasks, the NOT_AFFECTED_VERSION check is to not add the affected entry in the first place, while the other two versions (`end-of-life` and `unfixed`) are to remove the fixed version in the affected entry, but the overall affected entry is still added.

I'll add some comments inline to clarify this, but I don't think there's a good way to put the code in the same place without complicating it.

another-rex: Actually, I can add an extra argument, I'll try that

another-rex: More grouped now in f6be2d94e8a159d03682495360a84a99e2849ef7

andrewpollock (approved):
LGTM, thanks!

PR #1702, opened 2023-10-06
Perform a tree-less clone when analysing repos to optimise bandwidth/disk/time

andrewpollock:
https://github.blog/2020-12-21-get-up-to-speed-with-partial-clone-and-shallow-clone/


oliverchang (approved):
We'll need to monitor this carefully in test, in case there's some subtle breakage caused by this.

andrewpollock:
> We'll need to monitor this carefully in test, in case there's some subtle breakage caused by this.

Ack, I'm going to defer merging this until I'm back...

PR #1705, opened 2023-10-09
Translate cGit and GitWeb URLs to a Repo that is cloneable

andrewpollock:
I had previously not considered the import process and the need for the `repo` in the generated records to be cloneable, and was focusing on human-usable repo URLs, which crashes analysis.

### DIFF
--- a/vulnfeeds/cves/versions.go
+++ b/vulnfeeds/cves/versions.go
@@ -729,52 +729,6 @@ func Commit(u string) (string, error) {
 	return "", fmt.Errorf("Commit(): unsupported URL: %s", u)
 }
 
-// Some Git hosting has different human-accessible URLs for Git web interfaces to what is cloneable by Git and requires translation.
-// e.g.
-// the cGit-hosted repo:
-//
-//	https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git [web browseable]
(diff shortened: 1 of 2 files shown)


vulnfeeds/cves/versions.go
```diff
@@ -707,7 +729,53 @@ func Commit(u string) (string, error) {
...
+// the cGit-hosted repo:
+//
+//	https://git.savannah.gnu.org/cgit/emacs.git [web browseable]
+//	https://git.savannah.gnu.org/git/emacs.git [cloneable]
+//
+// the GitWeb-hosted repo:
+//
+//	https://git.gnupg.org/cgi-bin/gitweb.cgi?p=libksba.git [web browseable]
+//	git://git.gnupg.org/libksba.git [cloneable]
+func maybeRemapRepository(repo string, link string) (remappedRepo string) {
```
oliverchang, line 748: Is this function no longer required?

andrewpollock: Oh yeah, thanks, no it's not, I'll remove it.

andrewpollock: Removed in d36a83b

oliverchang (approved):

PR #1706, opened 2023-10-09
Support denylisting by Vendor, and denylist netapp after three different product strikes

andrewpollock:
Support denylisting by vendor, regardless of product, and denylist netapp, a third product has caused misattribution.

Denylist `netapp:cloud_backup`, causes misattribution to OpenSSH


oliverchang (approved):

vulnfeeds/cpp/main.go
```diff
@@ -87,6 +87,8 @@ var VendorProductDenyList = []VendorProduct{
 	{"netapp", "ontap_select_deploy_administration_utility"},
```
oliverchang, line 87: I almost wonder if we should completely blocklist anything "netapp"

andrewpollock: Done so in 2d5d59d

oliverchang (approved):

Issue #1569, opened 2023-08-19
Compute all transitive aliases for OSV entries


fingeromer:
It seems like when I search [CVE-2023-36258](https://osv.dev/list?ecosystem=&q=CVE-2023-36258) I got to results - PYSEC & GHSA:
<img width="1724" alt="image" src="https://github.com/google/osv.dev/assets/112797342/26a877c6-fa22-424e-8572-175cc12bfb69">

but when I search in the API: 
```
{      
    "package": {
       "purl":"pkg:pypi/langchain@0.0.266"
      }
    }
```

I got only the PYSEC entry:
```
 {
            "id": "PYSEC-2023-98",
            "details": "An issue in langchain v.0.0.199 allows an attacker to execute arbitrary code via the PALChain in the python exec method.",
            "aliases": [
                "CVE-2023-36258"
            ],
            "modified": "2023-08-18T15:49:49.212164Z",
            "published": "2023-07-03T21:15:00Z",
            "references": [
                {
                    "type": "REPORT",
                    "url": "https://github.com/hwchase17/langchain/issues/5872"
                }
            ],
            "affected": [
                {
                    "package": {
                        "name": "langchain",
                        "ecosystem": "PyPI",
                        "purl": "pkg:pypi/langchain"
                    },
                    "ranges": [
                        {
                            "type": "ECOSYSTEM",
                            "events": [
                                {
                                    "introduced": "0"
                                }
                            ]
                        }
                    ],
                    "versions": [
                        "0.0.1",
                        "0.0.10",
                        "0.0.100",
                        "0.0.101",
                        .
                        .
                        .
                        .
                    ],
                    "database_specific": {
                        "source": "https://github.com/pypa/advisory-database/blob/main/vulns/langchain/PYSEC-2023-98.yaml"
                    }
                }
```

It means a lot because it seems like for `langchain` the vulnerabilities in PYSEC are never fixed:
<img width="681" alt="image" src="https://github.com/google/osv.dev/assets/112797342/106fab25-4fb7-46c6-8dba-e2aa5a4995ec">

While in GHSA they are (or at least include the 'last affected version':
<img width="775" alt="image" src="https://github.com/google/osv.dev/assets/112797342/b35a72c0-2a0c-4ec6-8c6d-be62aa0f8e44">

andrewpollock:
In-progress assessment:

[PYSEC-2023-98](https://osv.dev/vulnerability/PYSEC-2023-98)

```
[...]

PR #1733, opened 2023-10-18
Add OSV alias computation cron job.

hogo6002:
Add an alias computation cron job to solve issues like https://github.com/google/osv.dev/issues/1569.


docker/alias/alias_computation.py
```diff
@@ -0,0 +1,96 @@
...
+    bug_ids.append(bug_id)
+
+    aliases = bugs_map.get(bug_id, [])
+    bugs_of_alias = aliases_map.get(bug_id, [])
+    to_visit.update(set(bugs_of_alias + aliases) - visited)
+
+  return sorted(bug_ids)
+
+
+def main():
```
michaelkedar, line 67: Can you add a few comments to describe the algorithm a bit

docker/alias/alias_computation.py
```diff
@@ -0,0 +1,96 @@
...
+  """Updates all alias groups in the datastore."""
+  bugs = osv.Bug.query(osv.Bug.aliases > '').fetch()
+  all_alias_group = osv.AliasGroup.query().fetch()
```
michaelkedar, line 70: I think you can remove the `.fetch()` for both of these and just iterate over the query directly.

I believe `.fetch()` loads the whole of every bug and alias group into memory at once, and it would stay in memory for the entire run of the job. Iterating over the query directly should use less memory since it won't need to store the unnecessary details of each bug.

docker/alias/alias_computation.py
```diff
@@ -0,0 +1,96 @@
...
+    bug_ids = _compute_bugs_ids(to_visit, visited)
+    _update_group(bug_ids, alias_group)
+
+  for bug_id in bugs_map:
+    if bug_id not in visited:
+      to_visit = {bug_id}
+      bug_ids = _compute_bugs_ids(to_visit, visited)
+      _create_alias_group(bug_ids)
+
+  all_alias_group = osv.AliasGroup.query().fetch()
```
michaelkedar, line 89: Is this last query needed/on purpose?

docker/alias/alias_computation_test.py
```diff
@@ -0,0 +1,130 @@
...
+  def test_basic(self):
+    """Tests basic case."""
+    osv.AliasGroup(
+        bug_ids=['aaa-123', 'aaa-124'],
+        last_modified=datetime.datetime(2023, 1, 1),
+    ).put()
+    alias_computation.main()
+    bug_ids = osv.AliasGroup.query(
+        osv.AliasGroup.bug_ids == 'aaa-123').get().bug_ids
+    self.assertEqual(['aaa-123', 'aaa-124'], bug_ids)
```
michaelkedar, line 41: Hmm, so do AliasGroups stay around even if none of the IDs are Bugs in the database?
I don't think we want that (though I don't know if it would happen in practice)

docker/alias/alias_computation_test.py
```diff
@@ -0,0 +1,130 @@
...
+    osv.AliasGroup(
+        bug_ids=['ccc-123'],
+        last_modified=datetime.datetime(2023, 1, 1),
+    ).put()
+    alias_computation.main()
+    alias_group = osv.AliasGroup.query(
+        osv.AliasGroup.bug_ids == 'ccc-123').get()
+    self.assertIsNone(alias_group)
+
+
```
michaelkedar, line 120: Can you also test for the case where we need to split an alias group into two?
e.g. have `A -> B -> C -> D` and remove the `B -> C` alias to get `A -> B` and `C -> D`

osv/models.py
```diff
@@ -666,6 +666,15 @@ def to_vulnerability(self, include_source=False):
...
       credits_.append(cr)
+    related_bugs = Bug.query(Bug.related == self.db_id).fetch()
+    if related_bugs:
+      related_bug_ids = [bug.db_id for bug in related_bugs]
+      self.related = sorted(list(set(related_bug_ids + self.related)))
+
+    alias_group = AliasGroup.query(AliasGroup.bug_ids == self.db_id).get()
+    if alias_group:
+      self.aliases = sorted(list(set(alias_group.bug_ids) - {self.db_id}))
+      modified = max(self.last_modified, alias_group.last_modified)
```
michaelkedar, line 677: nit: Make a new variable instead of modifying `self.related` and `self.aliases` here, we don't want to accidentally change what's stored in the raw aliases/related

docker/alias/run_tests.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash -x
```
michaelkedar, line 1: Also add an `alias-tests` target to the Makefile (and add it to `all-tests`)

michaelkedar (commented):
Looks good!
A few comments:

docker/alias/alias_computation.py
hogo6002: No, it was for debugging purpose. I will remove it.

docker/alias/alias_computation_test.py
hogo6002: Added Bugs to database.

deployment/clouddeploy/gke-workers/environments/oss-vdb/alias-computation.yaml
```diff
@@ -0,0 +1,16 @@
...
+    spec:
+      template:
+        spec:
+          containers:
+          - name: alias-computation
+            env:
+            - name: GOOGLE_CLOUD_PROJECT
+              value: oss-vdb
+            args:
+              - "--bucket=osv-vulnerabilities"
```
hogo6002, line 16: remove it

deployment/clouddeploy/gke-workers/environments/oss-vdb-test/alias-computation.yaml
```diff
@@ -0,0 +1,17 @@
...
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          containers:
+          - name: alias-computation
+            env:
+            - name: GOOGLE_CLOUD_PROJECT
+              value: oss-vdb-test
+            args:
```
hogo6002, line 15: no need

osv/models.py
```diff
@@ -666,14 +666,23 @@ def to_vulnerability(self, include_source=False):
       if credit.type:
         cr.type = vulnerability_pb2.Credit.Type.Value(credit.type)
       credits_.append(cr)
+    related_bugs = Bug.query(Bug.related == self.db_id).fetch()
```
hogo6002, line 669: move to a different PR

docker/alias/alias_computation.py
```diff
@@ -0,0 +1,135 @@
...
+  allow_list = {allow_entry.bug_id for allow_entry in allow_list_query}
+  deny_list = {deny_entry.bug_id for deny_entry in deny_list_query}
+
+  # For each bug, add its aliases to the maps and ignore invalid bugs.
+  for bug in bugs:
+    if bug.db_id not in allow_list and (bug.db_id in deny_list or
+                                        len(bug.aliases) > VULN_ALIASES_LIMIT):
+      logging.info('%s has too many listed aliases, skipping computation.',
+                   bug.db_id)
+      continue
```
michaelkedar, line 105: nit: Don't do this log if it's in the deny list. That would also make the logic a bit clearer:
```suggestion
    if bug.db_id in deny_list:
        continue
    if len(bug.aliases) > VULN_ALIASES_LIMIT and bug.db_id not in allow_list:
      logging.info('%s has too many listed aliases, skipping computation.',
                   bug.db_id)
      continue
```

docker/alias/alias_computation.py
```diff
@@ -0,0 +1,135 @@
...
+      aliases_map.setdefault(alias, []).append(bug.db_id)
+
+  visited = set()
+
+  # For each alias group, re-compute the bug IDs in the group and update the
+  # group with the computed bug IDs.
+  for alias_group in all_alias_group:
+    bug_id = alias_group.bug_ids[0]  # AliasGroups contain more than one bug.
+    if bug_id in visited:
+      alias_group.key.delete()
```
michaelkedar, line 119: We talked about this offline, but also add a comment to explain why we delete the AliasGroup if we've already visited the bug (i.e. because this means two AliasGroups have been combined)

docker/alias/alias_computation.py
```diff
@@ -0,0 +1,135 @@
...
+"""OSV alias computation."""
+import datetime
+import logging
+
+from google.cloud import ndb
+
+import osv
+import osv.logs
+
+ALIAS_GROUP_VULN_LIMIT = 32
```
michaelkedar, line 24: 32 is quite large (for reference, we currently only have 16 data sources across all ecosystems)
Maybe take it down to 16 or so?

michaelkedar (approved):
A few small comments but otherwise this all LGTM

docker/alias/alias_computation.py
```diff
@@ -0,0 +1,138 @@
...
+from google.cloud import ndb
+
+import osv
+import osv.logs
+
+ALIAS_GROUP_VULN_LIMIT = 16
+VULN_ALIASES_LIMIT = 5
+
+bugs_map = {}
+aliases_map = {}
```
oliverchang, line 28: Can we avoid these global variables?

oliverchang: Also, can we rename these to be more descriptive with explanatory comments? i.e. 

```
# Mapping of ID to direct aliases for that bug. 
bug_direct_aliases = {}

# Reverse map of aliases to the bug. 
aliases_to_bug = {}
```

docker/alias/alias_computation.py
```diff
@@ -0,0 +1,138 @@
...
+      continue
+    if len(bug.aliases) > VULN_ALIASES_LIMIT and bug.db_id not in allow_list:
+      logging.info('%s has too many listed aliases, skipping computation.',
+                   bug.db_id)
+      continue
+    if bug.withdrawn:
+      continue
+    bugs_map[bug.db_id] = bug.aliases
+    for alias in bug.aliases:
+      aliases_map.setdefault(alias, []).append(bug.db_id)
```
oliverchang, line 111: Do we also want to have a 

```
aliases_map.setdefault(bug.db_id, []).append(alias)
```

?

Could this also help consolidate the two maps into one?

oliverchang (commented):
nice!!!

docker/alias/alias_computation.py
hogo6002: Thanks, merged two maps into one.

oliverchang (approved):
LGTM with some minor its. OK to merge after!

another-rex (approved):

Issue #1734, opened 2023-10-18
UI does not render the `related` field


hogo6002:
**Is your feature request related to a problem? Please describe.**
Currently, osv.dev does not show the `related` field, and this can be confusing in some cases.

**Describe the solution you'd like**
Add the `related` field to web UI

PR #1737, opened 2023-10-19
Add `related` field to web UI.

hogo6002:
Fix issue https://github.com/google/osv.dev/issues/1734
image

### DIFF
--- a/gcp/appengine/frontend_handlers.py
+++ b/gcp/appengine/frontend_handlers.py
@@ -579,3 +579,10 @@ def package_in_ecosystem(package):
   if ecosystem in osv.ecosystems.package_urls:
     return osv.ecosystems.package_urls[ecosystem] + package['name']
   return ''
+
+
+@blueprint.app_template_filter('osv_has_bug')
+def osv_has_bug(vuln_id):
+  """Checks if an osv bug exists for the given ID."""
+  bug = osv.Bug.get_by_id(vuln_id)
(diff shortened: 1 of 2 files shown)


michaelkedar (commented):
Looks good!
We probably also want to display the vulns that list the current vuln as their related also.
So in your example, CVE-2020-23804's page should also show DLA-3620-1 as related.

hogo6002:
> DLA-3620-1

I am thinking this computation should be done by PR https://github.com/google/osv.dev/pull/1733 right? Once that PR merged, CVE-2020-23804's related field will have DLA-3620-1 and then show it here.

oliverchang (approved):

gcp/appengine/frontend_handlers.py
```diff
@@ -579,3 +579,9 @@ def package_in_ecosystem(package):
   if ecosystem in osv.ecosystems.package_urls:
     return osv.ecosystems.package_urls[ecosystem] + package['name']
   return ''
+
+
+@blueprint.app_template_filter('osv_has_bug')
```
cuixq, line 584: rename it to `osv_has_vuln`?

cuixq (approved):

gcp/appengine/frontend_handlers.py
hogo6002: Thanks!

PR #1754, opened 2023-10-26
fix sanitized-package error

michaelkedar:
Dependabot broke the Pipfile.lock in #1748 - it doesn't reference to the osv package correctly.


hogo6002 (approved):
Thanks!

another-rex (approved):

Issue #1694, opened 2023-10-04
Remove duplicated package names on the list page


hogo6002:
Some OSV entries show repeated package name on the list page. It's caused by the same package has multiple affected ranges.
![image](https://github.com/google/osv.dev/assets/39108850/34a50253-94f5-42f0-8fb6-41cc6971c841)
![image](https://github.com/google/osv.dev/assets/39108850/b0088638-301f-4021-8e65-2f99647bf551)
![image](https://github.com/google/osv.dev/assets/39108850/1bd0d15d-82c2-43ca-8e67-3d9262fe7813)

PR #1757, opened 2023-10-26
Remove duplicated package names on the list page.

hogo6002:
Fix issues https://github.com/google/osv.dev/issues/1694
Before:
image
After:
image

### DIFF
--- a/gcp/appengine/frontend_handlers.py
+++ b/gcp/appengine/frontend_handlers.py
@@ -595,7 +595,7 @@ def list_packages(vuln_affected: list[dict]):
 
   for affected in vuln_affected:
     for affected_range in affected.get('ranges', []):
-      if affected_range['type'] in ('ECOSYSTEM', 'SEMVER'):
+      if affected_range['type'] in ['ECOSYSTEM', 'SEMVER']:
         package_entry = affected['package']['ecosystem'] + '/' + affected[
             'package']['name']
         if package_entry not in packages:
--- a/gcp/appengine/frontend_handlers.py
+++ b/gcp/appengine/frontend_handlers.py
@@ -585,3 +585,24 @@ def package_in_ecosystem(package):
 def osv_has_vuln(vuln_id):
   """Checks if an osv vulnerability exists for the given ID."""
   return osv.Bug.get_by_id(vuln_id)
+
+
+@blueprint.app_template_filter('list_packages')
+def list_packages(vuln_affected: list[dict]):
+  """Lists all affected package names without duplicates,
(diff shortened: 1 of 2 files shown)


gcp/appengine/frontend_handlers.py
```diff
@@ -585,3 +585,24 @@ def package_in_ecosystem(package):
...
+  packages = []
+
+  for affected in vuln_affected:
+    for affected_range in affected['ranges']:
+      if affected_range['type'] in ('ECOSYSTEM', 'SEMVER'):
+        package_entry = affected['package']['ecosystem'] + '/' + affected[
+            'package']['name']
+        if package_entry not in packages:
+          packages.append(package_entry)
+      elif affected_range['repo'] == 'GIT':
```
another-rex, line 603: I think this should still be 'type' right?

gcp/appengine/frontend_handlers.py
```diff
@@ -585,3 +585,24 @@ def package_in_ecosystem(package):
 def osv_has_vuln(vuln_id):
   """Checks if an osv vulnerability exists for the given ID."""
   return osv.Bug.get_by_id(vuln_id)
+
+
+@blueprint.app_template_filter('list_packages')
+def list_packages(vuln_affected: list[dict]):
```
another-rex, line 591: Can you double check if this type is dict? (e.g. printing out the type with `type(vuln_affected)`)

gcp/appengine/frontend_handlers.py
```diff
@@ -585,3 +585,24 @@ def package_in_ecosystem(package):
...
+  """Lists all affected package names without duplicates,
+  remaining the original order."""
```
another-rex, line 593:
```suggestion
  remaining in the original order."""
```

another-rex (commented):
Thanks! added some minor comments.

gcp/appengine/frontend_handlers.py
hogo6002: type(vuln_affected) shows list, and type(affected) shows dict. So I think vulnerability.affected is a list[dict].

gcp/appengine/frontend_handlers.py
hogo6002: Thanks!

gcp/appengine/frontend_handlers.py
another-rex: ah cool, thanks!

another-rex (approved):
LGTM!

PR #1765, opened 2023-10-30
Explicitly specify security-events: write

andrewpollock:
This is required to get this to work correctly


.github/workflows/osv-scanner-pr.yml
```diff
@@ -26,4 +26,6 @@ permissions: read-all
 
 jobs:
   scan-pr-attempt:
```
another-rex, line 28:
```suggestion
  scan-pr:
```

another-rex (approved):

PR #1771, opened 2023-10-31
Ease record reimport

andrewpollock:
From time to time, particularly for OSV records converted from the NVD,
improvements in the heuristics will result in existing records needing
to be reimported. This tool eases that select reimport process.


oliverchang (approved):
neat!

PR #1772, opened 2023-10-31
bugfix: use all available repos, not the last one

andrewpollock:
Correct a bug where the last detected repo (for CVEs with multiple CPEs) is the one used by ensuring all repos for all the CPEs on a CVE are included in the per-CVE repo list.


oliverchang (approved):

oliverchang:
Do you have an example CVE that this would fix?

andrewpollock:
> Do you have an example CVE that this would fix? 

Yes, this discovery came about as a result of investigating why CVE-2021-3541 wasn't present, when I was re-reading https://osv.dev/blog/posts/using-the-determineversion-api/

Issue #1759, opened 2023-10-27
Inconsistent Ecosystem Handling in OSV Query API for Vulnerability and Malware Reports


abhisek:
**Describe the bug**

The OSV query API works by taking `ecosystem`, `name` and `version` as input as documented [here](https://google.github.io/osv.dev/post-v1-query/). This works for querying vulnerabilities. However with the `ecosystem` input set, OSV API do not report any malicious indicators that may be available for the package. OSV API however reports malicious indicators if the `ecosystem` input parameter is skipped.

**To Reproduce**

The following query correctly returns the malicious indicators associated with the package `z-dsm` in `npm` ecosystem

```bash
curl -d \
  '{"package": {"name": "z-dsm"}, "version": "1.0.1"}' \
  "https://api.osv.dev/v1/query"
```

However when `ecosystem` is added in the above query, it does not return any result

```bash
curl -d \
  '{"package": {"name": "z-dsm", "ecosystem": "npm"}, "version": "1.0.1"}' \
  "https://api.osv.dev/v1/query"
```

**Expected behaviour**

The OSV API should be consistent for both vulnerabilities and malware reports since they use the same API and response schema. For example, a query such as below should respond with all vulnerabilities and malicious indicators

```
curl -d \
  '{"package": {"name": "z-dsm", "ecosystem": "npm"}, "version": "1.0.1"}' \
  "https://api.osv.dev/v1/query"
```

oliverchang:
Thanks for the report! It looks like we're not handling SemVer eccosystems properly when exact versions are specified. 

The issue is here:

https://github.com/google/osv.dev/blob/4ef0664dc07dae1834dc16fe9267c9c73ac84716/gcp/api/server.py#L838

We're not matching on exact versions the way we do them here: 

https://github.com/google/osv.dev/blob/4ef0664dc07dae1834dc16fe9267c9c73ac84716/gcp/api/server.py#L750

https://github.com/google/osv.dev/blob/4ef0664dc07dae1834dc16fe9267c9c73ac84716/gcp/api/server.py#L764

oliverchang:
@cuixq wanna give a stab at fixing this?

abhisek:
Awesome. That was super fast. Thanks for resolving this @oliverchang @cuixq

oliverchang:
[...]

PR #1779, opened 2023-11-02
dedupe bugs returned in API response

cuixq:
https://github.com/google/osv.dev/issues/1759


another-rex (approved):
LGTM

michaelkedar (approved):
LGTM, but we should still also be querying the exact versions field in semver ecosystems also

Issue #1766, opened 2023-10-30
Indexer: don't index third party dependencies. 


oliverchang:
https://github.com/gemini-testing/png-img seems to cause matches against it instead of libpng.

oliverchang:
(from https://github.com/google/osv-scanner/pull/621)

another-rex:
Hmm... there's probably a more general approach here where we ignore these "third_party" and similar folders from the initial indexing, could reduce a lot of false positives. 

Also detecting a non C/C++ package manifest file and avoiding these repositories when indexing will help.

oliverchang:
> Hmm... there's probably a more general approach here where we ignore these "third_party" and similar folders from the initial indexing, could reduce a lot of false positives.

+1 that's a great idea! In the meantime though, does it seem reasonable ot just remove the config for this repo and delete all relevant index entities?

another-rex:
Yep can do

another-rex:
png-img has now been removed from the indexer config and datastore hashes.

PR #1780, opened 2023-11-02
Indexer: don't index vendored libraries.

oliverchang:
This can cause bad matches against libraries that depend on the correct library we're trying to identify (#1766).


andrewpollock (approved):

another-rex:
Pushed https://github.com/google/osv.dev/pull/1780/commits/fd17178287784de7abcab551e8d6a3f307197e6e which deletes buckets that doesn't match the document version **after** adding the new buckets in, avoiding any downtime.

susperius (approved):

PR #1783, opened 2023-11-03
gcp/api: also query by generic version for semver ecosystems

cuixq:
Previously, we only queried by semver if a semver ecosystem is set in request. However, this would cause us miss some bugs, see #1759.

In this PR, we query by generic version for all cases to make sure we don't miss any bugs.


gcp/api/server.py
```diff
@@ -852,16 +849,17 @@ def query_by_version(context: QueryContext,
...
+                                                ecosystem, purl, version)
+  for bug in new_bugs:
+    if bug not in bugs:
+      bugs.append(bug)
 
-    # Trying both is too difficult/ugly with paging
-    # Our documentation states that this is an invalid query
-    # context.service_context.abort(grpc.StatusCode.INVALID_ARGUMENT,
-    #                               'Ecosystem not specified')
+  # Trying both is too difficult/ugly with paging
```
michaelkedar, line 859: Always doing both queries for semver ecosystems would mess up pagination for those requests.
Though I don't really know if we can fix it, and I don't know how many semver packages actually trigger pagination in practice.

michaelkedar (approved):
LGTM

gcp/api/server.py
```diff
@@ -852,16 +849,17 @@ def query_by_version(context: QueryContext,
                                          ecosystem, purl, version)
     bugs.extend(new_bugs)
 
-    new_bugs, _ = yield _query_by_generic_version(context, query, package_name,
-                                                  ecosystem, purl, version)
-    for bug in new_bugs:
-      if bug not in bugs:
-        bugs.append(bug)
+  # Try querying by generic version for all cases.
+  new_bugs, _ = yield _query_by_generic_version(context, query, package_name,
```
another-rex, line 853: This will break paging for generic only queries.

Can you add a check here for if it's a generic only query and return the next page token.

+1 to Michael's comment about this still breaking paging even with that fixed since we can't support multiple paging tokens without a huge jump in complexity. I'm not sure what the solution is, maybe something else to motivate a db migration.

another-rex (changes requested):

gcp/api/server.py
cuixq: See https://github.com/google/osv.dev/pull/1804 for the change. Hope I understand the issue correctly!

Actually I am thinking maybe we should call both `_query_by_semver` and `query_by_generic_version` for all cases, and move the checking if a system is semver or not to `_query_by_semver`. Thus we don't need to check the conditions here (now it's a bit complicated).

For the paging thing, maybe we should open an issue for it?

PR #1791, opened 2023-11-06
Update datastore-to-vulnerability conversion

hogo6002:
https://github.com/google/osv.dev/pull/1733 added a new cron job to compute aliases. Updating bug datastore to vulnerability conversion to display computed aliases and related vulnerabilities.


osv/models.py
```diff
@@ -667,13 +667,24 @@ def to_vulnerability(self, include_source=False):
         cr.type = vulnerability_pb2.Credit.Type.Value(credit.type)
       credits_.append(cr)
 
+    related_bugs = Bug.query(Bug.related == self.db_id).fetch()
+    related_bug_ids = [bug.db_id for bug in related_bugs]
+    related = sorted(list(set(related_bug_ids + self.related)))
+
+    alias_group = AliasGroup.query(AliasGroup.bug_ids == self.db_id).get()
+    aliases = self.aliases
```
michaelkedar, line 675: I think we should use an empty list if there's no AliasGroup, in case it declares too many or we have it on the denylist.

osv/models.py
```diff
@@ -667,13 +667,24 @@ def to_vulnerability(self, include_source=False):
...
+      modified.FromDatetime(max(self.last_modified, alias_group.last_modified))
+
     result = vulnerability_pb2.Vulnerability(
         schema_version=SCHEMA_VERSION,
         id=self.id(),
         published=published,
         modified=modified,
-        aliases=self.aliases,
-        related=self.related,
+        aliases=aliases,
```
michaelkedar, line 686: With this change I think we should remove the alias computation logic on the frontend also:
https://github.com/google/osv.dev/blob/master/gcp/appengine/frontend_handlers.py#L238

michaelkedar (commented):
Looks good - just a few minor things

gcp/api/server.py
```diff
@@ -518,6 +518,28 @@ def to_response(b):
     context.service_context.abort(grpc.StatusCode.INVALID_ARGUMENT,
                                   'Invalid query.')
 
+  if include_details:
```
michaelkedar, line 521: nit: maybe worth adding a comment here to explain why we need to do this separately

michaelkedar (approved):
LGTM

gcp/api/server.py
```diff
@@ -528,7 +553,7 @@ def to_response(b):
 def bug_to_response(bug, include_details=True):
   """Convert a Bug entity to a response object."""
   if include_details:
-    return bug.to_vulnerability(include_source=True)
+    return bug.to_vulnerability(include_source=True, include_alias=False)
```
another-rex, line 556: Can you have the include alias arg be passed in so this function is not assuming the caller already set aliases.

gcp/api/server.py
```diff
@@ -99,7 +99,7 @@ def GetVulnById(self, request, context: grpc.ServicerContext):
       context.abort(grpc.StatusCode.PERMISSION_DENIED, 'Permission denied.')
       return None
 
-    return bug_to_response(bug)
+    return bug.to_vulnerability(include_source=True)
```
another-rex, line 102: And this can still use the bug to response func with the change to allow include aliases arg to be passed in

another-rex (commented):
Mostly LGTM, just one comment

gcp/api/server.py
```diff
@@ -518,17 +518,43 @@ def to_response(b):
...
+
   if next_page_token:
     next_page_token = next_page_token.urlsafe()
     logging.warning('Page size limit hit, response size: %s', len(bugs))
 
   return bugs, next_page_token
 
 
-def bug_to_response(bug, include_details=True):
+def bug_to_response(bug, include_details=True, include_alias=False):
```
oliverchang, line 553: Whenn would include_details != include_alias? Can we just consolidate these two into a single flag?

hogo6002: The `include_alias` flag here is for use in `bug.to_vulnerability()` function. It only indicates whether we want to retrieve aliases within that function. If `include_alias` is false, aliases may be retrieved asynchronously through other methods.
For example, when we run `query_by_version/package` query, we want the `bug.to_vulnerability()` function to include the source but not the aliases. So we can retrieve the aliases for those vulnerabilities asynchronously afterward, making the query more efficient.(include_source=True, include_alias=False)
But when we use `GetVulnById()` function, we only get one vulnerability at a time. So we can just get everything we need from the `bug.to_vulnerability()` function. (include_source=True, include_alias=True)
I will add some comments here to make it clearer.

PR #1792, opened 2023-11-06
Validate reference links before using them

andrewpollock:
This eliminates valid-looking references that have linkrot


vulnfeeds/cves/versions.go
```diff
@@ -739,6 +743,33 @@ func Commit(u string) (string, error) {
 	return "", fmt.Errorf("Commit(): unsupported URL: %s", u)
 }
 
+// Detect linkrot via HEAD request with exponential backoff.
+func ValidateLink(link string) (err error) {
+	backoff := retry.NewExponential(1 * time.Second)
+	if err := retry.Do(context.Background(), retry.WithMaxRetries(3, backoff), func(ctx context.Context) error {
+		resp, err := http.Head(link)
```
oliverchang, line 750: Do we need to set any timeout values anywhere?

andrewpollock: > Do we need to set any timeout values anywhere?

That's line 748 above. I'll add a comment to make that clearer.

andrewpollock: Sorry, I misunderstood. Looking at https://pkg.go.dev/net/http#Client, it defaults to https://pkg.go.dev/net/http#RoundTripper, which uses a 30 second timeout, which seems reasonable to me.

oliverchang (approved):

PR #1804, opened 2023-11-09
gcp/api: return next page token if query by generic version

cuixq:
Addressing comment here https://github.com/google/osv.dev/pull/1783#discussion_r1387363131

### DIFF
--- a/gcp/api/server.py
+++ b/gcp/api/server.py
@@ -864,8 +864,9 @@ def query_by_version(context: QueryContext,
       bugs, _ = yield _query_by_semver(context, query, package_name, ecosystem,
                                        purl, version)
 
-      new_bugs, _ = yield _query_by_generic_version(context, query, package_name,
-                                                ecosystem, purl, version)
+      new_bugs, _ = yield _query_by_generic_version(context, query,
+                                                    package_name, ecosystem,
+                                                    purl, version)
       for bug in new_bugs:
         if bug not in bugs:
           bugs.append(bug)
--- a/gcp/api/server.py
+++ b/gcp/api/server.py
@@ -849,8 +849,13 @@ def query_by_version(context: QueryContext,
     bugs.extend(new_bugs)
 
   # Try querying by generic version for all cases.
-  new_bugs, next_page_token = yield _query_by_generic_version(
+
+  new_bugs, generic_page_token = yield _query_by_generic_version(
       context, query, package_name, ecosystem, purl, version)
+
+  if ecosystem and not is_semver:
+    next_page_token = generic_page_token
+
   for bug in new_bugs:
     if bug not in bugs:
       bugs.append(bug)
--- a/gcp/api/server.py
+++ b/gcp/api/server.py
@@ -835,9 +835,8 @@ def query_by_version(context: QueryContext,
   if ecosystem:
     if is_semver:
       # Ecosystem supports semver only.
-      bugs, next_page_token = yield _query_by_semver(context, query,
-                                                     package_name, ecosystem,
-                                                     purl, version)
+      bugs, _ = yield _query_by_semver(context, query, package_name, ecosystem,
+                                       purl, version)
   else:
     logging.warning("Package query without ecosystem specified")
     # Unspecified ecosystem. Try semver first.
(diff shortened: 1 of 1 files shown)


another-rex:
Ah, so how the code previously worked is 

- if you are querying a semver ecosystem, return the paging token for the semver query
- if you are querying a non semver ecosystem, return the paging token for the nonsemver
- if you are querying without an ecosystem (unsupported/invalid by documentation), we don't return a page token.

Now that for semver queries, we are also making the generic query, I think we should return the token for the semver query (this is the case currently/after #1783), since most of the results will be semver I believe?

And for generic ecosystem queries, we should maintain the same behavior and return the generic page token.

oliverchang:
> Ah, so how the code previously worked is
> 
> * if you are querying a semver ecosystem, return the paging token for the semver query
> * if you are querying a non semver ecosystem, return the paging token for the nonsemver
> * if you are querying without an ecosystem (unsupported/invalid by documentation), we don't return a page token.
> 
> Now that for semver queries, we are also making the generic query, I think we should return the token for the semver query (this is the case currently/after #1783), since most of the results will be semver I believe?

+1. It will be much much more likely for the semver query to yield more results than exact version matches. 
> 
> And for generic ecosystem queries, we should maintain the same behavior and return the generic page token.

+1.

gcp/api/server.py
```diff
@@ -850,8 +850,13 @@ def query_by_version(context: QueryContext,
     bugs.extend(new_bugs)
 
   # Try querying by generic version for all cases.
-  new_bugs, _ = yield _query_by_generic_version(context, query, package_name,
-                                                ecosystem, purl, version)
+
+  new_bugs, generic_page_token = yield _query_by_generic_version(
+      context, query, package_name, ecosystem, purl, version)
+
+  if ecosystem and not is_semver:
```
oliverchang, line 857: We could potentialy do something like this to try and still support page tokens in the majority of the SemVer cases:

```python
if bugs and new_bugs:
   # Both queries returned results, we can't support a page token. 
  next_page_token = None
else if new_bugs:
  # The first (SemVer) query returned no results, so we can return a page token here.
  next_page_token = generic_page_token
# Otherwise, the first SemVer query returned results and the generic query returned no results. The page token should already be set to the one from the SemVer query. 
```

This also changes the behaviour slightly for the no-ecosystem-specified case in that we may return a page token in those cases, instead of never returning a page token. 

However, this will only work if either:
- We can distinguish which query a page token is for, or
- Passing an incompatible page token to a query is the same as passing no token (and there are no errors etc raised)

If none of these are possible, we may have to remove page tokens completely for SemVer queries. 

@another-rex @michaelkedar  WDYT?

another-rex: IIRC 2 is not the case, it will throw an exception if called with the wrong page token.

oliverchang: Hmm, in that case, we either: 
1. ignore wrong page token exceptions and proceed with the next query. If both queries raise exceptions, then we re-raise the error. 
2. we remove paging for SemVer completely. we can just move the _query_by_generic_version call inside the query_by_semver function, and remove the page token completely. 

Let's give 1 a shot and see if it works? If not, or if it's too brittle, we can just go with 2.

cuixq: I think we can check if an ecosystem is set or not to decide if we return a `None` page token?

gcp/api/server.py
```diff
@@ -876,8 +876,24 @@ def query_by_version(context: QueryContext,
...
+      context, query, package_name, ecosystem, purl, version)
+
+  if ecosystem:
+    if bugs and new_bugs:
+      # Both queries returned results, we can't support a page token.
+      next_page_token = None
+    elif new_bugs:
+      # The semver query returned no results, so we can return a page token
+      # here.
+      next_page_token = generic_page_token
```
oliverchang, line 889: I'm not sure htis will work as is, because if we try this token against the semver query first, we may get an exception.

Let's just simplify all of this and drop support for pagination for SemVer for now. This logic is getting a little hard to follow. We can just move the `_query_by_generic_version`call into `_query_by_semver` and have it return no page tokens. We can restore the `_query_by_generic_version` call also in the `else` branch after line 866, and have that be the only case that returns a token.

Thinking more, there's potentially a way for us to support this better through our current semver_indexes, but either way it's going to be extremely unlikely to hit this pagination limit in practice.

cuixq: I am a bit concerned to move `_query_by_generic_version` call into `_query_by_semver`, so I:
 - restore the `_query_by_generic_version` call for non-semver ecosystem
 - instead of calling `_query_by_generic_version` once at the all for all cases, I add it to semver ecosystem to make the workflow clear

Basically I revoked https://github.com/google/osv.dev/pull/1783 but add an extra call of `_query_by_generic_version` for semver ecosystem.

oliverchang (approved):
Thanks!!

PR #1805, opened 2023-11-09
Improve exporter performance

hogo6002:
Fix https://github.com/google/osv.dev/pull/1791
Replace query with https://cloud.google.com/appengine/docs/legacy/standard/python/ndb/queries#map

### DIFF
--- a/docker/exporter/exporter.py
+++ b/docker/exporter/exporter.py
@@ -91,14 +91,34 @@ class Exporter:
 
     zip_path = os.path.join(tmp_dir, 'all.zip')
     with zipfile.ZipFile(zip_path, 'w', zipfile.ZIP_DEFLATED) as zip_file:
-      for bug in osv.Bug.query(osv.Bug.ecosystem == ecosystem):
+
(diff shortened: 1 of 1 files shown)


michaelkedar (approved):

docker/exporter/exporter.py
```diff
@@ -91,15 +91,35 @@ def _export_ecosystem_to_bucket(self, ecosystem, tmp_dir):
...
         if not bug.public or bug.status == osv.BugStatus.UNPROCESSED:
-          continue
+          return
 
         file_path = os.path.join(tmp_dir, bug.id() + '.json')
-        osv.write_vulnerability(
-            bug.to_vulnerability(include_source=True), file_path)
+        vulnerability = bug.to_vulnerability(
+            include_source=True, include_alias=False)
+        alias_group = yield osv.get_aliases_async(vulnerability.id)
```
oliverchang, line 103: Can we put all of this into a `bug.to_vulnerability_async` or something similar that can be reused everywhere?

PR #1808, opened 2023-11-10
Handle link decay (where the repo URL directs elsewhere)

andrewpollock:
Currently targeting the known use case of a GitHub rename (i.e. the redirect remains on the same Git repo hosting provider and doesn't do something really funky like redirect from a GitWeb URL to a GitHub one.


oliverchang (approved):
nice!

PR #1813, opened 2023-11-15
Improve fuzzy version matching

andrewpollock:
When doing prefix matching, do not do it versions with a "pre-release" textual suffix, to avoid undesired matches.

Addresses multiple converted CVEs in Apache InLong that were showing as unfixed (their repo has some strange release tagging practices).

### DIFF
--- a/vulnfeeds/git/versions.go
+++ b/vulnfeeds/git/versions.go
@@ -16,6 +16,7 @@ package git
 
 import (
 	"fmt"
+	"regexp"
 	"strings"
 
 	"golang.org/x/exp/maps"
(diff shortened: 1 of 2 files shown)


vulnfeeds/git/versions.go
```diff
@@ -16,25 +16,36 @@ package git
...
 func fuzzyVersionToCommit(normalizedVersion string, repo string, commitType cves.CommitType, normalizedTags map[string]NormalizedTag) (ac cves.AffectedCommit, b bool) {
-	candidateTags := []string{}
+	candidateTags := []string{} // the subset of normalizedTags tags that might be appropriate to use as a fuzzy match for normalizedVersion.
+	// Keep in sync with the regex in cves.NormalizeVersion()
+	var validVersionText = regexp.MustCompile(`(?i)(?:rc|alpha|beta|preview)\d*`)
+
 	for _, k := range maps.Keys(normalizedTags) {
-		if strings.HasPrefix(k, normalizedVersion) {
+		// "1.8.0-RC0" shouldn't be considered a fuzzy match for "1.8.0"
+		if strings.HasPrefix(k, normalizedVersion) && !validVersionText.MatchString(k) {
```
oliverchang, line 37: Will this exclude legitimate rc/beta/alpha etc matches? i.e. ifr both the normalized tag and the `normalizedVersion` have -rc1 or something similar?

oliverchang: Maybe the check here should be: either both the normalized tag and version have a pre-release string tag (i.e. matches `validVersionText`), or none do?

andrewpollock: Yeah I think you're right. I also found a case sensitivity issue that was mishandling exact-match checks with versions with text in them. All addressed in 193a37b

oliverchang (approved):

Issue #1776, opened 2023-11-01
Data quality issue with CVE-2023-25621


andrewpollock:
**The CVE ID**
https://services.nvd.nist.gov/rest/json/cves/2.0?cveId=CVE-2023-25621

**Describe the data quality issue observed**
The repository is correctly identified:
```
"apache:sling_i18n": [
"https://github.com/apache/sling-org-apache-sling-i18n"
]
```

The tag format is not currently supported:

```
$ git ls-remote -t https://github.com/apache/sling-org-apache-sling-i18n | head -5
0e1fc40ddbb7f6171004c0390b8abb26d9a6d2bf        refs/tags/org.apache.sling.i18n-2.0.2
5743ab867ab5028d5564a49eace146893c5c47bb        refs/tags/org.apache.sling.i18n-2.0.2^{}
579198472b3d8b64333eb58a836597063e93423b        refs/tags/org.apache.sling.i18n-2.0.4
f19eade890dc1242badf0d06b94f79ffcc2c9f3e        refs/tags/org.apache.sling.i18n-2.0.4^{}
0fdb8d95fd302cf2fa9766d8b15a823b5d233f60        refs/tags/org.apache.sling.i18n-2.1.0
```

**Suggested changes to record**
The record needs to be successfully converted.

**Additional context**
https://nvd.nist.gov/vuln/search/results?form_type=Advanced&results_type=overview&isCpeNameSearch=true&seach_type=all&query=cpe:2.3:a:apache:sling_i18n:-:*:*:*:*:*:*:* suggests CVE-2023-25621 is the only CVE affected by this gap.

PR #1814, opened 2023-11-16
Add support for normalizing some more obscure tags

andrewpollock:
These have been observed to be involved with CVE conversions that are falsely reporting as unfixed or failing to convert entirely

Support tags that have the repo name in the middle (as seen causing CVE-2023-4366{6,7,8})
Support tags that have the Java package name in them (fixes: #1776)

### DIFF
--- a/vulnfeeds/git/repository.go
+++ b/vulnfeeds/git/repository.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"net/url"
 	"path"
+	"regexp"
 	"sort"
 	"strings"
 	"time"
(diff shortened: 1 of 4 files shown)


vulnfeeds/git/repository.go
```diff
@@ -154,6 +155,34 @@ func RepoTags(repoURL string, repoTagsCache RepoTagsCache) (tags Tags, e error)
...
+	prenormalizedTag := strings.TrimPrefix(strings.ToLower(tag), reponame)
+	// Deal with the reponame being in the *middle* of the tag like 'hudson-yui2-2800'
+	if strings.Contains(prenormalizedTag, reponame) {
+		_, after, found := strings.Cut(prenormalizedTag, reponame)
+		if found {
+			prenormalizedTag = after
+		}
+	}
+	if javaPackageRegex.MatchString(prenormalizedTag) {
+		prenormalizedTag = javaPackageRegex.FindStringSubmatch(prenormalizedTag)[1]
```
oliverchang, line 179: nit: Can we just call `FindStringSubmatch` once, and check for nil? Rather than doing hte initial `MatchString` check.

oliverchang (approved):

vulnfeeds/git/repository.go
andrewpollock: I think that would overall be worse for conciseness, as I wouldn't be able to iteratively reuse the `prenormalizedTag` variable the way I currently am, because it could get turned to `nil` by the call to `FindStringSubmatch`.

PR #1823, opened 2023-11-21
Migrate backup cron job from app engine to gke

hogo6002:
- Migrating 'Backup' cron from app engine to gke. 
- Adding it to 'oss-vdb-test' first for testing purpose.


docker/cron/backup/Dockerfile
```diff
@@ -0,0 +1,19 @@
+# Copyright 2023 Google LLC
```
oliverchang, line 1: Should we just have a single "cron" Dockerfile that can execute all the cron jobs?

hogo6002: yeah, I agree with it. I will create a base cron job docker and override the entrypoint in each k8s yaml.

deployment/clouddeploy/gke-workers/base/backup.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: batch/v1
```
michaelkedar, line 1: Will need to add this file to the `resources` section of the base `kustomization.yaml` file in order for the patch in the staging instance to work.

However, that will also deploy it to production. You could copy this file into the `resources` section of `oss-vdb-test/kustomization.yaml`, and not add it to the base file.

Another option (that I haven't really tried/not sure if it works) is to deploy it to production but keep it [suspended](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#schedule-suspension), with the production yaml file looking like:
```yaml
apiVersion: batch/v1
kind: CronJob
metadata:
  name: backup
spec:
  suspend: true
  jobTemplate:
    spec:
      template:
        spec:
          containers:
          - name: backup
```

docker/cron/backup/backup.py
```diff
@@ -0,0 +1,34 @@
...
+
+_PROJECT = 'oss-vdb-test'
+_BACKUP_BUCKET = 'osv-backup'
+
+
+def main():
+  """Create a Datastore backup."""
+  client = ds_admin.DatastoreAdminClient()
+  client.export_entities(
+      project_id=_PROJECT, output_url_prefix=f'gs://testing-{_BACKUP_BUCKET}')
```
michaelkedar, line 28: #1825 is going to create the bucket named `osv-test-backup`.

Project & bucket name should probably also come from a environment variable/command line argument, like it exporter does it, since it needs to be different between staging/prod

deployment/build-and-stage.yaml
```diff
@@ -98,6 +98,16 @@ steps:
...
+# Build/push cron job images.
+- name: gcr.io/cloud-builders/docker
+  args: ['build', '-t', 'gcr.io/oss-vdb/backup:latest', '-t', 'gcr.io/oss-vdb/backup:$COMMIT_SHA', '.']
+  dir: 'docker/backup'
+  id: 'build-backup'
+  waitFor: ['build-worker']
+- name: gcr.io/cloud-builders/docker
+  args: ['push', '--all-tags', 'gcr.io/oss-vdb/backup']
+  waitFor: ['build-backup', 'cloud-build-queue']
+
```
michaelkedar, line 110: Need to also add the image to the `gcloud deploy` command in this file
```
...
exporter=gcr.io/oss-vdb/exporter:$COMMIT_SHA,\
alias-computation=gcr.io/oss-vdb/alias-computation:$COMMIT_SHA,\
backup=gcr.io/oss-vdb/backup:$COMMIT_SHA,\
...
```

michaelkedar (commented):
Nice work

deployment/clouddeploy/gke-workers/environments/oss-vdb-test/kustomization.yaml
```diff
@@ -13,3 +13,4 @@ patches:
 - path: gen-cperepos-map.yaml
 - path: nvd-cve-osv.yaml
 - path: alias-computation.yaml
+- path: backup.yaml
```
michaelkedar, line 16: This will have to go under the `resources` section at the top, not the `patches` section

docker/cron/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+FROM gcr.io/oss-vdb/worker
+
+COPY . /usr/local/bin/
+RUN chmod 755 /usr/local/bin/*.py
```
michaelkedar, line 18: Does this work with the current folder structure, since it'd get copied as `backup/backup.py`?

deployment/clouddeploy/gke-workers/environments/oss-vdb-test/backup.yaml
```diff
@@ -0,0 +1,30 @@
...
+          - name: cron
+            image: cron
+            imagePullPolicy: Always
+            env:
+            - name: GOOGLE_CLOUD_PROJECT
+              value: "oss-vdb-test"
+            - name: BACKUP_BUCKET
+              value: "osv-test-backup"
+            - name: ENTRYPOINT
+              value: "./backup/backup.py"
```
michaelkedar, line 22: I think this needs to be set as the [command](https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/) instead.

Also, you copy backup into `/usr/local/bin/` in the Dockerfile, while I don't think is the workdir of the container, so I don't think using the local path works as-is.
Maybe change the Dockerfile to copy the code into `./` instead?

hogo6002: Thanks Michael! I still kept python files under `usr/local/bin` but updated file path here.

michaelkedar (approved):
LGTM

PR #1826, opened 2023-11-22
Fix query parsing error on Unicode characters

michaelkedar:
`urllib.parse.parse_qs()` throws an exception on bytes containing unicode characters (even if %-encoded).
We were calling it to remove `page` parameters coming from outside of the turbo frame.

Changed this to work with `request.args` directly, and pass the filtered query params into [`url_for()`](https://flask.palletsprojects.com/en/3.0.x/api/#flask.url_for) as keyword arguments, which are treated as query parameters for the redirect.

Had to make a list of accepted query parameters to sanitize the call to `url_for()`

### DIFF
--- a/gcp/appengine/frontend_handlers.py
+++ b/gcp/appengine/frontend_handlers.py
@@ -167,18 +167,24 @@ def docs():
   return redirect('https://google.github.io/osv.dev')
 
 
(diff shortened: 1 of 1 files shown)


another-rex (approved):
LGTM, tested it locally

oliverchang (approved):

PR #1833, opened 2023-11-27
Rework everything to use NVD API 2.0 format data

andrewpollock:
Added addtional instrumentation to assist with troubleshooting failures

Make API 2.0 downloading more fault tolerant:

Retry individual pages if a 503 is encountered or a potentially malformed page is served.

Add jq to the Docker images as it is used for
slicing up the retrieved output by year.

Simply how test data is managed for ease of expansion.


vulnfeeds/cves/nvd2.go
```diff
@@ -0,0 +1,399 @@
...
+import "github.com/atombender/go-jsonschema/pkg/types"
+
+type Config struct {
+	Operator string `json:"operator,omitempty" yaml:"operator,omitempty" mapstructure:"operator,omitempty"`
+	Negate bool `json:"negate,omitempty" yaml:"negate,omitempty" mapstructure:"negate,omitempty"`
+	Nodes []Node2 `json:"nodes" yaml:"nodes" mapstructure:"nodes"`
+}
+
+// CPE match string or range
+type CpeMatch struct {
```
oliverchang, line 22: nit: these acronyms should all be uppercase per https://google.github.io/styleguide/go/decisions#initialisms

vulnfeeds/cvss30/cvss30.go
```diff
@@ -0,0 +1,733 @@
...
+	TemporalScore *ScoreType `json:"temporalScore,omitempty" yaml:"temporalScore,omitempty" mapstructure:"temporalScore,omitempty"`
+
+	// TemporalSeverity corresponds to the JSON schema field "temporalSeverity".
+	TemporalSeverity *SeverityType `json:"temporalSeverity,omitempty" yaml:"temporalSeverity,omitempty" mapstructure:"temporalSeverity,omitempty"`
+
+	// UserInteraction corresponds to the JSON schema field "userInteraction".
+	UserInteraction *UserInteractionType `json:"userInteraction,omitempty" yaml:"userInteraction,omitempty" mapstructure:"userInteraction,omitempty"`
+
+	// VectorString corresponds to the JSON schema field "vectorString".
+	VectorString string `json:"vectorString" yaml:"vectorString" mapstructure:"vectorString"`
```
oliverchang, line 141: These structs are **incredibly** complicated. 

Do we actually need all of them? Or do we only need the `VectorString` here? If so, we can just replace all of this with a much simpler struct containing only this item, and omit everything else.

andrewpollock: Bear in mind:

1) this file is generated by go-jsonschema, and hasn't required any modifications (yet)
2) everything is appearing to work as-is

The struct is complicated because the schema is complicated. What harm is having a fully functional complicated struct? It's future-proof.

oliverchang: It's harder to review and maintain, and adds a dependency on a tool which may or may not be maintained in the future when we need it again. 

I suggest we follow YAGNI here. If this all changes in the future it's possible everything will have to be re-done anyway (e.g. if CVE 5.0 is adopted by the NVD in their output).

vulnfeeds/cves/nvd2.go
andrewpollock: Addressed in 4be1eeb

vulnfeeds/cvss30/cvss30.go
andrewpollock: Removed in 779c090

vulnfeeds/cves/nvd2.go
```diff
@@ -0,0 +1,418 @@
...
+	}
+	if len(plain.References) > 500 {
+		return fmt.Errorf("field %s length: must be <= %d", "references", 500)
+	}
+	*j = CVEItem(plain)
+	return nil
+}
+
+// (hand generated).
+type DefCveItem struct {
```
oliverchang, line 346: nit: s/Cve/CVE/.

Also, can we have a more descriptive comment? It's also not clear what "Def" is short for here.

vulnfeeds/cves/nvd2.go
```diff
@@ -0,0 +1,418 @@
...
+
+	// Tags corresponds to the JSON schema field "tags".
+	Tags []string `json:"tags,omitempty" yaml:"tags,omitempty" mapstructure:"tags,omitempty"`
+
+	// Url corresponds to the JSON schema field "url".
+	Url string `json:"url" yaml:"url" mapstructure:"url"`
+}
+
+// UnmarshalJSON implements json.Unmarshaler.
+func (j *Reference) UnmarshalJSON(b []byte) error {
```
oliverchang, line 243: Are all of these custom `UnmarshalJSON` functions actually all required? It's not clear that they offer much on top of the existing, out-of-the-box JSON unmarshalling in Go. 

The only difference appears to be validation of required fields, but this seems unnecessary for our use case as a consumer.  

Having these seems like it would just add cognitive load for anyone reading/modifying this code in the future.

oliverchang (approved):
Thanks! This largely LGTM.  I have some remaining comments, but we can address them in a different PR if you want to get some bake time for this in the test instance.

vulnfeeds/vulns/vulns.go
```diff
@@ -320,14 +360,14 @@ func (v *Vulnerability) ToYAML(w io.Writer) error {
 	return encoder.Encode(v)
 }
 
-func timestampToRFC3339(timestamp string) (string, error) {
-	t, err := cves.ParseTimestamp(timestamp)
-	if err != nil {
-		return "", err
-	}
+// func timestampToRFC3339(timestamp string) (string, error) {
```
another-rex, line 363: Can we remove this completely?

vulnfeeds/vulns/vulns.go
```diff
@@ -524,48 +564,49 @@ func unique[T comparable](s []T) []T {
...
-		Details: cves.EnglishDescription(cve.CVE),
-		Aliases: extractAliases(id, cve.CVE),
+		Details: cves.EnglishDescription(cve),
+		Aliases: extractAliases(id, cve),
 	}
 	var err error
 	var notes []string
-	v.Published, err = timestampToRFC3339(cve.PublishedDate)
+	v.Published = cve.Published.Format(time.RFC3339)
 	if err != nil {
```
another-rex, line 599: Error checks here are no longer necessary / incorrect right?

vulnfeeds/vulns/vulns.go
andrewpollock: Yes, I missed this one when I was cleaning up, thanks.

vulnfeeds/cves/nvd2.go
andrewpollock: If there's no interest in retaining (re)generatability, we can rename away from the generated names. I agree this one's unfortunately named due to the way the underlying schema is expressed.

vulnfeeds/cves/nvd2.go
andrewpollock: If there's no interest in retaining (re)generatability, we can rename the generated names, and remove other generated code as desired. 

The idea is this file is generated, and becomes an implementation detail, not something that needs to be maintained (but sure, read).

I pity the fool who has to regenerate the structs against a new schema and then decide should and shouldn't be ripped out/maintained from the future diff. That's the sort of cognitive load I wouldn't be wishing upon someone...

andrewpollock: For a custom type to be unmarshallable, it needs to implement `json.Unmarshaler`, which means it needs to have its own `UnmarshalJSON()`

vulnfeeds/cpp/run_cve_to_osv_generation
```diff
@@ -24,12 +24,24 @@ echo "Commencing conversion run"
...
 echo "Downloading NVD CVE dumps"
 mkdir -p "${WORK_DIR}/nvd"
-/usr/local/bin/download-cves --cvePath "${WORK_DIR}/nvd"
+APIKEY="$(gcloud --project "$GOOGLE_CLOUD_PROJECT" secrets versions access latest --secret=nvd-api --format='get(payload.data)' | base64 -d)"
+/usr/local/bin/download-cves --apiKey "$APIKEY" --cvePath "${WORK_DIR}/nvd"
+
+echo "Splitting monolithic file into years"
+for (( YEAR = 2002 ; YEAR <= $(date +%Y) ; YEAR++ ))
+do
+  cat "${WORK_DIR}/nvd/nvdcve-2.0.json" \
```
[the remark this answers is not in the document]
andrewpollock: It's my plan to decouple the data acquisition (and splitting) from both this and `combine-to-osv` in the interests of reducing runtime, so both scripts will shrink back down a bit in the very near future.

vulnfeeds/cves/nvd2.go
andrewpollock: Renamed for better readability in https://github.com/google/osv.dev/pull/1833/commits/e5bed10b2bd6973ad2a063fffc38780533f8cd3b

vulnfeeds/cves/nvd2.go
oliverchang: Given the immaturity of tools that do this, I would hypothesize that it's much easier to maintain and manually create the structs we need containing only the fields and other bits that we need. The amount of code that then needs to be reviewed/maintained would likely be very small. 

re `UnmarshaJSON`, that's only required if you need special behaviour for the unmarshalling that isn't just moving around primitive types. Other than the NVDTime UnmarshalJSON, it looks like most of these are unnecessary.

PR #1836, opened 2023-11-28
Migrate all cron jobs from gae to gke.

hogo6002:


deployment/clouddeploy/gke-workers/environments/oss-vdb/process-results.yaml
```diff
@@ -0,0 +1,27 @@
...
+  concurrencyPolicy: Forbid
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          containers:
+          - name: cron
+            image: cron
+            imagePullPolicy: Always
+            command: ["/usr/local/bin/make_bugs_public/make_bugs_public.py"]
```
another-rex, line 16: This looks like the wrong file?

docker/cron/make_bugs_public/make_bugs_public.py
```diff
@@ -0,0 +1,69 @@
...
+      make_affected_commits_public(bug)
+
+  if to_mark_public:
+    ndb.put_multi(to_mark_public)
+
+  return 0
+
+
+if __name__ == '__main__':
+  sys.exit(main())
```
michaelkedar, line 69: I think this needs to be wrapped in an ndb client context (e.g. in the [exporter](https://github.com/google/osv.dev/blob/d039ca1b38f56a6c959eee0982afb69d1e56a65a/docker/exporter/exporter.py#L143-L147)).
And also set up logging

docker/cron/process_results/process_results.py
```diff
@@ -0,0 +1,111 @@
...
+          source_id=key_id,
+          allocated_id=cur_id)
+    finally:
+      counter.put()
+
+  return 0
+
+
+if __name__ == '__main__':
+  sys.exit(main())
```
michaelkedar, line 111: Also wrap this with ndb client context.

deployment/clouddeploy/gke-workers/environments/oss-vdb/make-bugs-public.yaml
```diff
@@ -0,0 +1,24 @@
...
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          containers:
+          - name: cron
+            image: cron
+            imagePullPolicy: Always
+            command: ["/usr/local/bin/make_bugs_public/make_bugs_public.py"]
+            resources:
```
michaelkedar, line 17: You should define the `GOOGLE_CLOUD_PROJECT` env here as well - I think ndb uses it

michaelkedar (approved):

docker/cron/make_bugs_public/make_bugs_public.py
```diff
@@ -0,0 +1,72 @@
...
+
+import logging
+import sys
+from google.cloud import ndb
+import requests
+
+import monorail
+import osv
+
+_MONORAIL_ACCOUNT = 'service@oss-vdb.iam.gserviceaccount.com'
```
oliverchang, line 25: Note that this likely needs some IAM changes to add our compute engine service account to the list of accounts that can impersonate this. I just did this for oss-vdb.

oliverchang (approved):
Thanks Holly!

PR #1847, opened 2023-11-29
Fix shell commands being difficult to copy

jess-lowe:
When shell commands are copied from the codeblocks, the leading '$' is copied as well, which is awkward for the setup process. This commit removes the leading '$'s from the code blocks and changes the incline shell code comments to code blocks so they can also be copied a lot easier.


cuixq (approved):

andrewpollock:
See also https://github.com/orgs/community/discussions/35615

PR #1850, opened 2023-11-30
Use the fully fledged struct for downloading

andrewpollock:
The hand-constructed NVD API 2.0 struct was naively done and appears to lead to nil pointer panics under certain conditions.

### DIFF
--- a/vulnfeeds/cmd/download-cves/main.go
+++ b/vulnfeeds/cmd/download-cves/main.go
@@ -54,7 +54,7 @@ func main() {
 // Pages are offset based, this assumes the default (and maximum) page size of PageSize
 // Maintaining the recommended 6 seconds betweens calls is left to the caller.
 // See https://nvd.nist.gov/developers/vulnerabilities
-func downloadCVE2WithOffset(APIKey string, offset int) (page *cves.NVDCVE2, err error) {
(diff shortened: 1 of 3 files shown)


oliverchang (approved):

andrewpollock:
> do you have more details on the actual panics?

I saw a crash yesterday that related to this and didn't investigate further, but assumed there must have been some malformed JSON:

```
ERROR 2023-11-30T06:04:40.502525813Z [resource.labels.containerName: nvd-mirror] panic: runtime error: invalid memory address or nil pointer dereference [signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x983216] goroutine 1 [running]: main.downloadCVE2WithOffset({0x7fff84b72d58, 0x24}, 0x0) /src/cmd/download-cves/main.go:106 +0x6d6 main.downloadCVE2({0x7fff84b72d58, 0x24}, {0x7fff84b72d87, 0x8}) /src/cmd/download-cves/main.go:122 +0x1af main.main() /src/cmd/download-cves/main.go:42 +0xdd
```

I'd been meaning to excise this struct when I added the other one, but ran into some issues at the time that I was able to work through with fresh eyes on a second attempt.

PR #1853, opened 2023-12-03
Fix: Fuzzy searching by version

### LINKED ISSUE #1516: Normalize package versions for querying
Currently, for non-semver ecosystems, version numbers in queries only work if they are exactly what is listed on the package manager.

Some ecosystems allow for multiple ways to refer to a specific version e.g. in PyPI installing `django==4.2.0` will install [`Django==4.2`](https://pypi.org/project/Django/4.2/). (in fact, you can specify any number of trailing `.0`'s and it will install the same version)
Querying the OSV API for django 4.2.0 yields no results, even though django 4.2 does have vulnerabilities.

We need to normalize versions both when ingesting and on the API when querying so we can match these cases.

jess-lowe:
Handles the edgecase where query version has trailing zeros.

Closes #1516 & #1673 

Tested using the following queries:
``` console
curl -d   '{"package": {"purl": "pkg:pypi/cryptography"}, "version": "38.0.0"}'   "http://localhost:8080/v1/query"

curl -d   '{"package": {"purl": "pkg:nuget/SkiaSharp"}, "version": "2.80.3.0"}'   "http://localhost:8080/v1/query"

curl -d   '{"package": {"purl": "pkg:pypi/django"}, "version": "4.2.0"}'   "http://localhost:8080/v1/query"

```
-> integration tests also added to test between versions with trailing '0's.


another-rex:
Thanks! Can you add the 3 test cases into integration_tests.py so we can automate the tests?

gcp/api/server.py
```diff
@@ -797,11 +798,34 @@ def _query_by_generic_version(
...
+        version,
+        normalize=True):
+      results.append(bug)
+      context.total_responses.add(1)
 
+  if results:
+    return results, cursor
+
+  # Try again after canonicalizing + normalizing.
+  version = canonicalize_version(temp_version)
```
cuixq, line 826: `temp_version` is the version before normalization?

gcp/api/server.py
```diff
@@ -797,11 +798,34 @@ def _query_by_generic_version(
   if results:
     return results, cursor
 
+  temp_version = version
   # Try again after normalizing.
   version = osv.normalize_tag(version)
   query = base_query.filter(osv.Bug.affected_fuzzy == version)
```
another-rex, line 804: This query then loop is now repeated 3 different times, let's refactor it into it's own function and just call it 3 times. (Specifically 804 to 820)

another-rex (commented):
Hmm I initially thought we would need to do this per ecosystem, but since it looks like canonicalize version seems to work for multiple ecosystems, happy to go with this for now.

gcp/api/server.py
```diff
@@ -773,51 +774,68 @@ def _query_by_generic_version(
...
+@ndb.tasklet
+def query_by_generic_helper(results,
+                            cursor,
+                            context: QueryContext,
+                            base_query: ndb.Query,
+                            project: str,
+                            ecosystem: str,
+                            purl: PackageURL | None,
+                            version: str,
+                            is_normalized=False):
```
another-rex, line 821: nit: remove this default parameter and just specify it in the function call

gcp/api/server.py
```diff
@@ -773,51 +774,68 @@ def _query_by_generic_version(
...
+  results, cursor = yield query_by_generic_helper(results, cursor, context,
+                                                  base_query, project,
+                                                  ecosystem, purl,
+                                                  canonicalize_version(version),
+                                                  True)
+  return results, cursor
 
+
+@ndb.tasklet
+def query_by_generic_helper(results,
```
another-rex, line 813:
```suggestion
def query_by_generic_helper(results: list,
```

gcp/api/server.py
```diff
@@ -773,51 +774,68 @@ def _query_by_generic_version(
...
+                                                  base_query, project,
+                                                  ecosystem, purl, version)
 
   if results:
     return results, cursor
 
+    # page_token can be the token for this query, or the token for the one
+    # below. If the token is used for the normalized query below, this query
+    # must have returned no results, so will still return no results, fall
+    # through to the query below again.
```
another-rex, line 792: nit: move the indentation back here

another-rex (approved):
LGTM!

gcp/api/server.py
jess-lowe: done :)

PR #1855, opened 2023-12-04
Bump pandas to v2 in debian convert

michaelkedar:
Before #1832, `debian-converter` had `pandas==2.1.1` in the Pipfile.lock, despite the Pipfile specifying `pandas==1.5.3`. (how did this happen??)

It looks like `pandas==1.5.3` doesn't install properly in the docker container due to some problem with installing a compatible version of numpy (?)

Bumped pandas to the latest version, which seems to install when I was running it manually in the container.


hogo6002 (approved):
Thanks Michael!

PR #1857, opened 2023-12-04
Import `osv.logs` in cron

michaelkedar:
Python doesn't automatically load these submodules (all other `osv.X` seem to work as side effects of other imports)


hogo6002 (approved):
LGTM. Thanks!

PR #1863, opened 2023-12-05
use double dashes instead of single dash in license blog

josieang:


hayleycd (changes requested):
~~I'm not sure why, but this broke the docs link.~~

It looks ok now.

hayleycd (approved):

PR #1866, opened 2023-12-07
Update styles.scss

hayleycd:
I believe this should remove the external link image in the blog posts but keep them in the main site. Both Kara and I feel that the little arrows are distracting in the blog. 

We should confirm this on staging before release.


another-rex:
I don't think this change quite works currently, it's actually kind of difficult for CSS to do this "not in class" matching in general (e.g. https://stackoverflow.com/questions/31268382/css-selector-for-something-not-inside-something-else).

> It is currently not possible to construct a CSS selector that matches elements that are not descendants of specific elements ...

I updated it so that only the vuln page and home page has the arrow on links.

another-rex (approved):

PR #1875, opened 2023-12-17
Fix timestamp parsing for CVEs.

oliverchang:
```
Failed to decode JSON in "cve_jsons/nvdcve-2.0-2002.json": parsing time "2023-12-17T22:15:05.71Z": extra text: "Z"
```


another-rex (approved):

another-rex:
Looks like the fix is not that easy, sometimes the Z is missing, e,g,: https://services.nvd.nist.gov/rest/json/cves/2.0?cveId=CVE-2022-32746

hogo6002 (approved):

oliverchang:
> Looks like the fix is not that easy, sometimes the Z is missing, e,g,: https://services.nvd.nist.gov/rest/json/cves/2.0?cveId=CVE-2022-32746

Weird, the JSON from the nvd mirror does include these for the same CVEs, but that might be because of how the nvd downloader encodes it. 

Either way, I'll make this handle both to be safe. Thanks for the callout!

another-rex:
One more thing, can you add the logic to `cve.go` as well, it has the same issue with parsing disputed dates.

oliverchang:
> One more thing, can you add the logic to `cve.go` as well, it has the same issue with parsing disputed dates.

done!

PR #1895, opened 2024-01-01
Add dependency-management-data to README

jamietanna:
As I'd not spotted this was different to the third party docs added in
2a4022c818232b32573841a6f506796a25d13120.


another-rex (approved):
Thanks!

Issue #1235, opened 2023-04-19
Add support for a REST API endpoint as a datasource


andrewpollock:
Rocky Linux has started publishing OSV records via a REST API interface, which is not a supported consumption mechanism at present.

There's value in supporting a feed via a REST API, as this gives future data publishers more publication options than just a Git repo or a GCS bucket.

oliverchang:
This will also be useful for curl, who are currently publishing their OSV feed at https://curl.se/docs/vuln.json

jess-lowe:
Closed in #1901

PR #1901, opened 2024-01-07
Adding support for a REST API endpoint as a datasource

jess-lowe:
As mentioned in #1235, there is a need for being able to pull data from REST API endpoints.

Full documentation will follow on how this can be used to contribute. 

Testcases utilise a localhost instance to mock the behaviour of the external HTTP request.


docker/importer/importer.py
```diff
@@ -452,6 +453,63 @@ def thread_init():
...
+  def _process_updates_rest(self, source_repo: osv.SourceRepository):
+    """Process updates from REST API."""
+    logging.info("Begin processing REST: %s", source_repo.name)
+    import_time_now = utcnow()
+    request = requests.head(source_repo.rest_api_url, timeout=60)
+    if request.status_code != 200:
+      logging.error('Failed to fetch REST API: %s', request.status_code)
+      return
+    last_modified = datetime.datetime.strptime(request.headers['Last-Modified'],
+                                               '%a, %d %b %Y %H:%M:%S %Z')
```
oliverchang, line 465: Can you put the format in a CONSTANT further up in the file?

```python
HTTP_LAST_MODIFIED_FORMAT =  '...'
```

docker/importer/importer.py
```diff
@@ -452,6 +453,63 @@ def thread_init():
...
+    # Check whether endpoint has been modified since last update
+    if last_modified < source_repo.last_update_date:
+      logging.info('No changes since last update.')
+      return
+    request = requests.get(source_repo.rest_api_url, timeout=60)
```
oliverchang, line 470: can you put 60 in a constant further up in the file?

docker/importer/importer.py
```diff
@@ -452,6 +453,63 @@ def thread_init():
...
+    # Get all vulnerabilities from the REST API. (CURL approach)
+    vulns = request.json()
```
oliverchang, line 472: Can you convert them into Vulnerability objects using https://github.com/google/osv.dev/blob/20deb6f580eae41473176d15789efac1381aa2b3/osv/sources.py#L123 ?

This way you can also avoid parsing the modified timestamp yourself.

docker/importer/importer_test.py
```diff
@@ -554,6 +560,92 @@ def test_import_override(self, unused_mock_time: mock.MagicMock,
...
+    self.httpd = http.server.HTTPServer(SERVER_ADDRESS, MockDataHandler)
+    thread = threading.Thread(target=self.httpd.serve_forever)
+    thread.start()
+    self.source_repo.last_update_date = datetime.datetime(2020, 1, 1)
+    self.source_repo.put()
+    imp = importer.Importer('fake_public_key', 'fake_private_key', self.tmp_dir,
+                            importer.DEFAULT_PUBLIC_LOGGING_BUCKET, 'bucket',
+                            False)
+    imp.run()
+    mock_publish.assert_called()
```
cuixq, line 612: maybe also assert what is called?

docker/importer/importer.py
```diff
@@ -452,6 +453,63 @@ def thread_init():
...
+    # Create tasks for changed files.
+    for vuln in vulns:
+      import_failure_logs = []
+      last_modified = datetime.datetime.strptime(vuln['modified'],
+                                                 '%Y-%m-%dT%H:%M:%S.%fZ')
+
+      if last_modified < source_repo.last_update_date:
+        continue
+      try:
+        if source_repo.link[-1] != '/':
```
oliverchang, line 483: nit: can we just enforce that these should always end with '/' by throwing an exception at the start of all of this if this is not the case?

I think we already make this assumption everywhere else

docker/importer/importer.py
```diff
@@ -452,6 +453,63 @@ def thread_init():
...
+          source_repo.link += '/'
+        single_request = requests.get(
+            source_repo.link + vuln['id'] + source_repo.extension, timeout=60)
```
oliverchang, line 486: use a constant for the 60.

docker/worker/worker.py
```diff
@@ -389,6 +390,28 @@ def _source_update(self, message):
         return
 
       repo = None
+    elif source_repo.type == osv.SourceRepositoryType.REST_ENDPOINT:
+      vulnerabilities = []
+      if source_repo.link[-1] != '/':
```
oliverchang, line 395: same comment here: let's just enforce and make the assumption that these links must always end in /.

docker/worker/worker.py
```diff
@@ -389,6 +390,28 @@ def _source_update(self, message):
...
+        source_repo.link += '/'
+      for vuln in path:
+        url = source_repo.link + vuln + source_repo.extension
+        request = requests.get(url, timeout=60)
```
oliverchang, line 399: put the 60 in a constant somewhere.

docker/importer/importer_test.py
oliverchang, line 612: Can you assert the arguments that are called here as well? i.e. assert_called_with

docker/importer/importer.py
```diff
@@ -452,6 +453,63 @@ def thread_init():
...
+        continue
+      except Exception:
+        logging.error('Failed to parse %s', vuln['id'])
+        import_failure_logs.append('Failed to parse vulnerability "' +
+                                   vuln['id'] + '"')
+        continue
+    if vulns_to_update:
+      self._request_analysis_external(
+          source_repo, osv.sha256_bytes(source_repo.rest_api_url.encode()),
+          vulns_to_update)
```
oliverchang, line 504: Hmm, I'm not sure passing this (a list) works here. WE need to request individual updates for every vuln to update, rather than this single one.

docker/importer/importer.py
```diff
@@ -452,6 +453,63 @@ def thread_init():
...
+        logging.info('Entry does not have an OSV entry: %s', vuln['id'])
+        continue
+      except Exception:
+        logging.error('Failed to parse %s', vuln['id'])
+        import_failure_logs.append('Failed to parse vulnerability "' +
+                                   vuln['id'] + '"')
+        continue
+    if vulns_to_update:
+      self._request_analysis_external(
+          source_repo, osv.sha256_bytes(source_repo.rest_api_url.encode()),
```
oliverchang, line 503: this should be the sha256 of  the individual vuln

oliverchang:
nice work btw!

docker/mock_test_handler.py
```diff
@@ -0,0 +1,44 @@
...
+  def do_GET(self):  # pylint: disable=invalid-name
+    """Serve a mock GET request."""
+    try:
+      with open(self.file_path, 'r') as f:
+        data = json.load(f)
+      if self.path != '/':
+        found = False
+        # find the vulnerability to mock serve the new page
+        for vuln in data:
+          if vuln['id'] == (self.path.split('/')[-1]).split('.json')[0]:
```
cuixq, line 20: `vuln['id']` has the prefix of `CURL` but the file name in link starts with `CVE`?

docker/worker/worker.py
cuixq, line 395: also here

docker/worker/worker_test.py
cuixq, line 1: single quotes are changed to double quotes - is this a thing related to the formater?

jess-lowe: Yeah I think the auto-formatter did it. I'll change them back.

docker/importer/importer.py
```diff
@@ -452,6 +453,63 @@ def thread_init():
 
     logging.info("Finished processing bucket: %s", source_repo.name)
 
+  def _process_updates_rest(self, source_repo: osv.SourceRepository):
+    """Process updates from REST API."""
+    logging.info("Begin processing REST: %s", source_repo.name)
```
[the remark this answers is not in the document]
jess-lowe: Done :)

docker/importer/importer.py
```diff
@@ -452,8 +456,64 @@ def thread_init():
...
+      if last_modified < source_repo.last_update_date:
+        continue
+      try:
+        vuln_location = source_repo.link + vuln['id'] + source_repo.extension
+        single_request = requests.get(vuln_location, timeout=_TIMEOUT_SECONDS)
+        _ = osv.parse_vulnerability_from_dict(single_request.json(),
+                                              source_repo.key_path,
+                                              self._strict_validation)
+        self._request_analysis_external(
+            source_repo, osv.sha256_bytes(vuln_location.encode()),
```
another-rex, line 491: The sha256 should be of the response body data from the single_request similar to bucket. 

The `git` version encodes the path because the commit hash already makes sure that the data doesn't change, but that doesn't apply for a REST endpoint.

docker/importer/importer.py
```diff
@@ -452,8 +456,64 @@ def thread_init():
...
+    for vuln in vulns:
+      import_failure_logs = []
+      last_modified = datetime.datetime.strptime(vuln['modified'],
+                                                 _OSV_MODIFIED_FORMAT)
+
+      if last_modified < source_repo.last_update_date:
+        continue
+      try:
+        vuln_location = source_repo.link + vuln['id'] + source_repo.extension
+        single_request = requests.get(vuln_location, timeout=_TIMEOUT_SECONDS)
```
another-rex, line 486: We can use a `ThreadPoolExecutor` here to make the HTTP requests concurrent similar to how buckets are processed. 
Let's just add a TODO comment for it here for now, and update this in a followup PR. This way we can make sure it works without concurrency and measure how long it takes.

docker/importer/importer_test.py
```diff
@@ -554,6 +561,117 @@ def test_import_override(self, unused_mock_time: mock.MagicMock,
...
+  def setUp(self):
+    tests.reset_emulator()
+    self.maxDiff = None  # pylint: disable=invalid-name
+    self.tmp_dir = tempfile.mkdtemp()
+
+    tests.mock_datetime(self)
+    self.mock_repo = tests.mock_repository(self)
+    warnings.filterwarnings("ignore", "unclosed", ResourceWarning)
+
+    storage_patcher = mock.patch('google.cloud.storage.Client')
```
another-rex, line 578: Can you test what removing the storage patcher will do? I think this is only used for the bucket tests.

another-rex (commented):
Looks good! Added some comments

docker/importer/importer_test.py
jess-lowe: It's giving 403 Forbidden errors - something about POST permissions. I believe it is acting as the storage location for logging? _process_updates_git also uses a storage client to store logs by the looks of it (replace_importer_log) in importer.py.

docker/importer/importer.py
jess-lowe: Added TODO :)

docker/importer/importer.py
jess-lowe: Makes sense! Just updated it now to use request.text to encode. :)

docker/importer/importer_test.py
another-rex: Ah that makes sense, let's leave it in then.

docker/importer/importer.py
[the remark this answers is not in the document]
jess-lowe: It now raises an exception if the link exists but isn't in the right format.

docker/importer/importer.py
jess-lowe: This is updated now to request individual updates for every vuln. (Also fixes sha256 uniqueness problem)

docker/importer/importer_test.py
jess-lowe: Because there is ~24 vulns called here, I've now asserted the number called. This test expects them all to be updated so it now checks how many vulns are in the given file to compare and assert. For the filtered results (in the other testcase), they have all the mock calls asserted.

docker/importer/importer.py
oliverchang: Did you give this a try? Fine to skip if this ends up being a lot slower. This can potentially just simplify/reduce some duplication of logic with timestamp parsing etc.

docker/importer/importer.py
```diff
@@ -452,8 +456,65 @@ def thread_init():
...
+      import_failure_logs = []
+      last_modified = datetime.datetime.strptime(vuln['modified'],
+                                                 _OSV_MODIFIED_FORMAT)
+
+      if last_modified < source_repo.last_update_date:
+        continue
+      try:
+        #TODO(jesslowe): Use a ThreadPoolExecutor to parallelize this
+        vuln_location = source_repo.link + vuln['id'] + source_repo.extension
+        single_request = requests.get(vuln_location, timeout=_TIMEOUT_SECONDS)
```
oliverchang, line 487: nit: `single_vuln` would be more descriptive as a variable name. `single_request` can be confusing as the variable itself is not exactly a "request", but rather the response (i.e. the vuln).

docker/importer/importer.py
jess-lowe: Done!

docker/importer/importer.py
jess-lowe: Yeah, that works a lot better. Changed :)

another-rex (approved):
LGTM!

PR #1920, opened 2024-01-18
Add information on querybatch response ordering

noqcks:
ref: https://github.com/google/osv.dev/issues/321#issuecomment-1129453918


noqcks:
cc: @oliverchang

PR #1922, opened 2024-01-19
ignore_last_import_time implementation for REST source path

jess-lowe:
ignore_last_import_time is needed in case files need to be reimported


docker/importer/importer.py
```diff
@@ -502,7 +507,8 @@ def _process_updates_rest(self, source_repo: osv.SourceRepository):
 
     replace_importer_log(storage.Client(), source_repo.name,
                          self._public_log_bucket, import_failure_logs)
-
+    if ignore_last_import_time:
```
another-rex, line 510: Let's move this up to 463 to keep the behavior similar to the bucket import function (and add another source_repo.put()). We want to have ignore last import time set back to false even if there's an exception happening between 463 and 513.

another-rex (commented):
Just a minor comment, otherwise LGTM!

docker/importer/importer.py
jess-lowe: done :) 👍🏻

another-rex (approved):

PR #1931, opened 2024-01-25
modify job permissions

cuixq:
Some of the jobs has unnecessary `read-all` access, this PR modifies the default access to none and a job can overwrite the permission.


another-rex (approved):
Thanks!

michaelkedar (approved):

PR #1936, opened 2024-01-29
Rename/unify CVE vulnfeed project

another-rex:
Rename all the CVE vulnfeed project names into one name: `cpe-repo-gen`.

Currently the NVD cve vulnfeed project is two binaries with 5 different names (which was quite confusing to figure out when debugging):

- `cperepos` (dir path: vulnfeeds/cmd/cperepos/)
- `cpe-repo-gen` (docker image name)
- `gen-cperepos-map` (yaml file name and kubernetes worker and container name)

and the NVD to OSV conversion exe has:

- `cpp` (dir path: vulnfeeds/cpp)
- `nvd-cve-osv` (image/kubernetes worker name)

This PR unifies then down to two names, `cpe-repo-gen` and `nvd-cve-osv`. I also moved `nvd-cve-osv` under the `cmd/` directory since it's a binary (from `vulnfeeds/cpp` to `vulnfeeds/cmd/nvd-cve-osv`)

I'm not sure if changing the worker name will require some manual work when deploying?


michaelkedar:
> I'm not sure if changing the worker name will require some manual work when deploying?

IIRC, Cloud Deploy will make a new workload with the new name, but won't delete the old one with the original name, so that'll have to be manually deleted.

vulnfeeds/cmd/cpe-repo-gen/cpe-repo-gen_map.sh
michaelkedar, line 1: nit: If you're going to add the `.sh` extension here, you could also add it to `run_cve_to_osv_generation` also

michaelkedar (approved):

vulnfeeds/cmd/cpe-repo-gen/cpe-repo-gen_map.sh
another-rex: Ah yep you're right, updated that now!

vulnfeeds/cmd/nvd-cve-osv/run_cve_to_osv_generation.sh
```diff
@@ -36,7 +36,7 @@ IN_SCOPE_YEARS=($IN_SCOPE_YEARS)
 for YEAR in "${IN_SCOPE_YEARS[@]}"; do
```
andrewpollock, line 36: While we're here, this addresses a little embarrassing oversight that I'll further tidy up in a followup PR

```suggestion
for (( YEAR = 2002 ; YEAR <= $(date +%Y) ; YEAR++ )); do
```

andrewpollock (approved):
LGTM, let's address the little whoopsie for 2024 while the patient is on the table...

PR #1937, opened 2024-01-29
Script for reputting bugs 

jess-lowe:
In order to propagate 'GIT' ecosystem

- This would be one of the most cost-effective ways (instead of reimporting everything) to make sure that the bugs with Git ranges are given the 'git' ecosystem tag


tools/datafix/reput_bugs.py
```diff
@@ -0,0 +1,82 @@
...
+
+  result = list(query.fetch())
+
+  print(f"Retrieved {len(result)} bugs to examine for reputting")
+
+  # Chunk the results to reput in acceptibly sized batches for the API.
+  for batch in range(0, len(result), MAX_BATCH_SIZE):
+    try:
+      with client.transaction():
+        ndb.put_multi_async([
```
another-rex, line 64: Can you add a comment here clarifying that this will run the `_pre_put_hook()` func under Bug?

another-rex (approved):
Minor nit, otherwise LGTM!

tools/datafix/reput_bugs.py
jess-lowe: Sure thing! Done :)

PR #1939, opened 2024-01-29
Open Sourcing SourceRepositories to sync with datastore. 

jess-lowe:
Includes:

- A script for downloading the sourcerepositories to populate
- the SourceRepository information (non-dynamic) in yaml files at the root
- A script for updating the datastore with the local information. 

> After this, the source.yaml and source_test.yaml are the source of truth. **If they are not there, they will be DELETED from Datastore**


another-rex:
I notice that a lot of the fields in the SourceRepositoryTest kind are non-existant rather than empty/false/null. I wonder if that will cause problems when being loaded in by the python code.

tools/sourcerepo-sync/source_sync.py
```diff
@@ -0,0 +1,109 @@
...
+    for ds_repo in ds_repos:
+      # if it exists, check if it needs to be updated
+      if repo['name'] == ds_repo['name']:
+        repo_found = True
+        if args.verbose:
+          print(f'Found source repository {repo["name"]}')
+        ds_repos.pop(ds_repos.index(ds_repo))
+        update_sourcerepo(repo, ds_repo, args, client, args.kind)
+
+      if repo_found:
```
another-rex, line 69: This line might have the wrong indentation?

tools/sourcerepo-sync/source_sync.py
```diff
@@ -0,0 +1,109 @@
...
+        print(f'New source repository {repo["name"]}')
+      key = client.key(args.kind, repo['name'])
+      entity = datastore.Entity(key=key)
+      entity.update(repo)
+      if not args.dryrun:
+        client.put(entity)
+
+  # If the source repo is not in the local yaml, delete it
+  for ds_repo in ds_repos:
+    if ds_repo['name'] not in [repo['name'] for repo in local_sourcerepos]:
```
another-rex, line 83: I think it'll be clearer to move `[repo['name'] for repo in local_sourcerepos]` out to line 80 or so as a variable `local_sourcerepo_names`. You can also change the outer square brackets to `{ }` to create a set for O(1) lookups. (Not that it matters for the scale we are looping through, just good practice.)

tools/sourcerepo-sync/source_sync.py
```diff
@@ -0,0 +1,109 @@
...
+      key = client.key(args.kind, ds_repo['name'])
+      if not args.dryrun:
+        client.delete(key)
+
+
+def update_sourcerepo(repo, ds_repo, args, client, kind):
+  """Check the attributes of the source repo and update if needed."""
+  change_flag = False
+  for attr in repo:
+    if attr in ds_repo:
```
another-rex, line 95: Invert these three if statements to reduce nesting (i.e. `if attr not in ds_repo: continue`)

tools/sourcerepo-sync/source_sync.py
```diff
@@ -0,0 +1,109 @@
...
+        repo_found = True
+        if args.verbose:
+          print(f'Found source repository {repo["name"]}')
+        ds_repos.pop(ds_repos.index(ds_repo))
+        update_sourcerepo(repo, ds_repo, args, client, args.kind)
+
+      if repo_found:
+        continue
+    # if it doesn't exist in the datastore, create it
+    if not repo_found:
```
another-rex, line 72: This if statement is no longer necessary if the above one is moved out one indent.

tools/sourcerepo-sync/source_sync.py
```diff
@@ -0,0 +1,109 @@
...
+  if args.project == 'oss-vdb-test':
+    file = 'source_test.yaml'
+  elif args.project == 'oss-vdb':
+    file = 'source.yaml'
+  else:
+    print('Invalid project')
+    return
+  local_sourcerepos = []
+  with open(os.path.join(sys.path[-1] + '/', file), 'r') as f:
+    local_sourcerepos = yaml.load(f, Loader=yaml.FullLoader)
```
another-rex, line 47: Can you clarify/add a comment here as to why the Loader has to be manually specified?

tools/sourcerepo-sync/source_downloader.py
```diff
@@ -0,0 +1,67 @@
...
+    print('Invalid project')
+    return
+
+  client = datastore.Client(project=args.project)
+  query = client.query(kind=args.kind)
+  results = list(query.fetch())
+  if args.verbose:
+    print(f'Retrieved {len(results)} sourcerepos')
+  sources = ''
+  for result in results:
```
another-rex, line 46: Let's use python objects or dictionaries here + yaml.dump instead of manually writing strings, it'll make a more maintainable in the future if we decide to add more fields.

tools/sourcerepo-sync/source_sync.py
jess-lowe: It's just the arguments yaml.load() requires - do you still want me to comment this?

Either way, this did make me realise that CLoader is a thing that exists that supposedly runs a lot faster than FullLoader, so I've swapped them out :).

tools/sourcerepo-sync/source_sync.py
jess-lowe: Done now, I believe :)

tools/sourcerepo-sync/source_downloader.py
jess-lowe: done :)

tools/sourcerepo-sync/source_sync.py
jess-lowe: Done :)

tools/sourcerepo-sync/source_sync.py
another-rex: Hmm I think you can exclude the Loader argument and it'll default to the `Loader` (I can't figure out what FullLoader is, literally doesn't exist in the documentation :confused: ?? ). 

I think CLoader requires LibYAML library bindings, which we should avoid to simplify the build process.

Also can you see if you can use `safe_load` here as well?

tools/sourcerepo-sync/source_sync.py
another-rex: You can invert the next line down as well :) (118)

andrewpollock:
I'm late to the party on this, and haven't looked at the implementation yet, but I'm very excited to see it coming to fruition!

One of the more common "user journeys" is needing to trigger a reimport of a source, by setting `ignore_last_import_time` to `true`. I haven't looked closely at this PR to see how that would get handled.

Will that still be possible via direct edit to the entry in Datastore, or will it require a commit? I'm assuming by the absence of this in the YAML files in this PR the answer is "direct edit".

This would be problematic to do via a commit because the value gets reset back after the import has been triggered, causing Datastore and the file in the repo to diverge.

tools/sourcerepo-sync/source_sync.py
jess-lowe: Yes! `safe_load` works

another-rex (approved):
LGTM

another-rex:
> I'm late to the party on this, and haven't looked at the implementation yet, but I'm very excited to see it coming to fruition!
> 
> One of the more common "user journeys" is needing to trigger a reimport of a source, by setting `ignore_last_import_time` to `true`. I haven't looked closely at this PR to see how that would get handled.
> 
> Will that still be possible via direct edit to the entry in Datastore, or will it require a commit? I'm assuming by the absence of this in the YAML files in this PR the answer is "direct edit".
> 
> This would be problematic to do via a commit because the value gets reset back after the import has been triggered, causing Datastore and the file in the repo to diverge.

I believe this specifically avoids changing these dynamic variables, so ignore_last_import_time can still be switched on manually in the datastore interface. @jess-lowe Can you confirm?

jess-lowe:
> I'm late to the party on this, and haven't looked at the implementation yet, but I'm very excited to see it coming to fruition!
> 
> One of the more common "user journeys" is needing to trigger a reimport of a source, by setting `ignore_last_import_time` to `true`. I haven't looked closely at this PR to see how that would get handled.
> 
> Will that still be possible via direct edit to the entry in Datastore, or will it require a commit? I'm assuming by the absence of this in the YAML files in this PR the answer is "direct edit".
> 
> This would be problematic to do via a commit because the value gets reset back after the import has been triggered, causing Datastore and the file in the repo to diverge.

Yeah, currently it is the direct edit approach. I've purposefully left out the dynamic fields (like `last_synced_hash`, `last_update_date` and `ignore_last_import_time`) as they will be frequently updated or changing them in the config file will mess with the configuration. This script only syncs the attributes that are in the source file. 

It may be worth writing some sort of protection script to prevent a pull request accidentally including a dynamic field, or also handle flipping `import_last_import_time` back to false, but for now, direct edit is the approach.

andrewpollock:
> It may be worth writing some sort of protection script to prevent a pull request accidentally including a dynamic field

I suggest loudly failing if any of the dynamic fields are encountered in the input YAML file.

jess-lowe:
> > It may be worth writing some sort of protection script to prevent a pull request accidentally including a dynamic field
> 
> I suggest loudly failing if any of the dynamic fields are encountered in the input YAML file.

Done! Now on Line 116-119 :) It actually helped me catch a strangler ignore_last_import_time field so definitely worth putting it in there.

PR #1942, opened 2024-01-30
Refactored Reput Script and GIT ecosystem 

jess-lowe:
Needed to be able to run on prod as well.


tools/datafix/reput_bugs.py
```diff
@@ -14,40 +14,12 @@
...
-      "--project",
-      action="store",
-      dest="project",
-      default="oss-vdb-test",
-      help="GCP project to operate on")
-  args = parser.parse_args()
-
-  client = datastore.Client(project=args.project)
+def main(arg) -> None:
+  client = datastore.Client(project=arg.project)
```
another-rex, line 18: Huh I just noticed this is using both datastore.Client and ndb. Can we just use ndb for the query? 

e.g. Bug.query(...)

tools/datafix/reput_bugs.py
```diff
@@ -14,40 +14,12 @@
...
-  parser.add_argument(
-      "--project",
-      action="store",
-      dest="project",
-      default="oss-vdb-test",
-      help="GCP project to operate on")
-  args = parser.parse_args()
-
-  client = datastore.Client(project=args.project)
+def main(arg) -> None:
```
another-rex, line 17: nit: arg -> args

tools/datafix/reput_bugs.py
andrewpollock: FYI, I think there's a degree of cargo-culting from other scripts I've written, and from memory, utilising models.py and the NDB interface versus the Datastore interface greatly complicated things from a Python dependency/packaging perspective? 

Like for starters, I think the script can't live in this directory and would need to be run from within a Pipenv as it has a dependency on the osv Python package as a result.

The initial script I'd written was to solve a particular problem, and it Just Worked, and subsequently became the skeleton for other such Datastore manipulation scripts. I didn't wish to spend the time revisiting the overall implementation.

That's the back story to the existing implementation. Happy to be corrected on the complexity.

another-rex: Hmm I kind of see pros and cons to using ndb, it's a good point that having the osv as a dependency complicates running the script, but since the script lives in this repository anyway, I don't think it matters too much since you'll have the library right there.

The main benefit of using ndb is that we get (partial) type checking, so that if we change/update the type of Bug in the future the script will fail instead of doing something unexpected, and refactoring tools can automatically update the script as well.

andrewpollock: > The main benefit of using ndb is that we get (partial) type checking, so that if we change/update the type of Bug in the future the script will fail instead of doing something unexpected, and refactoring tools can automatically update the script as well.
> 
> I think that's enough of a benefit to outweigh the need to have the python environment loaded when running the script.

It certainly sounds like a compelling argument. Am I right in the code would need to move under the osv directory, or am I misunderstanding that nuance of Python dependencies?

I'd initially wanted to continue using the tools directory for standalone, you know, tools... These have no business winding up in the osv Python package.

another-rex (approved):
LGTM!

osv/models.py
```diff
@@ -374,14 +374,14 @@ def _pre_put_hook(self):  # pylint: disable=arguments-differ
         for pkg in self.affected_packages
         if pkg.package.ecosystem
     }
-
-    for pkg in self.affected_packages:
-      for r in pkg.ranges:
-        if r.type == 'GIT':
-          ecosystems_set.add('GIT')
+    if not ecosystems_set:
```
another-rex, line 377:
```suggestion
    # Only attempt to add the Git ecosystem if there are no existing ecosystems present
    if not ecosystems_set:
```

another-rex (approved):

another-rex (approved):

PR #1949, opened 2024-01-31
Add invalid repos based on pivot table analysis of 2024 FixUnesolvable

andrewpollock:
These all stood out as not the canonical software repo based on a quick eyeball of the now available 2024 conversion run results.


another-rex (approved):

oliverchang (approved):
LGTM! But can we explore some kind of automation (or ML) to automatically filter out these this year? It doesn't seem sustainable to have you doing this all the time.

cuixq (approved):

hogo6002 (approved):

andrewpollock:
> LGTM! But can we explore some kind of automation (or ML) to automatically filter out these this year? It doesn't seem sustainable to have you doing this all the time.

Yeah I intend to, eventually. It's far from the top of the list though.

There's a very big trade off between the amount of engineering time and effort it will take to implement such a thing to a good enough standard of accuracy, versus the amount of effort it takes to perform the analysis and create a PR like this one, particularly done periodically to keep on top of the ones that creep in.

Once there's a rigorous way to use the GitHub APIs, introspecting the repos will assist greatly with automation.

PR #1950, opened 2024-01-31
Remove erroneously added repo

andrewpollock:
PR #1949 added this repo in error, so remove it again.

Thanks to @another-rex for picking up on it.


andrewpollock:
I ran the full list through

```
for repo in $(cat /tmp/check_me | grep github.com); do 
echo $repo
gh api /repos/${repo#https://github.com/}/languages
done
```

and nothing else that has languages detected stood out as not belonging

another-rex (approved):

PR #1965, opened 2024-02-04
SourceRepository Sync 

jess-lowe:
Added some documentation around what each configuration means for SourceRepository and set up cloud build trigger bash script to run the source repository updater no-dry-run.

### DIFF
--- a/osv/models.py
+++ b/osv/models.py
@@ -793,7 +793,7 @@ class SourceRepository(ndb.Model):
   bucket = ndb.StringProperty()
   # Vulnerability data not under this path is ignored by the importer.
   directory_path = ndb.StringProperty()
-  # Last synced hash for SourceRepositoryType.GIT.
+  # Last synced hash for SourceRepositoryType.GIT & REST.
   last_synced_hash = ndb.StringProperty()
   # Last date recurring updates were requested.
   last_update_date = ndb.DateTimeProperty()
--- a/osv/models.py
+++ b/osv/models.py
@@ -793,7 +793,7 @@ class SourceRepository(ndb.Model):
   bucket = ndb.StringProperty()
   # Vulnerability data not under this path is ignored by the importer.
   directory_path = ndb.StringProperty()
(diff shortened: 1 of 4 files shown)


osv/models.py
```diff
@@ -793,7 +793,7 @@ class SourceRepository(ndb.Model):
   bucket = ndb.StringProperty()
   # Vulnerability data not under this path is ignored by the importer.
   directory_path = ndb.StringProperty()
-  # Last synced hash for SourceRepositoryType.GIT.
+  # Last synced hash for SourceRepositoryType.GIT & REST.
```
another-rex, line 796: Where is `last_synced_hash` used for REST?

tools/sourcerepo-sync/source_repo_default.yaml
```diff
@@ -1,21 +1,41 @@
...
+  # The username to use for SSH auth for SourceRepositoryType.GIT.
   repo_username: null
+  # The API endpoint for SourceRepositoryType.REST_ENDPOINT that 
+  # returns all of the vulnerability files to be updated (minimum id and modified dates)
   rest_api_url: null
+  # SourceRepositoryType - 0 for git, 1 for GCS, 2 for REST API
   type: 0
-  versions_from_repo: False
+  # Whether to populate "affected[].versions" from Git ranges.
+  versions_from_repo:
```
another-rex, line 41: This is missing a True/False value

another-rex (commented):
Nice! Let's merge this after the release tomorrow.

osv/models.py
jess-lowe: Ah maybe it's not used. Got it confused with the sha256 val. Removing.

tools/sourcerepo-sync/source_repo_default.yaml
jess-lowe: fixed.

another-rex (approved):

PR #1973, opened 2024-02-07
Revert "Test source repo sync update trigger"

another-rex:
Reverts google/osv.dev#1972

### DIFF
--- a/source_test.yaml
+++ b/source_test.yaml
@@ -266,16 +266,3 @@
   key_path: 'OSV'
   repo_username: 'git'
 
-- name: 'test'
-  versions_from_repo: True
-  type: 0
-  ignore_patterns: ['allowlist.json']
-  repo_url: 'blah'
-  detect_cherrypicks: False
-  extension: '.json'
-  db_prefix: 'TEST-'
-  ignore_git: False
-  link: 'blah'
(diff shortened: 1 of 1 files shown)


andrewpollock (approved):

source_test.yaml
```diff
@@ -266,16 +266,3 @@
   key_path: 'OSV'
   repo_username: 'git'
 
-- name: 'test'
-  versions_from_repo: True
-  type: 0
-  ignore_patterns: ['allowlist.json']
-  repo_url: 'blah'
```
andrewpollock, line 273: I'm assuming this is the culprit for a recent spate of

`ValueError: Source repository link must end with /` that just showed up in staging...

jess-lowe: Yup. Sorry about that 😂

jess-lowe (approved):
LGTM

source_test.yaml
andrewpollock: Potential cherry-on-top here would be to add some validation so that invalidity doesn't get checked in :-)

jess-lowe: Just added a pr to merge to this that should be able to fix that :)

PR #1974, opened 2024-02-07
Validate 'link' attribute has the ending '/'

jess-lowe:


andrewpollock (approved):
LGTM, thanks!

Are there any other opportunities for validation? This code is still going to be run post-merge, though right? So while it's stopping the rot from propagating further down the pipeline, it's not stopping it from being checked in in the first place.

jess-lowe:
> LGTM, thanks!
> 
> Are there any other opportunities for validation? This code is still going to be run post-merge, though right? So while it's stopping the rot from propagating further down the pipeline, it's not stopping it from being checked in in the first place.

Yeah true, maybe I can add a testcase to check for it so we know the build will fail before it merges?

another-rex:
Oh oops

another-rex:
Can you retarget this to

PR #1978, opened 2024-02-09
Add checks to ensure bad source repo changes are detected before push

jess-lowe:
This will just dry run the datastore updates.


tools/sourcerepo-sync/source_sync.py
```diff
@@ -46,93 +53,107 @@ def main() -> None:
...
-      change_flag = True
-    if args.verbose:
-      name = repo['name']
-      print(f'Found diff in {name}: {attr} - {repo[attr]} != {ds_repo[attr]}')
-    entity.update({attr: repo[attr]})
+  validate_repository(local_sourcerepos)
 
-  if change_flag and not args.dryrun:
-    client.put(entity)
+  if not args.validate:
```
another-rex, line 126: Let's invert this to 
```
if args.validate:
  return
```

tools/sourcerepo-sync/run_tests.sh
```diff
@@ -0,0 +1,20 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+export PIPENV_IGNORE_VIRTUALENVS=1
+pipenv sync
+
+pipenv run python source_sync.py --kind SourceRepository --project oss-vdb --file ../../source.yaml --dry-run --verbose --validate
```
another-rex, line 19: Is the dry-run flag here just to be extra safe?

tools/sourcerepo-sync/source_sync.py
```diff
@@ -46,93 +53,107 @@ def main() -> None:
...
+    if args.verbose:
+      print(f'Validated {len(repository)} source repositories')
 
-  local_sourcerepos_names = {repo['name'] for repo in local_sourcerepos}
-  # If the source repo is not in the local yaml, delete it
-  for ds_repo in ds_repos:
-    if ds_repo['name'] not in local_sourcerepos_names:
+  def create_sourcerepo(repo):
+    """Create a new source repo."""
+    if args.file.startswith('../../'):
```
another-rex, line 83: I think it's better to check for if the working directory is `sourcerepo-sync` instead.

another-rex (commented):
Thanks! Added some comments

tools/sourcerepo-sync/run_tests.sh
jess-lowe: Yeah.

tools/sourcerepo-sync/source_sync.py
jess-lowe: Done.

another-rex (approved):
LGTM

cloudbuild.yaml
```diff
@@ -82,6 +82,12 @@ steps:
     - CLOUDBUILD=1
   waitFor: ['init']
 
+- name: 'gcr.io/oss-vdb/ci'
+  id: 'sourcerepo-sync-tests'
+  dir: 'tools/sourcerepo-sync'
+  args: ['bash', '-ex', 'run_tests.sh']
+  waitFor: ['init']
```
michaelkedar, line 89: @jess-lowe @another-rex I think this has a race condition when running `pipenv sync` with `lib-tests` (since they use the same environment)

It might need to be changed to
```yaml
waitFor: ['init', 'lib-tests']
```
(and maybe move this below lib-tests & add a comment  so the relation is easier to see)

PR #1983, opened 2024-02-12
Fix cloudbuild race condition issue.

jess-lowe:
Fix w.r.t: https://github.com/google/osv.dev/pull/1978#discussion_r1485746544


michaelkedar (approved):

michaelkedar:
Thanks for this!

PR #1995, opened 2024-02-16
Fix blog code block formatting

another-rex:
Fix blog code block css for the latest CURL update post


cuixq (approved):

oliverchang:
While we're here, could we add a link to https://google.github.io/osv.dev/rest-api-contribution/ from the blog post?

PR #2001, opened 2024-02-19
Add placeholder ecosystem for Ubuntu.

oliverchang:


osv/ecosystems/_ecosystems.py
```diff
@@ -111,6 +112,10 @@ def get(name: str) -> Ecosystem:
     # TODO(michaelkedar)
     return OrderingUnsupportedEcosystem()
 
+  if name.startswith('Ubuntu:'):
+    # TODO(michaelkedar)
```
andrewpollock, line 116: Maybe don't namedrop Michael in the copypasta?

andrewpollock (approved):

michaelkedar (approved):

osv/ecosystems/_ecosystems.py
oliverchang: Done, I've replaced all occurrences with "unassigned", which is much more accurate

PR #2005, opened 2024-02-21
Use pygit2 enums

michaelkedar:
pygit2 1.14.0 deprecated the `GIT_*` constants in favour of new `pygit2.enums` (which also broke our usage of `pygit2.credentials.GIT_*`).


oliverchang (approved):
thanks!!

andrewpollock (approved):
It's annoying that this breakage appears to have made it through to production undiscovered in staging?

Is the timeline of how the breakage snuck in understood? Are there any gaps or deficiencies in how we're operating that need to be closed to prevent a repeat of such breakage?

michaelkedar:
> It's annoying that this breakage appears to have made it through to production undiscovered in staging?
> 
> Is the timeline of how the breakage snuck in understood? Are there any gaps or deficiencies in how we're operating that need to be closed to prevent a repeat of such breakage?

I think the credentials code is only run on repos that need authentication, and the only one that needs it is the oss-fuzz repo, which isn't on the testing environment.

PR #2014, opened 2024-02-26
Revert spicy-sections sha back

hogo6002:
Not sure why the package sha is updated while the version remains the same.
PR https://github.com/google/osv.dev/pull/2012/files


michaelkedar (approved):
This has happened before #1909 
I wonder where renovate is getting the hash from :thinking:

PR #2019, opened 2024-02-27
Backup AffectedCommits in a separate bucket.

oliverchang:
The existing backup doesn't easily distinguish between entity types. This makes it easier for internal consumers to use our AffectedCommits data.

Using a different bucket here so we can set a different lifecycle policy (3 days TTL) than our existing backup bucket (60 days TTL).

### DIFF
--- a/deployment/terraform/environments/oss-vdb-test/main.tf
+++ b/deployment/terraform/environments/oss-vdb-test/main.tf
@@ -3,13 +3,15 @@ module "osv_test" {
 
   project_id = "oss-vdb-test"
 
-  public_import_logs_bucket     = "osv-test-public-import-logs"
-  vulnerabilities_export_bucket = "osv-test-vulnerabilities"
-  logs_bucket                   = "osv-test-logs"
-  cve_osv_conversion_bucket     = "osv-test-cve-osv-conversion"
-  debian_osv_conversion_bucket  = "osv-test-debian-osv"
-  backups_bucket                = "osv-test-backup"
-  backups_bucket_retention_days = 5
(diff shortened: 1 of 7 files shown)


andrewpollock (approved):

another-rex (approved):
Does terraform need to be updated to add the new bucket?

oliverchang:
> Does terraform need to be updated to add the new bucket?

Forgot about that. Thanks for reminding me!

deployment/terraform/environments/oss-vdb-test/main.tf
```diff
@@ -3,13 +3,15 @@ module "osv_test" {
...
-  backups_bucket                = "osv-test-backup"
-  backups_bucket_retention_days = 5
+  public_import_logs_bucket                      = "osv-test-public-import-logs"
+  vulnerabilities_export_bucket                  = "osv-test-vulnerabilities"
+  logs_bucket                                    = "osv-test-logs"
+  cve_osv_conversion_bucket                      = "osv-test-cve-osv-conversion"
+  debian_osv_conversion_bucket                   = "osv-test-debian-osv"
+  backups_bucket                                 = "osv-test-backup"
+  backups_bucket_retention_days                  = 5
+  affected_commits_backups_bucket                = "osv-test-affected-commits"
```
oliverchang, line 13: These buckets already exist. Remind me if that going to cause problems @michaelkedar  ?

michaelkedar: I don't remember if buckets cause problems for these.
I guess we could just try and just `terraform import` if it fails.

michaelkedar (approved):

Issue #1467, opened 2023-07-18
Set up a cleanup cron job that checks all the entries in the database for existance


another-rex:
Checks the source field in the database to see if they still exist (does not return 404), if not, mark it as withdrawn.

rajbhoyar729:
Can you assign me this issue i want to work on it

another-rex:
Hey @rajbhoyar729, sorry missed this comment for a long time. This is more of a internal task as you would need access to the database backend to be able to adequately  work on it.

another-rex:
#1859 is another example of this being potentially useful.

rajbhoyar729:
Can you guide me how can I do it

andrewpollock:
@rajbhoyar729 your enthusiasm for contributing to OSV has been noticed 🥇 
This particular issue is difficult for external contributors who don't have access to our GCP infrastructure (that's what we cooked up the `needs-gcp-infra-access` to.

Please take a look at [the issues tagged with `good first issue`](https://github.com/google/osv.dev/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22). These are more suitable for folks like yourself. You may also like to [contribute to OSV Scanner](https://github.com/google/osv-scanner/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22)

Stay enthusiastic!

andrewpollock:
Manually ran this in Production:

```
5220 Bugs in Datastore considered deleted from GCS for cve-osv
23 Bugs in Datastore considered deleted from GCS for debian-dsa
19 Bugs in Datastore considered deleted from GCS for go
```

andrewpollock:
I think we can call this done finally.

PR #2030, opened 2024-03-05
Handle GCS bucket record deletion

andrewpollock:
Detect and plumb through to worker for deletion from Datastore Bugs that are no longer present in GCS.

Related changes (#2023, #2029) to the NVD CVE generation cause CVEs no longer being generated (due to changes in the heuristics like that made in #1707) to remain existing in GCS.

This PR addresses cases like this and the need identified in #1467 by adding a deletion phase to the importing of new/updated records. The functionality is flag-protected, it won't go live in Production until a new `--delete` flag is included so in the execution of the importer.

Incidental changes:

- use the GCS bucket `directory_path` to efficiently filter the blobs returned when listing bucket contents
- make blob retrieval resilient to blob generation change between blob listing and blob retrieval (this can happen if `combine-to-osv` happens to have run in between these two points in time)
- fix a behavior inconsistency with schema validation not being performed when `ignore_last_import_time` is in effect (addressing head scratching TODO from @michaelkedar)
- tidy up the existing tests, making them more readable and debuggable
- add a slow manual test against live data in staging to validate real-world behavior and run time (this adds ~13 minutes to an import run on just the CVE GCS bucket)


cloudbuild.yaml
```diff
@@ -50,6 +50,7 @@ steps:
   dir: docker/importer
   args: ['bash', '-ex', 'run_tests.sh']
   env:
+    - CLOUD_BUILD=1
```
michaelkedar, line 53: Also add this to the tests in [`staging.yaml`](https://github.com/google/osv.dev/blob/c938a662b814a54f43deb71f0ab6ae682515504a/deployment/staging.yaml#L18-L19)

(more tech debt: we need to remove/consolidate that with `build-and-stage.yaml` eventually :slightly_smiling_face:)

andrewpollock: Like in 01b8388 ?

docker/importer/importer.py
```diff
@@ -536,6 +655,23 @@ def process_updates(self, source_repo: osv.SourceRepository):
...
+    """Process source record deletions by withdrawing them."""
+    if source_repo.type == osv.SourceRepositoryType.GIT:
+      # TODO: To be implemented.
+      return
+
+    if source_repo.type == osv.SourceRepositoryType.BUCKET:
+      self._process_deletions_bucket(source_repo)
+      return
+
+    if source_repo.type == osv.SourceRepositoryType.REST_ENDPOINT:
```
oliverchang, line 668: We should add a TODO to refactor some of this code, before we even implement the missing support for REST/git.
 
Rather than have all these if statements for importing/deleting, we should start defining some SourceRetriever classes that define the source-type-specific ways to retrieve records.

i.e. something along the lines of:

```python

class SourceRetriver

  def list() -> list[str]:
    """Return list of vuln locators for the source."""

  def get(locator: str) -> Vulnerability:
    """Return a vulnerability for the given locator."""
```

This way we can consolidate the import and deletion logic, and handle the differences via this abstraction.

docker/importer/importer.py
```diff
@@ -455,6 +476,98 @@ def thread_init():
...
+          blob.name, blob.bucket,
+          generation=None).download_as_bytes(_client_store.storage_client)
+
+      try:
+        vulns = osv.parse_vulnerabilities_from_data(
+            blob_bytes,
+            os.path.splitext(blob.name)[1],
+            strict=self._strict_validation)
+        for vuln in vulns:
+          return vuln.id
```
oliverchang, line 531: Should we be returning **all** vulns here and handling them accordingly? It seems incorrect to make this assumption here when in other parts (e.g. the worker), we handle this case: https://github.com/google/osv.dev/blob/5a8605df882d8a24c086655338e5f43fbddfe627/docker/worker/worker.py#L385

docker/importer/importer.py
```diff
@@ -455,6 +476,98 @@ def thread_init():
...
+
+    with ThreadPoolExecutor(
+        _BUCKET_THREAD_COUNT, initializer=thread_init) as executor:
+      vulns_in_gcs = [
+          v for v in executor.map(convert_blob_to_vuln, listed_blobs) if v
+      ]
+
+      # diff
+      vulns_to_delete = list(
+          filter(lambda v: v[0] not in vulns_in_gcs, vuln_ids_for_source))
```
oliverchang, line 554: per https://google.github.io/styleguide/pyguide.html#210-lambda-functions

Can we rewrite this as


```python
[v for v in vuln_ids_for_source if v[0] not in vulns_in_gcs]
```

docker/importer/importer.py
```diff
@@ -455,6 +476,98 @@ def thread_init():
...
+    logging.info("Begin processing bucket for deletions: %s", source_repo.name)
+
+    # Get all the existing non-withdrawn Bug IDs for
+    # source_repo.name in Datastore
+    query = osv.Bug.query()
+    query = query.filter(osv.Bug.source == source_repo.name)
+    result = list(query.fetch(keys_only=False))
+    result.sort(key=lambda r: r.id())
+    vuln_ids_for_source = [
+        (r.id(), r.source_id.split(':')[1]) for r in result if not r.withdrawn
```
oliverchang, line 500: nit: use a namedtuple here for readability.

docker/importer/importer.py
```diff
@@ -455,6 +476,98 @@ def thread_init():
...
+      _client_store.storage_client = storage.Client()
+
+    with ThreadPoolExecutor(
+        _BUCKET_THREAD_COUNT, initializer=thread_init) as executor:
+      vulns_in_gcs = [
+          v for v in executor.map(convert_blob_to_vuln, listed_blobs) if v
+      ]
+
+      # diff
+      vulns_to_delete = list(
```
oliverchang, line 553: Does this still have to be indented under the context of the ThreadPoolExecutor?

docker/importer/importer.py
```diff
@@ -536,6 +655,23 @@ def process_updates(self, source_repo: osv.SourceRepository):
     logging.error('Invalid repo type: %s - %d', source_repo.name,
                   source_repo.type)
 
+  def process_deletions(self, source_repo: osv.SourceRepository):
+    """Process source record deletions by withdrawing them."""
+    if source_repo.type == osv.SourceRepositoryType.GIT:
+      # TODO: To be implemented.
```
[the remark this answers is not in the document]
andrewpollock: That's right, I remember that... Shall I reinstate https://github.com/google/osv.dev/pull/2031/commits/5fd32773fae35b32b9280e8a4c8e380a4647cdea as part of this PR and add a similar safeguard?

docker/importer/importer.py
```diff
@@ -455,6 +476,98 @@ def thread_init():
 
     logging.info('Finished processing bucket: %s', source_repo.name)
 
+  def _process_deletions_bucket(self, source_repo: osv.SourceRepository):
+    """Process deletions from a GCS bucket source.
+
+    This validates the continued existence of every Bug in Datastore (for the
+    given source) against every bug currently in that source's GCS bucket,
+    calculating the delta. The bugs determined to have been
+    deleted from GCS are then flagged for deletion from Datastore by the worker.
```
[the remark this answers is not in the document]
andrewpollock: That was an imprecise statement, firmed up in 851548b

Looking at https://github.com/google/osv.dev/blob/3a4727be7e1b7fa966447ec5f4a2408997324491/docker/worker/worker.py#L420-L436, I keep conflating `osv.BugStatus.INVALID` as "withdrawn" and that's not the case...

Looking at the behemoth that is the `_pre_put_hook()` for a Bug in https://github.com/google/osv.dev/blob/3a4727be7e1b7fa966447ec5f4a2408997324491/osv/models.py#L355 I see that we set withdrawn bugs as INVALID, but not INVALID bugs as withdrawn, so, should I

- adjust the _pre_put_hook() to be reciprocal when a Bug is marked as INVALID but not withdrawn, and set a withdrawn timestamp?
- adjust _handle_deleted() to set a `withdrawn` timestamp?
- something else?

docker/importer/importer.py
```diff
@@ -455,6 +476,98 @@ def thread_init():
...
+      # diff
+      vulns_to_delete = list(
+          filter(lambda v: v[0] not in vulns_in_gcs, vuln_ids_for_source))
+
+      # sanity check: deleting >10% of the records for source in Datastore is
+      # probably worth flagging for review.
+      if len(vulns_to_delete) / len(vuln_ids_for_source) > 10.0:
+        logging.error(
+            'Cowardly refusing to delete %d missing records from '
+            'GCS for: %s', len(vulns_to_delete), source_repo.name)
```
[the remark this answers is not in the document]
andrewpollock: Done in 0709f39

docker/importer/importer.py
andrewpollock: TODO added in 5fc725a

docker/importer/importer.py
andrewpollock: Done in e79c4b9

docker/importer/importer.py
andrewpollock: Done in 3bfdddd

docker/importer/importer.py
andrewpollock: I believe so, due to the scope of the variables involved?

docker/importer/importer.py
oliverchang: I think all we need to do here is to s/convert_blob_to_vuln/convert_blob_to_vulns/, have it return all the vulns, and handle that accordingly by concatenating them all into `vulns_in_gcs` ?

docker/importer/importer.py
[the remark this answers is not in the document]
oliverchang: Let's do "adjust _handle_deleted() to set a withdrawn timestamp?", as that feels like the more appropriate place for us to slap on a timestamp. _handle_deleted is the explicit context where we are making a determination about when the withdrawing is happening, as opposed to any random put().

docker/importer/importer.py
oliverchang: I think there's only function level scoping in Python. the "with" does not create its own scope.

andrewpollock: Yeah, wow, I've had that wrong in my head like forever... Fixed in d5ece18

docker/importer/importer.py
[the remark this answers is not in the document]
andrewpollock: As I  was creating [4568e92](https://github.com/google/osv.dev/pull/2030/commits/4568e92b30f607bdd560e673cf74dc2fe230c425) I also realized there's already a whole other `invalid` pubsub message and now I'm wondering if that is more appropriate to use? Ugh.

docker/importer/importer.py
andrewpollock: Sorry, I'm confused. That's my understanding of what it *is* doing currently, via the TheadPoolExecutor?

Are you saying to not use `osv.parse_vulnerabilities_from_data()` to do this?

andrewpollock: I believe I've addressed this in 635c4f3

docker/importer/importer.py
```diff
@@ -455,6 +481,104 @@ def thread_init():
...
+        return None
+
+    # Setup storage client
+    def thread_init():
+      _client_store.storage_client = storage.Client()
+
+    with ThreadPoolExecutor(
+        _BUCKET_THREAD_COUNT, initializer=thread_init) as executor:
+      vulns_in_gcs = [
+          v for v in executor.map(convert_blob_to_vulns, listed_blobs) if v
```
oliverchang, line 558: Is this correct? 

`convert_blobs_to_vulns` returns a list[str] now, which means that `vulns_in_gcs` is now a list of lists? 

Do we need to concatenate all the results here instead? i.e. `sum(v for v in executor.map(convert_blob_to_vulns, listed_blobs), [])`

docker/importer/importer.py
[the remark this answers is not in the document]
andrewpollock: I'll separate any reinstatement of the functionality removed in #829 to minimize the complexity of this PR.

Filed #2113 to track that work.

docker/importer/importer.py
```diff
@@ -455,6 +481,104 @@ def thread_init():
...
+    storage_client = storage.Client()
+    # Get all of the existing records in the GCS bucket
+    # (to get their IDs for checking against Datastore)
+    listed_blobs = list(
+        storage_client.list_blobs(
+            source_repo.bucket, prefix=source_repo.directory_path))
+
+    import_failure_logs = []
+
+    def convert_blob_to_vulns(blob: storage.Blob) -> Optional[Tuple[str]]:
```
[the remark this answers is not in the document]
andrewpollock: Done in fabd741

docker/importer/importer.py
andrewpollock: This has been overhauled in fabd741, so the comment may not still be applicable. I'm still testing the current code.

docker/importer/importer.py
```diff
@@ -455,6 +531,107 @@ def thread_init():
...
+
+    If the delta is too large, something undesirable has been assumed to have
+    happened and further processing is aborted.
+
+    Args:
+      source_repo: the osv.SourceRepository being operated on
+      threshold: the percentage delta considered safe to delete
+    """
+
+    logging.info("Begin processing bucket for deletions: %s", source_repo.name)
```
oliverchang, line 552: nit: single quotes for strings.

oliverchang (commented):
this generally looks good, thanks! 

how's the performance of this running on the staging CVE bucket? if this ends up being slow and impacts our freshness SLO, we may need to consider moving this to e.g. a separate "cleanup" cron job instead of sticking this in the importer.

docker/importer/importer.py
andrewpollock: Fixed in 39f0941

andrewpollock:
> how's the performance of this running on the staging CVE bucket? if this ends up being slow and impacts our freshness SLO, we may need to consider moving this to e.g. a separate "cleanup" cron job instead of sticking this in the importer.

Yeah I think based on my testing so far, while not running on the same sized GKE node as it would in production, running it as part of the regular importing run isn't looking tenable. I think we'll need to have a separate say daily cronjob that runs it in deletion mode. Good thing I've got all of that behind a flag :-)

docker/importer/importer.py
andrewpollock: I think this is no longer relevant.

docker/importer/importer.py
```diff
@@ -170,7 +183,11 @@ def run(self):
       if source_repo.name == 'oss-fuzz':
         self.process_oss_fuzz(source_repo)
 
-      self.process_updates(source_repo)
+      self.validate_source_repo(source_repo)
+      if not self._delete:
+        self.process_updates(source_repo)
+      if self._delete:
```
oliverchang, line 189: nit: just if / else would simplify this.

```python
if self._delete:
  self.process_deletions(...)
else:
  self.process_updates(...)
```

docker/importer/importer.py
```diff
@@ -258,6 +275,46 @@ def schedule_regular_updates(self, repo, source_repo: osv.SourceRepository):
...
+    blob_bytes = storage.Blob(
+        blob.name, blob.bucket, generation=None).download_as_bytes(client)
+
+    vuln_ids = []
+    try:
+      vulns = osv.parse_vulnerabilities_from_data(
+          blob_bytes,
+          os.path.splitext(blob.name)[1],
+          strict=self._strict_validation)
+    except Exception:
```
oliverchang, line 312: Should we log an error here?

docker/importer/importer.py
```diff
@@ -598,6 +802,15 @@ def main():
       '--strict_validation',
       help='Fail to import entries that does not pass validation',
       default=False)
+  parser.add_argument(
+      '--delete',
+      help='Propagate record deletions from source to Datastore',
```
oliverchang, line 807: Should we clarify that this will skip the usual import mode? i.e. this is running in a deletion only mode?

oliverchang (approved):
LGTM with some minor nits.

docker/importer/importer.py
andrewpollock: Added in 993d924

docker/importer/importer.py
andrewpollock: Clarified in 94f2e87

docker/importer/importer.py
andrewpollock: Simplified in 2abd779

oliverchang (approved):
LGTM

PR #2031, opened 2024-03-05
Enhance human link support

andrewpollock:
Support more complex human-friendly upstream bug URLs by allowing templating rather than simple appending of the the bug ID to a URL.

This allows AlmaLinux and Curl human-friendly links to be added to the frontend UI in addition the what's already there.

Also add an FAQ about data fixes, which can be referenced from the bug template


gcp/appengine/frontend_handlers.py
```diff
@@ -300,7 +300,8 @@ def add_source_info(bug, response):
   response['source'] = source_repo.link + source_path
   response['source_link'] = response['source']
   if source_repo.human_link:
```
another-rex, line 302: Can we rename this to `human_link_template`?

another-rex (commented):
Thanks! Mostly LGTM, we just need to be careful/aware that "code" is now in the source.yaml files, and is being executed.

gcp/appengine/frontend_handlers.py
andrewpollock: > Can we rename this to `human_link_template`?

I'm unsure? 

I can rename it at https://github.com/google/osv.dev/blob/5a8605df882d8a24c086655338e5f43fbddfe627/osv/models.py#L831 but I'm unclear of the ramifications on the existing data in Datastore of doing so, and how to handle that transition?

Alternatively, I could add a new property called `human_link_template` and adjust this PR to support both templated links and the previous behaviour?

gcp/appengine/frontend_handlers.py
```diff
@@ -366,7 +367,7 @@ def osv_query(search_string, page, affected_only, ecosystem):
     query = query.filter(osv.Bug.search_indices == search_string.lower())
 
   if affected_only:
-    query = query.filter(osv.Bug.has_affected == True)  # pylint: disable=singleton-comparison
+    query = query.filter(osv.Bug.has_affected is True)  # pylint: disable=singleton-comparison
```
oliverchang, line 370: does this actually work with ndb? 

I don't think this worked in the past, unless a more recent version of ndb has changed this.

source_test.yaml
```diff
@@ -9,6 +9,7 @@
   extension: '.json'
   db_prefix: 'ALBA-'
   ignore_git: False
+  human_link: 'https://errata.almalinux.org/{{ bug.ecosystem[1].split(":")[1] }}/{{ bug.id().replace(":", "-", 1) }}.html'
```
oliverchang, line 12: Hm, this introduces a dependency on ndb/our Datastore entity format in a way that isn't immediately obvious. if we change `Bug` in the future, we'd have to remember to update this.

Can we instead just support simpler string replacements here?

i.e. we can define "{{BASE_ECOSYSTEM}}", "{{BUG_ID}}" as strings that can be used here in the yaml. Our source updating implementation can just do the replacement on the fly. 

This way, all the assumptions about the bug/Datastore entity structure are made in code, which can also be more easily tested.

gcp/appengine/frontend_handlers.py
another-rex: Ah good point, let's not rename this for now then, as long as people are aware that this is a template.

gcp/appengine/frontend_handlers.py
andrewpollock: No, I blindly acted on a warning VS Code spat out, my bad. Fixed in [45432a6](https://github.com/google/osv.dev/pull/2031/commits/45432a66c566af0072022832cb3a526e56d9be22)

source_test.yaml
andrewpollock: > if we change `Bug` in the future, we'd have to remember to update this.

Is adding some Big Fat Comments in models.py sufficient to address this concern?

> we can define "{{BASE_ECOSYSTEM}}", "{{BUG_ID}}" as strings that can be used here in the yaml. Our source updating implementation can just do the replacement on the fly.

I think you're saying to pass specific kwargs into the template, rather than the whole bug? I had started doing that, and then decided to pass in the whole bug for maximum future flexibility without further changes, but happy to add those two kwargs for now instead if I'm understanding correctly?

oliverchang: Yes, I think kwargs would be preferred here. The flexibility comes at a cost of testability and potentially making things a bit brittle.

andrewpollock: One challenge with doing this generically is the values and how they're arrived at vary by ecosystem, so it shifts ecosystem-specific logic further away from the ecosystem definitions and into the handler code. That was part of the reasoning behind not doing it this way in the first place...

andrewpollock: Feedback address in https://github.com/google/osv.dev/pull/2031/commits/5fd32773fae35b32b9280e8a4c8e380a4647cdea

gcp/appengine/frontend_handlers.py
```diff
@@ -300,7 +300,15 @@ def add_source_info(bug, response):
...
-    response['human_source_link'] = source_repo.human_link + bug.id()
+    base_ecosystem = bug.ecosystem[0]
+    bug_id = bug.id()
+    if base_ecosystem == 'AlmaLinux':
+      base_ecosystem = bug.ecosystem[1].split(':')[1]
+      bug_id = bug_id.replace(':', '-', 1)
+    if base_ecosystem == 'GIT' and bug_id.startswith('CURL-'):
+      bug_id = bug_id.removeprefix('CURL-')
+    response['human_source_link'] = render_template_string(
+        source_repo.human_link, BASE_ECOSYSTEM=base_ecosystem, BUG_ID=bug_id)
```
oliverchang, line 311: I think there's a happy medium here for us to use flask/jinja filters here for the very ecosystem-specific bits (e.g. https://jinja.palletsprojects.com/en/3.0.x/templates/#jinja-filters.replace).

This way we don't have ecosystem-specific code here, and we don't expose internal Bug entity details in the yaml.

source_test.yaml
oliverchang: I believe there's a happy medium here. See my other comment!

gcp/appengine/frontend_handlers.py
oliverchang: To be clear about what I mean here, this would look something like:

```
{{ BUG_ID | replace("CURL-", "") }}
```

etc

andrewpollock: Got it. Addressed in f1239d5

oliverchang (approved):

PR #2034, opened 2024-03-06
Update Go version to 1.22.1

cuixq:
This PR updates Go version to 1.22.1 to resolve a few vulnerabilities. 

This PR also updates `google.golang.org/protobuf` to 1.33.0 for the same purpose.


andrewpollock (approved):
I'm not sure if https://github.com/google/osv.dev/blob/5a8605df882d8a24c086655338e5f43fbddfe627/.github/workflows/lint.yaml#L47 should also be updated, it might be worthwhile to keep these PRs comprehensively touching all the places the Go version is referenced

andrewpollock:
@michaelkedar I know we don't routinely build the ci Docker image because it takes a while, but would it be feasible to automatically build it only when content under https://github.com/google/osv.dev/tree/master/docker/ci changes? Otherwise there's the risk of us failing to do that step and our testing environment diverges from our runtime environment.

michaelkedar:
> @michaelkedar I know we don't routinely build the ci Docker image because it takes a while, but would it be feasible to automatically build it only when content under https://github.com/google/osv.dev/tree/master/docker/ci changes? Otherwise there's the risk of us failing to do that step and our testing environment diverges from our runtime environment.

We *could* have a job that builds and pushes the ci image when it changes.
The thing I don't know is how the order of operations would work.

I *think* we'd want something like:
On PR:
- If ci image changes -> rebuild and temporarily tag it for PR -> use new image to run tests.
- Otherwise, use existing ci image to run tests

Then on merge:
- If ci image had changed -> rebuild or re-label PR image to 'latest' -> use that to run tests.
- Otherwise, continue using existing ci image

But I don't really have an idea on how possible/feasible this kind of workflow is in Cloud Build.

michaelkedar (approved):

hogo6002 (approved):

another-rex (approved):

PR #2047, opened 2024-03-12
Add some logging to API queries

michaelkedar:
Log the type of query the API gets (`git`, `commit`, `ecosystem`, `invalid`) so we can better debug long-running queries.

Trying out [structured json logging](https://cloud.google.com/python/docs/reference/logging/latest/std-lib-integration#logging-json-payloads) in GCP to maybe make it more easily searchable.

### DIFF
--- a/gcp/api/server.py
+++ b/gcp/api/server.py
@@ -126,6 +126,24 @@ class OSVServicer(osv_service_v1_pb2_grpc.OSVServicer,
 
     version.
     """
+
+    # Log some information about the query with structured logging
+    qtype, ecosystem, versioned = query_info(request.query)
+    if qtype == 'ecosystem':
+      logging.info(
+          'QueryAffected for ecosystem "%s"',
+          ecosystem,
+          extra={
+              'json_fields': {
+                  'details': {
+                      'ecosystem': ecosystem,
+                      'versioned': versioned == 'versioned'
+                  }
+              }
+          })
+    else:
+      logging.info('QueryAffected for %s', qtype)
+
     page_token = None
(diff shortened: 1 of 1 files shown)


andrewpollock (approved):
Spiffy!

gcp/api/server.py
```diff
@@ -228,6 +282,23 @@ def Watch(self, request, context: grpc.ServicerContext):
...
+  If 'ecosystem', second two return values are the ecosystem name,
+  then 'versioned' or 'versionless' depending if the 'version' field is set.
+  Otherwise, last two return values are None.
+  """
+  if query.WhichOneof('param') == 'commit':
+    return 'commit', None, None
+  if not query.HasField('package'):
+    return 'invalid', None, None
+  if query.package.purl:
+    return 'purl', None, None
```
another-rex, line 297: Can we parse the purl here and turn it into an ecosystem query as well, since it's handled the same way dadtastore query-wise.

another-rex (commented):
I think we should parse the purl as well, otherwise LGTM!

gcp/api/server.py
oliverchang: +1 !

michaelkedar: purls seem to be handled differently?
https://github.com/google/osv.dev/blob/dcb6a15f10cc452b4fbde415e2cb1c1932878049/gcp/api/server.py#L882-L895

another-rex: Ah you're right, I got confused with the second filtering part parsing the git ecosystem. Let's keep it as is then.

PR #2052, opened 2024-03-13
Allow for json fields in python logs

michaelkedar:
The way Error Reporting information was being added to logs was preventing us from adding additional `json_fields` to our logs.
Changed from inserting Error Reporting information in a Log Record Factory to a Log Filter.
See: https://docs.python.org/3/howto/logging-cookbook.html#using-filters-to-impart-contextual-information

Also put the API logged fields back into the json data.

### DIFF
--- a/osv/logs.py
+++ b/osv/logs.py
@@ -17,30 +17,29 @@ import logging
 from google.cloud import logging as google_logging
 
 
-def setup_gcp_logging(service_name):
-  """Set up GCP logging and error reporting."""
+class _ErrorReportingFilter:
+  """
+  A logging filter that adds necessary json fields to error logs so that they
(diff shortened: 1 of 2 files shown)


osv/logs.py
```diff
@@ -17,30 +17,29 @@
 from google.cloud import logging as google_logging
 
 
-def setup_gcp_logging(service_name):
-  """Set up GCP logging and error reporting."""
+class _ErrorReportingFilter:
+  """
+  A logging filter that adds necessary json fields to error logs so that they
```
andrewpollock, line 22: Nit: Capitalism acronyms

```suggestion
  A logging filter that adds necessary JSON fields to error logs so that they
```

andrewpollock (approved):

PR #2053, opened 2024-03-13
Increase max instances

another-rex:
Increase the max instance count, default is 100.

We are currently running into the max instance count (or almost running into it, there is 97 instances running at the moment).


michaelkedar (approved):

oliverchang:
Do we need any quota increases for this for our GCP project?

PR #2055, opened 2024-03-14
Add trace to API logs

michaelkedar:
Adds the trace fields to most of the logs produced by the API, so it's easier correlate the logs to their requests.

Relevant links:
https://cloud.google.com/trace/docs/setup#force-trace
https://cloud.google.com/logging/docs/agent/logging/configuration#special-fields

### DIFF
--- a/gcp/api/server.py
+++ b/gcp/api/server.py
@@ -102,6 +102,31 @@ def ndb_context(func):
   return wrapper
 
 
+def trace_log_fields(context: grpc.ServicerContext) -> dict:
+  """Makes the json_field needed to associate a log with the request trace."""
+  fields = {}
+  trace_context = dict(
+      context.invocation_metadata()).get('x-cloud-trace-context')
+  if trace_context is None:
+    return fields
+
+  parts = trace_context.split('/')
+  trace_id = parts[0]
+  # We don't set the GOOGLE_CLOUD_PROJECT env var explicitly, and I can't find
+  # any confirmation on whether Cloud Run will set automatically.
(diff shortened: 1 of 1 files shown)


gcp/api/server.py
```diff
@@ -102,6 +102,31 @@ def wrapper(*args, **kwargs):
...
 
 
+def trace_log_fields(context: grpc.ServicerContext) -> dict:
+  """Makes the json_field needed to associate a log with the request trace."""
+  fields = {}
+  trace_context = dict(
+      context.invocation_metadata()).get('x-cloud-trace-context')
+  if trace_context is None:
+    return fields
+
```
another-rex, line 112:
```suggestion

# Trace context header example: "X-Cloud-Trace-Context: TRACE_ID/SPAN_ID;o=TRACE_TRUE"
```

another-rex (approved):
Nice! This will make looking at the logs so much easier!

oliverchang (approved):
LGTM, but is there a way we can  generalize this by doing this in https://github.com/google/osv.dev/blob/master/osv/logs.py instead? 

Perhaps we can set some kind of thread local storage to propagate the context to the logs.py `filter` function? 

Can be a follow up PR.

michaelkedar:
> LGTM, but is there a way we can generalize this by doing this in https://github.com/google/osv.dev/blob/master/osv/logs.py instead?
> 
> Perhaps we can set some kind of thread local storage to propagate the context to the logs.py `filter` function?
> 
> Can be a follow up PR.

Ack. Will look into this in another PR.
Though, I'm not sure this would belong in logs.py - the API would be the only thing to use it.

PR #2057, opened 2024-03-15
Set up second Cloud Run backend for batch queries (part 1)

michaelkedar:
Make a duplicate Cloud Run API backend with no concurrent requests for processing batch queries.
Don't route traffic to it yet - the traffic routing is controlled by terraform which is run before Cloud Deploy and we don't want it to completely take down our batch queries if something goes wrong.

### DIFF
--- a/deployment/clouddeploy/osv-api/run-batch.yaml
+++ b/deployment/clouddeploy/osv-api/run-batch.yaml
@@ -1,0 +1,26 @@
+apiVersion: serving.knative.dev/v1
+kind: Service
+metadata:
+  name: osv-grpc-backend-batch
+spec:
+  template:
+    metadata:
+      annotations:
+        autoscaling.knative.dev/maxScale: '100'
+    spec:
+      containers:
+      - image: osv-server
+        resources:
+          limits:
+            memory: 4Gi
+        startupProbe:
+          grpc:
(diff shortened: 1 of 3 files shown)


deployment/clouddeploy/osv-api/run-batch.yaml
```diff
@@ -0,0 +1,26 @@
+apiVersion: serving.knative.dev/v1
+kind: Service
+metadata:
+  name: osv-grpc-backend-batch
+spec:
+  template:
+    metadata:
+      annotations:
+        autoscaling.knative.dev/maxScale: '100'
```
oliverchang, line 9: should we bump this up a bit more?

oliverchang (approved):

another-rex (approved):

deployment/clouddeploy/osv-api/run-batch.yaml
michaelkedar: Currently, we hover around 100 instances serving all requests with 5 concurrency. Batch queries are < 5% of all our queries so I'd assume it'd use <25 instances (give or take).

I'll leave it at the default 100, but will keep an eye on it in case my dodgy maths it too dodgy.

PR #2064, opened 2024-03-18
Another attempt to fix logging traces

michaelkedar:
Hopefully this one works...


oliverchang (approved):

oliverchang:
hmm, looking around, we do this here for ClusterFuzz: https://github.com/google/clusterfuzz/blob/8516e38f6dcb383d3b2b97ca8c3e672954f4d7f4/src/clusterfuzz/_internal/metrics/logs.py#L509. not sure if this applies though because the logging mechanism we're using is quite different.

Issue #2095, opened 2024-04-09
dpkg repo down, causing our tests to be blocked


oliverchang:
```
--- FAIL: TestExtractGitCommit (35.11s)
    versions_test.go:605: test "Valid cGit commit URL": extractGitCommit for "https://git.dpkg.org/cgit/dpkg/dpkg.git/commit/?id=faa4c92debe45412bfcf8a44f26e827800bb24be" ('\x01') errored unexpectedly: &errors.errorString{s:"unable to determine validity of \"https://git.dpkg.org/cgit/dpkg/dpkg.git\": Head \"https://git.dpkg.org/cgit/dpkg/dpkg.git\": dial tcp 2.152.202.106:443: i/o timeout"}
    versions_test.go:608: test "Valid cGit commit URL": extractGitCommit for "https://git.dpkg.org/cgit/dpkg/dpkg.git/commit/?id=faa4c92debe45412bfcf8a44f26e827800bb24be" was incorrect, got: cves.AffectedCommit{Repo:"", Introduced:"", Fixed:"", Limit:"", LastAffected:""}, expected: cves.AffectedCommit{Repo:"https://git.dpkg.org/cgit/dpkg/dpkg.git", Introduced:"", Fixed:"faa4c92debe45412bfcf8a44f26e827800bb24be", Limit:"", LastAffected:""}
FAIL
```

oliverchang:
The repo is back.

PR #2096, opened 2024-04-09
Disable test to unblock deployments.

oliverchang:
See https://github.com/google/osv.dev/issues/2095


oliverchang:
Hmm, actually looks like https://git.dpkg.org/cgit/dpkg/dpkg.git came back as soon as I uploaded this.

oliverchang:
Yep, no longer necessary.

Issue #2089, opened 2024-04-06
Mageia vulnerabilities available in OSV


dfandrich:
The Mageia distribution now exports its vulnerability reports in OSV format. Here are some key URLs:

Source URL: `https://advisories.mageia.org/<ID>.html`
OSV Formatted URL: `https://advisories.mageia.org/<ID>.json`
Index of vulnerabilities: https://advisories.mageia.org/vulns.json
Mageia security advisories home: https://advisories.mageia.org/

We're using a PURL format analogous to the Fedora one. The index format is compatible with the one used by Go, except that the advisory modification time is not currently easily accessible in our infrastructure so it's left off.

I'm not sure what the most relevant link is for "How to contribute", but our Bugzilla instance is at https://bugs.mageia.org/ and we have a wiki page on ways to contribute to the distribution at https://wiki.mageia.org/en/Contributing.

oliverchang:
Thanks @dfandrich ! https://advisories.mageia.org/vulns.json doesn't seem to work at the moment. Is this expected to be live soon?

Also, would you be able to contribute an OSV schema definition here: https://ossf.github.io/osv-schema/#affectedpackage-field to define the ecosystem/package naming rules?

dfandrich:
I think you checked that URL during the time our servers were down due to a cooling issue in the datacentre. I can create a PR on the schema definition.

dfandrich:
I've also created #2107 on source_test.yaml. That one probably isn't useful right now without changes to the code, but it's a starting point for discussion. All but two sources (that return all advisories in a single giant file) seem to use a cloud storage directory API to retrieve a list. Mageia currently has a REST endpoint to get a list of IDs, then each desired one must be  retrieved in turn.

andrewpollock:
Leaving some notes here for future reference:

- https://advisories.mageia.org/vulns.json emits vulnerability IDs
- https://advisories.mageia.org/MGASA-2024-0129.json emits a single vulnerability in OSV format

Comparing with the two existing REST sources:

- https://curl.se/docs/vuln.json
- https://packages.wolfi.dev/os/osv/all.json

which emit an array of all the vulnerabilities.

@dfandrich how difficult would it be to stand up another endpoint that in essence
- fetches all the IDs from https://advisories.mageia.org/vulns.json
- iterates over them building up an array
- emits the array
?

dfandrich:
It wouldn't be difficult to set up an endpoint for all advisories, but it's
[...]

PR #2107, opened 2024-04-12
Add Mageia to source_test.yaml (#2089)

dfandrich:
No other source seems to use the style of index Mageia has (not even Go,
which is was derived from), so this probably isn't useful as-is without
code changes.

### DIFF
--- a/source_test.yaml
+++ b/source_test.yaml
@@ -170,6 +170,19 @@
   editable: False
   repo_username: 'git'
 
+- name: mageia
+  versions_from_repo: False
+  rest_api_url: 'https://advisories.mageia.org/vulns.json'
+  type: 2
+  directory_path: .
+  detect_cherrypicks: False
+  extension: .json
+  db_prefix: MGASA-
+  ignore_git: True
+  human_link: 'https://advisories.mageia.org/{{ BUG_ID }}.html'
+  link: https://advisories.mageia.org/
+  editable: False
+
 - name: 'malicious-packages'
   versions_from_repo: False
   type: 0


source_test.yaml
```diff
@@ -170,6 +170,19 @@
   editable: False
   repo_username: 'git'
 
+- name: mageia
+  versions_from_repo: False
+  rest_api_url: 'https://advisories.mageia.org/vulns.json'
+  type: 999
```
oliverchang, line 176: this should be `type: 2`.

dfandrich: Done

oliverchang:
@dfandrich is there any chance you'd be able to sign the CLA: https://github.com/google/osv.dev/pull/2107#issuecomment-2050884629? 

We can't merge this sadly unless you do this. Alternatively, if you prefer, we can independently commit a similar change ourselves.

dfandrich:
> @dfandrich is there any chance you'd be able to sign the CLA: #2107 (comment)?

Sorry, I completely missed this request. I'll look at it when I get a chance.

andrewpollock:
> Sorry, I completely missed this request. I'll look at it when I get a chance.

Circling back on this @dfandrich

dfandrich:
I didn't see a way to approve this without creating a Google account. But, it seems unlikely to me that this PR is even copyrightable in the first place. Just in case it is, I hereby dedicate this commit to the public domain, or under the Creative Commons CC0 license (at the choice of the reader, or where applicable). So, go ahead and use it if you'd like anyway.

dfandrich:
It doesn't look like this ever made it into prod. Is it working in staging?

andrewpollock:
@dfandrich thanks for flagging this, it looks like a PR to merge it into staging never got created, I've just sent out #3099

Issue #1048, opened 2023-02-17
Add RSS link to blog


hayleycd:
Currently we have an RSS feed, but we do not have a link that shows up on the blog.

ZhangChen199102:
I'd like to pick this issue up.
(leaving a comment here as I'm unable to assign issue to myself)

andrewpollock:
Thanks @ZhangChen199102 please ensure you've reviewed https://github.com/google/osv.dev/blob/master/CONTRIBUTING.md and any feedback on our first time contributor onboarding is very welcome :-)

PR #2115, opened 2024-04-18
Add RSS link to the page

ZhangChen199102:
This PR introduces a rss link to the site.

Issue: #1048 

i'm not quite sure about if we should just add a link tag or display a RSS icon beside the Github icon at the top right corner:
image
or maybe we want both of them? Please let me know your suggestion.


another-rex:
I think that's a good idea, though we might also want something more obvious. Can you have the RSS icon only show up for the blog page? (You can add a check for the URL in the Jinja2 template I believe)

I think it'll be confusing if it also shows on the vulnerability list page, people might think it's for subscribing to vuln advisories.

ZhangChen199102:
> I think that's a good idea, though we might also want something more obvious. Can you have the RSS icon only show up for the blog page? (You can add a check for the URL in the Jinja2 template I believe)
> 
> I think it'll be confusing if it also shows on the vulnerability list page, people might think it's for subscribing to vuln advisories.

Thanks, I've added the RSS icon to the nav bar, and it only shows for the blog pages.
Results:
image
image

Mobile version:
image

oliverchang:
Awesome! Could we also move the rss feed icon to be the left of the github icon? That way, there's more consistency with the positioning of the github icon (always the far right).

ZhangChen199102:
> Awesome! Could we also move the rss feed icon to be the left of the github icon? That way, there's more consistency with the positioning of the github icon (always the far right).

Updated RSS icon position for both of desktop and mobile views:
image

image

another-rex (approved):

Issue #2103, opened 2024-04-11
combine-to-osv doesn't reflect last modified time of component parts


andrewpollock:
**Describe the bug**
Currently, the `modified` time of an OSV record emitted by `combine-to-osv` reflects the modified time of the underlying CVE record.

This means that if one of the `PackageInfo` parts combined has changed, the resultant OSV record is not routinely re-imported (see also https://github.com/google/osv.dev/blob/master/tools/datafix/reimport_gcs_record.py)

**Expected behaviour**
- When one of the `PackageInfo` part files has changed, the OSV record is re-imported.
- The `modified` field reflects the most recent modification to one of the parts that compromise the OSV record

**Additional context**
One possible solution is for https://github.com/google/osv.dev/blob/216c120c52f7ee7c310aba51cbaa57418b64ed2f/vulnfeeds/cmd/combine-to-osv/main.go#L93-L108 to also return the most recent timestamp of all of the parts files loaded

hogo6002:
Thanks Andrew for recording this issue. 
I just noticed that almost all the Debian part files will be modified each time run. This is because it uses a map to store all Debian releases, so the order is not the same every time when it outputs to the `PackageInfos` file. 
I will need to fix this to avoid unnecessary updating. Also checked other parts output, Alpine doesn't have this issue, but I not sure about nvd (all last modified dates are today).

PR #2120, opened 2024-04-19
Sort Debian OSV pkgInfos

hogo6002:
- Sorts the Debian OSV outputs to remain consistent each runs (https://github.com/google/osv.dev/issues/2103)
- Fixes the `AffectedVersions`
- modifies `first_package_finder.py` to support future Debian releases, like Debian 13.

### DIFF
--- a/vulnfeeds/cmd/debian/main.go
+++ b/vulnfeeds/cmd/debian/main.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"path"
 	"sort"
+	"strconv"
 
 	"github.com/google/osv/vulnfeeds/cves"
 	"github.com/google/osv/vulnfeeds/utility"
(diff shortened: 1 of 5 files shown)


vulnfeeds/cmd/debian/main.go
```diff
@@ -97,9 +97,18 @@ func updateOSVPkgInfos(pkgName string, cveId string, releases map[string]Release
 	if value, ok := osvPkgInfos[cveId]; ok {
 		pkgInfos = value
 	}
-	for releaseName, release := range releases {
+
+	// Sorts releases to ensure pkgInfos remain consistent between runs.
+	var keys []string
+	for name := range releases {
```
another-rex, line 103: Can you sort this by the release order? E.g. have debian 10 before debian 11 before debian 12. I think you can just use the `debianReleaseMap` as the key to sort by.

another-rex (approved):
LGTM

vulnfeeds/cmd/debian/main_test.go
```diff
@@ -18,11 +18,12 @@ func Test_generateDebianSecurityTrackerOSV(t *testing.T) {
 	_ = json.NewDecoder(file).Decode(&decodedDebianData)
 
 	debianReleaseMap := make(map[string]string)
```
andrewpollock, line 20: Nit: Please either sort alphabetically by key, or numerically by the version.

andrewpollock (approved):

Issue #1984, opened 2024-02-12
Data quality issue for some GIT range : ranges containing overriding events 


RomainLefeuvre:
**Describe the bug**
Thanks again for your work ! 

We detected vulnerabilities having some "overriding" event : 
https://osv.dev/vulnerability/CVE-2018-5407
![image](https://github.com/google/osv.dev/assets/33905216/fecf03a6-9b28-4838-848b-d2c5456b190f)
The same commit is both an introduced and fixed event. 

Not sure if the specification authorize such case of if it is an inconsistency.
Some other vulnerabilities that have been logged for the same issue : 

[...]

PR #2127, opened 2024-04-21
Sanity check to prevent duplicate commits

andrewpollock:
This adds a check for the same commit being added to more than once when converting versions to commits.

#1984 surfaced that OpenSSL CVEs are being converted with the same commit as both `introduced` and `fixed` (due to incorrect normalization of OpenSSL versions). While the underlying defect is being addressed, add some sanity checking so that if the same commit hash can't be added to a GIT range being generated when converting versions to commits.


vulnfeeds/cves/versions.go
```diff
@@ -88,50 +97,82 @@ func (vi *VersionInfo) HasLastAffectedVersions() bool {
...
+		if ac.Repo == repo && ac.LastAffected != "" {
+			LastAffectedCommits = append(LastAffectedCommits, ac.Fixed)
 		}
 	}
 	return LastAffectedCommits
 }
 
+// Check if the same commit appears in multiple elements of the AffectedCommits array.
+// See https://github.com/google/osv.dev/issues/1984 for more context.
+func (vi *VersionInfo) Duplicated(candidate AffectedCommit) (err error) {
```
oliverchang, line 146: nit:  is the `err` expected to be used anywhere? it doesn't appear to be used above. If it's not used, I suggest following YAGNI and just returning a bool.

andrewpollock: Addressed in [0bfb1a4](https://github.com/google/osv.dev/pull/2127/commits/0bfb1a4d4902cf9fbf7593f03863dd37b8724ccb)

oliverchang (approved):

PR #2130, opened 2024-04-23
Workaround for Bioconductor test breakage

andrewpollock:
Enumeration seems to be failing for Bioconductor 3.19:

```
$ GET https://packagemanager.posit.co/__api__/repos/4/packages/a4?bioc_version=3.19
{"code":4,"error":"Package 'a4' not found","payload":null}
```

but works for 3.18 and 3.1 (the bounds of the array returned by `get_bioc_versions()`)

/cc @tylfin for visibility

### DIFF
--- a/osv/ecosystems/bioconductor.py
+++ b/osv/ecosystems/bioconductor.py
@@ -59,7 +59,7 @@ class Bioconductor(Ecosystem):
     versions = []
     # Currently breaking on 3.19,
     # see https://github.com/google/osv.dev/pull/1477/files#r1575458933
-    bioc_versions = bioc_versions[1:]
+    bioc_versions = bioc_versions.remove('3.19')
     for version in bioc_versions:
       response = requests.get(
           url.format(package=package, bioc_version=version),
--- a/osv/ecosystems/bioconductor.py
+++ b/osv/ecosystems/bioconductor.py
@@ -59,7 +59,10 @@ class Bioconductor(Ecosystem):
     versions = []
     # Currently breaking on 3.19,
     # see https://github.com/google/osv.dev/pull/1477/files#r1575458933
-    bioc_versions = bioc_versions.remove('3.19')
+    try:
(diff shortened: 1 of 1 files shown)


osv/ecosystems/bioconductor.py
```diff
@@ -57,6 +57,9 @@ def _enumerate_versions(self,
     """Helper method to enumerate versions from a specific URL."""
 
     versions = []
+    # Currently breaking on 3.19,
+    # see https://github.com/google/osv.dev/pull/1477/files#r1575458933
+    bioc_versions = bioc_versions[1:]
```
oliverchang, line 62: Rather than assume [0] is 3.19, can we just do a bioc_versions.remove() ?

andrewpollock: Addressed in 7e9644a

osv/ecosystems/bioconductor.py
```diff
@@ -59,7 +59,7 @@ def _enumerate_versions(self,
     versions = []
     # Currently breaking on 3.19,
     # see https://github.com/google/osv.dev/pull/1477/files#r1575458933
-    bioc_versions = bioc_versions[1:]
+    bioc_versions = bioc_versions.remove('3.19')
```
oliverchang, line 62: can we wrap this in a try except (catching ValueError per https://docs.python.org/3/tutorial/datastructures.html#more-on-lists) in case it's not there?

oliverchang (approved):

osv/ecosystems/bioconductor.py
andrewpollock: Done in [1b1f6f3](https://github.com/google/osv.dev/pull/2130/commits/1b1f6f31e031c24f9ed61baa66b766cc25f48ce5) + [4728b50](https://github.com/google/osv.dev/pull/2130/commits/4728b5044faf129ffd24a985c74cdf31040a3dde)

PR #2133, opened 2024-04-24
Add a second importer cronjob to run in delete mode

andrewpollock:
This runs the code introduced in #2030 that takes many hours to complete and isn't practical to run inline with regular imports.

Addresses #1467


deployment/clouddeploy/gke-workers/base/kustomization.yaml
```diff
@@ -2,6 +2,7 @@ resources:
 - workers.yaml
 - scaler.yaml
 - importer.yaml
+- importer-deleter.yaml
```
michaelkedar, line 5: Adding this to the base kustomization will add it to the production environment (which I don't think you intended?)

What we have been doing is temporarily adding the file as a resource to the test instance's kustomization.yaml. (kinda like what [debian-cve has done](https://github.com/google/osv.dev/blob/aad1acbb33fc9d54fa6366681d50ccc9da552117/deployment/clouddeploy/gke-workers/environments/oss-vdb-test/kustomization.yaml#L3))

There might be a smarter way to do this - maybe by adding the production patch with [`$patch: delete`](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-api-machinery/strategic-merge-patch.md#delete-directive)? Though I don't really know how that works.

andrewpollock: Ah yes, thanks for picking up on that. Addressed in 0f29c46

deployment/clouddeploy/gke-workers/environments/oss-vdb-test/importer-deleter.yaml
```diff
@@ -0,0 +1,18 @@
...
+        spec:
+          containers:
+          - name: importer-deleter
+            env:
+            - name: GOOGLE_CLOUD_PROJECT
+              value: oss-vdb-test
+            args:
+              # TODO(michaelkedar): ssh secrets
+              # TODO(michaelkedar): single source of truth w/ terraform config
+              - "--public_log_bucket=osv-test-public-import-logs"
```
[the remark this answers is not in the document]
andrewpollock: It looks like the reverse?

```
$ kubectl kustomize environments/oss-vdb-test/
.
.
.
apiVersion: batch/v1
kind: CronJob
metadata:
  name: importer-deleter
spec:
  concurrencyPolicy: Forbid
  jobTemplate:
    spec:
      template:
        spec:
          containers:
          - args:
            - --delete
            - --deletion_threshold=10
            image: importer
            imagePullPolicy: Always
            name: importer-deleter
            resources:
              limits:
                cpu: 1
                memory: 2G
              requests:
                cpu: 1
                memory: 1G
            securityContext:
              privileged: true
            volumeMounts:
            - mountPath: /work
              name: ssd
          nodeSelector:
            cloud.google.com/gke-nodepool: importer-pool
          restartPolicy: OnFailure
          tolerations:
          - key: workloadType
            operator: Equal
            value: importer-pool
          volumes:
          - hostPath:
              path: /mnt/disks/ssd0
            name: ssd
  schedule: '* */6 * * *'
```

I'll consolidate all of the args into the staging-specific file

deployment/clouddeploy/gke-workers/environments/oss-vdb-test/importer-deleter.yaml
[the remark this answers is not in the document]
andrewpollock: Done in [0f29c46](https://github.com/google/osv.dev/pull/2133/commits/0f29c46b1a48a124d3f9bc51471f0642ac3daa47)

deployment/clouddeploy/gke-workers/environments/oss-vdb-test/importer-deleter.yaml
```diff
@@ -0,0 +1,42 @@
...
+              privileged: true
+            resources:
+              requests:
+                cpu: 1
+                memory: "1G"
+              limits:
+                cpu: 1
+                memory: "2G"
+          nodeSelector:
+            cloud.google.com/gke-nodepool: importer-pool
```
michaelkedar, line 37: I think this `importer-pool` has only one node, which iirc was done so that the main importer could have some storage persist.

Could @another-rex confirm if this might be a problem?

michaelkedar (approved):

andrewpollock:
Hmm, I think this is going to require additional tweaking based on an attempt at manually running the job: it looks like it will make the existing importer job unschedulable for the duration due to CPU limitations in the cluster?

deployment/clouddeploy/gke-workers/environments/oss-vdb-test/importer-deleter.yaml
another-rex: Yes, it only has one node with only 2 CPU, which means the deleter and the importer can't be scheduled at the same time.

another-rex:
Does the importer deleter need the checked out git repos on disk? The main reason the importer is in its own pool is to avoid having to perform a full pull of the git repositories every time it runs.

andrewpollock:
> Does the importer deleter need the checked out git repos on disk? The main reason the importer is in its own pool is to avoid having to perform a full pull of the git repositories every time it runs

It doesn't, yet. Eventually deletion detection for Git repo sources will get reinstated as well, so it might be significant then?

Trying to think of the best places to put some signposting to avoid that becoming a problem later...

deployment/clouddeploy/gke-workers/environments/oss-vdb-test/importer-deleter.yaml
andrewpollock: Removed `nodeSelector` in [6f1885e](https://github.com/google/osv.dev/pull/2133/commits/6f1885e6e814b2a7aea43c3bb6bb37c490a62816)

michaelkedar (approved):

deployment/clouddeploy/gke-workers/environments/oss-vdb-test/importer-deleter.yaml
andrewpollock: Could you PTAL as of [5caf9c1](https://github.com/google/osv.dev/pull/2133/commits/5caf9c15a93c3037accffcafe23fba9af057b677)?

deployment/clouddeploy/gke-workers/environments/oss-vdb-test/importer-deleter.yaml
```diff
@@ -0,0 +1,45 @@
...
+  schedule: "* */6 * * *"
+  concurrencyPolicy: Forbid
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          tolerations:
+          - key: workloadType
+            operator: Equal
+            value: default-pool
```
michaelkedar, line 15: This isn't necessary.

The toleration was to allow jobs to be run on node pools with taints (e.g. the importer pool prevents anything without the `importer-pool` workloadType from being run on it)

The default poll has no such taints.

deployment/clouddeploy/gke-workers/environments/oss-vdb-test/importer-deleter.yaml
```diff
@@ -0,0 +1,45 @@
...
+              privileged: true
+            resources:
+              requests:
+                cpu: 1
+                memory: "10G"
+              limits:
+                cpu: 1
+                memory: "11G"
+          nodeSelector:
+-           cloud.google.com/gke-nodepool: default-pool
```
michaelkedar, line 40: This also isn't explicitly needed - all the other node pools have taints so it will be de facto scheduled on the default-pool

michaelkedar (approved):
LGTM, though some minor nits

another-rex (approved):

deployment/clouddeploy/gke-workers/environments/oss-vdb-test/importer-deleter.yaml
andrewpollock: Removed in [56f0ee2](https://github.com/google/osv.dev/pull/2133/commits/56f0ee2af11f582ba31a0ae4ca96198fe86f943e)

deployment/clouddeploy/gke-workers/environments/oss-vdb-test/importer-deleter.yaml
andrewpollock: Removed in 71dcb05b

andrewpollock:
> LGTM, though some minor nits

Thanks, deloused.

Issue #381, opened 2022-04-06
V2 UI: Handle overflowing tabs on vulnerability details better


oliverchang:
Currently, we set the spicy-sections tab bar to be collapsed instead if:
- The screen is <=500px wide. 
- This hacky heuristic says yes: https://github.com/google/osv/blob/8de7697b3b8a73e79a73ec34f17ef0fa842cfbb2/gcp/appengine/frontend_handlers.py#L342

The collapsed view has some issues:
- Everything is collapsed by default. We should look at expanding all sections by default. 

We should also see if there's a better way to handle this.

ZhangChen199102:
I'm happy to work on this issue.

oliverchang:
Mirroring this comment here: https://github.com/google/osv.dev/pull/2134#issuecomment-2095129375

ZhangChen199102:
Thanks @oliverchang, it's caused by turbo as it prevent full reload of the page, so `window.onload` event won't happen in that case. I've opened a PR for this issue: https://github.com/google/osv.dev/pull/2198

PR #2134, opened 2024-04-24
V2 UI: Handle overflowing tabs on vulnerability details better

ZhangChen199102:
Issue: #381 

Since [spicy-sections](https://github.com/tabvengers/spicy-sections) doesn't have a feature of expanding all collapsed sections by default, I decided to programmatically click on those package headers when the page is loaded. But this method is bit of awkward, so let me know if you have a better idea.

Mobile view:
mobile-view

An alternative way of implementing similar feature is using `exclusive-collapse` mode for `spicy-section`, but in that mode only the first section is expanded by default, and users are only able to expand one section (which is kind of similar with the tab-bar mode). Here is a demo of different modes of Spicy Section, and you just need to click on the "exclusive-collapse" button to check exclusive-collapse mode:
https://spicy-sections.glitch.me/programatic.html

### DIFF
--- a/gcp/appengine/frontend3/src/templates/vulnerability.html
+++ b/gcp/appengine/frontend3/src/templates/vulnerability.html
@@ -220,4 +220,19 @@
     {{ vulnerability.id }} - OSV
   </template>
 </turbo-stream>
+
+<script>
+  /*
+  We use `spicy-section` to make the packages content collapsible for mobile view,
+  but the issue is it collapse the content by default but we want to expaneded after
+  the page is loaded. So we decide to programmatically click on header of collapsed
+  packages after the page is loaded, and the content will be visible.
+  */
+  window.addEventListener('load', function() {
+    const package_headers = document.querySelectorAll('.package-header[affordance="collapse"]')
+    package_headers.forEach((header) => {
+      header.click()
+    });
+  })
+</script>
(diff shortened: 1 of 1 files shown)


gcp/appengine/frontend3/src/templates/vulnerability.html
```diff
@@ -220,4 +220,13 @@ <h3 class="mdc-layout-grid__cell--span-3">
     {{ vulnerability.id }} - OSV
   </template>
 </turbo-stream>
+
+<script>
+  window.addEventListener('load', function() {
```
oliverchang, line 225: Can you add some comments here explaining this?

oliverchang:
@ZhangChen199102 can you please merge the latest master into your branch?

another-rex (approved):

oliverchang:
Hmm, there's some strange behaviour with this. 

If I load https://test.osv.dev/vulnerability/ALSA-2024:1818 directly, then all the sections are expanded. 

If I instead navigate to the vulnerability via https://test.osv.dev/list?ecosystem=AlmaLinux&q=ALSA-2024%3A1818, they do not expand. I wonder if this is some side effect caused by turbo?

oliverchang:
@ZhangChen199102 can you please take a look?

PR #2137, opened 2024-04-26
Update the title to display summary instead of id

### LINKED ISSUE #2078: Details Page Does Not Display Vulnerability Summary
**Describe the bug**

The details page for vulnerabilities currently does not include the summary description. This makes it hard to quickly grasp what the vulnerability is about.

**To Reproduce**
1. Go to https://osv.dev/vulnerability/GHSA-6q4w-9x56-rmwq
2. Try to quickly grasp what this vulnerability is about.

**Expected behaviour**

The summary description that is displayed in the database list should be displayed prominently.

**Screenshots**

image

vs

image

zahraaalizadeh:
Prior to this change, in the vulnerability page, id of the vulnerability was displayed as title.

This change updates the title to display the human readable summary for the title, and adds ID as an data description.

resolves #2078

example 1 - Before:
before1

After:
after1

Example 2 - Before:
Screenshot 2024-04-26 at 3 40 44 pm

After:
after2

### DIFF
--- a/gcp/appengine/frontend3/src/styles.scss
+++ b/gcp/appengine/frontend3/src/styles.scss
@@ -633,6 +633,9 @@ dl.vulnerability-details,
     grid-column: 2;
   }
 
+  dt.summary, dd.summary {
+    margin-top: 16px;
+  }
 
   dd.details {
     overflow-wrap: break-word;
--- a/gcp/appengine/frontend3/src/styles.scss
+++ b/gcp/appengine/frontend3/src/styles.scss
@@ -637,6 +637,10 @@ dl.vulnerability-details,
     margin-top: 16px;
   }
 
+  dd.muted {
+    color: $osv-grey-600;
+  }
+
   dd.details {
     overflow-wrap: break-word;
 
--- a/gcp/appengine/frontend3/src/templates/vulnerability.html
+++ b/gcp/appengine/frontend3/src/templates/vulnerability.html
@@ -62,8 +62,8 @@
           <dd>{{ vulnerability.published }}</dd>
           <dt>Modified</dt>
           <dd>{{ vulnerability.modified }}</dd>
-          <dt>Summary</dt>
-          <dd>{{ vulnerability.summary }}</dd>
+          <dt class="summary">Summary</dt>
+          <dd class="summary">{{ vulnerability.summary }}</dd>
           <dt>Details</dt>
           <dd class="details">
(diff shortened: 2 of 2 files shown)


oliverchang:
Thanks! Could we instead add the summary to the main part of the page, above "details" ? 

IMO the title should remain the vulnerability ID. `summary` can get pretty long.

zahraaalizadeh:
@oliverchang Sometimes details include a Summary section too. Adding a field like _Summary_ or _Brief summary_ might seem duplicated? WDYT?

oliverchang:
> @oliverchang Sometimes details include a Summary section too. Adding a field like _Summary_ or _Brief summary_ might seem duplicated? WDYT?

I think that's OK. We should accurately render what's in the source JSON.

zahraaalizadeh:
@oliverchang I updated it as per suggestion. Now there a summary field that comes before details.

another-rex:
It is starting to look a bit busy, can you add a gap between summary and the Modified field above it (similar to the gap between references and details). This way we have 3 mini sections separated with some padding:
- the metadata
- the text (summary and description)
- references

Also, for some entries summary is empty, in those cases I think we should just hide the summary field (though that might be confusing, maybe a greyed out `[none]` or something similar could be better)

oliverchang (approved):

hogo6002:
It seems not all entries have `summary` field. Example: https://test.osv.dev/vulnerability/CVE-2024-34397 Screenshot 2024-05-10 at 2 58 54 PM

andrewpollock:
> It seems not all entries have `summary` field

This is definitely to be expected for CVEs, which don't have clean correlation between OSV's `summary` versus `details` fields. I think the current treatment isn't too bad. An improvement would be to omit the **Summary** field entirely from the per-vulnerability page.

The vulnerability listing page is deferring to the full record in the absence of a summary...

zahraaalizadeh:
@hogo6002 Displaying the summary as grayout `[none]` is intentional based on [this](https://github.com/google/osv.dev/pull/2137#issuecomment-2089319481) comment.

Please let me know if you think we could handle it better in the vulnerability page.

andrewpollock:
> An improvement would be to omit the **Summary** field entirely from the per-vulnerability page.

Apologies, that was a poorly formed statement... I meant to say it could be omitted when it is empty/undefined, but that inconsistency could also be visually jarring and may be worse than what we have now.

Issue #1697, opened 2023-10-05
Improve browser navigation (back button) on osv.dev


another-rex:
When pressing the back button on the `osv.dev/list` page it often will get stuck in a halfway state. Example:

![image](https://github.com/google/osv.dev/assets/106129829/ce984bfd-c935-4985-b35a-02654bd3def5)

Replication steps:

1. get on the `/list` page
2. Press `Android`
3. Press `Alpine`
4. Navigate back in the browser

Notice a few inconsistencies:
- List is still blurred, the blur class isn't removed properly
- URL and Title is correctly "Android" but "Alpine" is still highlighted in the filter
- The items in the list is correctly "Android"

SonuBardai:
It looks like `turbo-frame` applies a `busy` boolean attribute [[ref](https://turbo.hotwired.dev/reference/frames#html-attributes)] when the new data is being fetched. The `busy` attribute is what triggers the [styles](https://github.com/google/osv.dev/blob/ceb042141e4e7e4317502fdbf2008e60105bda1f/gcp/appengine/frontend3/src/styles.scss#L531-L538) to add the blur filter. Pressing the back button of the browser gets the last cached version of the page, which was the state when `busy` was set. So far, there's a few solutions I can think: 
- There should be some solution within turbo-frame to resolve this and listen for events when busy is turned off. We add some styles to revert the blur when this happens, so this version of the page is what gets cached. 
- Run some JS that will look for busy items and remove them when the user navigates.


https://github.com/google/osv.dev/assets/67749330/f5494a72-0b95-42e4-bf21-2bb03b31430d

- The incorrect filter highlighting also seems like a weird caching bug, although I can't say for sure. I'm not familiar with turbo-frame, which seems to be what we're using for the filters too.

andrewpollock:
@SonuBardai great analysis, thank you. 

Is a quick (interim?) fix to adjust the cacheability of the page?

SonuBardai:
Apologies for the delayed response @andrewpollock. I would say yes, adjusting the caching should fix it, or if there's a way to configure turbo-frame to reset on receiving the redirect response then that would fix this as well.

ZhangChen199102:
I'm happy to work on this issue

PR #2144, opened 2024-04-30
Improve browser navigation back for vulnerability list page

ZhangChen199102:
Issue: #1697 

As mentioned in the issue, the vulnerability list page gets stuck after apply filters and click on the browser's back button, I've done some research on this but can't find a nice solution, and I feel this could be a weird bug of Turbo. So I just disabled the Turbo cache behavior for vulnerability list page, and it just reloads the page after click the back button.

Please let me know if you have a better suggestion.


another-rex (approved):
Thanks, I think this is a reasonable fix, we can find other solutions later if we discover this to be a performance problem.

another-rex:
Can you rebase this onto the latest master branch? (or give us permission to modify the PR). The test error should be fixed there.

PR #2146, opened 2024-04-30
Fix querying by commit integration test more

andrewpollock:
The additional matching CVE has now been rejected, and the text of the description has changed. I suspect we haven't seen the last of this yet, as a rejected CVE should be resulting in a withdrawn OSV record...

(I'm investigating why that isn't the case, but want to unbreak the test)

### DIFF
--- a/gcp/api/integration_tests.py
+++ b/gcp/api/integration_tests.py
@@ -97,21 +97,13 @@ class IntegrationTests(unittest.TestCase,
 
   _VULN_31745 = {
       'id': 'CVE-2024-31745',
-      'details':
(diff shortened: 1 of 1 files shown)


oliverchang (approved):

andrewpollock:
I confused disputed CVEs with rejected ones.

https://github.com/google/osv.dev/blob/aad1acbb33fc9d54fa6366681d50ccc9da552117/vulnfeeds/cmd/combine-to-osv/main.go#L119-L128 withdraws _disputed_ CVEs.

I think we also need to be reflecting _rejected_ ones as withdrawn. I'll file an issue for this.

PR #2148, opened 2024-05-01
Revert spicy-sections integrity again

michaelkedar:
Repeat of #2014 and #1909
It'd be nice if we could work out how to stop this from happening...


hogo6002 (approved):
LGTM

Issue #1013, opened 2023-02-09
Website latency reduction


andrewpollock:
I'll use this issue to collect some of the opportunities to improve page load time based on my attempts to measure latency for defining an SLO:

- https://tools.pingdom.com/#618e1f3848c00000

zahraaalizadeh:
I'm planning to work on this one.

zahraaalizadeh:
I just assessed the performance using Page Speed Insight:
https://pagespeed.web.dev/analysis/https-osv-dev/rsqqsi0hrn?form_factor=mobile

and it indicates potential room for improvement on the mobile version.

zahraaalizadeh:
https://github.com/google/osv.dev/pull/2221

PR #2151, opened 2024-05-01
Add Cache-Control to response

zahraaalizadeh:
This change introduces Cache-Control headers to enhance website performance by improving caching efficiency. The addition of these headers aims to reduce unnecessary HTTP requests and ensure that clients can cache content appropriately, thereby improving page load speeds and reducing server load.

Cache-Control headers have been added with dynamic values based on the request path, enabling browser and intermediate caches to store static assets for optimized durations. The durations have been set as follows:

Static content: 1 week
Blog module: 1 day
Other links: 1 hour
I am open to updating these values based on your feedback.

Issue: https://github.com/google/osv.dev/issues/1013

Before:
before

After:
after_blog
after_list

### DIFF
--- a/gcp/appengine/frontend_handlers.py
+++ b/gcp/appengine/frontend_handlers.py
@@ -99,6 +99,18 @@ def add_cors_headers(response):
   return response
 
 
+@blueprint.after_request
+def add_cache_control_headers(response):
+  """Add Cache-Control headers."""
+  if 'static' in request.path:
+    response.headers['Cache-Control'] = 'public, max-age=604800'  # 1 week
+  elif 'blog' in request.path:
+    response.headers['Cache-Control'] = 'public, max-age=86400'  # 1 day
+  else:
+    response.headers['Cache-Control'] = 'public, max-age=3600'  # 1 hour
+  return response
+
+
 @blueprint.route('/v2/')
 def index_v2():
   return redirect('/')


oliverchang:
Thanks for this change.

We shouldn't be setting any form of cache-control on /list and /vulnerability/, as freshness is extremely important there. 

It's also unclear how much Cache-Control would help on our other pages, given that there's not many and they're not very heavyweight. 

I believe https://github.com/google/osv.dev/issues/1013 would require some more detailed benchmarking to identify other general ways to reduce latency. e.g. reducing unnecessary JS dependencies, static assets, improving compression etc.

zahraaalizadeh:
Thanks @oliverchang for your insights. I tested https://osv.dev/ with the Full Page test in Pingdom Tools. This specific test was done on May 2. The web page took 1670 ms to load, used 18 requests, and weighed in at 346.0 KB.
The Google Page Speed performance grade for this web page is 83/100. More deets here: https://tools.pingdom.com/#63ce324ad7800000

I started with adding the cache header as it was on top of the list. I can update my PR to just keep it for blog.

oliverchang:
> Thanks @oliverchang for your insights. I tested https://osv.dev/ with the Full Page test in Pingdom Tools. This specific test was done on May 2. The web page took 1670 ms to load, used 18 requests, and weighed in at 346.0 KB. The Google Page Speed performance grade for this web page is 83/100. More deets here: https://tools.pingdom.com/#63ce324ad7800000
> 
> I started with adding the cache header as it was on top of the list. I can update my PR to just keep it for blog page though. WDYT?

I believe we are already  returning Cache-Control on our `/static/` resources, which is where this likely matters the most. 

I think we can just drop this PR for now.

PR #2158, opened 2024-05-06
Update last modified time of combine-to-osv entries

hogo6002:
Fixes https://github.com/google/osv.dev/issues/2103

The current `modfied` time of a combine-to-osv entry only reflects the CVE modified time without checking if any `parts` has been changed. Updates code to set the `modified` field to the most recent modification time of parts' components or the original CVE entry.

- Uses `os.Stat(file).ModTime()` to check the last modified time of each part.
- Maintains a map to record the most recent modification of each part file associated with each CVE ID.
- Updates `PackageInfo`'s `modified` field if the part modified time is later than the original CVE's modified time.

It may not be the best implementation, but its failure wouldn't affect the logic of any existing code.

### DIFF
--- a/vulnfeeds/cmd/combine-to-osv/main.go
+++ b/vulnfeeds/cmd/combine-to-osv/main.go
@@ -3,9 +3,13 @@ package main
 import (
 	"encoding/json"
 	"flag"
+	"fmt"
+	"net/http"
+	"net/url"
 	"os"
 	"path"
 	"strings"
+	"time"
 
 	"github.com/google/osv/vulnfeeds/cves"
 	"github.com/google/osv/vulnfeeds/utility"
@@ -17,6 +21,7 @@ const (
 	defaultPartsInputPath = "parts"
 	defaultOSVOutputPath  = "osv_output"
 	defaultCVEListPath    = "."
+	defaultInputBucket    = "osv-test-cve-osv-conversion"
 )
 
 var Logger utility.LoggerWrapper
@@ -30,6 +35,7 @@ func main() {
 	partsInputPath := flag.String("partsPath", defaultPartsInputPath, "Path to CVE file")
 	osvOutputPath := flag.String("osvOutputPath", defaultOSVOutputPath, "Path to CVE file")
 	cveListPath := flag.String("cveListPath", defaultCVEListPath, "Path to clone of https://github.com/CVEProject/cvelistV5")
+	inputBucket := flag.String("inputBucket", defaultInputBucket, "Input bucket")
 	flag.Parse()
 
 	err := os.MkdirAll(*cvePath, 0755)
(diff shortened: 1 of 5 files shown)


vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -90,25 +141,26 @@ func loadInnerParts(innerPartInputPath string, output map[cves.CVEID][]vulns.Pac
 //
 // ## Returns
 // A mapping of "CVE-ID": []<Affected Package Information>
-func loadParts(partsInputPath string) map[cves.CVEID][]vulns.PackageInfo {
+func loadParts(partsInputPath string, inputBucket string) (map[cves.CVEID][]vulns.PackageInfo, map[cves.CVEID]time.Time) {
```
another-rex, line 144: Can you update the doc comment above with the new parameters and return type

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -42,13 +48,46 @@ func main() {
...
+	modifiedTime, err := time.Parse(time.RFC3339, jsonData.Updated)
+	if err != nil {
+		return emptyTime, err
+	}
+	return modifiedTime, err
+}
+
 // loadInnerParts loads second level folder for the loadParts function
-func loadInnerParts(innerPartInputPath string, output map[cves.CVEID][]vulns.PackageInfo) {
+func loadInnerParts(innerPartInputPath string, output map[cves.CVEID][]vulns.PackageInfo, inputBucket string, cvePartsModifiedTime map[cves.CVEID]time.Time) {
```
another-rex, line 90: Can you update the doc comment with what each parameters does, now that there are quite a few parameters.

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -74,6 +114,17 @@ func loadInnerParts(innerPartInputPath string, output map[cves.CVEID][]vulns.Pac
 
 		Logger.Infof(
 			"Loaded Item: %s", entryInner.Name())
+
+		// Updates the latest OSV parts modified time of each CVE
+		modifiedTime, err := getModifiedTime(filePath, inputBucket)
+		if err != nil {
+			Logger.Warnf("Failed to get modified time of %s: %s", filePath, err)
```
another-rex, line 121: Log the inputBucket as well here.

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -42,13 +48,46 @@ func main() {
...
 	allCves := loadAllCVEs(*cvePath)
-	allParts := loadParts(*partsInputPath)
-	combinedData := combineIntoOSV(allCves, allParts, *cveListPath)
+	allParts, cveModifiedMap := loadParts(*partsInputPath, *inputBucket)
+	combinedData := combineIntoOSV(allCves, allParts, *cveListPath, cveModifiedMap)
 	writeOSVFile(combinedData, *osvOutputPath)
 }
 
+// getModifiedTime gets the modification time of a given GCS file
+func getModifiedTime(innerPath string, bucket string) (time.Time, error) {
```
another-rex, line 57: Do we need to query GCS for the modified time? I think there's a way to copy the modified time when downloading the records (it might already be the case / is default behaviour), and we can just read it straight from disk.

another-rex (commented):
Thanks!

vulnfeeds/cmd/combine-to-osv/main.go
hogo6002: Ah, good point! I will update this

another-rex (approved):
minor comment, otherwise LGTM!

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -48,46 +43,30 @@ func main() {
...
 	combinedData := combineIntoOSV(allCves, allParts, *cveListPath, cveModifiedMap)
 	writeOSVFile(combinedData, *osvOutputPath)
 }
 
-// getModifiedTime gets the modification time of a given GCS file
-func getModifiedTime(innerPath string, bucket string) (time.Time, error) {
-	// Uses the GCS JSON API to get the modified date, details: https://cloud.google.com/storage/docs/json_api/v1/objects/get
-	baseUrl := "https://storage.googleapis.com/storage/v1"
-	filePath := fmt.Sprintf("%s/b/%s/o/%s", baseUrl, bucket, url.PathEscape(innerPath))
+// getModifiedTime gets the modification time of a given file
```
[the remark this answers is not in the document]
hogo6002: Updated. I also did some local testing, and the result matches with what we think.

PR #2167, opened 2024-05-07
Add parallel exporter

another-rex:
Exporter currently take too long to complete (around 1 hour and 10 minutes), and will only get longer to run as the OSV database expands.

This change parallelizes the exporter along each ecosystem, separating the ecosystem export portion of the script from the selection of each ecosystem. This roughly reduced the time taken down to how long the longest ecosystem takes to export, 13 minutes or so.

### DIFF
--- a/docker/exporter/export_runner.py
+++ b/docker/exporter/export_runner.py
@@ -68,7 +68,7 @@ def spawn_ecosystem_exporter(work_dir: str, bucket: str, eco: str):
   ])
   return_code = proc.wait()
   if return_code != 0:
-    logging.error(f"Export of {eco} failed with Exit Code: {return_code}")
+    logging.error("Export of %s failed with Exit Code: %d", eco, return_code)
 
 
 if __name__ == '__main__':
--- a/docker/exporter/export_runner.py
+++ b/docker/exporter/export_runner.py
@@ -39,8 +39,9 @@ def main():
       default=DEFAULT_EXPORT_BUCKET)
   parser.add_argument(
       '--processes',
-      help='Maximum number of parallel exports',
-      default=DEFAULT_EXPORT_PROCESSES)
+      help='Maximum number of parallel exports, default to number of cpu cores',
+      # If 0 or None, use the DEFAULT_EXPORT_PROCESSES value
+      default=os.cpu_count() or DEFAULT_EXPORT_PROCESSES)
   args = parser.parse_args()
 
   tmp_dir = os.path.join(args.work_dir, 'tmp')
--- a/docker/exporter/export_runner.py
+++ b/docker/exporter/export_runner.py
@@ -49,7 +49,7 @@ def main():
   os.environ['TMPDIR'] = tmp_dir
 
   query = osv.Bug.query(projection=[osv.Bug.ecosystem], distinct=True)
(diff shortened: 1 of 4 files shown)


docker/exporter/export_runner.py
```diff
@@ -0,0 +1,77 @@
...
+  parser.add_argument(
+      '--work_dir', help='Working directory', default=DEFAULT_WORK_DIR)
+  parser.add_argument(
+      '--bucket',
+      help='Bucket name to export to',
+      default=DEFAULT_EXPORT_BUCKET)
+  parser.add_argument(
+      '--processes',
+      help='Maximum number of parallel exports',
+      default=DEFAULT_EXPORT_PROCESSES)
```
andrewpollock, line 43: It might be tidier to default to `os.cpu_count()` here, that way this just does the right/intended thing if the number of CPUs is increased in the Kubernetes job spec?

docker/exporter/exporter.py
```diff
@@ -139,19 +140,13 @@ def main():
       '--bucket',
       help='Bucket name to export to',
       default=DEFAULT_EXPORT_BUCKET)
+  parser.add_argument('--ecosystem', required=True, help='Ecosystem to upload')
```
andrewpollock, line 143: If I'm understanding this correctly, the intended use of this flag is either an ecosystem name, or the special string "list", which retains the old behaviour? Please call out this "list" value in the flag help text.

andrewpollock (approved):
LGTM, a few minor things.
Curious as to why 7?

docker/exporter/export_runner.py
another-rex: Good point, updated to use os.cpu_count() by default, and falling back to DEFAULT_EXPORT_PROCESSES if failing to get CPU count.

docker/exporter/exporter.py
another-rex: `list` value will upload the ecosystem.txt ecosystem list. Called this out in the help message now!

another-rex:
Ended up at 7 because that's the maximum number of cores for a pod a node with 8 cores can support (since some CPU is being used up by metrics and logging driver). It doesn't really matter too much now. I'll switch this to 6 to get a more round number.

andrewpollock (approved):
Awesome performance improvement

docker/exporter/export_runner.py
```diff
@@ -0,0 +1,78 @@
...
+      # If 0 or None, use the DEFAULT_EXPORT_PROCESSES value
+      default=os.cpu_count() or DEFAULT_EXPORT_PROCESSES)
+  args = parser.parse_args()
+
+  tmp_dir = os.path.join(args.work_dir, 'tmp')
+  os.makedirs(tmp_dir, exist_ok=True)
+  os.environ['TMPDIR'] = tmp_dir
+
+  query = osv.Bug.query(projection=[osv.Bug.ecosystem], distinct=True)
+  ecosystems = [bug.ecosystem[0] for bug in query if bug.ecosystem] + ["list"]
```
oliverchang, line 52: nit: use single quotes for all strings in this file for consistency.

oliverchang (approved):
awesome stuff!

PR #2174, opened 2024-05-08
Update Go version to 1.22.2

cuixq:
This resolves the vulnerabilities reported on Go 1.22.1.


andrewpollock (approved):
I think from memory the last time I did this, I had to build the Docker image locally to bootstrap, have you been able to do that successfully?

cuixq:
I think we still need to do this manually - at least I did this in March last time.

PR #2180, opened 2024-05-08
Configure GCLB in terraform

michaelkedar:
Because importing existing resources is awkward (and will end up deleting/recreating most things if our names don't line up), I've decided to put all the Load Balancer / Certificate stuff into terraform now. I've tried to match what was set up manually on staging as closely as possible. I also added an IPv6 address, since our App Engine currently also has IPv6.

The google-managed certificate stuff is now set up in terraform, but configuring the DNS `CNAME`/`A`/`AAAA` records still needs to be done manually. I've added an output to the terraform to display all the records that need to be created.

The existing Load Balancer and Certificate stuff needs to be deleted from staging before merging this, and the DNS records need to be updated after.


deployment/terraform/modules/osv/website.tf
```diff
@@ -32,4 +33,122 @@ resource "google_cloud_run_service_iam_binding" "website" {
...
+  ssl                 = true
+  certificate_map     = google_certificate_manager_certificate_map.website.id
+
+  load_balancing_scheme = "EXTERNAL_MANAGED"
+
+  backends = {
+    default = {
+      groups = [
+        {
+          # TODO(michaelkedar): Replace with Cloud Run when ready
```
andrewpollock, line 97: The picture I had in my head was we could have both back ends ~now, with the Cloud Run backend drained of any traffic. Is that possible to do?

andrewpollock (approved):
LGTM

deployment/terraform/modules/osv/website.tf
michaelkedar: I think I've managed to work out how to do this with c650a4d

PR #2194, opened 2024-05-09
Prepare second importer cronjob for production

andrewpollock:
Increase run frequency to 3 hourly


oliverchang (approved):

michaelkedar (approved):

andrewpollock:
FYI I added d8e9db57 create the CronJob suspended so it doesn't immediately run after the release (I want to supervise it closely)

PR #2196, opened 2024-05-09
Fix modified time of nvd part entries

hogo6002:
The `gcloud storage rsync --checksums-only` command always updates the modified time on GCS objects once a file has been touched, even if the file content remains the same. Updates it to `gsutil rsync` command to keep the original modified time on GCS objects when nothing has been changed.


andrewpollock (approved):
Sadly LGTM

PR #2198, opened 2024-05-10
Fix collapsed packages when navigate to vulnerability page

ZhangChen199102:
Issue: https://github.com/google/osv.dev/issues/381

Initially the page listens the window onload event, and programatically clicks on the headers so the sections will be expanded. But since we use Turbo, it will request the new page by using fetch request rather than doing a full reload, so the `load` event won't be triggered.

In this PR we make the script listens `turbo:load` event which always happens when visiting a new page.

Result:
May-09-2024 16-34-20

### DIFF
--- a/gcp/appengine/frontend3/src/templates/vulnerability.html
+++ b/gcp/appengine/frontend3/src/templates/vulnerability.html
@@ -228,8 +228,8 @@
   the page is loaded. So we decide to programmatically click on header of collapsed
   packages after the page is loaded, and the content will be visible.
   */
-  window.addEventListener('load', function() {
-    const package_headers = document.querySelectorAll('.package-header[affordance="collapse"]')
+  document.addEventListener('turbo:load', function() {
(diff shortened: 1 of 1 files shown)


gcp/appengine/frontend3/src/templates/vulnerability.html
```diff
@@ -228,8 +228,8 @@ <h3 class="mdc-layout-grid__cell--span-3">
   the page is loaded. So we decide to programmatically click on header of collapsed
   packages after the page is loaded, and the content will be visible.
   */
-  window.addEventListener('load', function() {
-    const package_headers = document.querySelectorAll('.package-header[affordance="collapse"]')
+  document.addEventListener('turbo:load', function() {
+    const package_headers = document.querySelectorAll('.package-header[affordance="collapse"][aria-expanded="false"]')
```
ZhangChen199102, line 232: `turbo:load` happens multiple times but we just want to click the headers once, so only click them when `aria-expanded` is false

oliverchang (approved):

PR #2201, opened 2024-05-12
Disable boringssl OSS-Fuzz -> OSV import

### LINKED ISSUE #2178: Disable automatic OSS-Fuzz -> OSV import for BoringSSL
In principle, this import is great and we'd love to have it. However, until the following issues are resolved, the tradeoffs are not viable. I would much *much* rather we lose the OSV import than lose the fuzzers, so please disable the import.
* https://github.com/google/oss-fuzz/issues/11925
* https://github.com/google/oss-fuzz/issues/11939
* https://github.com/google/osv.dev/issues/2176
* https://github.com/google/osv.dev/issues/2177

oliverchang:
Fixes #2178. 

BoringSSL maintainers would like more control / vetting of entries. Disable the import for now until we have a better system in place to enable that.

### DIFF
--- a/docker/worker/oss_fuzz.py
+++ b/docker/worker/oss_fuzz.py
@@ -39,6 +39,7 @@ UNKNOWN_COMMIT = 'unknown'
 # Large projects which take way too long to build.
 # TODO(ochang): Don't hardcode this.
 PROJECT_DENYLIST = {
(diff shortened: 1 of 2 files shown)


andrewpollock (approved):

evverx:
Is it possible to turn this off for avahi, elfutils, systemd and libbpf too?

Issue #2152, opened 2024-05-01
Support easy access to the json version of a vulnerability


G-Rath:
**Is your feature request related to a problem? Please describe.**
"I'm always frustrated when ..." I keep having to lookup the API endpoint for getting the JSON of an osv 😅 

**Describe the solution you'd like**
Having a link to the underlying JSON data that is used by the api, or something like supporting adding `.json` to the end of the URL to show the JSON version.

**Describe alternatives you've considered**
Writing the api path on a sticky note, but I'm concerned if I add too many more of those my monitor will fall off my desk due to the weight :/

another-rex:
A current two click workaround is going to osv.dev/CVE-2023-45802 and then clicking the import source link, which should give you the JSON.

Though allowing .json to be appended directly to the URL is probably still a good idea.

G-Rath:
I almost mentioned that but I didn't - the import source is useful but not quite the same since part of the APIs role is cleaning up/improving/whatevering the osv from the original source and sometimes that isn't even json (e.g. https://osv.dev/vulnerability/PYSEC-2024-50 the source is YAML).

oliverchang:
Can you file a feature request to support a `.json` URL suffix ? We can probably easily redirect to `https://api.osv.dev/v1/vulns/ID`

G-Rath:
Is there an actual other repository you'd like an issue opened against, or was that just a case of Friday-afternoon-brain? 😅

oliverchang:
Friday-afternoon-brain :) My bad!

PR #2207, opened 2024-05-14
Create get vulnerability details API

ZhangChen199102:
Issue: #2152 

This PR introduces a new GET vulnerability details API

### DIFF
--- a/gcp/appengine/frontend_handlers.py
+++ b/gcp/appengine/frontend_handlers.py
@@ -260,11 +260,15 @@ def vulnerability_json_redirector(potential_vuln_id):
     return None
 
   vuln = osv_get_by_id(potential_vuln_id)
-  if vuln:
-    return redirect(f'https://api.osv.dev/v1/vulns/{potential_vuln_id}')
+  if not vuln:
+    abort(404)
+    return None
 
-  abort(404)
-  return None
+  if utils.is_prod():
+    api_url = 'api.osv.dev'
+  else:
+    api_url = 'api.test.osv.dev'
+  return redirect(f'https://{api_url}/v1/vulns/{potential_vuln_id}')
 
 
 def bug_to_response(bug, detailed=True):
--- a/gcp/appengine/frontend_handlers.py
+++ b/gcp/appengine/frontend_handlers.py
@@ -224,7 +224,14 @@ def list_vulnerabilities():
 def vulnerability(vuln_id):
   """Vulnerability page."""
   vuln = osv_get_by_id(vuln_id)
-  return render_template('vulnerability.html', vulnerability=vuln)
+
+  if utils.is_prod():
+    api_url = 'api.osv.dev'
+  else:
+    api_url = 'api.test.osv.dev'
+
+  return render_template(
+      'vulnerability.html', vulnerability=vuln, api_url=api_url)
 
 
 @blueprint.route('/<potential_vuln_id>')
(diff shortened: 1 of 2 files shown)


gcp/appengine/frontend_handlers.py
```diff
@@ -237,102 +237,6 @@ def vulnerability_redirector(potential_vuln_id):
   return None
 
 
-def bug_to_response(bug, detailed=True):
```
ZhangChen199102, line 240: Moving these functions to `utils.py` so they can be used by both of `api_handlers` and `frontend_handlers`.

gcp/appengine/api_handlers.py
```diff
@@ -0,0 +1,47 @@
+# Copyright 2021 Google LLC
```
ZhangChen199102, line 1: Create a new app for API

another-rex:
Unless I'm misunderstanding the goal, rather than putting API code in the frontend as well, I think we can just put a link to the GET API endpoint on the page, or add a new handler to redirect requests ending in .json to the API endpoint.
https://google.github.io/osv.dev/get-v1-vulns/

E.g. https://api.osv.dev/v1/vulns/CVE-2023-45802

andrewpollock:
> or add a new handler to redirect requests ending in .json to the API endpoint.

That was my understanding of the easiest way forward based on https://github.com/google/osv.dev/issues/2152#issuecomment-2092104090

ZhangChen199102:
So this time I just added a JSON Data field, with a GET vulnerability details link on the vulnerability details page:

image

andrewpollock:
> So this time I just added a JSON Data field

That's cool, but doesn't address the UX intention behind #2152 (as I understand it).

Today, there's a convenience redirect where:
https://OSV.dev/GHSA-75mx-chcf-2q32 redirects to https://osv.dev/vulnerability/GHSA-75mx-chcf-2q32

An additional convenience redirect would be for:
https://OSV.dev/GHSA-75mx-chcf-2q32.json to redirect to https://api.osv.dev/v1/vulns/GHSA-75mx-chcf-2q32

Without having tried this, I think a variation on:
https://github.com/google/osv.dev/blob/7428293389102994be1aa90499ed18c11f5fa19a/gcp/appengine/frontend_handlers.py#L230-L242 would accomplish this redirection:

- support `.json` in the URL handler
- if the vulnerability ID exists, redirect to the existing JSON API endpoint, per my example above

andrewpollock (commented):
This LGTM, have you been able to do any testing to confirm it works as intended?
FYI, I wasn't averse to highlighting the availability of the API from the individual vulnerability pages on the website in addition to this. Feel free to reinstate that.

gcp/appengine/frontend_handlers.py
```diff
@@ -242,6 +242,24 @@ def vulnerability_redirector(potential_vuln_id):
...
+@blueprint.route('/vulnerability/<potential_vuln_id>.json')
+def vulnerability_json_redirector(potential_vuln_id):
+  """Convenience redirector for /VULN-ID.json and /vulnerability/VULN-ID.json to
+  https://api.osv.dev/v1/vulns/VULN-ID.
+  """
+  if not _VALID_VULN_ID.match(potential_vuln_id):
+    abort(404)
+    return None
+
+  vuln = osv_get_by_id(potential_vuln_id)
```
another-rex, line 255: minor nit: can you change the `if vuln:` to `if not vuln:` to have the final return statement be the redirect, and the if branches be the aborts.

another-rex (approved):
LGTM! One minor note would be this will always redirect to the production API, even when running on test.osv.dev. Not sure if there's an easy fix for that or not, if not I think that behaviour is fine.

ZhangChen199102:
>This LGTM, have you been able to do any testing to confirm it works as intended?

Yes, I've tested http://127.0.0.1:8000/vulnerability/GHSA-vw63-824v-qf2j.json and http://127.0.0.1:8000/GHSA-vw63-824v-qf2j.json on local, and both redirects me to https://api.test.osv.dev/v1/vulns/GHSA-vw63-824v-qf2j

ZhangChen199102:
>One minor note would be this will always redirect to the production API, even when running on test.osv.dev. Not sure if there's an easy fix for that or not, if not I think that behaviour is fine.

Thanks, I've updated this and now it will redirect to `api.test.osv.dev` in local and test environment, and `api.osv.dev` for prod.

andrewpollock (approved):

PR #2211, opened 2024-05-15
Migrate website traffic entirely to Cloud Run

michaelkedar:
I will delete all our app engine stuff probably in the next week or two, in case we really need to roll back.


andrewpollock (approved):
Woohoo!

oliverchang (approved):
YAY!!

PR #2213, opened 2024-05-15
Fix integration tests withdrawn CVEs

andrewpollock:
The cleanup of stale GCS-based records changed the CVE record landscape.


andrewpollock:
Although inspecting these CVEs, they *should* be converting. Yay, another bug found...

andrewpollock:
> Yay, another bug found...

FYI I expect these to return after https://github.com/google/osv.dev/pull/2214 makes its way into Production...

michaelkedar (approved):

Issue #2184, opened 2024-05-09
Navigating away from a blog post attempts to get the blog's images from the new page


michaelkedar:
**Describe the bug**
On the website, if you are view a blog post with images and navigate to another page, the browser will attempt to get the images from the blog post relative to the page you've just navigated to, which will 404. 

**To Reproduce**
Steps to reproduce the behaviour:
Monitoring the browsers network requests:
1. Go to a blog post that has images (e.g. "Announcing Guided Remediation in OSV-Scanner")
2. Click on a link to another page on the osv.dev site (e.g. Click "Blog" in the navbar)
3. Observe 404-ed requests made to get images from the new page.

**Expected behaviour**
The browser should not be trying to fetch blog images when leaving the blog's page.

**Screenshots**
![image](https://github.com/google/osv.dev/assets/19356069/9ca208df-165f-4caf-915a-8d2910e904c6)

**Additional context**
Chrome's network browser is telling me the initiator is from `@hotwired/turbo`, which seems to be interacting with how images are relatively loaded from the hugo-generated blog pages.

PR #2217, opened 2024-05-16
Fix 404 error when navigating away from blog page

ZhangChen199102:
Issue: #2184

Turbo always tries to cache the page after a link, but after navigate to a new page, it can't get the images of the blogs cause they use relative path, e.g.
https://github.com/google/osv.dev/blob/2e1117e5c9360f755d46165eaf2f2bf6b576d09f/gcp/appengine/blog/content/posts/announcing-guided-remediation-in-osv-scanner/index.md?plain=1#L36

This PR disables the cache feature of Turbo Drive which is a quick solution, but this will reduce the performance when navigate back to the blog (content won't be stored and it will load a fresh page).

Another solution is improving the images path for each blog, for example use `/static/blog/posts/announcing-guided-remediation-in-osv-scanner/gr_main.png` instead of `gr_main.png`. But by going this way, we need to keep in mind to use absolute path when adding a new blog.

Please let me know your thought, or any other solution.

### DIFF
--- a/gcp/appengine/frontend3/src/templates/blog_post.html
+++ b/gcp/appengine/frontend3/src/templates/blog_post.html
@@ -1,5 +1,6 @@
 {% extends 'base.html' %}
 {% set active_section = 'blog' %}
+{% set disable_turbo_cache = 'true' %}
 
 {% block content %}
 <div class="blog-post-page">


michaelkedar (approved):
I think removing turbo's caching is fine.

> Another solution is improving the images path for each blog, for example use `/static/blog/posts/announcing-guided-remediation-in-osv-scanner/gr_main.png` instead of `gr_main.png`. But by going this way, we need to keep in mind to use absolute path when adding a new blog.

I don't like having to specify the absolute path manually, but I wonder if there's a way to automatically convert these paths from relative to absolute in hugo?

ZhangChen199102:
> I think removing turbo's caching is fine.
> 
> > Another solution is improving the images path for each blog, for example use `/static/blog/posts/announcing-guided-remediation-in-osv-scanner/gr_main.png` instead of `gr_main.png`. But by going this way, we need to keep in mind to use absolute path when adding a new blog.
> 
> I don't like having to specify the absolute path manually, but I wonder if there's a way to automatically convert these paths from relative to absolute in hugo?

I've went through Hugo's doc, and don't think they provide a method to automatically convert paths to absolute path.

Issue #2176, opened 2024-05-08
Request OSS-Fuzz project owner approval before importing OSS-Fuzz reports


davidben:
Automating vulnerability reports from OSS-Fuzz is great, but the unfortunate fact is that OSS-Fuzz is very noisy. It's quite frequent that infrastructure changes cause false positives (see https://github.com/google/oss-fuzz/issues/11881).

Prior to OSV, this was fine because false positive reports could simply be ignored, and then OSS-Fuzz will automatically close them later. Fuzzing is super valuable, so we were more than happy to tolerate a moderate false positive rate, in exchange for avoiding false negatives, **provided that false positives were low cost**.

OSV completely throws this equation off. As soon as a bad report is imported into OSV, a ton of busywork is triggered downstream. The more OSV is used, the worse this will become. See, for example, https://github.com/google/oss-fuzz-vulns/pull/37 where every imported report was incorrect.

Fuzzers are incredibly valuable, and we do not want to be forced to shut off fuzzers because OSS-Fuzz's false vs true positive tradeoffs are no longer viable.

I think the most natural fix here is that, prior to importing from OSS-Fuzz, OSV should ping the project owners and provide an automated way for project owners to modify the imported report. If the maintainers don't respond after a short period of time, it can import automatically. But maintainers who are active *need* to have some way to deal with this noise, or we'll have to stop fuzzing.

evverx:
I came here from https://github.com/google/oss-fuzz/issues/11925 and while I agree that it certainly makes sense to let maintainers vet OSV entries before they are published I think it would be nice if all the entries were published anyway. There are projects that can reject things left and right and it would still be useful to be able get the "raw" feed to double-check.

(more generally I think that the problem is that OSV consumers can be a bit overzealous in that everything is treated as "vulnerability" without vetting anything. Maybe there should be a way to say that some entries are generated based on automated tools and should be vetted. If they are confirmed by maintainers consumers can be confident that entries are vulnerabilities. I proposed "confidence" in another context https://github.com/google/osv.dev/issues/918 but it can probably be applied here)

andrewpollock:
> Maybe there should be a way to say that some entries are generated based on automated tools and should be vetted.

This is a very interesting thought: being able to weight (or even filter) vulnerabilities by their source

davidben:
[...]

PR #2222, opened 2024-05-17
fix: update the OSS-Fuzz denylist

evverx:
It's a follow-up to https://github.com/google/osv.dev/pull/2201. The comments point to https://github.com/google/osv.dev/issues/2178 because the reason is the same.

### DIFF
--- a/docker/worker/oss_fuzz.py
+++ b/docker/worker/oss_fuzz.py
@@ -39,10 +39,15 @@ UNKNOWN_COMMIT = 'unknown'
 # Large projects which take way too long to build.
 # TODO(ochang): Don't hardcode this.
 PROJECT_DENYLIST = {
+    'avahi',  # https://github.com/google/osv.dev/issues/2178
     'boringssl',  # https://github.com/google/osv.dev/issues/2178
+    'dbus-broker',  # https://github.com/google/osv.dev/issues/2178
+    'elfutils',  # https://github.com/google/osv.dev/issues/2178
     'ffmpeg',
     'imagemagick',
+    'libbpf',  # https://github.com/google/osv.dev/issues/2178
     'libreoffice',
+    'systemd',  # https://github.com/google/osv.dev/issues/2178
 }
 
 REPO_DENYLIST = {


oliverchang:
Thanks for the change. Is there a list of false positive OSV entries you can point to as justification for this case?

evverx:
https://github.com/google/oss-fuzz/issues/7434, https://github.com/google/oss-fuzz-vulns/pull/25, https://github.com/google/oss-fuzz-vulns/pull/35. At some point I stopped keeping track of them.

oliverchang:
Thanks for the examples! 

> https://github.com/google/oss-fuzz/issues/7434

Given the brittleness of MSan, I'm inclined for us to just start stop importing MSan issues by default period.

> https://github.com/google/oss-fuzz-vulns/pull/25

This looks like some infrastructure issue that should've hopefully been fixed since.

> https://github.com/google/oss-fuzz-vulns/pull/35

This looks like a legitimate OSS-Fuzz environment issue. 

> I'd also point out things like https://lists.thekelleys.org.uk/pipermail/dnsmasq-discuss/2022q1/016161.html

This wasn't us :) And we still have no idea who was submitting these CVEs.. particularly during an early time when this pipeline wasn't very mature.

More generally, if we a) stop doing this automatically for all MSan issues, and b) improve infrastructure reliably to a point where you no longer see missing fixes, and implement the UX fixes in https://github.com/google/osv.dev/issues/2176#issuecomment-2105139193, would you be open to turning this back on again? 

It's important for us to not turn this off by default for all OSS-Fuzz projects, because we've more often than not observed projects fixing vulnerabilities without ever notifying downstream consumers or requesting a CVE (because it's a lot of work).

evverx:
> More generally, if we a) stop doing this automatically for all MSan issues, and b) improve infrastructure reliably to a point where you no longer see missing fixes, and implement the UX fixes in https://github.com/google/osv.dev/issues/2176#issuecomment-2105139193, would you be open to turning this back on again?

I think it depends. It would still be necessary to filter out reliability bugs that have nothing to do with security to prevent them from ending up in the OSV database and I'm not sure who is going to do that. Though if
> When OSV imports an issue, if the issue was not triaged, include a big warning that this may not be a real security issue due to frequent OSS-Fuzz false positives

was implemented I think they would less likely to trigger overzealous vulnerability scanners.

> we've more often than not observed projects fixing vulnerabilities without ever notifying downstream consumers or requesting a CVE (because it's a lot of work)

I understand that but I'm not sure how it can fixed. Downstream consumers should be able to deal with that in general.

oliverchang:
> > When OSV imports an issue, if the issue was not triaged, include a big warning that this may not be a real security issue due to frequent OSS-Fuzz false positives
> 
> was implemented I think they would less likely to trigger overzealous vulnerability scanners.

I'm thinking this can be a bit somewhere inside the `database_specific` section of the relevant imported OSV entry. If a maintainer has acknowledged the bug somewhere (e.g. via the issue tracker), then this would be true.

evverx:
> this can be a bit somewhere inside the database_specific section of the relevant imported OSV entry

As far as I understand it would work like "confidence" from https://github.com/google/osv.dev/issues/2176#issuecomment-2103129279 so I think it should help. Though I suspect this bit is likely to be false in a lot of entries because (most?) projects aren't particularly interested in vetting OSV entries.

evverx:
Looks like the linter isn't happy. I'll reformat it and force-push the commit.

evverx:
I've just force-pushed the commit. The linter should be happy now hopefully. I was going to add more projects but decided not to partly because they are unlikely to attract that sort of attention and partly because I'm kind of hoping some sort of automated solution to this issue should pop up eventually and it would be nice if it could learn from false positives too.

evverx:
Looks like another linter isn't happy:
```
No release type found in pull request title "update the OSS-Fuzz denylist"
```
I'm not exactly sure what it should be. I'll add "fix" because it's a (temporary) fix from my perspective but if that title is wrong it can be changed by someone else.

oliverchang (approved):

Issue #2216, opened 2024-05-16
Handle very long fields on vulnerabilities pages


oliverchang:
Currently https://osv.dev/vulnerability/USN-6766-2 looks like this:
![image](https://github.com/google/osv.dev/assets/759062/068d3dd8-4d86-4cd4-a0be-6720c4a83069)

We should collapse these by default when there's too many, and let the user click through to expand instead.

oliverchang:
@ZhangChen199102 @zahraaalizadeh is this something one of you can help with?

PR #2223, opened 2024-05-21
Collapse fields on vulnerability page when it's too long

ZhangChen199102:
Issue: #2216

Result:
May-21-2024 15-13-51

Does not render a toggle button if it's a short list:
image

### DIFF
--- a/gcp/appengine/frontend3/src/templates/vulnerability.html
+++ b/gcp/appengine/frontend3/src/templates/vulnerability.html
@@ -53,7 +53,7 @@
           {% if vulnerability.related -%}
           <dt>Related</dt>
           <dd>
-            <ul class="aliases">
+            <ul class="aliases expandible-list">
               {% for related in vulnerability.related -%}
               <li>
                 {% if related | osv_has_vuln -%}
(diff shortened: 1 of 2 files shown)


gcp/appengine/frontend3/src/templates/vulnerability.html
```diff
@@ -38,7 +38,7 @@ <h1 class="title">{{ vulnerability.id }}</h1>
           {% if vulnerability.related -%}
           <dt>Related</dt>
           <dd>
-            <ul class="aliases">
+            <ul class="aliases expandible-list">
```
ZhangChen199102, line 41: We can make any `<ul>...</ul>` collapsible on this page by adding `expandible-list` class.

gcp/appengine/frontend3/src/templates/vulnerability.html
```diff
@@ -230,17 +230,51 @@ <h3 class="mdc-layout-grid__cell--span-3">
...
     package_headers.forEach((header) => {
       header.click()
     });
-  })
+
+    /*
+    Collapse fields if they have too many items. You can enable this feature by adding
+    an 'expandible-list' class to lists with long value.
+    */
+    const EXPANDIBLE_LIST_DEFAULT_LENGTH = 5
```
ZhangChen199102, line 249: Currently it shows at most 5 items by default, but we can change this number later.

andrewpollock: Rather than hard-coding a number, is it possible to take the size of the window into consideration? I think it's important for someone viewing a vulnerability record to be able to visually reason about all of the affected packages in a single page (where the window is reasonably large)

andrewpollock: This was more about expanding the affected packages, than collapsing the references (which I definitely think should be collapsed by default if more than say 5)

andrewpollock (approved):
LGTM, let's submit after this week's release to review in staging

ZhangChen199102:
Thanks @andrewpollock, can you please help me to merge this PR, I just realised that I lost the write access

Issue #1005, opened 2023-01-31
Add button to vulnerability UI to link to how to suggest an improvement


oliverchang:
On our detailed vulnerability page, we could link to a source-specific page for how a user can suggest an improvement.

E.g. for https://osv.dev/vulnerability/GHSA-2cxf-6567-7pp6, we can add a link to https://github.com/advisories/GHSA-2cxf-6567-7pp6/improve

andrewpollock:
In a similar vein to this, I think some of our feedback has come via downstream consumers of our GCS-exported data.

Include a "README" or "FEEDBACK" or file of this nature in

1. the zip files
2. the per-ecosystem GCS bucket directories

that serves as a signpost to these pages that created as part of this issue.

PR #2227, opened 2024-05-22
Add suggest improvement button to vulnerability page

ZhangChen199102:
Issue: #1005

Result:
image

image

### DIFF
--- a/gcp/appengine/frontend3/src/templates/vulnerability.html
+++ b/gcp/appengine/frontend3/src/templates/vulnerability.html
@@ -9,8 +9,20 @@
 <div class="vulnerability-page">
   <div class="mdc-layout-grid">
     <div class="mdc-layout-grid__inner">
-      <div class="mdc-layout-grid__cell--span-12">
-        <h1 class="title">{{ vulnerability.id }}</h1>
+      <div class="mdc-layout-grid__cell--span-12 vulnerability-title">
+        <h1 class="title">
+          {{ vulnerability.id }}
+        </h1>
+        {% if vulnerability.human_source_link and vulnerability.human_source_link.startswith("https://github.com/advisories/") -%}
+          <a class="vulnerability-improvement-link" href="{{ vulnerability.human_source_link }}/improve">
+            Suggest an improvement
+          </a>
+        {% else -%}
+          <a class="vulnerability-improvement-link" href="https://google.github.io/osv.dev/faq/#ive-found-something-wrong-with-the-data">
+            See a problem?
+          </a>
+        {% endif -%}
+      </div>
       </div>
       <div class="mdc-layout-grid__cell--span-12">
         <dl class="vulnerability-details">
--- a/gcp/appengine/frontend3/src/styles.scss
+++ b/gcp/appengine/frontend3/src/styles.scss
@@ -609,6 +609,22 @@ mwc-icon-button.mdc-data-table__sort-icon-button {
       font-size: $osv-heading-size-mobile;
     }
   }
+
+  .vulnerability-title {
+    display: flex;
+    justify-content: space-between;
+    align-items: center;
+
+    .vulnerability-improvement-link {
+      font-family: $osv-heading-font-family;
+    }
+
+    @media (max-width: $osv-mobile-breakpoint) {
+      .vulnerability-improvement-link {
+        display: none;
+      }
+    }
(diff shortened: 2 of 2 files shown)


andrewpollock:
I think suggesting improvements from a mobile device is probably going to be a dreadful user experience, so I suggest not presenting the button on small screens.

For large screens, could we try utilizing the left margin for actions? Bring the button up to be beside the **Source** field.
For

gcp/appengine/frontend3/src/templates/vulnerability.html
```diff
@@ -221,6 +221,14 @@ <h3 class="mdc-layout-grid__cell--span-3">
   </div>
   {% endfor -%}
   </spicy-sections>
+
+  {% if vulnerability.human_source_link and vulnerability.human_source_link.startswith("https://github.com/advisories/") %}
```
andrewpollock, line 225: Please add an else case for the other sources that creates a button titled "See a problem?" that links to https://google.github.io/osv.dev/faq/#ive-found-something-wrong-with-the-data

gcp/appengine/frontend3/src/templates/vulnerability.html
```diff
@@ -283,6 +283,14 @@ <h3 class="mdc-layout-grid__cell--span-3">
       </div>
       {% endfor -%}
     </spicy-sections>
+  
+    {% if vulnerability.human_source_link and vulnerability.human_source_link.startswith("https://github.com/advisories/") -%}
```
ZhangChen199102, line 287: I found that not every vulnerability has a github advisory (e.g.https://osv.dev/vulnerability/GO-2024-2571), so I think we might want to check if the `human_source_link` starts with 'https://github.com/advisories/'. Please let me know if this is wrong.

ZhangChen199102:
> I think suggesting improvements from a mobile device is probably going to be a dreadful user experience, so I suggest not presenting the button on small screens.
>
>For large screens, could we try utilizing the left margin for actions? Bring the button up to be beside the Source field.
For

Thanks @andrewpollock, I've tried moving the button next to Source, but I found the UI is bit of Inconsistent (probably because the height of the button):
image

So I tried moving below of Source and it looks better I think:
image

Please let me know your thought

andrewpollock:
I was actually thinking about having it in the left gutter next to the Source title text

andrewpollock:
I was actually thinking about having it in the left gutter next to the Source title text itself, not the link

ZhangChen199102:
@andrewpollock do you mean something like this:
WX20240527-100619@2x

Or maybe:
left-side-button

andrewpollock:
@ZhangChen199102 yes, the second option in https://github.com/google/osv.dev/pull/2227#issuecomment-2132519497 was what I had in mind. 

You'll need to adjust the overall height of each row so that when the button gets rendered it doesn't make the Source line vertical spacing look unusual.

Another possibility is to de-couple it from the Source row and put the button to the left of the identifier instead...

ZhangChen199102:
@andrewpollock thanks, my only concern is when the screen gets narrower (but still wider than mobile devices screen), there will be very limited space between the text body and browser's border, so I wonder where should we put the button in this case:
image

andrewpollock:
> my only concern is when the screen gets narrower

Is it possible to progressively not render components in this case? Or if we go with putting the button to the side of the identifier, it could go on the right hand side of the identifier at all times. There's guaranteed to be plenty of screen real estate on that side...

ZhangChen199102:
I've moved the button next to the title:
image

Also I can move the button to very right side if you think they're too closed, e.g.
image

andrewpollock:
I like the right-justified button (next to the ID) best of what I've seen so far.

andrewpollock:
We talked about this in our team meeting, and agree that the right-justified button on the same line as the identifier is preferred. We would like to de-emphasize the button a little, and were thinking about the ecosystem buttons on https://osv.dev/list

ZhangChen199102:
Thanks @andrewpollock, here's the latest result:
image

And the button will be hidden for mobile device:
image

andrewpollock (approved):
LGTM!

ZhangChen199102:
@another-rex here’s the button without border and background color:
image

image

another-rex:
Let's go with the link option (no border/bg color)!

PR #2241, opened 2024-05-28
Fix another bug in alpine enumeration

another-rex:
Turns out alpine enumeration currently works by accident, relying on git to bundle the changes to package version and package revision in the same diff hunk, which is not always the case.

This caused enumeration to break for some packages (e.g. busybox https://osv.dev/vulnerability/CVE-2023-42366 , 3.16.1 only have r11).

This change fixes this by separating out the "start" search for versions and revisions, so changes to both are displayed, not just changes to the first element found (which is usually `pkgver`, therefore it was relying on changes in `pkgrel` to be in the same diff hunk in git to be shown).


osv/ecosystems/alpine.py
```diff
@@ -141,14 +141,17 @@ def _get_versions(branch: str, package: str) -> typing.List[str]:
 
     relative_path = os.path.relpath(directories[0], checkout_dir)
 
-    regex_test = '\\|'.join([
-        '\\(' + x.removeprefix('+') + '\\)'
-        for x in Alpine._PKGVER_ALIASES + Alpine._PKGREL_ALIASES
-    ])
+    regex_test_aliases = '\\|'.join(
```
michaelkedar, line 144: nit: Name this `regex_test_versions`?

osv/ecosystems/alpine.py
```diff
@@ -141,14 +141,17 @@ def _get_versions(branch: str, package: str) -> typing.List[str]:
...
 
+    regex_test_revisions = '\\|'.join(
+        ['\\(' + x.removeprefix('+') + '\\)' for x in Alpine._PKGREL_ALIASES])
+
+    # https://git-scm.com/docs/git-log/2.33.1#Documentation/git-log.txt--Lltstartgtltendgtltfilegt
     stdout_data = subprocess.check_output([
         'git', 'log', '--oneline', '-L',
-        '/' + regex_test + '/,+1:' + relative_path
+        '^/' + regex_test_aliases + '/,+1:' + relative_path, '-L',
+        '^/' + regex_test_revisions + '/,+1:' + relative_path
```
michaelkedar, line 154: nit: Will the linter let you format this as
```
'git', 'log', '--oneline',
'-L', '^/' + ...,
'-L', '^/' + ...,
],
```
So the -L flag is together with its value

michaelkedar (approved):
LGTM, with some tiny, tiny nits

andrewpollock (commented):
Could you add a test for something that is currently failing that no longer fails with this change?

osv/ecosystems/alpine.py
another-rex: I tried, the linter keeps undoing my attempts :(

andrewpollock (commented):
I'm not sufficiently familiar with this code to offer a meaningful review of this change without spending additional time to familiarize myself with it. I'm happy to defer to Michael's LGTM. I'd suggest adding some more docstrings to help unfamiliar readers get up to speed with what this code is trying to accomplish.

another-rex:
Adding tests for this is quite tough, as the old version actually works 90% of the time with no predictable way (afaik) to trigger the issue. I'll definitely add some more comments to explain what this file is doing though.

PR #2246, opened 2024-05-29
Remove withdrawn entry

another-rex:
Withdrawn entry has now been removed from the API calls.

@andrewpollock is this correct? Is this caused by https://github.com/google/osv.dev/pull/2133 ?


another-rex:
Wait it looks like this has already been fixed and I'm just on a older branch...

PR #2247, opened 2024-05-29
Fixed UI Affected Packages Ecosystems label being in uppercase

### LINKED ISSUE #2240: Show ecosystem case in osv.dev
**Describe the bug**
Currently on osv.dev vulnerability details page, the ecosystem is styled to always show uppercase. 
Example of the above

However, the ecosystem field is case sensitive, so this should not be styled uppercase.

jess-lowe:
Removed the text-transform property so affected packages ecosystems should be in normal case:
image

Closes #2240

P.S Hey again :D


another-rex (approved):
Thanks Jess!

PR #2248, opened 2024-05-29
Manually bump requests as renovatebot doesn't seem to update them correctly

another-rex:
Manually bump requests to resolve https://osv.dev/vulnerability/GHSA-9wx4-h78v-vm56 as renovatebot doesn't seem to reliably update them. (And causes other breaking changes which will be fixed in followup PRs)


docker/worker/Pipfile
```diff
@@ -11,7 +11,8 @@ PyYAML = "*"
 google-cloud-storage = "*"
 redis = "*"
 osv = {editable = true, path = "./../.."}
-pygit2 = ">=1.14.0"
+pygit2 = "==1.14.1"
```
another-rex, line 14: Pinning is needed here as 1.15.0 causes breaking changes. This will be resolved in a followup PR.

cuixq (approved):

michaelkedar (approved):

PR #2249, opened 2024-05-29
Remove copy pasta

andrewpollock:
This was causing reader confusion about the expected behaviour of this test case.


another-rex (commented):
Can you also add a check for expectFailure actually correctly failing?

another-rex (approved):

andrewpollock:
> Can you also add a check for expectFailure actually correctly failing?

I just made 2f364993 but I think what you're actually asking for is already covered by https://github.com/google/osv.dev/blob/2f3649933e1292bac39e99bfbbcef350eea5eb79/vulnfeeds/cves/versions_test.go#L545-L565 and https://github.com/google/osv.dev/blob/2f3649933e1292bac39e99bfbbcef350eea5eb79/vulnfeeds/cves/versions_test.go#L461-L466 and 

https://github.com/google/osv.dev/blob/2f3649933e1292bac39e99bfbbcef350eea5eb79/vulnfeeds/cves/versions_test.go#L476-L481

?

another-rex:
2f36499 is what I mostly meant, just showing a message to make it clear that it is actually a success when it's supposed to have failed, which means that cases like this one where expect failure is accidentally set to true will be caught rather than just ignored.

PR #2257, opened 2024-05-30
Remove App Engine backend from GCLB

michaelkedar:
This just removes App Engine from the load balancer. I'll manually apply this on prod after it's merged.

I still need to understand what happens to Datastore if we delete App Engine entirely.

### DIFF
--- a/deployment/terraform/modules/osv/website.tf
+++ b/deployment/terraform/modules/osv/website.tf
@@ -94,31 +94,6 @@ module "gclb" {
   url_map        = google_compute_url_map.website.id
 
   backends = {
-    appengine = {
-      groups = [
-        {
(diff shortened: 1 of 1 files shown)


another-rex (approved):
Do we also need to delete 
```
resource "google_project_service" "app_engine" {
```
in gcp_apis.tf? Or does this need to get removed first

michaelkedar:
> Do we also need to delete
> 
> ```
> resource "google_project_service" "app_engine" {
> ```
> 
> in gcp_apis.tf? Or does this need to get removed first

That's what enabled App Engine on the Google Cloud Project, and since I'm not removing App Engine yet (just the Load Balancer config) it should probably stay
(Not that removing it will do anything anyway, since it won't disable the api when destroyed)

Issue #1117, opened 2023-03-14
Gracefully handle negative page offsets 


andrewpollock:
As seen in a review of the logs, someone or something is feeding negative page offsets, which results in undesired noise in the logs:

```
Traceback (most recent call last):
  File "/layers/google.python.pip/pip/lib/python3.10/site-packages/flask/app.py", line 2528, in wsgi_app
    response = self.full_dispatch_request()
  File "/layers/google.python.pip/pip/lib/python3.10/site-packages/flask/app.py", line 1825, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/layers/google.python.pip/pip/lib/python3.10/site-packages/flask/app.py", line 1823, in full_dispatch_request
    rv = self.dispatch_request()
  File "/layers/google.python.pip/pip/lib/python3.10/site-packages/flask/app.py", line 1799, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)
  File "/srv/frontend_handlers.py", line 171, in list_vulnerabilities
    results = osv_query(query, page, False, ecosystem)
  File "/srv/frontend_handlers.py", line 377, in osv_query
    bugs, _, _ = query.fetch_page(
  File "/layers/google.python.pip/pip/lib/python3.10/site-packages/google/cloud/ndb/query.py", line 1214, in wrapper
    return wrapped(self, *dummy_args, _options=query_options)
  File "/layers/google.python.pip/pip/lib/python3.10/site-packages/google/cloud/ndb/utils.py", line 121, in wrapper
    return wrapped(*args, **new_kwargs)
  File "/layers/google.python.pip/pip/lib/python3.10/site-packages/google/cloud/ndb/utils.py", line 153, in positional_wrapper
    return wrapped(*args, **kwds)
  File "/layers/google.python.pip/pip/lib/python3.10/site-packages/google/cloud/ndb/query.py", line 2281, in fetch_page
    return self.fetch_page_async(None, _options=kwargs["_options"]).result()
  File "/layers/google.python.pip/pip/lib/python3.10/site-packages/google/cloud/ndb/tasklets.py", line 210, in result
    self.check_success()
  File "/layers/google.python.pip/pip/lib/python3.10/site-packages/google/cloud/ndb/tasklets.py", line 157, in check_success
    raise self._exception
  File "/layers/google.python.pip/pip/lib/python3.10/site-packages/google/cloud/ndb/tasklets.py", line 330, in _advance_tasklet
[...]

PR #2258, opened 2024-05-30
Handle negative page for vulnerability list page

ZhangChen199102:
Issue: #1117

Automatically redirect to the first page if query string has a negative page number


another-rex (approved):
Thanks!

Issue #2255, opened 2024-05-30
Make it possible to visually evaluate a list of  vulnerabilities by severity


andrewpollock:
**Is your feature request related to a problem? Please describe.**
When viewing a list of vulnerabilities at https://osv.dev/list it is not possible to visually discern the severity of them

**Describe the solution you'd like**
- Sort by CVSS base score
- Have a graduated red background for each vulnerability's row based on severity

**Describe alternatives you've considered**
None, other than leaving as is.

**Additional context**
- https://pypi.org/project/cvss/ is probably relevant to the implementation for this

zahraaalizadeh:
@another-rex Do we want to consider sorting feature for severity here too? Or this issue is resolved now?

another-rex:
I think this is resolved for now. The sorting is a much more long term issue.

another-rex:
FIxed in https://github.com/google/osv.dev/pull/2262

Issue #2254, opened 2024-05-30
Calculate and display the CVSS base score


andrewpollock:
**Is your feature request related to a problem? Please describe.**
It's not possible to reason about the CVSS severity of a vulnerability from the individual vulnerability page

**Describe the solution you'd like**
Calculate and display the CVSS base score alongside the vector string. Potentially link out to something like FIRST or the NVD's CVSS calculator so the user can further understand the CVSS score.

**Describe alternatives you've considered**
None, other than leaving it

**Additional context**
Fun finds:
* you can directly link a CVSS 3.0 vector string in FIRST's CVSS calculator, e.g. https://www.first.org/cvss/calculator/3.0#CVSS:3.0/AV:L/AC:H/PR:H/UI:R/S:U/C:H/I:H/A:N (I took this string from GHSA-9wx4-h78v-vm56 and had to manually change the 3.1 to 3.0 to get it to work)
* you can directly link a CVSS 3.0 or 3.1 vector string to the NVD's CVSS calculator, e.g. https://nvd.nist.gov/vuln-metrics/cvss/v3-calculator?vector=AV:L/AC:H/PR:H/UI:R/S:U/C:H/I:H/A:N&version=3.1 for the same string as above
* https://pypi.org/project/cvss/ may or may not be relevant to the implementation of the solution here

PR #2262, opened 2024-06-03
Add severity column to vulnerability list

ZhangChen199102:
Issue: #2255

image

### DIFF
--- a/gcp/appengine/Pipfile.lock
+++ b/gcp/appengine/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "240619547f891be54f53c119bf3555fa120e4d6a61cd2b35d192968d5d7e563b"
+            "sha256": "82f643493aa6d0daa5fadae7962316d832d413f4ed9066958325825318d59786"
         },
         "pipfile-spec": 6,
         "requires": {
@@ -16,6 +16,14 @@
         ]
     },
     "default": {
+        "async-timeout": {
+            "hashes": [
+                "sha256:4640d96be84d82d02ed59ea2b7105a0f7b33abe8703703cd0ab0bf87c427522f",
+                "sha256:7405140ff1230c310e51dc27b3145b9092d659ce68ff733fb0cefe3ee42be028"
+            ],
+            "markers": "python_full_version <= '3.11.2'",
+            "version": "==4.0.3"
+        },
         "attrs": {
             "hashes": [
                 "sha256:935dc3b529c262f6cf76e50877d35a4bd3c1de194fd41f47a2b7ae8f19971f30",
@@ -296,7 +304,7 @@
                 "sha256:fd1abc0d89e30cc4e02e4064dc67fcc51bd941eb395c502aac3ec19fab46b519",
                 "sha256:ff8fa367d09b717b2a17a052544193ad76cd49979c805768879cb63d9ca50561"
             ],
-            "markers": "python_version >= '3.7'",
+            "markers": "python_full_version >= '3.7.0'",
             "version": "==3.3.2"
         },
         "click": {
@@ -628,7 +636,8 @@
         },
         "osv": {
             "editable": true,
-            "path": "./../.."
+            "markers": "python_version >= '3.7'",
+            "path": "../../"
         },
         "packaging": {
             "hashes": [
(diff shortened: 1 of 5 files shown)


gcp/appengine/frontend3/src/templates/list.html
```diff
@@ -133,6 +135,13 @@ <h1 class="title">Vulnerability Library</h1>
               <span class="tag fix-unavailable">No fix available</span>
               {%- endif -%}
             </span>
+            <span role="cell" class="vuln-table-cell mdc-data-table__cell hide-on-mobile">
+              <ul>
+                {% for score in (vulnerability.severity_scores) %}
+                <li>{{ score }}</li>
+                {% endfor %}
```
andrewpollock, line 142: I think it would be preferable to only output the highest available CVSS score, I.e. where CVSSv3 and CVSSv4 scores are provided, favour the v4 score.

andrewpollock (commented):
LGTMish (without seeing a preview), with some thoughts on handling multiple CVSS versions.

There's also the UX question of records that do not have any CVSS scores and how they would sort...

andrewpollock (approved):
LGTM

oliverchang:
> LGTM

This adds more width to our page :/ Should we evaluate removing some other columns first?

oliverchang:
e.g. fix available, and versions probably isn't that useful.

hogo6002:
Hey @ZhangChen199102 , thanks for implementing this feature! After discussing with the team, we have some thoughts on this change:
- Instead of adding a new column for the severity score, we can just move it into the "Fix Available" column and making it more of an "Attributes" column. This can reduce the number of columns and allow us to easily add other tags in the future.
- the current severity tag colour makes the text a bit hard to see. Maybe we can just use a white tag with black text, displaying the information like "Severity - 5.4 (MEDIUM)"?
- the "Affected versions" listed here aren't adding much information and make the page looks a bit crowded. We can remove this column entirely from https://osv.dev/list (this can be done in a separate PR, and it would also automatically resolve https://github.com/google/osv.dev/issues/2286)

zahraaalizadeh (commented):
Resolves: https://github.com/google/osv.dev/issues/2255,  https://github.com/google/osv.dev/issues/2286

Screenshot 2024-06-19 at 3 10 51 PM

another-rex:
Looks really good! Can you also try keeping the color, but for the darker colors have the text color switch from black to white to make it easier to read? I'm not sure if we'll go with that, but I think it would be nice to compare.

zahraaalizadeh:
With the current colors, switching the text color to white instead of black actually fails the contrast ratio for accessibility (checked using [this tool](https://webaim.org/resources/contrastchecker/)).

I’ve attached a sample output for reference, showing how the text looks with white colors on these backgrounds. Additionally, I’ve attached a photo demonstrating the appearance with black text on these colors.

As another alternative, I’ve used red-themed colors for high and critical severities to maintain the contrast.

Sample output with white text on colors( fails accessibility checks )
Screenshot 2024-06-20 at 4 34 33 PM

Black and white text on colors ( passes accessibility checks )
Screenshot 2024-06-20 at 4 41 52 PM

Red-themed colors for medium high and critical severities
Screenshot 2024-06-20 at 5 06 07 PM

another-rex:
I personally prefer the middle option, though can you try switching the orange (High) text color to white instead of black?

hogo6002:
> I personally prefer the middle option, though can you try switching the orange (High) text color to white instead of black?

+1, I also think the middle one looks better. The third one has a pink colour, which is the same as our "no fix available" tag.

zahraaalizadeh:
> I personally prefer the middle option, though can you try switching the orange (High) text color to white instead of black?

@another-rex, unfortunately, using white on orange fails the accessibility contrast checks. the result is in demonstrated on pic1 in my previous comment.

Screenshot 2024-06-25 at 9 39 04 AM

zahraaalizadeh:
Here is the updated colours:
Screenshot 2024-06-25 at 4 39 25 PM

another-rex (commented):
Styling LGTM! Just some minor comments.

gcp/appengine/Pipfile.lock
```diff
@@ -969,63 +978,12 @@
...
-                "sha256:e9e9d4e2e336c529d4c435baad846a181e39a982f823f7e4495ec0b0ec8538d2",
-                "sha256:f058a77ef0ece4e210bb0450e68408d4223f728b109764676e1a13537d056bb0",
-                "sha256:f1a4b358947a65b94e2501ce3e078bbc929b039ede4679ddb0460829b12f7375",
-                "sha256:f9b2cde1cd1b2a10246dbc143ba49d942d14fb3d2b4bccf4618d475c65464912",
-                "sha256:fe3390c538f12437b859d815040763abc728955a52ca6ff9c5d4ac707c4ad98e"
-            ],
-            "markers": "platform_python_implementation != 'PyPy'",
-            "version": "==0.22.0"
+            "markers": "python_version >= '3.7'",
+            "version": "==6.5.0"
```
zahraaalizadeh, line 986: The version of `whitenoise` specified in the Pipfile is `6.5.0`, so it makes sense that `pipenv lock` downgrades the version from `6.7.0` to `6.5.0`, as introduced by [this](https://github.com/google/osv.dev/commit/fa0672b233985ca2e57f752e1f91542f80b8b22f) commit.
However I'm unsure why it removed `zstandard`. I suspect it might be because I'm running it from a mac machine?

gcp/appengine/frontend_handlers.py
```diff
@@ -297,39 +297,41 @@ def bug_to_response(bug, detailed=True):
   return response
 
 
+def calculate_severity_details(
+    severity: dict) -> tuple[float | None, str | None]:
+  """Calculate score and rating of severity"""
```
zahraaalizadeh, line 302: I cherry-picked this commit from my other branch which adds[ link to cvss calculator.](https://github.com/google/osv.dev/issues/2254)
FYI - Once this PR was merged, I'll resolve the potential conflicts and open a seperate PR for the mentioned issue.

gcp/appengine/Pipfile.lock
another-rex: Hmm I'm guessing whitenoise added zstd support somewhere between 6.7.0 and 6.5.0. Since it looks like we are already using whitenoise 6.7.0, can you update our pipfile for whitenoise to ==6.7.0 (The only reason renovatebot didn't do it is because there's currently a dependency conflict in the Google cloud dependencies. )

another-rex (approved):

gcp/appengine/Pipfile.lock
zahraaalizadeh: Re `zstandard`, looking at `pipenv graph` it's a dependancy for `flask-compress` version `1.15` as the version in Pipfile is specified as `1.13` it's been removed.

zahraaalizadeh: @another-rex Should I upgrade`flask-compress` from `1.13` to `1.15` too?

hogo6002 (approved):
Looks really nice! Thanks Zahra!

PR #2263, opened 2024-06-03
Update terraform version

michaelkedar:
I'd like to use an [import block](https://developer.hashicorp.com/terraform/language/v1.5.x/import) to help import the datastore into terraform, which was introduced in v1.5.

Please also update your local terraform versions to at least 1.5.

1.5.7 isn't the latest version of terraform, but version 1.6 is when they changed their license and I don't really understand those ramifications :sweat_smile:

### DIFF
--- a/docker/terraform/README.md
+++ b/docker/terraform/README.md
@@ -18,10 +18,10 @@ steps:
 ```
 
 ## Updating Terraform Version
-`_TERRAFORM_VERSION` is defined in `cloudbuild.yaml`. Currently version 1.3.8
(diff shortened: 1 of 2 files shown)


cuixq (approved):

andrewpollock (approved):

hogo6002 (approved):

andrewpollock:
> Please also update your local terraform versions to at least 1.5.

That's `go install github.com/hashicorp/terraform@latest` for anyone else getting that way...

PR #2265, opened 2024-06-03
A smarter osv.dev website caching implementation

another-rex:
**Problem:**
Every 24 hours, a few unfortunate visitors will have osv.dev take incredibly long to load (over 30 seconds), as the ecosystem counts cache expires, and the website tries to count the number of advisories in each ecosystem.

**Solution:**
This PR adds a smarter refresh logic where at half the cache timeout duration (24 hours / 2 = 12 hours), the next request will update the cache in a background thread, while still returning results instantly from the existing cache. When updated, the old cache value is updated and the cache timeout reset. 

This also keeps track of whether an update is already in progress, and prevents python from spawning multiple threads to update the same thing.

---

Also some minor cleanup of main.py

### DIFF
--- a/gcp/appengine/frontend_handlers.py
+++ b/gcp/appengine/frontend_handlers.py
@@ -349,7 +349,8 @@ def osv_get_ecosystems():
 def osv_get_ecosystem_counts_cached():
   """Get count of vulnerabilities per ecosystem, cached"""
   # Check if we're already in ndb context, if not, put us in one
-  # We can sometimes not be in ndb context because caching runs in a separate thread
+  # We can sometimes not be in ndb context because caching
+  # runs in a separate thread
   if ndb.get_context(raise_context_error=False) is None:
     # Make sure this ndb.Client is the same as the one defined in main.py
     with ndb.Client().context():
--- a/gcp/appengine/frontend_handlers.py
+++ b/gcp/appengine/frontend_handlers.py
@@ -352,7 +352,8 @@ def osv_get_ecosystem_counts_cached():
   # We can sometimes not be in ndb context because caching
   # runs in a separate thread
   if ndb.get_context(raise_context_error=False) is None:
-    # IMPORTANT: Ensure this ndb.Client remains consistent with the one defined in main.py
+    # IMPORTANT: Ensure this ndb.Client remains consistent 
+    # with the one defined in main.py
     with ndb.Client().context():
       return osv_get_ecosystem_counts()
 
--- a/gcp/appengine/frontend_handlers.py
+++ b/gcp/appengine/frontend_handlers.py
@@ -352,7 +352,7 @@ def osv_get_ecosystem_counts_cached():
   # We can sometimes not be in ndb context because caching
   # runs in a separate thread
   if ndb.get_context(raise_context_error=False) is None:
-    # Make sure this ndb.Client is the same as the one defined in main.py
+    # IMPORTANT: Ensure this ndb.Client remains consistent with the one defined in main.py
     with ndb.Client().context():
       return osv_get_ecosystem_counts()
 
--- a/gcp/appengine/frontend_handlers.py
+++ b/gcp/appengine/frontend_handlers.py
@@ -28,7 +28,8 @@ from flask import render_template, render_template_string
 from flask import request
 from flask import url_for
 from flask import send_from_directory
-from werkzeug.utils import safe_join
+from werkzeug.security import safe_join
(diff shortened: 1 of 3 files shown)


gcp/appengine/frontend_handlers.py
```diff
@@ -345,11 +345,16 @@ def osv_get_ecosystems():
...
-# TODO: Figure out how to skip cache when testing
-@cache.instance.cached(
-    timeout=24 * 60 * 60, key_prefix='osv_get_ecosystem_counts')
+@cache.smart_cache("osv_get_ecosystem_counts", timeout=24 * 60 * 60)
 def osv_get_ecosystem_counts_cached():
   """Get count of vulnerabilities per ecosystem, cached"""
+  # Check if we're already in ndb context, if not, put us in one
+  # We can sometimes not be in ndb context because caching runs in a separate thread
+  if ndb.get_context(raise_context_error=False) is None:
+    # Make sure this ndb.Client is the same as the one defined in main.py
```
andrewpollock, line 354: Is this comment a TODO or a statement of what the following line is doing?

gcp/appengine/frontend_handlers.py
```diff
@@ -345,11 +345,16 @@ def osv_get_ecosystems():
                 key=str.lower)
 
 
-# TODO: Figure out how to skip cache when testing
-@cache.instance.cached(
-    timeout=24 * 60 * 60, key_prefix='osv_get_ecosystem_counts')
+@cache.smart_cache("osv_get_ecosystem_counts", timeout=24 * 60 * 60)
```
andrewpollock, line 348: Could we use https://pypi.org/project/humanize/ to make this timeout a little more human-friendly?

andrewpollock (approved):
🧠  LGTM with a couple of non-blockers. 🧠

gcp/appengine/frontend_handlers.py
another-rex: Just a statement of what the following line is doing. Right now we just call `ndb.Client()` with no arguments, but I can imagine in the future someone wants to add a argument to `ndb.Client()`, changes it in main.py, then gets really confused that the caching mechanism behaves differently. Not sure how to prevent this though.

andrewpollock: Oh, so it's more a warning to future modifiers? I didn't take that away from the comment as written, perhaps this instead?
```suggestion
    # IMPORTANT: Ensure this ndb.Client remains consistent with the one defined in main.py
```

andrewpollock (approved):

gcp/appengine/frontend_handlers.py
```diff
@@ -345,11 +345,17 @@ def osv_get_ecosystems():
...
 def osv_get_ecosystem_counts_cached():
   """Get count of vulnerabilities per ecosystem, cached"""
+  # Check if we're already in ndb context, if not, put us in one
+  # We can sometimes not be in ndb context because caching
+  # runs in a separate thread
+  if ndb.get_context(raise_context_error=False) is None:
+    # IMPORTANT: Ensure this ndb.Client remains consistent with the one defined in main.py
```
another-rex, line 355:
```suggestion
    # IMPORTANT: Ensure this ndb.Client remains consistent 
    # with the one defined in main.py
```

gcp/appengine/frontend_handlers.py
another-rex: Thanks, much clearer!

gcp/appengine/frontend_handlers.py
```diff
@@ -345,11 +345,18 @@ def osv_get_ecosystems():
...
-@cache.instance.cached(
-    timeout=24 * 60 * 60, key_prefix='osv_get_ecosystem_counts')
+@cache.smart_cache("osv_get_ecosystem_counts", timeout=24 * 60 * 60)
 def osv_get_ecosystem_counts_cached():
   """Get count of vulnerabilities per ecosystem, cached"""
+  # Check if we're already in ndb context, if not, put us in one
+  # We can sometimes not be in ndb context because caching
+  # runs in a separate thread
+  if ndb.get_context(raise_context_error=False) is None:
+    # IMPORTANT: Ensure this ndb.Client remains consistent
```
another-rex, line 355:
```suggestion
    # IMPORTANT: Ensure this ndb.Client remains consistent
```

michaelkedar (approved):

cuixq (approved):

PR #2287, opened 2024-06-06
Replicate the exception handling fix from #2283

andrewpollock:
Avoid any copy pasta fails if someone cribs from one of these for a future script and replicate the exception handling fix from #2283 to all the scripts.


hogo6002 (approved):
Thanks Andrew!

Issue #2281, opened 2024-06-05
Error importing osv in Python 3.9


javihernandez:
**Describe the bug**
Since last 0.0.21 update, on Python 3.9, I can no longer make use of `osv` as I get the following error:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/code/alws/utils/osv.py", line 5, in <module>
    from errata2osv import errata_to_osv
  File "/usr/local/lib/python3.9/site-packages/errata2osv.py", line 6, in <module>
    from src.models import *
  File "/usr/local/lib/python3.9/site-packages/src/models/__init__.py", line 1, in <module>
    from .converter import *
  File "/usr/local/lib/python3.9/site-packages/src/models/converter.py", line 4, in <module>
    from osv import Bug
  File "/usr/local/lib/python3.9/site-packages/osv/__init__.py", line 17, in <module>
    from .impact import *
  File "/usr/local/lib/python3.9/site-packages/osv/impact.py", line 31, in <module>
    from . import models
  File "/usr/local/lib/python3.9/site-packages/osv/models.py", line 31, in <module>
    from . import purl_helpers
  File "/usr/local/lib/python3.9/site-packages/osv/purl_helpers.py", line 42, in <module>
    def package_to_purl(ecosystem: str, package_name: str) -> str | None:
TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
```

**To Reproduce**
Steps to reproduce the behaviour:
1. Create a python 3.9 env
2. Install osv through any installation method
3. Try importing `purl_helpers`
**Expected behaviour**
I expect `purl_helpers` to be loaded without any problem

**Additional context**
If not wrong, type annotations like ` str | None` support was introduced in 3.10, and I see that, according to osv docs, it's supposed to work in `Python >= 3.7`.

hogo6002:
Yeah, we will drop support and require `Python >= 3.10`. I will update docs to reflect this. Thanks for reporting it!

Issue #1637, opened 2023-09-18
Case sensitivity of Ecosystem names for API calls


hayleycd:
While I was looking into #1635 I realized that we have not explicitly indicated the formats for the "ecosystem" parameter in the package object. I tried a sample call and got an error when I tried to use "pypi" for the ecosystem. "PyPI" however, did work. "Go" works but "go" doesn't. 

I am happy to update documentation to help people choose the correct ecosystem. But should the ecosystem parameter be case sensitive at all?

hayleycd:
package names also appear to be case sensitive

hayleycd:
This is a potentially related discussion: #1673 

It involves versions instead of package names/ecosystems but the question of how strict to be about matching is the same.

hayleycd:
Per discussion with @oliverchang, this is behaving as intended 
- We want ecosystem to be case sensitive
- We want the package and versioning to be sensitive as well because it varies among ecosystems. 

I will update docs to reflect case sensitivity

PR #2289, opened 2024-06-06
Update docs to reflect API case sensitivity and python version

hogo6002:
https://github.com/google/osv.dev/issues/1637: Add case sensitive requirement to API docs 
https://github.com/google/osv.dev/issues/2281: Update Python required version to 3.11+

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -51,6 +51,6 @@ setuptools.setup(
         # Include any JSON schemas.
         '': ['*.json'],
     },
-    python_requires='>=3.10',
+    python_requires='>=3.11',
     zip_safe=False,
 )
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -51,6 +51,6 @@
         # Include any JSON schemas.
         '': ['*.json'],
     },
-    python_requires='>=3.7',
+    python_requires='>=3.10',
```
andrewpollock, line 54: https://github.com/google/osv.dev/blob/master/CONTRIBUTING.md#prerequisites states Python 3.11, so for consistency we might want to make this that...

```suggestion
    python_requires='>=3.11',
```

andrewpollock (approved):
LGTM with a suggested consistency change

setup.py
hogo6002: Good point, thanks Andrew!

oliverchang (approved):

PR #2297, opened 2024-06-07
Set up sitemaps 

another-rex:
Set up sitemaps generation on a cronjob to a bucket.

The bucket is served via the load balancer on paths prefixed by `/sitemap_`

This PR contains:
- New bucket for prod and test
- New bucket backend for serving assets
- Load balancer config to redirect requests to the bucket backend when appropriate
- Cron config for prod and test

### DIFF
--- a/deployment/clouddeploy/gke-workers/base/generate-sitemap.yaml
+++ b/deployment/clouddeploy/gke-workers/base/generate-sitemap.yaml
@@ -1,0 +1,24 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: generate-sitemap
+spec:
+  schedule: "30 8 * * *"
+  concurrencyPolicy: Forbid
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          containers:
+          - name: generate-sitemap-cron
+            image: cron
+            imagePullPolicy: Always
+            command: ["/usr/local/bin/generate_sitemap/generate_sitemap.py", "--base_url", "$BASE_URL"]
+            resources:
+              requests:
+                cpu: 1
+                memory: "4G"
+              limits:
+                cpu: 1
+                memory: "6G"
+          restartPolicy: OnFailure
--- a/deployment/clouddeploy/gke-workers/environments/oss-vdb-test/generate-sitemap.yaml
+++ b/deployment/clouddeploy/gke-workers/environments/oss-vdb-test/generate-sitemap.yaml
@@ -1,0 +1,14 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: generate-sitemap
+spec:
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          containers:
+          - name: generate-sitemap-cron
+            env:
+            - name: BASE_URL
+              value: "https://test.osv.dev"
--- a/deployment/clouddeploy/gke-workers/environments/oss-vdb-test/generate-sitemap.yaml
+++ b/deployment/clouddeploy/gke-workers/environments/oss-vdb-test/generate-sitemap.yaml
@@ -12,3 +12,5 @@ spec:
             env:
             - name: BASE_URL
               value: "https://test.osv.dev"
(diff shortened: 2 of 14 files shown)


deployment/clouddeploy/gke-workers/base/generate-sitemap.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: generate-sitemap
+spec:
+  schedule: "30 8 * * *"
```
michaelkedar, line 6: Just to confirm, you want this to run once per day at 8:30?

deployment/clouddeploy/gke-workers/environments/oss-vdb-test/generate-sitemap.yaml
```diff
@@ -0,0 +1,14 @@
...
+metadata:
+  name: generate-sitemap
+spec:
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          containers:
+          - name: generate-sitemap-cron
+            env:
```
michaelkedar, line 12: Please also add a `GOOGLE_CLOUD_PROJECT` env

deployment/clouddeploy/gke-workers/environments/oss-vdb/generate-sitemap.yaml
```diff
@@ -0,0 +1,14 @@
...
+metadata:
+  name: generate-sitemap
+spec:
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          containers:
+          - name: generate-sitemap-cron
+            env:
```
michaelkedar, line 12: `GOOGLE_CLOUD_PROJECT` env

docker/cron/generate_sitemap/generate_sitemap.py
```diff
@@ -78,22 +77,8 @@ def generate_sitemap_for_ecosystem(ecosystem: str, base_url: str) -> None:
   tree.write(filename, encoding='utf-8', xml_declaration=True)
 
 
-def compress_file(file_path: str) -> str:
```
michaelkedar, line 81: Is there any particular reason you removed the compression?

michaelkedar (commented):
Nice! Just a couple of comments

deployment/clouddeploy/gke-workers/base/generate-sitemap.yaml
another-rex: Yes, that's the idea, I didn't want to run it at midnight as it might line up with other tasks causing a spike in performance.

docker/cron/generate_sitemap/generate_sitemap.py
another-rex: I'm not 100% sure how to get GCS to serve .gz files directly as their non compressed version (with a .xml extension) (and setting the content-type correctly).

And since the performance and storage cost is moved to GCS, it doesn't matter as much that there is more bandwidth being sent. And this is mostly here for search engines crawlers anyway, where latency is not as much of an issue.

gcp/appengine/frontend_handlers.py
```diff
@@ -115,6 +115,13 @@ def index():
       'home.html', ecosystem_counts=osv_get_ecosystem_counts_cached())
 
 
+@blueprint.route('/robots.txt')
+def robots():
+  response = make_response(f"Sitemap: {request.host_url}sitemap_index.xml\n")
+  response.mimetype = "text/plain"
```
michaelkedar, line 121: nit: single quotes for python strings.

michaelkedar (approved):

gcp/appengine/frontend_handlers.py
oliverchang: +1. please do this in a follow-up PR.

oliverchang (approved):

michaelkedar:
I've just noticed that the cron job doesn't actually upload the sitemaps to the gcp bucket :slightly_smiling_face:

PR #2280, opened 2024-06-05
Correct a bug with AddPkgInfo commit event assembly

andrewpollock:
While investigating an unexpected false positive for CVE-2019-14248 it was discovered that commit event ranges were being incorrectly assembled, and all of the events were being inserted into a single element, rather than one per element, i.e.

```
{
      "ranges": [
        {
          "type": "GIT",
          "repo": "https://github.com/netwide-assembler/nasm",
          "events": [
            {
              "introduced": "9a1216a1efa0ccb48e5df97acc763ea3de71e0ce",
              "last_affected": "74246c499ea4313fb8837977dc0c135fc50567c0"
            }
          ]
        }
      ]
}
```

instead of:

```
{
      "ranges": [
        {
          "type": "GIT",
          "repo": "https://github.com/netwide-assembler/nasm",
          "events": [
            {
              "introduced": "9a1216a1efa0ccb48e5df97acc763ea3de71e0ce",
            },
            {
              "last_affected": "74246c499ea4313fb8837977dc0c135fc50567c0"
            }
          ]
        }
      ]
}
```

which produced an invalid OSV record.

PR #2298, opened 2024-06-07
Prevent invalid empty Events from being added by AddPkgInfo()

andrewpollock:
While investigating some schema violations, I noticed that #2280 was adding empty Event structs when there were multiple non-`introduced` events for a repo.

Clean up some dead code that the static checker was complaining about.

### DIFF
--- a/vulnfeeds/vulns/vulns.go
+++ b/vulnfeeds/vulns/vulns.go
@@ -223,6 +223,7 @@ func (v *Vulnerability) AddPkgInfo(pkgInfo PackageInfo) {
 		}
 	}
 
+	// Aggregate commits by their repo, and synthesize a zero introduced commit if necessary.
 	if len(pkgInfo.VersionInfo.AffectedCommits) > 0 {
 		gitCommitRangesByRepo := map[string]AffectedRange{}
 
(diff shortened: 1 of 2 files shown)


vulnfeeds/vulns/vulns.go
```diff
@@ -224,41 +223,43 @@ func (v *Vulnerability) AddPkgInfo(pkgInfo PackageInfo) {
...
+						Event{
+							Introduced: "0",
+						},
+					)
+					hasAddedZeroIntroduced[ac.Repo] = true
+				}
 			}
 
-			if pkgInfo.VersionInfo.HasIntroducedCommits(ac.Repo) {
+			if ac.Introduced != "" && pkgInfo.VersionInfo.HasIntroducedCommits(ac.Repo) {
```
oliverchang, line 253: Are these `pkgInfo.VersionInfo.HasIntroducedCommits` (and similar) checks redundant?  

It seems to me that they will always return `true` if the first part of the condition (`ac.Introduced != ""`) is true?

oliverchang (approved):
LGTM with a question.

vulnfeeds/vulns/vulns.go
andrewpollock: Yes, you're right. Removing them in 642e6153 did not change the test results.

PR #2309, opened 2024-06-13
Actually upload the generated sitemaps (and perform logging)

another-rex:
Turns out, the original code wasn't doing any uploading 🤦

### DIFF
--- a/docker/cron/generate_sitemap/generate_sitemap.py
+++ b/docker/cron/generate_sitemap/generate_sitemap.py
@@ -14,6 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """Generate sitemap."""
+import logging
 import sys
 import os
 import osv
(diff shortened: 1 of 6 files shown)


docker/cron/generate_sitemap/generate_sitemap.py
```diff
@@ -79,6 +81,7 @@ def generate_sitemap_for_ecosystem(ecosystem: str, base_url: str) -> None:
 
 def generate_sitemap_index(ecosystems: set[str], base_url: str) -> None:
   """Generate a sitemap index."""
+  logging.info("Generating sitemap index.")
```
michaelkedar, line 84: nit: single quotes for python strings

michaelkedar (approved):

docker/cron/generate_sitemap/generate_sitemap.py
another-rex: Finally fixed haha!

PR #2310, opened 2024-06-13
Fix sitemap generate and upload script

another-rex:
The CWD is not in the same directory as the script. Use the script from the global PATH.

### DIFF
--- a/docker/cron/generate_sitemap/generate_and_upload.sh
+++ b/docker/cron/generate_sitemap/generate_and_upload.sh
@@ -8,7 +8,7 @@ BASE_URL_PATH="${BASE_URL:=https://test.osv.dev}"
 
 echo "Begin sitemap generation for $BASE_URL_PATH"
 
-./generate_sitemap.py --base_url $BASE_URL_PATH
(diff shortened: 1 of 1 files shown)


docker/cron/generate_sitemap/generate_and_upload.sh
```diff
@@ -8,7 +8,7 @@ BASE_URL_PATH="${BASE_URL:=https://test.osv.dev}"
 
 echo "Begin sitemap generation for $BASE_URL_PATH"
 
-./generate_sitemap.py --base_url $BASE_URL_PATH
+generate_sitemap.py --base_url $BASE_URL_PATH
```
michaelkedar, line 11: Doesn't this get copied to `/usr/local/bin/generate_sitemap/generate_sitemap.py`? I don't think that's in the PATH

another-rex: Huh, good point, I'm not sure why I was so sure it would be on the $PATH. I'll just put the full path there.

michaelkedar (approved):

PR #2311, opened 2024-06-13
Sort zip in exporter

another-rex:
Might fix https://github.com/google/osv-scanner/pull/1007#issuecomment-2164464087

Sort the entries before adding to the zip archive.


hogo6002 (approved):
LGTM, thanks Rex!!

andrewpollock:
This is a good find and a good call to address and may improve unnecessary churn and maybe eventual cache-ability.

Issue #2268, opened 2024-06-04
Update material web components to 1.0


another-rex:
Material web components on our frontend is currently on beta versions, but 1.0 has already been released: https://github.com/material-components/material-web/discussions/5004

Let's update them to the 1.0 release.

PR #2314, opened 2024-06-14
Update the project to use the new Material components

zahraaalizadeh:
Prior to this change, the deprecated `@material/mwc-*` was used in this project.

This change replaces it with `@material/web` components, and updates all relevant imports and usage to align with the new material design specifications, and adjusts styles to integrate with the new material components.

resolves https://github.com/google/osv.dev/issues/2268


gcp/appengine/frontend3/src/templates/list.html
```diff
@@ -14,9 +14,11 @@
...
       {{ column_name }}
     </div>
-    {% if is_sortable %}
-    <mwc-icon-button class="mdc-data-table__sort-icon-button" icon="arrow_upward" disabled
-      aria-label="Sort by {{ column_name }}" aria-describedby="{{ column_id }}-status-label"></mwc-icon-button>
+    {% if is_sorted %}
+    <md-icon-button class="mdc-data-table__sort-icon-button" disabled
+      aria-label="Sorted by {{ column_name }}" aria-describedby="{{ column_id }}-status-label">
+      <md-icon aria-hidden="false">arrow_upward</md-icon>
+    </md-icon-button>
```
zahraaalizadeh, line 21: As far as there is not sortable column, the if statement should check the `is_sorted` rather that `is_sortable`. 
This change also fixed the color of the arrow to be grey instead of black to give more visibility.
Before:
Screenshot 2024-06-12 at 3 30 44 PM


After:
Screenshot 2024-06-14 at 12 51 27 PM

gcp/appengine/frontend3/src/styles.scss
```diff
@@ -404,8 +403,21 @@ pre {
...
+  --md-filled-text-field-input-text-color: $osv-text-color;
+  --md-filled-text-field-leading-icon-color: $osv-text-color;
+  --md-filled-text-field-focus-label-text-color: $osv-text-color;
+  --md-filled-text-field-focus-input-text-color: $osv-text-color;
+  --md-filled-text-field-focus-leading-icon-color: $osv-text-color;
+  --md-filled-text-field-hover-label-text-color: $osv-text-color;
+  --md-filled-text-field-hover-input-text-color: $osv-text-color;
+  --md-filled-text-field-hover-leading-icon-color: $osv-text-color;
+  --md-filled-text-field-caret-color: $osv-text-color;
+  --md-filled-text-field-focus-caret-color: $osv-text-color;
```
zahraaalizadeh, line 420: Yep, There are many settings here! Another approach is to pass system colours, e.g. `--md-sys-color-on-surface`. I believe setting up this configuration makes it clear what the colour settings are, instead of relying on a few system colours, which will be used as fallbacks if the final colours are not specified. 
However, I'm open to your suggestions.

gcp/appengine/frontend3/src/styles.scss
```diff
@@ -416,12 +428,6 @@ pre {
   // https://github.com/material-components/material-web/commit/0c200ecf04a533b1dc56932c3979e3e5d8fb7d73
   position: relative;
 
-  .search-icon {
-    position: absolute;
-    top: 50%;
-    transform: translate(50%, -50%);
-  }
-
```
zahraaalizadeh, line 424: As new material supports leading slot, we don't need this class any more.

gcp/appengine/frontend3/src/styles.scss
```diff
@@ -1252,7 +1262,10 @@ dl.vulnerability-details,
       .icon {
         background: #fff;
         border-radius: 999px;
-        color: #d46d59;
+        --md-icon-button-icon-color: #d46d59;
+        --md-icon-button-hover-icon-color: #d46d59;
+        --md-icon-button-focus-icon-color: #d46d59;
+        --md-icon-button-pressed-icon-color: #d46d59;
```
zahraaalizadeh, line 1268: ♻️ The same as above comment on system colours.

gcp/appengine/frontend3/src/templates/list.html
```diff
@@ -149,7 +152,7 @@ <h1 class="title">Vulnerability Library</h1>
                   total_pages - page > 1 %}s{% endif %} left)</span>
                   {%- endif -%}
               </a>
-              <mwc-circular-progress class="next-page-indicator" indeterminate density="-4"></mwc-circular-progress>
+              <md-circular-progress class="next-page-indicator" indeterminate density="-4" aria-label="Page loading progress"></md-circular-progress>
```
zahraaalizadeh, line 155: Added `aria-label` to improve accessibility.

gcp/appengine/frontend3/src/styles.scss
another-rex: I think I prefer we set the system colours, and add the overrides as needed.

Main reasoning for this is while having a list of all the text colour variables is useful, there's no way to keep it up to date when they add new variables (which I believe they can do without changing the major version), so the list could be misleading, and you would have to look up the variables anyway.

zahraaalizadeh: Updated via [c7a9bea](https://github.com/google/osv.dev/pull/2314/commits/c7a9bea483be59f447aaaf8fe27bf0051712f093)

gcp/appengine/frontend3/src/styles.scss
zahraaalizadeh: updated [c7a9bea](https://github.com/google/osv.dev/pull/2314/commits/c7a9bea483be59f447aaaf8fe27bf0051712f093)

another-rex (approved):
LGTM, thanks!

PR #2322, opened 2024-06-18
Make last modified date export with the UTC offset

another-rex:
- Make the lastmod date export with the UTC offset (+00:00) (otherwise Google Search Console complains the date format is incorrect)
- Set the lastmod date to the modified date of each entry, so crawlers don't have to revisit every page every day.
- Add projection to avoid returning the entire bug.


gcp/appengine/index.yaml
```diff
@@ -80,6 +80,15 @@ indexes:
       direction: desc
 
   # ecosystem + search_indices queries is served by merging the two indexes above
+  
+  # To allow sitemap generation to project and sort on last_modified
+  - kind: Bug
```
another-rex, line 85: This index already exists on prod, but was not recorded here (and doesn't exist on test). I think I removed it when I did some cleanup and when we switched the frontend from sorting by last modified to published. 

I have now added this on test as well.

docker/cron/generate_sitemap/generate_sitemap.py
```diff
@@ -90,7 +110,8 @@ def generate_sitemap_index(ecosystems: set[str], base_url: str) -> None:
     loc = SubElement(sitemap, 'loc')
     loc.text = get_sitemap_url_for_ecosystem(ecosystem, base_url)
     lastmod = SubElement(sitemap, 'lastmod')
-    lastmod.text = datetime.datetime.now().isoformat()
+    # Make sure to set the timezone to UTC to add +00:00 when outputting iso
+    lastmod.text = last_mod_dict[ecosystem].astimezone(datetime.UTC).isoformat()
```
michaelkedar, line 114: same timezone conversion issue here

docker/cron/generate_sitemap/generate_sitemap.py
```diff
@@ -58,28 +63,43 @@ def get_sitemap_url_for_ecosystem(ecosystem: str, base_url: str) -> str:
...
+    # Make sure to set the timezone to UTC to add +00:00 when outputting iso
+    lastmod.text = last_modified.astimezone(datetime.UTC).isoformat()
 
   tree = ElementTree(urlset)
   tree.write(filename, encoding='utf-8', xml_declaration=True)
 
+  # Addition of year 2000 for edge cases where vulnerability is empty
+  return max([
+      last_mod for _, last_mod in vulnerability_and_dates[:_SITEMAP_URL_LIMIT]
+  ] + [datetime.datetime.fromisocalendar(2000, 1, 1)])
```
michaelkedar, line 98: nit: maybe 1970/1/1 (`datetime.datetime.fromtimestamp(0).astimezone(datetime.UTC)`) indicates emptiness better.

docker/cron/generate_sitemap/generate_sitemap.py
```diff
@@ -58,28 +63,43 @@ def get_sitemap_url_for_ecosystem(ecosystem: str, base_url: str) -> str:
...
-  for vuln in vulnerability_ids[:_SITEMAP_URL_LIMIT]:
+  for vuln_id, last_modified in vulnerability_and_dates[:_SITEMAP_URL_LIMIT]:
     url = SubElement(urlset, 'url')
     loc = SubElement(url, 'loc')
-    loc.text = f'{base_url}/vulnerability/{vuln}'
+    loc.text = f'{base_url}/vulnerability/{vuln_id}'
     lastmod = SubElement(url, 'lastmod')
-    lastmod.text = datetime.datetime.now().isoformat()
+    # Make sure to set the timezone to UTC to add +00:00 when outputting iso
+    lastmod.text = last_modified.astimezone(datetime.UTC).isoformat()
```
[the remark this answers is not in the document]
another-rex: datetimes from the datastore comes with the time zone information attached, so astimezone will convert it from that to UTC with the correct offset. (Though should we even be converting them?)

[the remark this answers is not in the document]
another-rex: Huh, looking at this more, I don't think I'm actually correct in the datetimes coming with time zone information.

michaelkedar (approved):

Issue #2336, opened 2024-06-24
Incorrect tooltip text on osv.dev when starting affected version is 0


another-rex:
**Describe the bug**
The tooltip text says the exact initial commit is unknown, even on entries without any commit information.

(E.g. https://osv.dev/vulnerability/GHSA-25hc-qcg6-38wj) 

**Expected behaviour**
It should say: "Unknown introduced version / All previous versions are affected". Perhaps change the word "version" to "commit" on affected ranges with commits instead of versions.

**Screenshots**
![image](https://github.com/google/osv.dev/assets/106129829/8c88f41c-31dc-44a1-89c7-0d553c770996)

PR #2341, opened 2024-06-25
Display the correct introduced event tooltip

zahraaalizadeh:
Prior to this change, when the introduced version in the affected range was 0, the tooltip displayed: "The exact introduced commit is unknown".

This change, updates the html block to be more readable, and displays a more descriptive tooltip based on the type of the range, indicating if it has versions or commits.

resolves: https://github.com/google/osv.dev/issues/2336

Sample output for GIT range:
Screenshot 2024-06-25 at 12 31 10 PM

Sample output for other type of ranges:
Screenshot 2024-06-25 at 12 30 53 PM

### DIFF
--- a/gcp/appengine/frontend3/src/templates/vulnerability.html
+++ b/gcp/appengine/frontend3/src/templates/vulnerability.html
@@ -226,17 +226,19 @@
                       {% set link = event | event_link -%}
                       {% if link -%}
                         <a href="{{ link }}">
+                          {{ event | event_value -}}
+                        </a>
                       {% elif event | event_type == 'Introduced' and event | event_value == '0' -%}
                         <div class="tooltip">
-                      {% endif -%}
-
-                      {{ event | event_value -}}
-
-                      {% if link -%}
-                        </a>
-                      {% elif event.get('introduced') == '0' -%}
-                        <span class="tooltiptext">The exact introduced commit is unknown</span>
+                          {{ event | event_value -}}
+                          {% if range.type == 'GIT' %}
+                            <span class="tooltiptext">Unknown commit / All previous commits are affected</span>
+                          {% else -%}
+                            <span class="tooltiptext">Unknown version / All previous versions are affected</span>
+                          {% endif -%}
                         </div>
+                      {% else -%}
+                        {{ event | event_value -}}
                       {% endif -%}
                     </div>
                   {% endfor -%}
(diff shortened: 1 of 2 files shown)


gcp/appengine/frontend3/src/templates/vulnerability.html
```diff
@@ -226,17 +226,19 @@ <h3 class="mdc-layout-grid__cell--span-3">
...
-
-                      {% if link -%}
-                        </a>
-                      {% elif event.get('introduced') == '0' -%}
-                        <span class="tooltiptext">The exact introduced commit is unknown</span>
+                          {{ event | event_value -}}
+                          {% if range.type == 'GIT' %}
+                            <span class="tooltiptext">Unknown commit / All previous commits are affected</span>
+                          {% else -%}
+                            <span class="tooltiptext">Unknown version / All previous versions are affected</span>
```
zahraaalizadeh, line 237: I slightly updated the suggested text for the tooltip to get it fit a single line. With long messages the introduced and fixed rows will be one line apart:
Screenshot 2024-06-25 at 12 17 20 PM

zahraaalizadeh (commented):
I

another-rex:
I think it's important to have the "introduced" in there, can we change the max-width in the css to make it fit on one line?

zahraaalizadeh (commented):
Desktop:
Screenshot 2024-06-26 at 7 19 53 PM

Mobile:
Screenshot 2024-06-26 at 7 19 40 PM

gcp/appengine/frontend3/src/templates/vulnerability.html
```diff
@@ -219,24 +219,26 @@ <h3 class="mdc-layout-grid__cell--span-3">
               <dd>
                 <div class="mdc-layout-grid__inner events">
                   {% for event in range.events -%}
-                    <div class="mdc-layout-grid__cell--span-3">
+                    <div class="mdc-layout-grid__cell--span-2">
```
another-rex, line 222: Are the layout width changes intentional? I do prefer to keep the width the same here if possible.

zahraaalizadeh: @another-rex Ah, yesterday, I had a afternoon brain and tried a few options, but none worked. Just re-tried and now it's been fixed.

another-rex (approved):

another-rex:
Thanks! I think just merging in the main branch should fix the PR check.

PR #2350, opened 2024-06-27
Remove relative-time lib to improve CLS

zahraaalizadeh:
issue: https://github.com/google/osv.dev/issues/1013

Currently, we use relative-time package to convert timestamp to a human readable relative time, when
displaying published time of the bug in the list page.

Using this approach is not optimised as it changes the texts in the loading time and leads to CLS issues.

This change replaces relative-time lib in favour of a new template tag which converts the published to a human readable relative time using a utility function.

Before:
Screenshot 2024-06-26 at 5 33 14 PM

After:
Screenshot 2024-06-26 at 6 29 22 PM

### DIFF
--- a/gcp/appengine/utils.py
+++ b/gcp/appengine/utils.py
@@ -14,10 +14,7 @@
 """Utils."""
 
 import os
-from datetime import datetime
-from dateutil.relativedelta import relativedelta
-from dateutil import tz
-from dateutil.parser import isoparse
+from datetime import datetime, UTC
 
 
 def is_prod():
@@ -48,22 +45,26 @@ def relative_time(value: datetime | str) -> str:
     """
   # Parse the ISO 8601 string to a datetime object
   if isinstance(value, str):
-    value = isoparse(value)
+    value = datetime.fromisoformat(value)
 
-  now = datetime.now(tz=tz.tzutc())
-  diff = relativedelta(now, value)
-  diff_seconds = (now - value).total_seconds()
+  now = datetime.now(tz=UTC)
+  diff = now - value
+  diff_seconds = diff.total_seconds()
+  diff_minutes = diff_seconds // 60
+  diff_hours = diff_seconds // (60 * 60)
+  diff_days = diff_seconds // ((60 * 60) * 24)
+  diff_weeks = diff_seconds // (((60 * 60) * 24) * 7)
 
-  if diff_seconds < 60:
+  if diff_minutes == 0:
     return "just now"
-  if diff_seconds < 3600:
-    return f"{diff.minutes} minute{'s' if diff.minutes > 1 else ''} ago"
-  if diff_seconds < 86400:
-    return f"{diff.hours} hour{'s' if diff.hours > 1 else ''} ago"
-  if diff_seconds < 172800:
+  if diff_hours == 0:
+    return f"{int(diff_minutes)} minute{'s' if diff_minutes > 1 else ''} ago"
+  if diff_days == 0:
+    return f"{int(diff_hours)} hour{'s' if diff_hours > 1 else ''} ago"
+  if diff_days == 1:
     return "yesterday"
-  if diff_seconds < 604800:
-    return f"{diff.days} day{'s' if diff.days > 1 else ''} ago"
+  if diff_weeks == 0:
+    return f"{int(diff_days)} day{'s' if diff_days > 1 else ''} ago"
   if value.year == now.year:
     return value.strftime("%d %b")
   return value.strftime("%d %b %Y")
--- a/gcp/appengine/utils.py
+++ b/gcp/appengine/utils.py
@@ -37,8 +37,14 @@ def relative_time(value: datetime | str) -> str:
         value: The input datetime as an ISO 8601 string or a datetime object.
 
     Returns:
-        str: A human-readable string representing the relative time 
-             (e.g., '2 days ago').
+        str: A human-readable string representing the relative time as:
+             - < 1 minute: Just now
+             - < 1 hour: x minutes ago
+             - < 1 day: x days ago
+             - < 2 day: yesterday
+             - < 7 days: x days ago
+             - >= 7 days: DD MMM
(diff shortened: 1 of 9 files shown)


another-rex (commented):
Good idea! 

I think for the relative date, anything longer than 7 days should just display the date that it was published (just month and day). (And anything not in the current year should also display the year as well).

So something like:
```
< 1 minute: Just now
< 1 hour: x minutes ago
< 1 day: x days ago
< 2 day: yesterday
< 7 days: x days ago
>= 7 days: DD MMM
Year is not current year: DD MMM YYYY
```

This is a bit different than the current behaviour in prod, which goes back to last months before starting to display the date, but I think last months is too inaccurate already, and 7 days is a good limit.

gcp/appengine/utils_test.py
```diff
@@ -0,0 +1,80 @@
...
+    self.assertEqual(
+        utils.relative_time(last_week), last_week.strftime("%d %b"))
+
+  @mock.patch("utils.datetime")
+  def test_date_without_year(self, mock_datetime):
+    """Check when the value is longer than 7 days but in the same year"""
+    now = datetime(2024, 6, 27, tzinfo=tzutc())
+    mock_datetime.now.return_value = now
+    date_within_year = now - timedelta(days=10)
+    self.assertEqual(
```
zahraaalizadeh, line 64: To avoid flaky tests on 7-10 Jan mocked this test.

gcp/appengine/utils.py
```diff
@@ -23,3 +27,43 @@ def is_prod():
...
+             - < 1 hour: x minutes ago
+             - < 1 day: x days ago
+             - < 2 day: yesterday
+             - < 7 days: x days ago
+             - >= 7 days: DD MMM
+             - Year is not current year: DD MMM YYYY
+    """
+  # Parse the ISO 8601 string to a datetime object
+  if isinstance(value, str):
+    value = isoparse(value)
```
another-rex, line 51: This can be done with `datetime.fromisoformat()` directly I believe?

gcp/appengine/utils.py
```diff
@@ -23,3 +27,43 @@ def is_prod():
...
+
+  now = datetime.now(tz=tz.tzutc())
```
another-rex, line 53: rather than using tz.tzutc(), we can just use the standard library datetime.UTC

gcp/appengine/utils_test.py
```diff
@@ -0,0 +1,80 @@
...
+        utils.relative_time(date_within_year),
+        date_within_year.strftime("%d %b"))
```
another-rex, line 66: Since this is already using a mocked date, can you write out the date as a string here? This makes it easy to see if it's in the right format or not.

gcp/appengine/utils_test.py
```diff
@@ -0,0 +1,80 @@
...
+
+  def test_date_with_year(self):
+    date_outside_year = self.now - timedelta(days=400)
+    self.assertEqual(
+        utils.relative_time(date_outside_year),
+        date_outside_year.strftime("%d %b %Y"))
```
another-rex, line 72: Same here.

gcp/appengine/frontend3/img/.DS_Store
another-rex, line 1: I think this file is unintentionally checked in?

gcp/appengine/utils.py
```diff
@@ -23,3 +27,43 @@ def is_prod():
...
+  diff = relativedelta(now, value)
+  diff_seconds = (now - value).total_seconds()
```
another-rex, line 55:
```suggestion
  diff = now - value
  diff_seconds = diff.total_seconds()
  diff_minutes = diff_seconds // 60
  diff_hours = diff_seconds // (60 * 60)
```

I think we can change this part to avoid needing to import a new library.

gcp/appengine/utils.py
```diff
@@ -23,3 +27,43 @@ def is_prod():
...
+  if diff_seconds < 60:
+    return "just now"
+  if diff_seconds < 3600:
+    return f"{diff.minutes} minute{'s' if diff.minutes > 1 else ''} ago"
+  if diff_seconds < 86400:
+    return f"{diff.hours} hour{'s' if diff.hours > 1 else ''} ago"
+  if diff_seconds < 172800:
+    return "yesterday"
+  if diff_seconds < 604800:
+    return f"{diff.days} day{'s' if diff.days > 1 else ''} ago"
```
another-rex, line 66: An alternative to having these precalculated numbers here is to just match by whether the next unit is zero: i.e.
```suggestion
  if diff_minutes == 0:
    return "just now"
  if diff_hours == 0:
    return f"{diff.minutes} minute{'s' if diff.minutes > 1 else ''} ago"
  if diff.days == 0:
    return f"{diff.hours} hour{'s' if diff.hours > 1 else ''} ago"
  if diff.days == 1:
    return "yesterday"
  if diff.days < 7:
    return f"{diff.days} day{'s' if diff.days > 1 else ''} ago"
```

(Though you might want to double check this, I'm not sure if this will work)

another-rex (commented):
Thanks! Gave some suggestions for how to avoid needing the extra library.

gcp/appengine/utils.py
zahraaalizadeh: Thanks! That's a great suggestion! I refactored it.

gcp/appengine/frontend3/img/.DS_Store
zahraaalizadeh: Oops. Maybe it'd be worth it to gitignore `.DS_Store` files?

gcp/appengine/utils.py
zahraaalizadeh: Nice! Yes it works!

gcp/appengine/frontend3/img/.DS_Store
another-rex: Yeah, we should add .DS_Store into the gitignore

another-rex (approved):

PR #2278, opened 2024-06-05
Fix integration tests

andrewpollock:
- the updated Debian conversion stuff hitting production caused more CVEs to match for existing queries

PR #2356, opened 2024-07-02
Fix integration tests

cuixq:
Update the expected responses in API integration tests:
 - add a new Go vulnerability to test response
 - versions that we test against are no longer affected

### DIFF
--- a/gcp/api/integration_tests.py
+++ b/gcp/api/integration_tests.py
@@ -294,11 +294,14 @@ class IntegrationTests(unittest.TestCase,
     go_2020_0001 = self._get('GO-2020-0001')
     go_2021_0052 = self._get('GO-2021-0052')
     ghsa_6vm3_jj99_7229 = self._get('GHSA-6vm3-jj99-7229')
+    ghsa_869c_j7wc_8jqv = self._get('GHSA-869c-j7wc-8jqv')
     ghsa_h395_qcrw_5vmq = self._get('GHSA-h395-qcrw-5vmq')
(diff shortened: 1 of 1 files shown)


gcp/api/integration_tests.py
```diff
@@ -485,12 +487,7 @@ def test_query_purl(self):
 
     self.assert_results_equal({'vulns': another_expected}, response.json())
 
-    expected_deb = [
-        self._get('CVE-2009-4487'),
-        self._get('CVE-2013-0337'),
-        self._get('DLA-3203-1'),
-        self._get('DSA-4921-1')
-    ]
+    expected_deb = [self._get('DLA-3203-1'), self._get('DSA-4921-1')]
```
cuixq, line 490: @andrewpollock I noticed you had a PR https://github.com/google/osv.dev/pull/2278 that added the two vulns that are going to be removed in this PR - do you know if this is expected?

andrewpollock: These CVEs started being introduced from the additional Debian conversion by @hogo6002 
If they're not converting any more we should figure out why that's the case

hogo6002: The CVE (https://osv.dev/vulnerability/CVE-2009-4487) entries are still on OSV.dev, but they no longer list this queried version(`1.14.2-2+deb10u3`) as an affected version.

cuixq: so is this something we need to worry about? or this is intended?

PR #2357, opened 2024-07-02
Poetry initial migration 

another-rex:
First step in migrating from `pipenv` to `poetry`.

This PR creates all the pyproject.toml and Poetry.lock files needed, however leaves almost all instances of pipenv calls intact. 

To try running with the poetry environment, use the USE_POETRY=true env flag, e.g. 
```
USE_POETRY=true make run-appengine
USE_POETRY=true make all-tests
```

Also in this PR:
- Fix Pipenv for datafix to included a required package
- Standardise how we are specifying required package versions. (Generally all upgrades are allowed till the next major version)
- Fix our run_test.sh scripts to all use bash, with -e and -x flags. Currently some of our run_test.sh tests will silently fail because of the lack of -e flag. (i.e., when there are multiple tests, we need to use the -e flag otherwise as long as the final test is ok, the script will return 0, even if other tests failed. (I think?)).


cuixq (approved):

michaelkedar (commented):
Can you change the Pipfiles' requirement version specs/pinning to match the changes you've done in the pyproject.tomls?

another-rex:
Updated the lockfiles, surprisingly none of the lockfiles got updated.

PR #2359, opened 2024-07-02
Add API support for non-enumerated ecosystems

hogo6002:
The current API queries for ecosystem-specific versions require pre-enumeration of all affected versions per package. This PR will allow API queries for all non-enumerated ecosystems by comparing the queried version against the affected range.

Changes only affects the **test instance**:
- Supports API queries for Ubuntu, Rocky Linux and AlmaLinux
- Non-semver ecosystems (e.g., Debian, Alpine) now switch to affected range checks instead of enumeration matching.
- Debian queries without specifying a release number may now return more vulnerabilities. This change is due to two reasons:
  - Pre-enumeration limitations: The previous method, which relied on pre-enumerating package versions, could sometimes fail to calculate affected versions (e.g., https://osv.dev/vulnerability/CVE-2016-3137).
  - New version range comparison: The updated logic compares across different Debian releases when a specific release number is not provided. For example, a query for version `1.14.2-2+deb10u3` will now also return vulnerability [DSA-5281-1 ](https://osv.dev/vulnerability/DSA-5281-1)(affected range: `0` to `1.18.0-6.1+deb11u3`), which was not included in previous results.

Follow-up PRs:
- Add support for Chainguard and Wolfi
- Require release numbers for Alpine queries (error message)
- Switch integration tests to test instance (address missing records, e.g., https://test.osv.dev/vulnerability/OSV-2023-890)
- Improve handling for Linux package queries without release numbers
- Handle Debian end-of-life releases


gcp/api/server.py
```diff
@@ -1064,7 +1066,10 @@ def query_by_version(context: QueryContext,
       for bug in new_bugs:
         if bug not in bugs:
           bugs.append(bug)
-
+    elif ecosystem_info.supports_comparing:
```
another-rex, line 1069: This needs to be a variable like is_semver above where we check if ecosystem_info is None or not.

osv/ecosystems/helper_base.py
```diff
@@ -118,6 +118,10 @@ def is_semver(self):
   def supports_ordering(self):
     return True
 
+  @property
+  def supports_comparing(self):
```
another-rex, line 122: What's the difference between this and supports_ordering right above this line?

another-rex (commented):
Thanks! General structure looks good, added some comments

osv/ecosystems/helper_base.py
hogo6002: I was thinking to support 4 Linux ecosystems first to check performance. And `supports_ordering` includes additional ecosystems like `Alpine`, so I created a new property here. But yeah, eventually we do want to apply this new approach to all non-semver ecosystems. I will update code to reuse `supports_ordering` and implement a configuration option to enable this new approach on the test instance only.

gcp/api/server.py
```diff
@@ -1049,9 +1051,11 @@ def query_by_version(context: QueryContext,
 
   ecosystem_info = ecosystems.get(ecosystem)
   is_semver = ecosystem_info and ecosystem_info.is_semver
+  supports_ordering = ecosystem_info and ecosystem_info.supports_ordering
 
   bugs = []
   next_page_token = None
+  project = getattr(_ndb_client, 'project', 'oss-vdb')
```
michaelkedar, line 1058: nit: Can you make a `get_gcp_project()` function for this line, with the comment from the `LogTraceFilter` that explains this.

gcp/api/server.py
```diff
@@ -1179,6 +1221,26 @@ def is_cloud_run() -> bool:
...
+    introduced_version = '0'
+    fixed_version = '99999'
+    for event in range.events:
+      event: osv.AffectedEvent
+      if event.type == 'introduced':
+        introduced_version = event.value
+      elif event.type == 'fixed':
+        fixed_version = event.value
+
+    if ecosystem.sort_key(version) >= ecosystem.sort_key(introduced_version) and ecosystem.sort_key(version) < ecosystem.sort_key(fixed_version):
```
michaelkedar, line 1238: A few things:
- You should use `None` (or similar) to represent a missing fixed version. `99999` might not parse in some ecosystems, or it could be smaller than a legitimate version e.g. CalVer like `20240704`
- You should also be checking for `last_affected` events.
- There can be multiple `introduced`/`fixed`/`last_affected` events in an affected range (and they're not guaranteed to be sorted)

You could probably just reuse the logic in `_is_semver_affected()`:
https://github.com/google/osv.dev/blob/6084fc7849a1e32dae4a9a09fd56e4b4417377d8/gcp/api/server.py#L862-L876

gcp/api/server.py
```diff
@@ -626,6 +621,10 @@ def do_query(query, context: QueryContext, include_details=True):
     purl_str = ''
 
   if ecosystem and not ecosystems.get(ecosystem):
+    if ecosystem.startswith('Alpine') and not ecosystem.startswith('Alpine:v'):
```
hogo6002, line 624: I'll move the Alpine release number check to a followup PR since it will affect both test and prod instances.

gcp/api/server.py
```diff
@@ -1094,6 +1102,44 @@ def query_by_version(context: QueryContext,
...
+      affected_package: osv.AffectedPackage
+      package = affected_package.package
+      package: osv.Package
+
+      package_ecosystem = package.ecosystem
+      if not has_release:
+        # If no release is specified,
+        # extracts only the ecosystem name (e.g., "Debian")
+        package_ecosystem = package_ecosystem.split(':')[0]
+      if ecosystem != package_ecosystem:
```
another-rex, line 1132:
```suggestion

      if ecosystem != package_ecosystem:
```

gcp/api/server.py
```diff
@@ -1179,6 +1225,47 @@ def is_cloud_run() -> bool:
...
+                 affected_package: osv.AffectedPackage) -> bool:
+  """Checks if a version is affected."""
+  affected = False
+  ecosystem_info = ecosystems.get(ecosystem)
+  queried_version = ecosystem_info.sort_key(version)
+
+  for r in affected_package.ranges:
+    r: osv.AffectedRange2
+
+    for event in osv.sorted_events(ecosystem, r.type, r.events):
```
another-rex, line 1248: nit, all  3 of these branches seem mutually exclusive, can we just add continue after each one? (or use elseif)

osv/ecosystems/_ecosystems.py
```diff
@@ -105,27 +108,24 @@
 def get(name: str) -> Ecosystem:
   """Get ecosystem helpers for a given ecosystem."""
 
-  if name.startswith('Debian:'):
-    return Debian(name.split(':')[1])
+  if name.startswith('Debian'):
+    return Debian(name.partition(':')[2])
 
   if name.startswith('Alpine:'):
     return Alpine(name.split(':')[1])
```
another-rex, line 115: Use partition here as well to keep things consistent

osv/ecosystems/debian_test.py
```diff
@@ -99,3 +100,11 @@ def test_cache(self, cache_mock: mock.MagicMock):
     debian.next_version('nginx', '1.13.5-1')
     cache_mock.get.assert_called_once()
     cache_mock.set.assert_called_once()
+
+  # def test_debian_version(self):
```
another-rex, line 104: Should these tests be uncommented?

another-rex (commented):
LGTM overall, just a couple of nits

andrewpollock:
> * Debian queries without release numbers may return more vulnerabilities than before due to the match method change.

Could you update the description to explicitly state that this is considered correct behaviour (if that's the case)?

osv/ecosystems/debian_test.py
```diff
@@ -19,6 +19,7 @@
 
 from .. import cache
 from .. import ecosystems
+from osv import ecosystems
```
another-rex, line 22: Is this change needed?

osv/ecosystems/debian_test.py
another-rex, line 1: Can you also add some tests for versions with +deb after it, especially cross debian versions as well (e.g. 1.12+deb12... compared to 1.12+deb11, or cross different versions). Essentially adding the versions where 

> Debian queries without release numbers may return more vulnerabilities than before due to the match method change.

as test cases.

another-rex (approved):

gcp/api/server.py
```diff
@@ -1094,6 +1102,47 @@ def query_by_version(context: QueryContext,
...
+      cursor = it.cursor_after()
+      break
+
+    bug: osv.Bug = it.next()
+    for affected_package in bug.affected_packages:
+      affected_package: osv.AffectedPackage
+      package = affected_package.package
+      package: osv.Package
+
+      package_ecosystem = package.ecosystem
```
andrewpollock, line 1127: If you use this, it makes checking for whether or not a release was specified unnecessary.

```suggestion
      package_ecosystem = package.ecosystem.partition(':')[0]
```

gcp/api/server.py
```diff
@@ -1179,6 +1228,45 @@ def is_cloud_run() -> bool:
   return os.getenv('K_SERVICE') is not None
 
 
+def get_gcp_project():
+  """Get the gcp project name."""
```
andrewpollock, line 1232: Case nit.

```suggestion
  """Get the GCP project name."""
```

andrewpollock (commented):
Is it possible to add some integration tests that confirm this is working as intended?

osv/ecosystems/_ecosystems.py
```diff
@@ -105,27 +108,24 @@
 def get(name: str) -> Ecosystem:
   """Get ecosystem helpers for a given ecosystem."""
 
-  if name.startswith('Debian:'):
-    return Debian(name.split(':')[1])
+  if name.startswith('Debian'):
+    return Debian(name.partition(':')[2])
```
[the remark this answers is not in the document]
michaelkedar: In  this case, I'm pretty sure [`str.partition`](https://docs.python.org/3/library/stdtypes.html#str.partition) always returns a tuple with length 3, so an `IndexError` should not be possible.

osv/ecosystems/_ecosystems.py
```diff
@@ -105,27 +108,24 @@
...
+    return RockyLinux()
 
   if name.startswith('Photon OS:'):
     # TODO(unassigned)
     return OrderingUnsupportedEcosystem()
 
-  if name.startswith('Ubuntu:'):
-    # TODO(unassigned)
-    return OrderingUnsupportedEcosystem()
+  if name.startswith('Ubuntu'):
```
michaelkedar, line 127: nit (and not really important): removing the `:` from all these `startswith` checks makes things like `Ubuntuabcd` `Debian!!!`, `Rocky Linux is an ecosystem`, etc. match these checks.

Maybe you could do something like `base_name, sep, version = name.partition(':')` first, and then check if `base_name` equals the ecosystem names

michaelkedar: And for a future refactor, all `Ecosystem` helpers can take a `version` in their constructors, and this whole function can just become something like:
```python
def get(name: str) -> Ecosystem:
  base_name, _, version = name.partition(':')
  return _ecosystems.get(base_name)(version)
```

osv/ecosystems/rocky_linux.py
```diff
@@ -0,0 +1,36 @@
...
+from ..third_party.univers.rpm import RpmVersion
+from .helper_base import Ecosystem
+
+
+class RockyLinux(Ecosystem):
+  """"Rocky Linux ecosystem"""
+
+  @property
+  def name(self):
+    return "Rocky Linux"
```
michaelkedar, line 25: Is this `name` still necessary?

michaelkedar (approved):
LGTM

osv/ecosystems/_ecosystems.py
hogo6002: I'll do this in a follow up PR where I will add some checks for invalid ecosystem such as query for Alpine without release number.

osv/ecosystems/rocky_linux.py
hogo6002: Not for this PR, but I think it would be better to keep it as the ecosystem name does not exactly match the class name.

gcp/api/server.py
hogo6002: I added some comments to explain the code logic here. 

If the user requires an ecosystem like `Debian:11`, we only want to return results with the same release number. So I think we shouldn't just get the base ecosystem name from the affected package, it depends on whether the queried ecosystem has a release.

hogo6002:
> Is it possible to add some integration tests that confirm this is working as intended?

It's a bit hard to add integration tests here because they currently use the production instance. I'd need to switch them to run on the test environment to execute the new code logic. However, the test instance is missing some records, which would cause the integration tests to fail.

So, I plan to switch the integration tests to use the test instance and rewrite some of the test results in a follow-up PR.

osv/ecosystems/_ecosystems.py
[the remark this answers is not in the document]
andrewpollock: Yeah good point.

PR #2360, opened 2024-07-02
Update timestamp on Chainguard blog post

andrewpollock:
Blog posts only get published once:

- a build happens
- the date of the post is in the past (at build time)

As we release a pre-existing build, having a post date too far in the future won't allow it to necessarily get released when desired.


hogo6002 (approved):

another-rex (approved):

osv/osv-schema
cuixq, line 1: I also have something similar to this in my local work repository - does anyone know what this is?

cuixq (approved):

michaelkedar (approved):

osv/osv-schema
andrewpollock: I want to say it's because of recent commits to https://github.com/ossf/osv-schema, which is submoduled into this repo. That repo doesn't see a lot of commits, so I think this just doesn't change very frequently.

PR #2366, opened 2024-07-03
Handle pagination of GHC enumeration

### LINKED ISSUE #2367: Haskell GHC ecosystem enumeration broken
**Describe the bug**
Enumeration of the Haskell GHC ecosystem is no longer working correctly

**To Reproduce**
1. `make lib-tests` @ [35f782f](https://github.com/google/osv.dev/commit/35f782ff4cc000ad4aef65add2feb672dfaa155b)
2. observe `osv.ecosystems.haskell_test.GHCEcosystemTest.test_next_version` fails

**Expected behaviour**
Test does not fail

**Screenshots**
```
$ make lib-tests 
./run_tests.sh
+ export PIPENV_IGNORE_VIRTUALENVS=1
+ PIPENV_IGNORE_VIRTUALENVS=1
+ python3 -m pipenv sync
Installing dependencies from Pipfile.lock (9c09ee)...
To activate this project's virtualenv, run pipenv shell.
Alternatively, run a command inside the virtualenv with pipenv run.
All dependencies are now up-to-date!
+ python3 -m pipenv run python -m unittest osv.bug_test
.
Ran 1 test in 0.000s

OK
+ python3 -m pipenv run python -m unittest osv.purl_helpers_test
......
Ran 6 tests in 0.000s

OK
+ python3 -m pipenv run python -m unittest osv.request_helper_test
.
Ran 1 test in 0.001s

OK
+ python3 -m pipenv run python -m unittest osv.semver_index_test
...
Ran 3 tests in 0.001s

OK
+ python3 -m pipenv run python -m unittest osv.impact_test
....
Ran 4 tests in 12.224s

OK
+ python3 -m pipenv run python -m unittest discover osv/ecosystems/ '*_test.py' .
.....F.........................
======================================================================
FAIL: test_next_version (osv.ecosystems.haskell_test.GHCEcosystemTest.test_next_version)
Test next_version.
  File "/usr/local/google/home/apollock/gosst/osv/osv.dev/osv/ecosystems/haskell_test.py", line 44, in test_next_version
    self.assertEqual('7.2.1', ecosystem.next_version('GHC', '7.0.4'))
AssertionError: '7.2.1' != '9.2.3'
- 7.2.1
+ 9.2.3


Ran 31 tests in 108.714s

FAILED (failures=1)
make: *** [Makefile:15: lib-tests] Error 1
```

**Additional context**
It would appear that the behaviour of the GitLab API has changed

https://github.com/google/osv.dev/blob/35f782ff4cc000ad4aef65add2feb672dfaa155b/osv/ecosystems/haskell.py#L71-L72

is now returning only 20 entries, whereas I _suspect_ the old behaviour of passing `?per_page=-1` was to return all of them (`?per_page=100` certainly returns 100, and 20 appears to the default page size). Implementing [pagination](https://docs.gitlab.com/ee/api/rest/#pagination) is probably the solution.

michaelkedar:
Fixes #2367 
We were using `per_page=-1` on the GitLab tag get request to get every tag for GHC enumeration. That recently stopped returning every tag.

Changed it to query for each page, per  https://docs.gitlab.com/ee/api/rest/index.html#pagination-link-header

### DIFF
--- a/osv/ecosystems/haskell.py
+++ b/osv/ecosystems/haskell.py
@@ -68,7 +68,6 @@ class Hackage(Ecosystem):
 class GHC(Ecosystem):
   """Glasgow Haskell Compiler (GHC) ecosystem."""
 
-  # FIXME: https://github.com/google/osv.dev/issues/2367
   _API_PACKAGE_URL = ('https://gitlab.haskell.org'
                       '/api/v4/projects/3561/repository/tags?per_page=100')
   """
--- a/osv/ecosystems/haskell.py
+++ b/osv/ecosystems/haskell.py
@@ -70,7 +70,7 @@ class GHC(Ecosystem):
 
   # FIXME: https://github.com/google/osv.dev/issues/2367
   _API_PACKAGE_URL = ('https://gitlab.haskell.org'
-                      '/api/v4/projects/3561/repository/tags?per_page=-1')
+                      '/api/v4/projects/3561/repository/tags?per_page=100')
   """
   Historical versions do not have tags in the Git repo, so we hardcode the
   list.  See https://github.com/google/osv.dev/pull/1463 for discussion.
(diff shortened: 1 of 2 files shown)


osv/ecosystems/haskell.py
```diff
@@ -69,7 +69,7 @@ class GHC(Ecosystem):
   """Glasgow Haskell Compiler (GHC) ecosystem."""
 
   _API_PACKAGE_URL = ('https://gitlab.haskell.org'
-                      '/api/v4/projects/3561/repository/tags?per_page=-1')
+                      '/api/v4/projects/3561/repository/tags?per_page=100')
```
cuixq, line 72: maybe add a comment saying that 100 is `max` for `per_page`?

cuixq (approved):

osv/ecosystems/haskell.py
michaelkedar: comment added

michaelkedar:
I had a brief look at the tags on the repo, and I've noticed that the latest tag is 9.6.2, whereas [haskell.org lists up to 9.6.6/9.8.2/9.10.1](https://www.haskell.org/ghc/download.html)

I wonder if we could instead parse the [names of the files here](https://downloads.haskell.org/~ghc/), but it doesn't seem consistent enough to automate :/

cuixq:
I am happy with the current implementation, the maintenance status of both websites is kind of unknown, they don't seem like a reliable source to me.

michaelkedar:
Personally, I think https://downloads.haskell.org/~ghc/ would be the most correct source of truth, since we're missing some tags as-is.

That said, we currently have zero GHC vulnerabilities in OSV. We can revisit this when it's actually needed :)

PR #2373, opened 2024-07-05
Fix tag styling on frontend

another-rex:
Because of how our font (Overpass?) looks, it causes the text to feel too high in a tag. This tries to add some manual padding to push it down.

Before:
image

After:

image


hogo6002 (approved):
Thanks!

Issue #2374, opened 2024-07-05
Alias missing for some CVEs in OSV


Moullisha:
**Describe the bug**
Some vulnerabilities in OSV do not mention alias whereas the source link has alias data.

**To Reproduce**
https://vuln.go.dev/ID/GO-2024-2947.json mentions two aliases whereas https://api.osv.dev/v1/vulns/GO-2024-2947 does not.
Similar behaviour observed in case of GO-2024-2918, GHSA-v6v8-xj6m-xwqh, GO-2024-2963

**Expected behaviour**
If alias is available in source data, same should be provided by OSV.

**Screenshots**

**Additional context**

another-rex:
So this is caused by our AliasGroup computation finding there were too many transitive aliases, likely because the GO entry refers to (CVE-2024-6104), and I see a large number of chain guard advisories also refer to CVE-2024-6104, so deciding to skip generating the alias group.

There are 2 separate issues here: 

1. When no alias groups are generated,  we should still always return the upstream aliases for this specific entry. 
2. Chain guard might not be correctly using aliases, and should be publishing related instead.

andrewpollock:
> 2\. Chain guard might not be correctly using aliases, and should be publishing related instead.

@cpanato are you the right person to review the Chainguard records being published? Refer to https://ossf.github.io/osv-schema/#aliases-field and https://ossf.github.io/osv-schema/#related-field

cpanato:
> > 2. Chain guard might not be correctly using aliases, and should be publishing related instead.
> 
> @cpanato are you the right person to review the Chainguard records being published? Refer to https://ossf.github.io/osv-schema/#aliases-field and https://ossf.github.io/osv-schema/#related-field

can be, what is the issue, i did not understand that, do you have an example?

andrewpollock:
> can be, what is the issue, i did not understand that, do you have an example?

In a nutshell, `aliases` is meant to be 1:1 under most circumstances, whereas `related` may be 1:many

https://ossf.github.io/osv-schema/#aliases-field

> Aliases should be considered symmetric

https://ossf.github.io/osv-schema/#related-field

> Related vulnerabilities are symmetric but not transitive.

(any feedback on how to make this aspect of the schema documentation more comprehensible is very welcome)

It sounds like there are many Chainguard advisories aliased to CVE-2024-6104, which tends to indicate something is wrong:

- CGA-23mp-53xj-rxq8
- CGA-27mv-v84v-3r2r
[...]

PR #2376, opened 2024-07-09
Assign source aliases if no alias group is defined

hogo6002:
Some vulnerabilities may lack listed aliases, even if their source data contains them. This occurs when an `alias_group` exceeds a maximum number of aliases, causing this group to be not added. For example: https://osv.dev/vulnerability/GO-2024-2947 

In such cases, assign the original source aliases to the `aliases` field to match with the source info.

Fixes: #2374

### DIFF
--- a/osv/models.py
+++ b/osv/models.py
@@ -688,7 +688,7 @@ class Bug(ndb.Model):
       credits_.append(cr)
 
     related = self.related
-    aliases = []
+    aliases = self.aliases
 
     if include_alias:
       related_bugs = Bug.query(Bug.related == self.db_id).fetch()


hogo6002:
This can also cause issues where a vulnerability lists different aliases with its alias. Also, the raw aliases may sometimes exceed the maximum number of vulnerabilities allowed. Not sure if we really want to implement this. @michaelkedar

michaelkedar:
Yeah, my main issue with doing this is that there's no indication when we do fall back to the raw aliases and, when we do, the returned alias lists become no longer symmetric. So it still can be confusing to consumers...

hogo6002:
This seems not the right approach, close the PR for now.

Issue #1821, opened 2023-11-20
Log the vulnerability being processed to assist with fault investigation when analysis fails


andrewpollock:
**Describe the bug**
Git repository analysis can and does fail from time to time, and when it does it happens fairly deep in the call path from where the vulnerability record is being processed, e.g:

```
subprocess.CalledProcessError: Command '['git', '-C', '/work/tmp/tmp8es_y16u/.git/', 'branch', '-r', '--contains', '8f6e1365b6dec0ae1aa0b0b15b102c9133f347e5']' returned non-zero exit status 129.

at .run ( /usr/local/lib/python3.11/subprocess.py:571 )
at .check_output ( /usr/local/lib/python3.11/subprocess.py:466 )
at ._branches_with_commit ( /env/osv/impact.py:387 )
at ._get_affected_range ( /env/osv/impact.py:185 )
at .get_affected ( /env/osv/impact.py:133 )
at ._analyze_git_ranges ( /env/osv/impact.py:582 )
at .analyze ( /env/osv/impact.py:666 )
at ._analyze_vulnerability ( /usr/local/bin/worker.py:447 )
at ._do_update ( /usr/local/bin/worker.py:482 )
at ._source_update ( /usr/local/bin/worker.py:402 )
at ._do_process_task ( /usr/local/bin/worker.py:564 )
```

which makes it hard/impossible to reason about the vulnerability record that triggered the failure to then take further actions, in appropriate. It also makes it difficult/impossible to even know what the repository is (at least without having to do further log archaeology). 

**To Reproduce**
N/A

**Expected behaviour**
This exception has additional information that includes the vulnerability ID of the vulnerability that is being analyzed. One possible way to do this is to handle this exception earlier in the call stack where the vulnerability information is still available, and explicitly log something there. This would make the error logging a bit cleaner as well.

**Screenshots**
```
subprocess.CalledProcessError: Command '['git', '-C', '/work/tmp/tmp8es_y16u/.git/', 'branch', '-r', '--contains', '8f6e1365b6dec0ae1aa0b0b15b102c9133f347e5']' returned non-zero exit status 129.

at .run ( /usr/local/lib/python3.11/subprocess.py:571 )
at .check_output ( /usr/local/lib/python3.11/subprocess.py:466 )
at ._branches_with_commit ( /env/osv/impact.py:387 )
at ._get_affected_range ( /env/osv/impact.py:185 )
at .get_affected ( /env/osv/impact.py:133 )
[...]

PR #2406, opened 2024-07-22
Annotate any exceptions with the vulnerability ID

andrewpollock:
Learned about https://docs.python.org/3/library/exceptions.html#BaseException.add_note via go/python-tips/067

and since I've been looking at this code recently, knew exactly where to use it.

Fixes: #1821


oliverchang (approved):
very neat! We should probably add this to other places in our worker too.

PR #2408, opened 2024-07-23
Exempt backlog label from stale treatment

andrewpollock:
https://docs.github.com/en/issues/using-labels-and-milestones-to-track-work/managing-labels#applying-a-label

> Anyone with triage access to a repository can apply and dismiss labels.

### DIFF
--- a/.github/workflows/staleness.yml
+++ b/.github/workflows/staleness.yml
@@ -14,7 +14,7 @@ jobs:
           stale-issue-label: stale
           stale-pr-label: stale
           remove-stale-when-updated: true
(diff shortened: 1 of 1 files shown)


G-Rath:
you pushed you commit before I hit enter 😅 

fwiw I did this in https://github.com/google/osv-scanner/pull/1131 too - I couldn't tell from the repo if it's actually correct, but I guess we'll now both find out 🤷

G-Rath (approved):

oliverchang (approved):

michaelkedar (approved):

PR #2409, opened 2024-07-23
Run `pipenv verify` before sync

michaelkedar:
Pipenv doesn't automatically check if the `Pipfile` matches the `Pipfile.lock` when running `pipenv sync`. Added `pipenv verify` commands in all the places to do the manifest/lockfile verification. This should prevent things like #2379 from passing our tests.

Poetry seems to automatically do this validation when you run `poetry install` (But there is a `poetry check` command as well?)

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -40,7 +40,7 @@ jobs:
       - name: Install pipenv
         run: pip install pipenv==2023.12.1
       - name: Set up pipenv
-        run: pipenv sync --dev
+        run: pipenv verify && pipenv sync --dev
       - name: Set up Go
         uses: actions/setup-go@v5
         with:
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ ifeq "$(USE_POETRY)" "true"
 	install-cmd := poetry install
 	run-cmd := poetry run
 else
-	install-cmd := pipenv sync
+	install-cmd := pipenv verify && pipenv sync
 	run-cmd := pipenv run
 endif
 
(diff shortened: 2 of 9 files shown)


.github/workflows/lint.yaml
```diff
@@ -40,7 +40,7 @@ jobs:
       - name: Install pipenv
         run: pip install pipenv==2023.12.1
       - name: Set up pipenv
-        run: pipenv sync --dev
+        run: pipenv verify && pipenv sync --dev
```
andrewpollock, line 43: I'm unsure if the `&&` here will result in `pipenv sync` silently not being run if the verify fails?

Makefile
```diff
@@ -16,7 +16,7 @@ ifeq "$(USE_POETRY)" "true"
 	install-cmd := poetry install
 	run-cmd := poetry run
 else
-	install-cmd := pipenv sync
+	install-cmd := pipenv verify && pipenv sync
```
andrewpollock, line 19: I'm unsure if the `&&` here will result in `pipenv sync` silently not being run if the verify fails?

andrewpollock (approved):
LGTM with uncertainty

Makefile
michaelkedar: The make command prints and errors if the verify fails:
```
...
Pipfile.lock is out-of-date. Run $ pipenv lock to update.
make: *** [Makefile:64: run-api-server] Error 1
```

Which seems reasonable to me

.github/workflows/lint.yaml
michaelkedar: I'm not 100% sure, but I assumed if `pipenv verify` fails here the whole workflow will fail, similar to the Makefile.

PR #2416, opened 2024-07-25
Utility to download OSV.dev source records

andrewpollock:
To aid with running OSV record lint checks against all of a source's records, pre-import

### DIFF
--- a/tools/source-sync/source_sync.py
+++ b/tools/source-sync/source_sync.py
@@ -17,12 +17,12 @@ from google.cloud.storage import retry
 
 
 class Error(Exception):
-  pass
+  """General purpose exception for errors."""
 
 
-def Sources(source: str) -> dict:
+def Sources(source_path: str) -> dict:
   """Return a dictionary of the contents of a source.yaml file."""
-  return yaml.safe_load(open(source))
+  return yaml.safe_load(open(source_path))
 
 
 def DownloadSource(source: dict, directory: str, verbose: bool = False) -> None:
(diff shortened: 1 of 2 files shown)


tools/source-sync/source_sync.py
```diff
@@ -0,0 +1,192 @@
...
+from urllib3.util.retry import Retry
+import requests
+from requests.adapters import HTTPAdapter
+import re
+import pygit2
+from google.cloud import storage
+from google.cloud.storage import retry
+
+
+class Error(Exception):
```
oliverchang, line 19: nit: use a docstring here (and then we can also get rid of the `pass`)

tools/source-sync/source_sync.py
```diff
@@ -0,0 +1,192 @@
...
+  pass
+
+
+def Sources(source: str) -> dict:
```
oliverchang, line 23: nit: s/source/source_path/ to be more explicit about what this argument means?

oliverchang (approved):

tools/source-sync/source_sync.py
andrewpollock: Done in https://github.com/google/osv.dev/pull/2416/commits/159714a2edc34dfdd1d0b92fb6f9cdddfce632ee

tools/source-sync/source_sync.py
andrewpollock: Done in https://github.com/google/osv.dev/pull/2416/commits/159714a2edc34dfdd1d0b92fb6f9cdddfce632ee

PR #2417, opened 2024-07-29
Use android-osv-test for test.

oliverchang:


andrewpollock (approved):
This this require any additional adjustment of records already present in Datastore?

cuixq (approved):

hogo6002 (approved):

michaelkedar (approved):

oliverchang:
> This this require any additional adjustment of records already present in Datastore?

Hmm, possibly for entries that only exist in android-osv and not android-osv-test, since they won't be cleaned up right? I can take a look after the initial import.

PR #2432, opened 2024-08-01
Ignore CVE IDs from Ubuntu

michaelkedar:
Ubuntu has recently published a bunch of records with CVE IDs which is causing problems for us.
Explicitly ignore them for now.

### DIFF
--- a/source.yaml
+++ b/source.yaml
@@ -287,7 +287,7 @@
 - name: 'ubuntu'
   versions_from_repo: False
   type: 0
-  ignore_patterns: []
+  ignore_patterns: ['CVE-.*']
   directory_path: 'osv'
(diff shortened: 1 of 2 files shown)


andrewpollock (approved):
Thanks! A followup to see if it's possible to do a negative pattern ignoring the unexpected db_prefix will be the more sustainable solution to this.

another-rex (approved):

hogo6002 (approved):
Thanks!

cuixq (approved):
shall we leave a TODO or open an issue to remind us of further investigation?

andrewpollock:
> shall we leave a TODO or open an issue to remind us of further investigation?

I'll file an issue

Issue #2433, opened 2024-08-01
Sustainably ensure invalidly prefixed records are not imported


andrewpollock:
**Describe the bug**
One OSV record provider can publish records with the prefix of another provider, causing them to be clobbered unexpectedly.

**To Reproduce**
Steps to reproduce the behaviour:
1. https://github.com/canonical/ubuntu-security-notices/commit/8383ab87af7ecb44dd9fae038097cc9c001c6091
2. Hilarity ensues

**Expected behaviour**
Such records are ignored by the importer

**Additional context**
Partial fix: https://github.com/google/osv.dev/pull/2432

andrewpollock:
@dodys FYI. It's on us to be a little less liberal in what we're accepting 🤭 😬 🤦

oliverchang:
Note: we probably need to turn the existing https://github.com/google/osv.dev/blob/f182e67869f4963d1cfc406b75953edd46eb6338/osv/models.py#L834 into a repeated field instead.

IIRC, NDB lets us do this in a backwards compatible way -- we can make the change to the ndb.Model with the existing singular field in Datastore, and have it work seamlessly.

andrewpollock:
I think everything that needed to be done for this is now done and live in Production.

PR #2434, opened 2024-08-01
Test ignoring unexpected records

andrewpollock:
Addresses #2433

### DIFF
--- a/source_test.yaml
+++ b/source_test.yaml
@@ -44,7 +44,7 @@
 - name: 'android'
   versions_from_repo: False
   type: 1
-  ignore_patterns: ['^(?!A-).*$']
+  ignore_patterns: ['^(?!A-).*$', '^(?!ASB-A-).*$', '^(?!PUB-ASB-A-).*$']
   detect_cherrypicks: False
   extension: '.json'
   bucket: 'android-osv-test'
--- a/source_test.yaml
+++ b/source_test.yaml
@@ -182,8 +182,7 @@
 - name: 'malicious-packages'
   versions_from_repo: False
   type: 0
-  ignore_patterns: ['^(?!MAL-).*$']
-  ignore_patterns: ['MAL-0000.*', 'config.yaml']
+  ignore_patterns: ['^(?!MAL-).*$', 'MAL-0000.*', 'config.yaml']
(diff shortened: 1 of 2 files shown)


source_test.yaml
```diff
@@ -44,6 +44,7 @@
 - name: 'android'
   versions_from_repo: False
   type: 1
+  ignore_patterns: ['^(?!A-).*$']
```
oliverchang, line 47: I don't think this one is correct. 

There may be other cases for us to consider. They also export PUB-A-*.json for example.

oliverchang (approved):

andrewpollock:
I haven't desk-checked this against https://github.com/google/osv.dev/blob/f182e67869f4963d1cfc406b75953edd46eb6338/osv/models.py#L836-L846 yet

source_test.yaml
andrewpollock: Good catch, thanks. Corrected in 203fec69

oliverchang (approved):

andrewpollock:
I'm feeling fairly confident about this approach now that I've added some test coverage in aaa03831

source_test.yaml
```diff
@@ -68,6 +70,7 @@
   versions_from_repo: False
   rest_api_url: 'https://curl.se/docs/vuln.json'
   type: 2
+  ignore_patterns: ['^(?!CURL-).*$']
```
oliverchang, line 73: actually, I just noticed this line is duplicated with line 74. It also isn't applicable for REST sources right?

oliverchang (approved):

source_test.yaml
andrewpollock: Thanks for catching that. Fixed in 8c34b659, and yes, by the looks of things, https://github.com/google/osv.dev/blob/f182e67869f4963d1cfc406b75953edd46eb6338/docker/importer/importer.py#L58 is not currently called from https://github.com/google/osv.dev/blob/f182e67869f4963d1cfc406b75953edd46eb6338/docker/importer/importer.py#L652

I've left the entries for the two REST sources as a statement of intent, and added a comment to make it obvious that it's not currently honoured.

oliverchang (approved):

PR #2435, opened 2024-08-01
Enforce db_prefix.

oliverchang:
- Make this a repeated field in SourceRepository. No Datastore migrations are necessary as NDB seamlessly handles translating from a singular field in Datastore to a list in ndb.Model.

- Remove old legacy code that constructed a Datastore key for `Bug` based on the source (`<source_name>:<osv ID>`) if we're importing an unseen record and the the db_prefix didn't match. This actually never got used, because all of our Bug queries check for `db_id` and bypass the Datastore key (https://github.com/google/osv.dev/blob/f182e67869f4963d1cfc406b75953edd46eb6338/osv/models.py#L337)

  Existing keys in Datastore with the existing format (`<source_name>:<osv ID>`) will remain unmodified (again, because our queries key on `db_id`). In the future, we can further clean this up by removing `db_id` and just using the Datastore key instead (which will likely also bring some minor performance improvements).

For #2433.

### DIFF
--- a/docker/worker/worker_test.py
+++ b/docker/worker/worker_test.py
@@ -1126,6 +1126,7 @@ class UpdateTest(unittest.TestCase, tests.ExpectationTest(TEST_DATA_DIR)):
     self.source_repo.ignore_git = False
     self.source_repo.versions_from_repo = False
     self.source_repo.detect_cherrypicks = False
+    self.source_repo.db_prefix.append('PYSEC-')
     self.source_repo.put()
 
     self.mock_repo.add_file(
@@ -1175,6 +1176,7 @@ class UpdateTest(unittest.TestCase, tests.ExpectationTest(TEST_DATA_DIR)):
     self.source_repo.ignore_git = False
     self.source_repo.versions_from_repo = False
     self.source_repo.detect_cherrypicks = False
+    self.source_repo.db_prefix.append('PYSEC-')
     self.source_repo.put()
 
     self.mock_repo.add_file(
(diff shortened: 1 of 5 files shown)


andrewpollock (commented):
You'll want to omit any changes to source.yaml until the code changes have hit Production, right? 
Perhaps also source_test.yaml until the code changes are live in Staging?

andrewpollock (commented):
This is a pretty neat (and final line of defense) way by doing it at the Datastore pre-put hook layer. I like how bulletproof (and noisy) this is.

oliverchang:
> You'll want to omit any changes to source.yaml until the code changes have hit Production, right? Perhaps also source_test.yaml until the code changes are live in Staging?

Yes, good call out, I'll split this out.

oliverchang:
> > You'll want to omit any changes to source.yaml until the code changes have hit Production, right? Perhaps also source_test.yaml until the code changes are live in Staging?
> 
> Yes, good call out, I'll split this out.

Broke out https://github.com/google/osv.dev/pull/2446 for this.

docker/worker/worker_test.py
```diff
@@ -1620,6 +1631,18 @@ def test_last_affected_git(self):
         [codecs.encode(commit, 'hex') for commit in affected_commits.commits],
     )
 
+  def test_invalid_prefix(self):
+    """Test attempting to create a bug with a invalid db_prefix."""
+    with self.assertRaises(ValueError):
+      osv.Bug(
```
another-rex, line 1637:
```suggestion
      # Default db_prefix is `OSV-`
      osv.Bug(
```

another-rex (approved):

PR #2445, opened 2024-08-02
Don't index versions from git tags if they exceed 5000.

oliverchang:
This is to avoid Datastore index limits for edge cases (e.g. CVE-2022-2816 which currently generates 14k+ versions).

For git matching, the primary supported method is also via commit hashes instead of version numbers.


docker/worker/worker_test.py
```diff
@@ -1644,6 +1644,37 @@ def test_invalid_prefix(self):
           source_of_truth=osv.SourceOfTruth.SOURCE_REPO,
       ).put()
 
+  def test_dont_index_too_many_git_versions(self):
+    """Test that we don't index too many versions from Git."""
+    self.source_repo.ignore_git = False
+    self.source_repo.versions_from_repo = True
+    # detect_cherrypicks should not cause result in cherrypick detection for
+    # `last_affected`, since equivalent `last_affected` across different
+    # branches likely no have relation to the actual vulnerable range.
```
andrewpollock, line 1653: I think this is how you intended to word this?

```suggestion
    # branches likely have no relation to the actual vulnerable range.
```

docker/worker/worker_test.py
```diff
@@ -1644,6 +1644,37 @@ def test_invalid_prefix(self):
...
+    self.source_repo.detect_cherrypicks = True
+    self.source_repo.put()
+
+    self.mock_repo.add_file(
+        'OSV-TEST-last-affected-01.yaml',
```
andrewpollock, line 1658: I'm not following this use of this file versus `docker/worker/testdata/UpdateTest_dont_index_too_many_git_versions.txt` added in this PR.

osv/models.py
```diff
@@ -429,11 +432,25 @@ def _pre_put_hook(self):  # pylint: disable=arguments-differ
...
+        if (not affected_package.package.ecosystem and
+            len(affected_package.versions) > _MAX_GIT_VERSIONS_TO_INDEX):
+          # Assume that if there is no ecosystem specified, then these versions
+          # were enumerated from Git.
+          #
+          # Mitigate cases where the Git repo tag matching results in too many
+          # versions to index for Datastore.
+          # It's OK to do this because the primary intended matching mechanism
+          # for Git is via commit hash matching instead.
+          logging.info(
```
andrewpollock, line 444: Could we also add/append to https://ossf.github.io/osv-schema/#affecteddatabase_specific-field here so it's explicit to a record consumer that we've made this decision here?

docker/worker/worker_test.py
oliverchang: This is an existing test input OSV file, not added by this PR. I'll add more comments here.

`docker/worker/testdata/UpdateTest_dont_index_too_many_git_versions.txt` is the fixture for recording the expected output.

osv/models.py
oliverchang: This has nothing to do with database_specific.

This is just an indexing side change for OSV.dev. The only visible changes for OSV producers is that more of their records will import, but they will fail to match via a version query to OSV.dev API.

docker/worker/worker_test.py
oliverchang: I'll remove this completely, since it's irrelevant to this test.

osv/models.py
andrewpollock: If I've read this correctly, for such records `affected.versions[]` will be completely empty (versus truncated to 5000)? I guess as an alternative to making this explicitly clear in the record itself, calling it out at https://google.github.io/osv.dev/faq/#what-does-osvdev-do-to-the-records-it-imports ?

The problem I frequently run into when reading the code that does the work is there's a lot of other "last-minute" things that go on in the pre-put hook in `models.py` that aren't immediately obvious and violate the principle of least surprise.

oliverchang: No, we are simply not indexing them in `affected_fuzzy` (which is not visible to users). We are not truncating them in the OSV record.

oliverchang: see https://github.com/google/osv.dev/blob/c5c3e96d0e9a0b56922bc7a32c27f0c4975396aa/osv/models.py#L274

oliverchang: And also the test output file: https://github.com/google/osv.dev/blob/17ee9228492825626dc0efb1d6d4eef1075ed00c/docker/worker/testdata/UpdateTest_dont_index_too_many_git_versions.txt#L15

oliverchang: And also the test output file: https://github.com/google/osv.dev/blob/17ee9228492825626dc0efb1d6d4eef1075ed00c/docker/worker/testdata/UpdateTest_dont_index_too_many_git_versions.txt#L15

andrewpollock: My bad, I misunderstood indexing as enumeration...

andrewpollock (approved):

PR #2454, opened 2024-08-06
feat: use GCS object update time instead of creation time

andrewpollock:
This should achieve the same outcome and is easier to manipulate (it doesn't require object recreation) and so should make triggering a reimport of specific records a little more efficient (doesn't require downloading and reuploading the object to reset the _creation_ time)

Tested in Staging:

Before:
```
$ gsutil stat gs://osv-test-cve-osv-conversion/osv-output/CVE-2024-7067.json
gs://osv-test-cve-osv-conversion/osv-output/CVE-2024-7067.json:
    Creation time:          Fri, 26 Jul 2024 16:53:59 GMT
    Update time:            Wed, 31 Jul 2024 03:52:06 GMT
    Storage class:          STANDARD
    Content-Length:         3073
    Content-Type:           application/json
    Metadata:
        goog-reserved-file-mtime:1722394200
    Hash (crc32c):          Lo3Xwg==
    Hash (md5):             rdI8478THJQOVCgblWW9UQ==
    ETag:                   COmK0tyVxYcDED4=
    Generation:             1722012839216489
    Metageneration:         62
```

After:
```
$ gsutil stat gs://osv-test-cve-osv-conversion/osv-output/CVE-2024-7067.json
gs://osv-test-cve-osv-conversion/osv-output/CVE-2024-7067.json:
    Creation time:          Fri, 26 Jul 2024 16:53:59 GMT
    Update time:            Tue, 06 Aug 2024 07:07:12 GMT
    Storage class:          STANDARD
    Content-Length:         3073
    Content-Type:           application/json
    Metadata:
        goog-reserved-file-mtime:1722394200
    Hash (crc32c):          Lo3Xwg==
    Hash (md5):             rdI8478THJQOVCgblWW9UQ==
    ETag:                   COmK0tyVxYcDED8=
    Generation:             1722012839216489
    Metageneration:         63
```

I have a sneaking suspicion this may facilitate the intentional reimporting of CVEs that get updated when one of their constituent parts changes (the fact that the original creation and update times were already divergent surprised me)

This also requires `combine-to-osv` to switch back to using `gsutil rsync` instead of `gcloud storage rsync` due to a subtle difference in the treatment of the GCS object modification time for otherwise unchanged objects. (see #2196 for additional context)


hogo6002:
I recall that we previously encountered an issue where `gcloud storage rsync --checksums-only` was repeatedly updating the modified time on GCS objects (https://github.com/google/osv.dev/pull/2196). We are still [using this command](https://github.com/google/osv.dev/blob/baa360d54f68001b594a0689f0d2f3b9ad31c806/vulnfeeds/cmd/combine-to-osv/run_combine_to_osv_convert.sh#L44) to write results to the `osv-output` folder. So I am wondering that our `osv-output` entries may always update their modification times even when there are no actual changes to the data. This may cause we re-import more times than we need if we switch checks from creation time to update time?

andrewpollock:
> I recall that we previously encountered an issue where `gcloud storage rsync --checksums-only` was repeatedly updating the modified time on GCS objects (#2196)

Ah, if that's the (unfortunate) culprit (which I dare say it will be) we may need to also migrate that away from `gcloud storage rsync` in favour of `gsutil rsync`... That difference in behaviour continues to annoy me :-(

hogo6002 (approved):
LGTM, thanks Andrew!

docker/importer/importer.py
```diff
@@ -463,7 +463,8 @@ def convert_blob_to_vuln(blob: storage.Blob) -> Optional[Tuple[str, str]]:
       if not _is_vulnerability_file(source_repo, blob.name):
         return None
       if not ignore_last_import_time and \
-         not blob.time_created > utc_last_update_date:
+   blob.updated is not None and \
+   not blob.updated > utc_last_update_date:
```
another-rex, line 467:
```suggestion
         blob.updated is not None and \
         not blob.updated > utc_last_update_date:
```

andrewpollock:
@hogo6002 if I followed #2611 correctly, it makes what's changing in this PR more viable?

hogo6002:
> @hogo6002 if I followed #2611 correctly, it makes what's changing in this PR more viable?

Yeah, I think as long as we use `gsutil`, it should be ok for us to merge this PR.

PR #2460, opened 2024-08-07
refactor: Migrate from Pipenv to poetry

another-rex:
This moves Pipenv to poetry, and removes Pipfile Pipfile.lock, and other mentions or Pipenv from the repository. 

Changes in addition to a direct migration:
- Appengine
  - Removes gcloud appengine related files, as we are now fully on cloud run via building a docker image (With the exception of index.yaml, which is for appengine datastore)
  - Add a `build.sh` script for building the docker image similar to other `build.sh` scripts.
  - Removed most dependencies in deployment as it's originally only needed for appengine (see the deleted `gcp/appengine/deploy.sh`)
- Python environment
  - For many of the docker images, pipenv installs to the **system** python dependencies. This is not really ideal, as even in a docker container, it is good practice to compartmentalise into separate virtualenvs. 
  - For the worker image, I have retained this behaviour, and will leave this migration to a future PR, as it affects a large number of different images based on it, e.g.:
    - ci
    - cron
    - exporter
    - importer
  - For other non worker based images, I have switched to using virtualenvs, this includes:
    - Appengine/Cloud Run
    - API
    - Debian (already uses virtualenvs)

TODO/looking for suggestions:

The poetry version is defined in every file that it is installed with `==`, is there a way to set the version number for the whole repository in one place?


another-rex:
It looks like we need an intermediary CI that has both poetry and Pipenv installed.

cloudbuild.yaml
```diff
@@ -26,6 +26,9 @@ steps:
   args: ['bash', '-ex', 'run_tests.sh']
   env:
     # Each concurrent test that uses the datastore emulator must have a unique port number
+    # lib-tests is the first test that starts the emulator. The first emulator call creates the
```
cuixq, line 29: this is the change in another PR to fix race condition right?

docker/deployment/Dockerfile
```diff
@@ -1,15 +1,8 @@
...
 RUN echo "deb [signed-by=/usr/share/keyrings/cloud.google.gpg] http://packages.cloud.google.com/apt cloud-sdk main" | tee -a /etc/apt/sources.list.d/google-cloud-sdk.list && \
     curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | gpg --dearmor -o /usr/share/keyrings/cloud.google.gpg && \
     apt-get update && apt-get install -y google-cloud-sdk
-
-RUN pip3 install pipenv==2023.12.1
-# Install the latest npm version 8 release (last release is in Feb 2023)
-# This will take precedence over the system installed version on the PATH
-RUN npm install -g npm@8
-
-COPY --from=gcr.io/oss-vdb/ci /root/go/bin/hugo /usr/local/bin/hugo
```
cuixq, line 15: is it intentional not to `npm` or copy `hugo`?

another-rex: Yes, these dependencies were originally included because we deployed the website through appengine using this deployment image. This is no longer necessary, as we build everything in appengine's own image now, and only use the deployment image to upload to cloudrun.

cloudbuild.yaml
another-rex: Yes, I split this change out into another PR to fix the race condition. I cannot merge that other PR before this one though, as that will overwrite the temporary CI image I am using which has both poetry and Pipenv installed.

cuixq (approved):

setup.py
```diff
@@ -33,7 +33,7 @@
         'Operating System :: OS Independent',
     ],
     install_requires=[
-        # TODO(ochang): Get this from the Pipfile.
+        # TODO(ochang): Migrate to using pyproject.toml directly.
```
another-rex, line 36:
```suggestion
        # TODO(rexpan): Migrate to using pyproject.toml directly.
```

docker/worker/Dockerfile
```diff
@@ -19,12 +19,18 @@ RUN apt-get update && apt-get upgrade -y
...
 # TODO(ochang): Just copy the entire project (needs a clean checkout).
-COPY docker/worker/Pipfile* /env/docker/worker/
-COPY setup.py Pipfile* README.md /env/
+COPY docker/worker/poetry.lock docker/worker/pyproject.toml /env/docker/worker/
+COPY setup.py poetry.lock pyproject.toml README.md /env/
 COPY osv /env/osv
-RUN cd /env/docker/worker && pipenv install --deploy --system
+
+# Set virtualenv creation to false to install globally
+RUN poetry config virtualenvs.create false
```
oliverchang, line 29: can we just set `POETRY_VIRTUALENVS_CREATE=false poetry install`  of these?

per https://python-poetry.org/docs/configuration/#virtualenvscreate

docker/worker-base/Dockerfile
```diff
@@ -35,15 +35,21 @@ RUN curl -sS https://www.python.org/ftp/python/${PYTHON_VERSION}/Python-${PYTHON
     make -j && \
     make install && \
     rm -rf /tmp/Python-${PYTHON_VERSION}
-RUN pip3 install pipenv==2023.12.1
+
+# Setup Poetry in its own virtual environment.
+# So when poetry changes the system dependencies, it doesn't uninstall itself.
```
oliverchang, line 40: This is pretty surprising. Do we know why poetry does this?

Looking at https://python-poetry.org/docs/cli/#options-2, it's not clear that this is the default behaviour.  `--sync` or `--remove-untracked` doesn't seem like default behaviour?

another-rex: The comment itself is probably not fully accurate, yes uninstalling itself completely is unlikely, but we could have dependencies that overlap with poetry's own dependencies, therefore changing the version unexpectedly.

This is the biggest/first warning on the installation instructions: https://python-poetry.org/docs/#installation

> Poetry should always be installed in a dedicated virtual environment to isolate it from the rest of your system. It should in no case be installed in the environment of the project that is to be managed by Poetry. This ensures that Poetry’s own dependencies will not be accidentally upgraded or uninstalled. (Each of the following installation methods ensures that Poetry is installed into an isolated environment.) In addition, the isolated virtual environment in which poetry is installed should not be activated for running poetry commands.

docker/worker/Dockerfile
another-rex: Good idea!

docker/worker-base/Dockerfile
oliverchang: Got it, thanks! Could you add this link and detail as a comment here?

another-rex: Done

osv/osv-schema
michaelkedar, line 1: What's happened here? -32 lines on the CHANGELOG.md seems like it's been downgraded.

another-rex: Oh... hmm must have somehow reverted the submodule version...

another-rex: Removed.

oliverchang (approved):

PR #2462, opened 2024-08-07
feat: add durable support for overriding the bucket

andrewpollock:
This is a more durable solution to the edge case we were trying to correct from, where the Bug in Datastore was from a GIT source (with a different path to the source file) than what we wanted to trigger reimporting for, and so various sanity checks and assumptions fail.

This adds support for overriding the bucket name (and optionally, the path within the bucket), which is normally self-determined from the existing record.

```
pipenv run python tools/datafix/reimport_gcs_record.py --verbose --no-dry-run --project oss-vdb --bucket cve-osv-conversion:osv-output CVE-2022-2816
```


andrewpollock:
This is instead of the quick hacks in 536d8403169585521583475f89d6ce5aa03e0a92

tools/datafix/reimport_gcs_record.py
```diff
@@ -150,6 +168,13 @@ def main() -> None:
       dest="tmpdir",
       default="/tmp",
       help="Local directory to copy to from GCS")
+  parser.add_argument(
+      "--bucket",
+      action="store",
+      dest="bucket",
+      default=None,
+      help=("Override the bucket name (and with a colon + path, the path) "
```
another-rex, line 176: Can you add an example here? The help text is a bit hard to parse. From looking at the source I think you mean 
```
bucket:optional/path/prefix/<individual-item-paths>
```

andrewpollock: I've expanded the help text in [a03a80e](https://github.com/google/osv.dev/pull/2462/commits/a03a80e5a7f2d40c6c17d8059def0824741dce5a)

another-rex (approved):

PR #2466, opened 2024-08-09
fix: datastore emulator startup race condition

another-rex:
The `start_datastore_emulator()` function that runs the gcloud emulator actually will run two commands.

First it will check whether `/builder/home/.config/gcloud/emulators/datastore` exists, if not it will run `cloud_datastore_emulator create` with the test project ID, which makes the directory. Then it will follow up with the actual datastore emulator command.

The problem is with the first command, if multiple datastore emulator tests are ran at the start without `/builder/home/.config/gcloud/emulators/datastore` existing, it will try to run `cloud_datastore_emulator create` multiple times, which will fail, as by the time it's ran, the directory would have started existing.

This fix makes it so one data store emulator is ran by itself at the start (`lib-tests`), before any concurrent tests start.

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -26,6 +26,9 @@ steps:
   args: ['bash', '-ex', 'run_tests.sh']
   env:
     # Each concurrent test that uses the datastore emulator must have a unique port number
+    # lib-tests is the first test that starts the emulator. The first emulator call creates the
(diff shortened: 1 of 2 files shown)


oliverchang (approved):

oliverchang:
thanks for debugging this!

cuixq (approved):

PR #2469, opened 2024-08-12
ci(workflow): address github.com/rhysd/actionlint findings

andrewpollock:
`branches` is invalid for `on.merge-group`


another-rex (approved):
FYI if we ever decide to re-enable merge groups (I believe it is now also supported by cloudbuild), we'll need to add the `[checks_requested]` line.

I think for now we have no plans to enable merge groups.

PR #2480, opened 2024-08-15
fix(combine-to-osv): attempt to work around execution latency problems

andrewpollock:
Given that we're seeing the GCS copy phase at the end of the execution
of combine-to-osv overrun the current CronJob:
- allow the CronJob to run concurrently, on the assumption that they
  will be operating on different sets of files concurrently, in the
  worst case scenario
- double the maximum runtime to 4h to prevent premature termination
- don't restart on failure, to advoid masking failures and on the
  assumption the next scheduled run will be "soon enough"

Pin Cloud SDK at [485.0.0](https://cloud.google.com/sdk/docs/release-notes#48500_2024-07-23) to test the hypothesis that something was introduced in [486.0.0](https://cloud.google.com/sdk/docs/release-notes#48600_2024-07-30) that caused a latency regression.

Also add some additional instrumentation to what gets logged so we have data on the scale of various operations.

### DIFF
--- a/deployment/clouddeploy/gke-workers/base/combine-to-osv.yaml
+++ b/deployment/clouddeploy/gke-workers/base/combine-to-osv.yaml
@@ -4,11 +4,11 @@ metadata:
   name: combine-to-osv
 spec:
   schedule: "30 */1 * * *"
-  concurrencyPolicy: Forbid
+  concurrencyPolicy: Allow # on the assumption the dominant part of the runtime is the GCS copy at the end
   failedJobsHistoryLimit: 3
   jobTemplate:
     spec:
-      activeDeadlineSeconds: 7200
+      activeDeadlineSeconds: 14400
       template:
         spec:
           tolerations:
@@ -30,7 +30,7 @@ spec:
                 memory: "8G"
           nodeSelector:
             cloud.google.com/gke-nodepool: highend
-          restartPolicy: OnFailure
+          restartPolicy: Never # on the assumption that the next scheduled run is soon enough
           volumes:
             - name: "ssd"
               hostPath:
--- a/vulnfeeds/cmd/combine-to-osv/Dockerfile
+++ b/vulnfeeds/cmd/combine-to-osv/Dockerfile
@@ -26,7 +26,7 @@ RUN go build -o combine-to-osv ./cmd/combine-to-osv/
 RUN go build -o download-cves ./cmd/download-cves/
 
 
-FROM gcr.io/google.com/cloudsdktool/google-cloud-cli:alpine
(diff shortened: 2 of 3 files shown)


another-rex (approved):
Do we also want to bump the timeout up by an hour or two to quickly resolve this, and figure out a longer term fix later?

andrewpollock:
> Do we also want to bump the timeout up by an hour or two to quickly resolve this, and figure out a longer term fix later?

Can we surgically do that on a Friday? Otherwise I'd say let's try to get Staging fixed today and do that on Monday?

another-rex:
If we just bump up the timeout in the yaml file and merge, it should get deployed to staging by clouddeploy I believe? We can leave the actual deployment to prod next week.

andrewpollock:
> If we just bump up the timeout in the yaml file and merge, it should get deployed to staging by clouddeploy I believe? We can leave the actual deployment to prod next week.

I'm all for that. Made a bunch of changes in 6fc009c, PTAL

deployment/clouddeploy/gke-workers/base/combine-to-osv.yaml
```diff
@@ -4,11 +4,11 @@ metadata:
   name: combine-to-osv
 spec:
   schedule: "30 */1 * * *"
-  concurrencyPolicy: Forbid
+  concurrencyPolicy: Allow # on the assumption the dominant part of the runtime is the GCS copy at the end
```
michaelkedar, line 7: Is there a risk we'll end up with two jobs trying to do the GCS copy at the same time (and would that be a problem)?

michaelkedar (approved):

deployment/clouddeploy/gke-workers/base/combine-to-osv.yaml
andrewpollock: > Is there a risk we'll end up with two jobs trying to do the GCS copy at the same time (and would that be a problem)?

Yes there is, and I'm of the opinion that it won't be a problem because:

- the source files being copied will be in discretely separate pods
- the concurrent jobs will be operating at different points in these discrete sets of files

i.e. newer job B will start at (expected different) file 1 (of ~39K) while older job B is already up to file 22K (of ~39K) and the set of files that actually require transferring (as opposed to getting their modification time updated) will be disjoint.

I'm happy to be proven wrong here.

michaelkedar: Sounds reasonable to me

michaelkedar (approved):

another-rex (approved):

PR #2388, opened 2024-07-11
Add security tracker links to references

hogo6002:
Includes Alpine and Debian security tracker links to vulnerability references on OSV.dev
Fixes https://github.com/google/osv.dev/issues/2284

PR #2481, opened 2024-08-16
fix: remove security tracker URL validation to reduce job runtime

hogo6002:
https://github.com/google/osv.dev/pull/2388 significantly increased `combine-to-osv` runtime and caused job failures. Remove the URL validation to reduce time.

Follow up PR: run all HTTP request parallelly.

### DIFF
--- a/vulnfeeds/cmd/combine-to-osv/main.go
+++ b/vulnfeeds/cmd/combine-to-osv/main.go
@@ -252,10 +252,5 @@ func addReference(cveId string, ecosystem string, convertedCve *vulns.Vulnerabil
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


another-rex:
A followup to this would be to reenable this, but just parallelise the combineIntoOSV function, which should speed this up by quite a bit, as I think most of the time is just waiting for a response. Go should make this pretty straightforward.

andrewpollock (approved):
From watching the current run in Staging with #2480 applied, yes, I think it's the combination of the additional runtime this introduces, plus GCS performance issues with the newer Cloud SDK that has tipped the total runtime over the previous 2h maximum... 

So let's merge this and review another run of it in Staging today to confirm the combination brings the run time back to something far more reasonable...

PR #2484, opened 2024-08-16
fix(build): pin all usage of cloud-sdk image

andrewpollock:
A recent apparent GCS performance regression in the Cloud SDK highlighted uncontrolled upgrading of the Cloud SDK Docker image. Pin everything to what #2480 pinned to for stability.

### DIFF
--- a/docker/terraform/Dockerfile
+++ b/docker/terraform/Dockerfile
@@ -6,7 +6,7 @@ ARG TERRAFORM_VERSION
 WORKDIR /build/
 RUN GOBIN=$(pwd) go install github.com/hashicorp/terraform@v${TERRAFORM_VERSION}
 
-FROM gcr.io/google.com/cloudsdktool/cloud-sdk:alpine
+FROM gcr.io/google.com/cloudsdktool/google-cloud-cli:alpine@sha256:d5da0344b23d03a6f2728657732c7a60300a91acaad9b8076c6fd30b1dfe1ff4
 
 COPY --from=GO_BUILD /build/terraform /usr/bin/terraform
 COPY entrypoint.bash /builder/entrypoint.bash
--- a/vulnfeeds/cmd/alpine/Dockerfile
+++ b/vulnfeeds/cmd/alpine/Dockerfile
@@ -25,7 +25,7 @@ COPY ./ /src/
 RUN go build -o alpine-osv ./cmd/alpine/
 
 
-FROM gcr.io/google.com/cloudsdktool/google-cloud-cli:alpine
+FROM gcr.io/google.com/cloudsdktool/google-cloud-cli:alpine@@sha256:d5da0344b23d03a6f2728657732c7a60300a91acaad9b8076c6fd30b1dfe1ff4
 
 WORKDIR /root/
 COPY --from=GO_BUILD /src/alpine-osv ./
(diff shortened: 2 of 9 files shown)


hogo6002 (approved):
Thanks Andrew!

vulnfeeds/cmd/alpine/Dockerfile
```diff
@@ -25,7 +25,7 @@ COPY ./ /src/
 RUN go build -o alpine-osv ./cmd/alpine/
 
 
-FROM gcr.io/google.com/cloudsdktool/google-cloud-cli:alpine
+FROM gcr.io/google.com/cloudsdktool/google-cloud-cli:alpine@@sha256:d5da0344b23d03a6f2728657732c7a60300a91acaad9b8076c6fd30b1dfe1ff4
```
another-rex, line 28: Is double `@` the correct syntax? I thought it was just one `@`

andrewpollock: 🤦‍♂️ thanks for catching that, fixed in [5d1c751](https://github.com/google/osv.dev/pull/2484/commits/5d1c75185134d7f0e95e729e3494b32872e8b9ea)

another-rex (approved):

docker/terraform/Dockerfile
```diff
@@ -6,7 +6,7 @@ ARG TERRAFORM_VERSION
 WORKDIR /build/
 RUN GOBIN=$(pwd) go install github.com/hashicorp/terraform@v${TERRAFORM_VERSION}
 
-FROM gcr.io/google.com/cloudsdktool/cloud-sdk:alpine
+FROM gcr.io/google.com/cloudsdktool/google-cloud-cli:alpine@sha256:d5da0344b23d03a6f2728657732c7a60300a91acaad9b8076c6fd30b1dfe1ff4
```
andrewpollock, line 9: @michaelkedar I standardized on the same image everywhere, can you confirm that's okay to do here?

vulnfeeds/tools/debian/Dockerfile
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM google/cloud-sdk:449.0.0-alpine
+FROM gcr.io/google.com/cloudsdktool/google-cloud-cli:alpine@sha256:d5da0344b23d03a6f2728657732c7a60300a91acaad9b8076c6fd30b1dfe1ff4
```
andrewpollock, line 15: @hogo6002 I standardized on the same image everywhere, can you confirm this is okay to do here?

docker/terraform/Dockerfile
michaelkedar: I'm not actually sure what's specifically required from this image.
The [original source of this Dockerfile](https://github.com/GoogleCloudPlatform/cloud-builders-community/blob/master/terraform/Dockerfile) used `gcr.io/cloud-builders/gcloud`, so if that had some special sauce needed for running in Cloud Build, it wasn't necessary.

So I'd say it's *probably* fine.

michaelkedar: Also note that this image isn't automatically built/pushed.

michaelkedar (approved):

vulnfeeds/tools/debian/Dockerfile
hogo6002: Yeah, i think this is fine.

Issue #2485, opened 2024-08-16
Missing CVSS_V4 Severity Type in osvSeverityType Enum in Swagger Definition


James-o-Howson:
It appears that the current Swagger definition provided by your API does not include the CVSS_V4 severity type in the osvSeverityType enum. However, when querying the API, the response includes severity entries with type set to CVSS_V4. This inconsistency between the API's actual behavior and the Swagger documentation is causing issues in our client code, as the missing enum value is not accounted for during code generation.

The swagger definition was retrieved from https://osv.dev/docs/osv_service_v1.swagger.json.

The following query object was used against `https://api.osv.dev/v1/query`:
```
var query = new V1Query
{
    Version = "4.1.2",
    Package = new OsvPackage
    {
        Name = "IdentityServer4",
        Ecosystem = "NuGet"
    }
};
```

The following `severity` array was returned in the response for vulnerability with `id` of `GHSA-55p7-v223-x366`
```
"severity": [
  {
    "type": "CVSS_V3",
    "score": "CVSS:3.1/AV:N/AC:L/PR:N/UI:R/S:C/C:L/I:N/A:N"
  },
  {
    "type": "CVSS_V4",
    "score": "CVSS:4.0/AV:N/AC:L/AT:N/PR:N/UI:A/VC:L/VI:N/VA:N/SC:L/SI:N/SA:N"
  }
]
```

You would expect the swagger definition made available to accurately reflect the APIs definition.

PR #2486, opened 2024-08-18
fix(api): add CVSS_V4 to API Swagger documentation

andrewpollock:
The addition of `CVSS_V4` in #2312 did not add it to the Swagger documentation.

Open to suggestions on how to guard against such divergence in the future.

Fixes: #2485


another-rex:
It looks like the swagger spec can be generated from the `build_swagger.py` script. We can make a workflow to run that and check the diff between the generated swagger and the one checked in.

another-rex (approved):

PR #2488, opened 2024-08-19
fix(build): install Poetry via apk instead of pip

andrewpollock:
The switch from `google/cloud-sdk:449.0.0-alpine` to `gcr.io/google.com/cloudsdktool/google-cloud-cli:alpine` was not a drop-in replacement:

This moved from Alpine 3.18.4 to 3.20.2 and 3.20.2 is more intolerant of the `pip install poetry==1.8.3` being performed:

```
error: externally-managed-environment
```

Install Poetry via `apk` instead.

### DIFF
--- a/vulnfeeds/tools/debian/Dockerfile
+++ b/vulnfeeds/tools/debian/Dockerfile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM gcr.io/google.com/cloudsdktool/google-cloud-cli:alpine@sha256:d5da0344b23d03a6f2728657732c7a60300a91acaad9b8076c6fd30b1dfe1ff4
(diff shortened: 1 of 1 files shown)


andrewpollock:
Please advise if there's a more appropriate forward-fix here.

another-rex (approved):

another-rex:
We can probably install poetry directly via `apk install poetry`

hogo6002 (approved):

another-rex:
Tested installing poetry directly, the docker images builds successfully, maybe let's try that first.

andrewpollock:
> We can probably install poetry directly via `apk install poetry`

Yeah good call, trying that in 2d304614 instead.

andrewpollock:
I guess the (small) trade off here is the hidden potential for Poetry version divergence?

PR #2492, opened 2024-08-19
feat(api-perf-test): revert API query to original method on test instance for A/B performance testing

hogo6002:
Temporarily revert API query to original method on test instance for A/B performance testing

### DIFF
--- a/gcp/api/server.py
+++ b/gcp/api/server.py
@@ -1063,8 +1063,13 @@ def query_by_version(context: QueryContext,
           bugs.append(bug)
     elif project == 'oss-vdb-test' and supports_ordering:
       # Query for non-enumerated ecosystems.
-      bugs, next_page_token = yield _query_by_comparing_versions(
-          context, query, ecosystem, version)
+
+      # Performance testing only
+      # TODO(gongh): revert change back after testing.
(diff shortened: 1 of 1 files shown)


gcp/api/server.py
```diff
@@ -1063,8 +1063,13 @@ def query_by_version(context: QueryContext,
...
       # Query for non-enumerated ecosystems.
-      bugs, next_page_token = yield _query_by_comparing_versions(
-          context, query, ecosystem, version)
+
+      # Performance testing only
+      # TODO(gongh): revert change back after testing.
+      # bugs, next_page_token = yield _query_by_comparing_versions(
+      #     context, query, ecosystem, version)
+      bugs, next_page_token = yield _query_by_generic_version(
+          context, query, package_name, ecosystem, purl, version)
```
cuixq, line 1072: since these two lines are the same as below, can we just comment out this block?

hogo6002: I feel updating here makes it easier to see the difference than commenting out the block. Both do the same thing, and I will revert this change back after the deployment. So I think it doesn't really matter

oliverchang (approved):

oliverchang (approved):

PR #2493, opened 2024-08-20
fix: bump exporter memory allowances

michaelkedar:
The large number and size of records coming from Ubuntu made our exporter job start exceeding the memory limit.
Throw some more memory at the problem until it goes away.


another-rex (approved):

andrewpollock (approved):
Is there still an active problem with the exporter?

michaelkedar:
> Is there still an active problem with the exporter?

Yeah, it seems like it fails to export some of the time. Might want to release this if there's no problems in staging.

PR #2494, opened 2024-08-20
build: pin remaining unpinned third-party Docker images

andrewpollock:
This avoids uncontrolled changes to build dependencies sneaking into our Docker image builds, and is considered best practice.

The Docker Hub expresses both an "index digest" and a "manifest digest". I have used the latter, as this seemed the correct thing to do.

### DIFF
--- a/docker/deployment/Dockerfile
+++ b/docker/deployment/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:22.04
+FROM ubuntu:22.04@sha256:fa17826afb526a9fc7250e0fbcbfd18d03fe7a54849472f86879d8bf562c629e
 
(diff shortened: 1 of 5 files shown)


cuixq (approved):

hogo6002:
I am not sure about the difference between an index hash and a manifest hash, but it seems the [Renovate bot always updates ](https://github.com/google/osv.dev/pull/2345/files)the hash to the [index hash,](https://hub.docker.com/layers/library/golang/1.22.4-alpine/images/sha256-6522f0ca555a7b14c46a2c9f50b86604a234cdc72452bf6a268cae6461d9000b) even when we provide the manifest hash.

hogo6002 (approved):

PR #2495, opened 2024-08-20
fix: add total ordering to Maven version

hogo6002:
Maven `Version` doesn't support `>=`, causing [version query](https://github.com/google/osv.dev/blob/30d55798022eabb000ecc091f03ec7d5a2002c4a/gcp/api/server.py#L1250) failures. Add the `total_ordering` function to support all kinds of ordering.

### DIFF
--- a/osv/ecosystems/maven.py
+++ b/osv/ecosystems/maven.py
@@ -14,6 +14,7 @@
 """Maven ecosystem helper."""
 
 import collections
+import functools
 import re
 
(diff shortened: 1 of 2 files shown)


michaelkedar (approved):

another-rex (approved):

cuixq (approved):

andrewpollock (approved):
Easy fix :-)

PR #2496, opened 2024-08-20
feat: API Dynamic page sizing when query exceeds 30 seconds

another-rex:
Do dynamic page sizing for when the query starts to exceed 30 seconds. 

Turns out past me already thought of this, just needed to uncomment the relevant lines.

### DIFF
--- a/gcp/api/server.py
+++ b/gcp/api/server.py
@@ -777,7 +777,8 @@ def query_by_commit(
   cursor = None
   while (yield it.has_next_async()):
     now = datetime.now()
-    if len(bug_ids) >= context.total_responses.page_limit() or (now - context.request_start_time) > _MAX_QUERY_TIME:
+    if len(bug_ids) >= context.total_responses.page_limit() or (
+        now - context.request_start_time) > _MAX_QUERY_TIME:
       cursor = it.cursor_after()
       break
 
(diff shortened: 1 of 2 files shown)


another-rex:
A significant amount of time is spend in converting the bug to response. When running locally, I managed to retrieve 93 entries before the time limit is hit, resulting in a 16mb file. The conversion on those 93 entries took 6 seconds. 

When running in the cloud, I expect to be able to retrieve more than 93 entries, therefore the conversion will also take longer. If the conversion takes longer than 30 seconds, this will still timeout. Requires testing this in staging.

hogo6002 (approved):

gcp/api/server.py
```diff
@@ -992,7 +998,9 @@ def query_by_generic_helper(results: list, cursor, context: QueryContext,
   query: ndb.Query = base_query.filter(osv.Bug.affected_fuzzy == version)
   it: ndb.QueryIterator = query.iter(start_cursor=context.page_token)
   while (yield it.has_next_async()):
-    if len(results) >= context.total_responses.page_limit():
+    now = datetime.now()
```
oliverchang, line 1001: nit: is it possible to write some kind of wrapper such that there's less  duplication of  this logic across all the places we iterate?

gcp/api/test_server.py
```diff
@@ -139,7 +139,7 @@ def start_esp(port, backend_port, credential_path, log_path):
         '--non_gcp',
         '--enable_debug',
         '--transcoding_preserve_proto_field_names',
-        '--envoy_connection_buffer_limit_bytes=10485760',
+        '--envoy_connection_buffer_limit_bytes=104857600',
```
oliverchang, line 142: is there some prod config we have to change here too?

another-rex: Yes, just discovered this when running on staging, we need to update the terraform config as well, I'll make a follow-up PR after testing on staging.

PR #2497, opened 2024-08-20
fix(api): Reduce dynamic paging time to 20 seconds

another-rex:
- Refactored the code to take in a cutoff time and place all the shared logic in a context method.
- Separate out batch query time limit and normal query time limit, as they have very different behavior. The batch query can and should be longer as it does not return all the details, the `to_response` step after the initial query takes much less time compared to the full response. Batch query also takes longer in general in the querying step compared to normal queries.
- Bump up the envoy_connection_buffer_limit_bytes by 10x. I'm not sure what impact this will have, but currently the responses made by the ubuntu query will exceed the previous limit (10mb to 100mb limit).

At 30 seconds, the current query might timeout on a fresh instance (fresh datastore connection, no JIT optimisations yet...etc.). On a warm instance, it took 51 seconds to return a 35mb json response. 

This reduces it down to 20 seconds to more reliably return.

### DIFF
--- a/gcp/api/server.py
+++ b/gcp/api/server.py
@@ -50,7 +50,8 @@ import osv_service_v1_pb2_grpc
 
 _SHUTDOWN_GRACE_DURATION = 5
 
-_MAX_QUERY_TIME = timedelta(seconds=20)
+_MAX_SINGLE_QUERY_TIME = timedelta(seconds=20)
+_MAX_BATCH_QUERY_TIME = timedelta(seconds=35)
 _MAX_BATCH_QUERY = 1000
 # Maximum number of responses to return before applying post exceeded limit
 _MAX_VULN_RESP_THRESH = 3000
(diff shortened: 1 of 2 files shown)


oliverchang (approved):

gcp/api/server.py
```diff
@@ -404,10 +405,14 @@ def page_limit(self) -> int:
...
   page_token: ndb.Cursor | None
-  request_start_time: datetime
+  request_cutoff_time: datetime
   # Use a dataclass to copy by reference
   total_responses: ResponsesCount
 
+  def should_break_page(self, response_count: int):
+    return response_count >= self.total_responses.page_limit(
+    ) or self.request_cutoff_time < datetime.now()
+
```
michaelkedar, line 415: (I know this has already been merged but) nit:
The way the brackets on `page_limit()` has been split across lines is yucky.
Also, IMO, `datetime.now() > self.request_cutoff_time` is a bit easier to understand.

another-rex: Yeah... this is made by the yapf linter to try and get it to fit within the max line width limit. Turns out if you add brackets you can split on multiple lines, I'll try that.

andrewpollock: A good prose docstring would also assist here...

PR #2498, opened 2024-08-20
fix: give way more memory to exporter

michaelkedar:
It's kind of concerning that it's still running out of memory on some Ubuntu exports - we probably should look at *why* it's using so much.
128GB should be way overkill, I just wanted to get a sense of how much it actually uses.

### DIFF
--- a/deployment/clouddeploy/gke-workers/base/exporter.yaml
+++ b/deployment/clouddeploy/gke-workers/base/exporter.yaml
@@ -31,7 +31,7 @@ spec:
                 memory: "16G"
               limits:
                 cpu: "6"
-                memory: "32G"
+                memory: "128G"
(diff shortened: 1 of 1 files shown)


another-rex (approved):
Maybe also bump the request memory up? In case it gets allocated to a node that already has 200GBs used up for some reason.  (Though this is pretty unlikely. )

andrewpollock (approved):

michaelkedar:
> Maybe also bump the request memory up? In case it gets allocated to a node that already has 200GBs used up for some reason. (Though this is pretty unlikely. )

I've bumped the requested memory to 48GB, which (I hope) should be more than enough for the Ubuntu export. I might change it down to the typical used amount later.

PR #2499, opened 2024-08-20
chore: make function line break prettier

another-rex:
Brackets to the rescue! Turns out in python you can add brackets to make line breaks where-ever you want.


michaelkedar (approved):
Thanks for humouring my petty style preferences :smile:

PR #2506, opened 2024-08-21
fix: Add docstring, fix issue with BadArgumentError

another-rex:


oliverchang (approved):

gcp/api/server.py
```diff
@@ -972,9 +977,9 @@ def _query_by_generic_version(
                                                   ecosystem, purl, version,
                                                   False)
 
-  # If no results is because of a page break, then there is no reason to pass further down
-  # as page break would still be in effect with the following queries,
-  # and would have to immediately return.
+  # If no results is because of a page break, then there is no reason
+  # to pass further down  # as page break would still be in effect with
```
andrewpollock, line 981: Remove errant hash

```suggestion
  # to pass further down as page break would still be in effect with
```

andrewpollock (approved):

gcp/api/server.py
```diff
@@ -995,12 +1011,22 @@ def query_by_generic_helper(results: list, cursor, context: QueryContext,
...
     if context.should_break_page(len(results)):
-      cursor = it.cursor_after()
+      # Because this helper function might be called multiple times
+      # we might break before the first result is even queried, raising
+      # a BadArgumentError
+      try:
+        cursor = it.cursor_after()
+      except ndb_exceptions.BadArgumentError:
+        # Don't set the cursor in this case and just return existing cursor
+        pass
```
michaelkedar, line 1029: If the query didn't originally have a `page_token`, will the API not include a page_token in its response, which would look like there are 0 results?

another-rex: Yes you're right, but to fix this requires quite a big reworking of cursor. I'll open a follow-up PR to solve this, and merge this one for now.

PR #2507, opened 2024-08-21
build: repin distroless image dependency

andrewpollock:
The previous hash was resulting in a weird error when pulling the image.

Verified this hash by pulling it locally first.


michaelkedar (approved):
Thanks!

PR #2509, opened 2024-08-22
build: pin to distroless image tagged latest

andrewpollock:
The UX of the UI for [gcr.io/distroless/base](https://gcr.io/distroless/base) is weird. Apparently the latest image is the one tagged with `latest`, which has an update time older than the image reported as most recently updated.

Are you confused? I'm confused.


docker/indexer/Dockerfile
```diff
@@ -24,7 +24,7 @@ RUN go mod download
 COPY ./ /build
 RUN CGO_ENABLED=0 ./build.sh
 
-FROM gcr.io/distroless/base@sha256:5c395fa86ad4eb0613bc9cae7412003f575a25ff9b5cb4845e37e175d2a50bba
+FROM gcr.io/distroless/base@sha256:1aae189e3baecbb4044c648d356ddb75025b2ba8d14cdc9c2a19ba784c90bfb9
```
andrewpollock, line 27: @another-rex reading https://github.com/GoogleContainerTools/distroless/blob/main/README.md#how-do-i-use-distroless-images I'm not even sure this is the right image to be using? Do you have any idea which would be a better choice from that list?

another-rex: I think we can go even smaller and use the static image, since go doesn't require libc, but I think we can stick with the base image since it's already very small. Let's pick the base-debian12 image.

another-rex (approved):

another-rex:
It is a bit weird the image is called distroless, and we immediately tag a distro on it.

andrewpollock:
> It is a bit weird the image is called distroless, and we immediately tag a distro on it.

🤷‍♂️ I just work here...

PR #2514, opened 2024-08-23
feat: Proper working cursor

another-rex:
The current paging/cursor implementation is slightly hacky and fails at various edge cases, and for many query types does not work at all, e.g. (queries without ecosystem, semver queries...). 

When initially implemented, the assumption was almost all API calls resulted in one datastore query. This is no longer the case. 

This refactor/rework adds:
- New cursor type
  -  which can represent 3 states (Start, End, InProgress) in a type safe way (through enums, thanks for the ideas @michaelkedar @DonggeLiu!)
  - Also stores the ndb.Cursor inside
  - Also stores which query number the ndb.Cursor is for when there are multiple queries
  - can serialise to a page token, and deserialise in a backward compatible fashion.
- Reworked logic to centralise cursor handling to make the code more readable and predicable:
  - Cursor state is stored in context, rather than passed through a tuple through return types
  - Query skips are determined next to each `query.iter()` function, where the datastore queries actually happens, rather than in outer functions.
- Context now accounts for the number of datastore queries that has been performed
  - This is used to know which query a cursor belongs to
- Minor refactors to `query_generic_helper()` to remove unnecessary arguments being passed in
- Add a skipIf on env variable to allow actually running the pagination tests locally.
- Modify the tests to run faster (by lowering the sleep duration haha), 
- Modify tests to take in unittest arguments of which tests to run from the second argument onwards (rather than having the gcloud service account always being the last argument)

### DIFF
--- a/gcp/api/cursor.py
+++ b/gcp/api/cursor.py
@@ -1,0 +1,126 @@
+# Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""OSV API server cursor implementation."""
+
+import base64
+from enum import Enum
+from typing import Self
+import typing
+
+from google.cloud import ndb
+import google.cloud.ndb.exceptions as ndb_exceptions
+
+_FIRST_PAGE_TOKEN = base64.urlsafe_b64encode(b'FIRST_PAGE_TOKEN').decode()
+# Use ':' as the separator as it doesn't appear in urlsafe_b64encode
+# (which is what is used for both _FIRST_PAGE_TOKEN, and ndb.Cursor.urlsafe())
+_METADATA_SEPARATOR = ':'
+
+
+class _QueryCursorState(Enum):
+  """
+  Stores the 3 states a query cursor can be in.
+  """
+
+  # The cursor has reached the end, no need to return to the user
+  ENDED = 0
+  # The cursor is at the very start of the query, no cursor needs to be
+  # set when querying ndb
+  STARTED = 1
+  # ndb.Cursor is set and in progress
+  IN_PROGRESS = 2
+
+
+class QueryCursor:
+  """
+  Custom cursor class that wraps the ndb cursor.
+  This cursor should be initialized every time a pageable 
+  query.iter() is called.
+
+  Allows us to represent the "starting" cursor.
+
+  This type could have 3 states encoded in _QueryCursorState.
+  If the current state is IN_PROGRESS, self.cursor will not be None.
+  """
+
+  cursor: ndb.Cursor | None = None
+  cursor_state: _QueryCursorState = _QueryCursorState.ENDED
+  query_number: int = 1
+
+  def __init__(self, query_number: int) -> None:
+    self.query_number = query_number
+    pass
+
+  @classmethod
+  def from_page_token(cls, page_token: str | None) -> Self:
+    """Generate a query cursor from a url safe page token."""
+    qc = cls(0)
+
+    if not page_token:
+      qc.cursor_state = _QueryCursorState.STARTED
+      qc.cursor = None
+      return qc
+
+    split_values = page_token.split(_METADATA_SEPARATOR, 1)
+    if len(split_values) == 2:
+      page_token = split_values[1]
+      try:
+        qc.query_number = int(split_values[0])
+      except ValueError as e:
+        raise ValueError('Invalid page token.') from e
+
+    if not page_token or page_token == _FIRST_PAGE_TOKEN:
+      qc.cursor_state = _QueryCursorState.STARTED
+      qc.cursor = None
+      return qc
+
+    qc.cursor = ndb.Cursor(urlsafe=page_token)
+    qc.cursor_state = _QueryCursorState.IN_PROGRESS
+    return qc
+
+  def update_from_iterator(self, it: ndb.QueryIterator) -> None:
+    try:
+      self.cursor = typing.cast(ndb.Cursor, it.cursor_after())
+      self.cursor_state = _QueryCursorState.IN_PROGRESS
+    except ndb_exceptions.BadArgumentError:
+      self.cursor = None
+      self.cursor_state = _QueryCursorState.STARTED
+
+  def get_cursor(self) -> ndb.Cursor | None:
+    if self.cursor_state == _QueryCursorState.IN_PROGRESS:
+      return self.cursor
+
+    return None
+
+  def ended(self) -> bool:
+    return self.cursor_state == _QueryCursorState.ENDED
+
+  def url_safe_encode(self) -> str | None:
+    """Create a url safe page token to pass back to the API caller"""
+    cursor_part: str = ''
+    match self.cursor_state:
+      case _QueryCursorState.STARTED:
+        cursor_part = _FIRST_PAGE_TOKEN
+      case _QueryCursorState.IN_PROGRESS:
+        # Assume that IN_PROGRESS means self.cursor is always set.
+        # Loudly throw an exception if this is not the case
+        cursor_part = self.cursor.urlsafe().decode()  # type: ignore
+      case _QueryCursorState.ENDED:
+        # If ENDED, we want to return None to not include
+        # a token in the response
+        return None
+
+    if self.query_number == 0:
+      return cursor_part
+
+    return str(self.query_number) + _METADATA_SEPARATOR + cursor_part
--- a/gcp/api/cursor.py
+++ b/gcp/api/cursor.py
@@ -50,10 +50,16 @@ class QueryCursor:
 
   This type could have 3 states encoded in _QueryCursorState.
   If the current state is IN_PROGRESS, self.cursor will not be None.
+
+  Attributes:
+    query_number: This cursor is specifically for the Nth ndb datastore
+      query in the current query request.
+    ndb_cursor: Get the internal ndb_cursor. This could be None.
+    ended: Whether this cursor is for a query that has finished returning data.
   """
 
-  cursor: ndb.Cursor | None = None
-  cursor_state: _QueryCursorState = _QueryCursorState.ENDED
+  _ndb_cursor: ndb.Cursor | None = None
+  _cursor_state: _QueryCursorState = _QueryCursorState.ENDED
   query_number: int = 1
 
   @classmethod
@@ -62,8 +68,8 @@ class QueryCursor:
     qc = cls()
 
     if not page_token:
-      qc.cursor_state = _QueryCursorState.STARTED
-      qc.cursor = None
+      qc._cursor_state = _QueryCursorState.STARTED
+      qc._ndb_cursor = None
       return qc
 
     split_values = page_token.split(_METADATA_SEPARATOR, 1)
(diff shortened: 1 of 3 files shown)


gcp/api/cursor.py
```diff
@@ -0,0 +1,121 @@
...
+      self.cursor = None
+      self.cursor_state = _QueryCursorState.STARTED
+
+  def get_cursor(self) -> ndb.Cursor | None:
+    if self.cursor_state == _QueryCursorState.IN_PROGRESS:
+      return self.cursor
+
+    return None
+
+  def ended(self) -> bool:
```
oliverchang, line 100: nit: make this a `@property` ?

And can you add a docstring?

gcp/api/cursor.py
```diff
@@ -0,0 +1,121 @@
...
+
+  def update_from_iterator(self, it: ndb.QueryIterator) -> None:
+    try:
+      self.cursor = typing.cast(ndb.Cursor, it.cursor_after())
+      self.cursor_state = _QueryCursorState.IN_PROGRESS
+    except ndb_exceptions.BadArgumentError:
+      self.cursor = None
+      self.cursor_state = _QueryCursorState.STARTED
+
+  def get_cursor(self) -> ndb.Cursor | None:
```
oliverchang, line 94: nit: make this a `@property`? And also maybe explicitly call this `ndb_cursor` to differentiate what this is returning compared to the class itself (`QueryCursor`).

gcp/api/cursor.py
```diff
@@ -0,0 +1,121 @@
...
+
+    qc.cursor = ndb.Cursor(urlsafe=page_token)
+    qc.cursor_state = _QueryCursorState.IN_PROGRESS
+    return qc
+
+  def update_from_iterator(self, it: ndb.QueryIterator) -> None:
+    try:
+      self.cursor = typing.cast(ndb.Cursor, it.cursor_after())
+      self.cursor_state = _QueryCursorState.IN_PROGRESS
+    except ndb_exceptions.BadArgumentError:
```
oliverchang, line 90: When can this happen, and why do we do this in such cases? Can you add a comment?

gcp/api/cursor.py
```diff
@@ -0,0 +1,121 @@
...
+
+  The default state is ENDED with no ndb.Cursor.
+
+  This type could have 3 states encoded in _QueryCursorState.
+  If the current state is IN_PROGRESS, self.cursor will not be None.
+  """
+
+  cursor: ndb.Cursor | None = None
+  cursor_state: _QueryCursorState = _QueryCursorState.ENDED
+  query_number: int = 1
```
oliverchang, line 57: Why not 0 as the default value?

gcp/api/cursor.py
```diff
@@ -0,0 +1,121 @@
...
+    qc.cursor_state = _QueryCursorState.IN_PROGRESS
+    return qc
+
+  def update_from_iterator(self, it: ndb.QueryIterator) -> None:
+    try:
+      self.cursor = typing.cast(ndb.Cursor, it.cursor_after())
+      self.cursor_state = _QueryCursorState.IN_PROGRESS
+    except ndb_exceptions.BadArgumentError:
+      self.cursor = None
+      self.cursor_state = _QueryCursorState.STARTED
```
oliverchang, line 92: Is it correct for this to be started when there is no cursor?

gcp/api/cursor.py
```diff
@@ -0,0 +1,121 @@
...
+  Custom cursor class that wraps the ndb cursor.
+  Allows us to represent the "starting" cursor.
+
+  The default state is ENDED with no ndb.Cursor.
+
+  This type could have 3 states encoded in _QueryCursorState.
+  If the current state is IN_PROGRESS, self.cursor will not be None.
+  """
+
+  cursor: ndb.Cursor | None = None
```
oliverchang, line 55: nit: should we make cursor, cursor_state private and prepend them with _?

oliverchang (commented):
nice!

gcp/api/cursor.py
```diff
@@ -0,0 +1,121 @@
...
+  ENDED = 0
+  # The cursor is at the very start of the query, no cursor needs to be
+  # set when querying ndb
+  STARTED = 1
+  # ndb.Cursor is set and in progress
+  IN_PROGRESS = 2
+
+
+class QueryCursor:
+  """
```
andrewpollock, line 45: Could you please add the attributes to the docstring: https://google.github.io/styleguide/pyguide.html#s3.8.4-comments-in-classes

gcp/api/cursor.py
```diff
@@ -0,0 +1,121 @@
...
+_METADATA_SEPARATOR = ':'
+
+
+class _QueryCursorState(Enum):
+  """
+  Stores the 3 states a query cursor can be in.
+  """
+
+  # The cursor has reached the end, no need to return to the user
+  ENDED = 0
```
andrewpollock, line 36: It seems to be pretty common to make the default value something invalid to catch unexpected unitialization: https://cloud.google.com/apis/design/design_patterns#enum_default_value

gcp/api/cursor.py
```diff
@@ -0,0 +1,121 @@
...
+
+  def url_safe_encode(self) -> str | None:
+    """Create a url safe page token to pass back to the API caller"""
+    cursor_part: str = ''
+    match self.cursor_state:
+      case _QueryCursorState.STARTED:
+        cursor_part = _FIRST_PAGE_TOKEN
+      case _QueryCursorState.IN_PROGRESS:
+        # Assume that IN_PROGRESS means self.cursor is always set.
+        # Loudly throw an exception if this is not the case
```
andrewpollock, line 111: If I understand this comment correctly, you're relying on `self.cursor.urlsafe().decode()` to raise an exception? Can you please expand the docstring to mention this? https://google.github.io/styleguide/pyguide.html#s3.8.3-functions-and-methods

gcp/api/cursor.py
andrewpollock: I'm guessing this is because the first query is `query_number = 1`? Count from zero or count from one?

gcp/api/server.py
```diff
@@ -739,11 +767,17 @@ def to_response(b):
       bugs[i].related[:] = sorted(
           list(set(related_bug_ids + list(bugs[i].related))))
 
-  if next_page_token:
-    next_page_token = next_page_token.urlsafe()
+  if context.query_counter < context.input_cursor.query_number:
+    # If the input cursor is for a query number that's greater than
+    # the number of queries performed, the cursor must be invalid
+    # (and there will be no results, as everything is skipped)
+    raise ValueError("Cursor is invalid/does not belong to this query")
```
andrewpollock, line 774: Do you want to include the `query_counter` and `input_cursor.query_number` values here to aid in any future debugging? (Or a log entry with them prior to raising the exception?)

gcp/api/server.py
```diff
@@ -781,21 +815,23 @@ def _datastore_normalized_purl(purl: PackageURL):
...
 
 @ndb.tasklet
-def query_by_commit(
-    context: QueryContext,
-    commit: bytes,
-    to_response: Callable = bug_to_response) -> tuple[list, ndb.Cursor]:
+def query_by_commit(context: QueryContext,
+                    commit: bytes,
+                    to_response: Callable = bug_to_response) -> list:
   """Query by commit."""
```
andrewpollock, line 821: Please improve this docstring while you're here: https://google.github.io/styleguide/pyguide.html#s3.8.3-functions-and-methods

gcp/api/server.py
```diff
@@ -935,17 +971,22 @@ def _query_by_semver(context: QueryContext, query: ndb.Query, package_name: str,
                      ecosystem: str, purl: PackageURL | None, version: str):
   """Query by semver."""
```
andrewpollock, line 972: Please improve this docstring while you're here: https://google.github.io/styleguide/pyguide.html#s3.8.3-functions-and-methods

gcp/api/server.py
```diff
@@ -967,66 +1008,51 @@ def _query_by_generic_version(
     version: str,
 ):
   """Query by generic version."""
```
andrewpollock, line 1010: Please improve this docstring while you're here: https://google.github.io/styleguide/pyguide.html#s3.8.3-functions-and-methods

gcp/api/server.py
```diff
@@ -967,66 +1008,51 @@ def _query_by_generic_version(
...
+  # Try without normalizing.
+  results = yield query_by_generic_helper(context, base_query, project,
+                                          ecosystem, purl, version, False)
 
-  results = []
+  # If there is results, then we should return with this query,
```
andrewpollock, line 1015: Nit: Grammar

```suggestion
  # If there are results, then we should return with this query,
```

gcp/api/server.py
```diff
@@ -1169,13 +1196,12 @@ def _query_by_comparing_versions(context: QueryContext, query: ndb.Query,
...
-  return bugs, cursor
+  return bugs
 
 
 @ndb.tasklet
 def query_by_package(context: QueryContext, package_name: str, ecosystem: str,
-                     purl: PackageURL | None,
-                     to_response: Callable) -> tuple[list, ndb.Cursor]:
+                     purl: PackageURL | None, to_response: Callable) -> list:
   """Query by package."""
```
andrewpollock, line 1205: Please improve this docstring while you're here: https://google.github.io/styleguide/pyguide.html#s3.8.3-functions-and-methods

andrewpollock (commented):
This is awesome

gcp/api/cursor.py
another-rex: It's a bit difficult to document why it's 1 here in a comment. It essentially makes another completely separate part of the code a bit easier to write. 

We call `context.should_skip_query()` every time we do a query, and we also want to increment the context.query_counter by 1 no matter what should_skip_query() returns, so we do the increment before calling should_skip_query()

`should_skip_query()` also checks the query_counter, so rather than having - 1 in a few separate places, I just made the query number start at 1.

another-rex: Now looking at it, maybe I should move the query_counter increment into should_skip_query(), since they are always paired together, but it also feels weird to be mutating the state in a "question" method.

gcp/api/cursor.py
another-rex: I think this case qualifies for the exception:
> An idiomatic name may be used for the 0 value

ENDED is equivalent to being None (currently in production it is represented just as None). This Enum is also private, and can only be used by the QueryCursor implementation, so I think it's alright to keep it as is.

gcp/api/cursor.py
another-rex: It will  raise an ArgumentError since urlsafe() will not exist on None. I'm not sure if I should be documenting this in the docstring since it's an implementation/programmer error of QueryCursor if that is ever the case. This is not something the caller should be accounting for, as it should never happen.

> You should not document exceptions that get raised if the API specified in the docstring is violated (because this would paradoxically make behavior under violation of the API part of the API).

gcp/api/server.py
another-rex: Good idea, added an error log.

gcp/api/cursor.py
another-rex: Good idea! done both.

gcp/api/cursor.py
another-rex: Added a comment:

```
      # This exception can happen when iterator has not begun iterating.
      # and it.next() is the very first element.
      #
      # In those cases, `cursor_after()`` would not be 'after' any element,
      # throwing this exception.

      # We represent this by setting the state to STARTED.
```

gcp/api/cursor.py
another-rex: made private

gcp/api/cursor.py
another-rex: Done!

gcp/api/server.py
another-rex: Added a bunch of docs to all those functions, though I feel like I probably missed some exceptions that could be raised. Also fixed the typing a bit on the arguments.

gcp/api/server.py
```diff
@@ -1080,68 +1181,81 @@ def query_by_version(context: QueryContext,
...
-  """Query by package."""
+                                 ecosystem: str, version: str) -> list:
+  """
+  Query by comparing versions.
+
+  Args:
+    context: QueryContext for the current query.
+    query: A partially completed ndb.Query object which only needs 
+      version filters to be added before query is performed.
+    ecosystem: Optional ecosystem of the package to query.
```
hogo6002, line 1238: For comparing_versions, `ecosystem` is required. It needs `ecosytem.sort_key()` to compare versions.

gcp/api/cursor.py
```diff
@@ -0,0 +1,148 @@
...
+        cursor_part = self._ndb_cursor.urlsafe().decode()  # type: ignore
+      case _QueryCursorState.ENDED:
+        # If ENDED, we want to return None to not include
+        # a token in the response
+        return None
+
+    if self.query_number == 0:
```
oliverchang, line 145: When is this ever true, given https://github.com/google/osv.dev/pull/2514/files#r1730697012 ?

gcp/api/cursor.py
oliverchang: IMO we shouldn't be coupling implementation details of another part of the code with design decisions here. 

+1 to moving the query_counter increment to `should_skip_query`. This way we also avoid the -1 right? 

Re naming, we could rename it to something like `preprocess_and_check_subquery()`, and invert the return value so it'll look something like:


```python
if not context.preprocess_and_check_subquery():
  return []
```

WDYT?

oliverchang (approved):
LGTM with some minor remaining comments. Awesome stuff!

gcp/api/cursor.py
```diff
@@ -0,0 +1,148 @@
...
+from google.cloud import ndb
+import google.cloud.ndb.exceptions as ndb_exceptions
+
+_FIRST_PAGE_TOKEN = base64.urlsafe_b64encode(b'FIRST_PAGE_TOKEN').decode()
+# Use ':' as the separator as it doesn't appear in urlsafe_b64encode
+# (which is what is used for both _FIRST_PAGE_TOKEN, and ndb.Cursor.urlsafe())
+_METADATA_SEPARATOR = ':'
+
+
+class _QueryCursorState(Enum):
```
michaelkedar, line 30: nit: Since it's the only thing that uses it , this could possibly go inside of `QueryCursor`:
```python
class QueryCursor:
  class _State(Enum):
    ENDED = 0
    # ...
    
  _cursor_state: _State = _State.ENDED

  @property
  def ended(self) -> bool:
    return self._cursor_state == QueryCursor._State.ENDED
````

That said, I don't think I have a preference either way.

another-rex: I think it's just a bit clearer to be private but at the top level to avoid having to write the class name repeatedly.

gcp/api/cursor.py
another-rex: Ah that should be 1

gcp/api/cursor.py
another-rex: Yeah I'll do that!

The most "correct" version of this is probably a function decorator which does the skipping and counting, but I think that disconnects the .iter() too much from the increment of the query counter.

another-rex: Hmm it'll avoid the -1 in that specific instances, but we'll still need -1 for two other places. I think it'll be clearer to define the query number as incrementing once before the start of the query, so everything that uses the query number can know that the first query is numbered "1". Rather than having inconsistent query number matching. I'll add that to the comments to make it clearer. 

I'll make the change to combine the +1 with the skip check though. 

In fact making this change made me aware of the need to check the query number when retrieving the ndb cursor which I am not currently doing.

gcp/api/cursor.py
another-rex: Actually removed this entirely, there is no reason to not return this as part of the page token.

gcp/api/cursor.py
another-rex: After implementing it, I think it's still much neater/clearer to keep it as two separate lines, as it avoids mutation happening inside an if statement, and clearly shows that the query number increments before anything related to the query begins.

gcp/api/cursor.py
andrewpollock: https://www.xkcd.com/2200/😆

gcp/api/cursor.py
oliverchang: SGTM

gcp/api/server.py
another-rex: Updated

PR #2515, opened 2024-08-23
fix: handle invalid version input in ecosystem sort_key

hogo6002:
Some ecosystem `sort_key` functions do not validate versions before converting them to their ecosystem-specific formats. This leads to several errors occur when using `sort_key` to compare user-input versions with affected ranges.

Add version validation before conversion to ensure no errors occur.


andrewpollock (approved):

another-rex (approved):
LGTM!

PR #2517, opened 2024-08-23
fix(nvd-mirror): use a more efficient technique to split NVD by year

andrewpollock:
To address apparent overrun issues (and also hopefully reduce the amount of RAM the job requires) use a more time and memory efficient technique to split the monolithic NVD dump by year.

This avoids creating multiple in-memory arrays of the same data

Refactor invocation for better readability


michaelkedar (approved):
I'm not too familiar with jq but LGTM

andrewpollock:
> I'm not too familiar with jq

I've manually built a container and hacked up a Kubernetes job YAML to run it in Staging currently and it is flying along using < 1% of the RAM allocated, by the looks of things :-)

PR #2518, opened 2024-08-23
feat(tools): add API performance test tool

hogo6002:
API perf test script:
- Retrieve all bugs from the datastore and write them into JSON files.
- Read bugs from the JSON files and generate queries randomly.
- Send the generated queries asynchronously to the test instance.


tools/api-performance-test/performance.py
```diff
@@ -0,0 +1,333 @@
...
+
+async def main():
+  """Main"""
+  if not os.path.exists(BUG_DIR):
+    # This will take around 10 mins
+    get_bugs_from_datastore()
+
+  # The `ecosystem_map` can be used to filter our queries for a
+  # specific ecosystem.
+  ecosystem_map, bug_map, project_map = load_all_bugs(BUG_DIR)
```
another-rex, line 305: I know it's called 'project' in the datastore, but I think it'll be clearer if we name this package instead of project (project makes me think of the gcp project.)

andrewpollock (changes requested):
This looks really cool!

To enhance the readability and maintainability of this, could you please:

- Expand the docstrings to include arguments and returns per https://google.github.io/styleguide/pyguide.html#s3.8.3-functions-and-methods
- Add type hints per https://google.github.io/styleguide/pyguide.html#s2.21-type-annotated-code

tools/api-performance-test/performance.py
```diff
@@ -0,0 +1,333 @@
...
+  vuln_query_ids = list(bug_map.keys())
+  package_query_ids = []
+  for project in project_map:
+    # Tests each project once.
+    package_query_ids.append(project_map[project].pop())
+  random.shuffle(package_query_ids)
```
another-rex, line 311: Generate a random seed at the top and use that fixed seed to do the shuffling, then print out / save the seed number to a file. This way if you encounter any weird behaviour you can replicate the order of the queries that was  sent.

tools/api-performance-test/performance.py
```diff
@@ -0,0 +1,333 @@
...
+
+
+def package_payload(request_id, bug_map):
+  """package/project query payload"""
+  package = random.choice(bug_map[request_id].projects)
+  ecosystem = random.choice(bug_map[request_id].ecosystems)
+  return {"package": {"name": package, "ecosystem": ecosystem}}
+
+
+def version_payload(request_id, bug_map):
```
another-rex, line 211: nit: Rename these payload generator functions to `build_version_payload` to more easily differentiate them from just a variable of the payload itself.

tools/api-performance-test/performance.py
```diff
@@ -0,0 +1,333 @@
...
+      self.ecosystems = list(bug_dict['ecosystem'])
+
+    # Use the `affected fuzzy` value as the query version.
+    # If no 'affected fuzzy' is present, assign a default value.
+    self.affected_fuzzy = bug_dict['affected_fuzzy']
+    if not self.affected_fuzzy:
+      self.affected_fuzzy = '1.0.0'
+
+
+def format_bug_for_output(bug):
```
another-rex, line 54: Can you add some type hints to function arguments and return values

another-rex (commented):
Looks pretty good!

It would be good to add a script to generate the JSON file as well, and at least add a sample JSON file with a few entries so users know how to generate their own JSON file to test with.

tools/api-performance-test/performance.py
hogo6002: Good idea, updated. Thanks!

tools/api-performance-test/performance.py
```diff
@@ -0,0 +1,518 @@
...
+        package_map[package].add(bug.db_id)
+      bug_map[bug.db_id] = bug
+
+
+def load_all_bugs() -> tuple[defaultdict, dict, defaultdict]:
+  """Loads bugs from JSON directory
+
+  Returns:
+    A defaultdict mapping ecosystem names to their bugs. For example:
+    {'Maven': (CVE-XXXX-XXXX, CVE-XXXX-XXXX), 'PyPI': ()}
```
another-rex, line 173: Is this suppose to look like:
```
{'Maven': ['CVE-XXXX-XXXX', 'CVE-XXXX-XXXX'], 'PyPI': []}
```

(Square brackets and quotes)?

tools/api-performance-test/performance.py
```diff
@@ -0,0 +1,518 @@
...
+      self.ecosystems = list(bug_dict['ecosystem'])
+
+    # Use the `affected fuzzy` value as the query version.
+    # If no 'affected fuzzy' is present, assign a default value.
+    self.affected_fuzzy = bug_dict['affected_fuzzy']
+    if not self.affected_fuzzy:
+      self.affected_fuzzy = '1.0.0'
+
+
+def format_bug_for_output(bug) -> dict[str:any]:
```
another-rex, line 60: pretty sure this should be `dict[str, any]`

another-rex (changes requested):
Just need to add the function argument types, and it should be good to merge! 

Could be useful to store the results to use to verify behavior changes, but that can be a follow up later.

tools/api-performance-test/performance.py
hogo6002: Updated

another-rex (approved):

andrewpollock (approved):

PR #2520, opened 2024-08-23
test: disable Debian ecosystem tests

andrewpollock:
snapshot.debian.org has an expired certificate that is causing intermittent failures depending on which IPv4 address is used.

https://bugs.debian.org/1079320


hogo6002 (approved):
Thanks Andrew!

PR #2527, opened 2024-08-26
feat(nvd-mirror): run the year splitting in parallel

andrewpollock:
This greatly speeds up the completion of this final stage of the mirroring and based on manual testing in Staging, comfortably fits within the existing RAM allowances for the job.


another-rex (approved):
TIL the wait command exists!

Issue #2401, opened 2024-07-18
Support API queries for non-enumerated ecosystems


hogo6002:
Enable OSV.dev API queries for specific versions in non-enumerated ecosystems (e.g., Ubuntu, Rocky Linux, AlmaLinux).

```
curl -d '{"package": {"name": "curl", "ecosystem": "Ubuntu"}, "version": "8.2.1-1ubuntu3.1"}' http://api.osv.dev/v1/query
```

oliverchang:
I think we can close this as done @hogo6002  ?

PR #2530, opened 2024-08-27
feat: support non-enumerated ecosystem version queries on prod instance

hogo6002:
Issue: https://github.com/google/osv.dev/issues/2401
Enable API queries for specific versions in non-enumerated ecosystems (Rocky Linux, AlmaLinux, Chainguard, and Wolfi) on the production instance. This change will not affect queries for other ecosystems.


another-rex (approved):
LGTM!

PR #2536, opened 2024-08-29
fix(datasources): correct Android ignore_pattern

andrewpollock:
Correctly permit all expected prefixes (PUB-A was being excluded in error).

Reorder and harmonize so that a side-by-side vimdiff or other comparison more easily surfaces inconsistencies and omissions.

**_Notable retained inconsistency: `rockylinux-rlsa` (prod) vs `rockylinux` (staging), as I think a wholesale source name change will be disruptive?_**

### DIFF
--- a/source_test.yaml
+++ b/source_test.yaml
@@ -193,7 +193,7 @@
 - name: 'malicious-packages'
   versions_from_repo: False
   type: 0
-  ignore_patterns: ['^(?!MAL-).*$']
+  ignore_patterns: ['MAL-0000.*', '^(?!MAL-).*$']
   directory_path: 'osv'
   repo_url: 'https://github.com/ossf/malicious-packages.git'
   detect_cherrypicks: False
--- a/source_test.yaml
+++ b/source_test.yaml
@@ -44,7 +44,7 @@
 - name: 'android'
   versions_from_repo: False
   type: 1
-  ignore_patterns: ['^(?!(PUB-)?ASB-A-).*$']
+  ignore_patterns: ['^(?!(ASB|PUB|A))-.*$']
   detect_cherrypicks: False
   extension: '.json'
   bucket: 'android-osv-test'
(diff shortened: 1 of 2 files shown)


source_test.yaml
```diff
@@ -209,13 +233,14 @@
   type: 0
   ignore_patterns: ['PYSEC-0000.*', '^(?!PYSEC-).*$']
   directory_path: 'vulns'
-  repo_url: 'https://github.com/pypa/advisory-database.git'
+  repo_url: 'ssh://github.com/pypa/advisory-database'
```
oliverchang, line 236: Should we use https for both prod and test? I'm not sure ssh creds exist in test... and they're not necessary in any case for this source.

source_test.yaml
```diff
@@ -181,7 +193,7 @@
 - name: 'malicious-packages'
   versions_from_repo: False
   type: 0
-  ignore_patterns: ['^(?!MAL-).*$', 'MAL-0000.*']
+  ignore_patterns: ['^(?!MAL-).*$']
```
oliverchang, line 196: should we bring back the MAL-0000 ? I think that's still needed.

oliverchang (changes requested):

doryiii:
Thanks Andrew!!

source_test.yaml
```diff
@@ -191,6 +203,18 @@
   link: 'https://github.com/ossf/malicious-packages/blob/main/'
   editable: False
 
+- name: 'test-oss-fuzz'
```
oliverchang, line 206: oss-fuzz won't work for test because the necessary creds aren't there. OSS-Fuzz is also a rare source where OSV pushes changes, so we definitely can't have both prod and test do this.

source_test.yaml
andrewpollock: I'm happy to harmonise in that direction, I had biased towards what was in Production as the base.

source_test.yaml
andrewpollock: This was purely a reordering of what was already present, so should this be removed entirely?

oliverchang: Ah, I see that `editable` is `False` here, which is what controls the write. Let's leave this in if this was just a reorder.

source_test.yaml
andrewpollock: Done in 9256d75d

source_test.yaml
andrewpollock: From memory, we found that having multiple items in this list did not achieve the desired effect. I now note that the `python` source is still in this state (in the reverse order) so I'll need to check how that's working out... I think the regex needs to be merged into a single one (in both cases).

oliverchang: Hmm, I think it should work as intended in this case, because we ignore the file if **any** of the patterns in the list match. 

MAL-0000- (and PYSEC-0000-) entries will exist and represent intermediate advisories that don't yet have a property ID assigned, so we need to ignore them.

oliverchang: ref: https://github.com/google/osv.dev/blob/34bd002b44baf7f0752614c43a1151f6f2aeca2d/osv/models.py#L855

andrewpollock: Ack. a5419a21 adjusts accordingly.

oliverchang (approved):

PR #2539, opened 2024-08-30
fix: Fix android ignore pattern subtle regex error

another-rex:
One of the patterns is just slightly different than the others 🥷 , complete coincidence that android also added some new irrelevant json files.


andrewpollock (approved):
https://xkcd.com/1171/ strikes again.

Issue #2189, opened 2024-05-09
OSV record providers have a machine-readable way to reason about existing published records that do not meet this quality bar


andrewpollock:
# Persona

Database operator publishing OSV records imported by OSV.dev, listed in [`source.yaml`](https://github.com/google/osv.dev/blob/master/source.yaml) (e.g. `ghsa`)

# Intended user journey

- Visit `https://api.osv.dev/v1/importfindings/{source}` (e.g. `https://api.osv.dev/v1/importfindings/ghsa`)
  - Receive JSON output in the format defined by [`osv/feedback`](https://github.com/google/osv.dev/blob/master/osv/feedback.json), describing which records are failing to be imported and why

This data can be incorporated into whatever presentation formats desired by the data publisher (any sort of "dashboard" is out of scope of this specific issue)

## Example

```
curl https://api.osv.dev/v1/importfindings/ghsa

{
  "invalid_records": [
    {
      "id": "GHSA-78hx-gp6g-7mj6",
      "finding_first_seen": "2024-05-03T08:15:00Z",
      "import_last_attempt": "2024-10-22T08:15:00Z",
      "findings": [
        "INVALID_RANGE"
      ]
    }
  ]
}
```

another-rex:
There is a related existing import status output in the bucket here: https://storage.googleapis.com/osv-public-import-logs/

andrewpollock:
> There is a related existing import status output in the bucket here: https://storage.googleapis.com/osv-public-import-logs/

Yeah, from memory, these get overwritten on each run of the Importer, so there's no long-term state for a given record.

andrewpollock:
Basic functionality exists:

```
$ for source in $(poetry run python tools/source-sync/source_sync.py --source_file source.yaml --list | tail -n +3 | awk '{ print $1 }'); do echo $source $(GET https://api.osv.dev/v1experimental/importfindings/
${source})
done
almalinux-alba {}
almalinux-alea {}
almalinux-alsa {}
android {}
bitnami {}
chainguard {}
curl {}
cve-osv {}
debian-dla {}
debian-dsa {}
debian-dtsa {}
ghsa {}
go {}
haskell {}
mageia {}
malicious-packages {}
oss-fuzz {}
psf {}
python {}
r {}
redhat {}
rockylinux-rlsa {}
rockylinux-rxsa {}
rust {}
suse {}
[...]

PR #2541, opened 2024-09-02
feat: machine readable record feedback JSON schema

andrewpollock:
First iteration of the JSON schema for encapsulating machine-readable feedback on individual OSV records.

Part of #2189

### DIFF
--- a/osv/feedback.json
+++ b/osv/feedback.json
@@ -71,13 +71,17 @@
                     "id": "GSD-2021-1001229",
                     "finding_first_seen": "2024-04-01T08:15:00Z",
                     "import_last_attempt": "2024-05-03T08:15:00Z",
-                    "findings": "INVALID_JSON"
+                    "findings": [
+                        "INVALID_JSON"
+                    ]
                 },
                 {
                     "id": "GSD-2022-1003535",
                     "finding_first_seen": "2024-04-01T08:15:00Z",
                     "import_last_attempt": "2024-05-03T08:15:00Z",
-                    "findings": "DELETED"
+                    "findings": [
+                        "DELETED"
+                    ]
                 }]
             ]
         }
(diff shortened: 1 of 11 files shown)


osv/feedback.json
```diff
@@ -0,0 +1,99 @@
...
+                }]
+            ]
+        }
+    },
+    "examples": [{
+        "invalid_records": [{
+            "id": "GSD-2021-1001229",
+            "finding_first_seen": "2024-04-01T08:15:00Z",
+            "import_last_attempt": "2024-05-03T08:15:00Z",
+            "findings": "INVALID_JSON"
```
another-rex, line 90: These examples seem invalid, this should be an array right?

osv/feedback.json
```diff
@@ -0,0 +1,99 @@
...
+                        "type": "string",
+                        "title": "The id of the OSV record",
+                        "examples": [
+                            "GSD-2021-1001229",
+                            "GSD-2022-1003535"
+                        ]
+                    },
+                    "import_last_attempt": {
+                        "type": "string",
+                        "title": "When an import for this record was last attempted in RFC3339",
```
another-rex, line 35: nit: Should we use ISO8601 instead?, it seems to be the more popular format from what I understand. (I think they are both very similar)

e.g. python has a toisoformat and fromisoformat which uses 8601. but I know golang only has RFC3339.

Apparently neither is a superset of the other...  https://ijmacd.github.io/rfc3339-iso8601/

osv/feedback.json
andrewpollock: Yes, I did an incomplete fix up of this late breaking change. Corrected in [8a7ca49](https://github.com/google/osv.dev/pull/2541/commits/8a7ca49d5dd223558f5d9a06fa3b797497f46936)

osv/feedback.json
andrewpollock: Yes, RFC3339 is indeed an annoying variant. https://xkcd.com/927/
I went with RFC3339 for consistency with the rest of the OSV Schema.

another-rex (approved):

PR #2542, opened 2024-09-03
fix(api): add package name check

hogo6002:
Add package name check before the affected version check, and filter out unmatched packages.

### DIFF
--- a/gcp/api/server.py
+++ b/gcp/api/server.py
@@ -1203,8 +1203,8 @@ def query_by_version(context: QueryContext,
           bugs.append(bug)
     elif supports_comparing:
       # Query for non-enumerated ecosystems.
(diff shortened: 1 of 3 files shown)


gcp/api/server.py
```diff
@@ -1231,14 +1231,16 @@ def query_by_version(context: QueryContext,
...
+                                 package_name: str, ecosystem: str,
+                                 version: str) -> list:
   """
   Query by comparing versions.
 
   Args:
     context: QueryContext for the current query.
     query: A partially completed ndb.Query object which only needs 
       version filters to be added before query is performed.
+    package_name: The queried package.
```
another-rex, line 1243: Mention that this is required as well in the comment.

michaelkedar (approved):

another-rex (approved):

Issue #2482, opened 2024-08-16
Combine all `not yet assigned` Debian Security Tracker issues into OSV


hogo6002:
OSV [doesn't combine issues ](https://github.com/google/osv.dev/blob/02575ef484fe7fd5b88e38f347163062d53df30c/vulnfeeds/cmd/debian/main.go#L114)with the `not yet assigned` urgency tag from the Debian Security Tracker because we thought the status of those issues might change soon. But we've observed that many vulns remain in this stage for a very long time, and it doesn't seem like their status will change.

Confirmed with the Debian Security Team via email that they no longer assign urgencies to security issues, only some legacy ones with low/medium/high tags. For all issues without an assigned urgency, we should treat them as regular security issues and merge them into OSV. 

A very large number of issues have `not yet assigned` tag (around 10k), but not all are valid entries (some may not affect any release)

carnil:
FWIW, we have updated our documentation for the current practice with https://salsa.debian.org/security-tracker-team/security-tracker/-/commit/ee436d666e0fb8cf845de22b012efe539d13d116

PR #2566, opened 2024-09-04
feat: combine more Debian security tracker data into OSV

hogo6002:
Combine Debian security issues without an urgency tag into OSV. Details: https://github.com/google/osv.dev/issues/2482

This will significantly increase entries to combine-to-osv, and should be merged only after [SDK performance testing](https://github.com/google/osv.dev/pull/2546).


andrewpollock (approved):
LGTM. I'll plan to conclude performance testing on Friday, and either roll the SDK further forward or back to the known-good version.

Issue #2543, opened 2024-09-03
Ingest SUSE/openSUSE OSV advisories


oliverchang:
Per https://github.com/ossf/osv-schema/issues/259, the SUSE/openSUSE ecosystems have been added to the OSV schema, and there is a feed available at https://ftp.suse.com/pub/projects/security/osv/

oliverchang:
For version ordering, SUSE seems to use RPM (https://en.opensuse.org/openSUSE:Package_versioning_guidelines).

For https://ftp.suse.com/pub/projects/security/osv/, we may need an `https://ftp.suse.com/pub/projects/security/osv/all.json` per https://google.github.io/osv.dev/rest-api-contribution/#1-a-url-pointing-to-a-rest-endpoint-containing-at-least-all-of-the-vulnerabilities-ids-and-date-modified. @msmeissn is this something that would be possible to add?

msmeissn:
i added an all.json file now.

oliverchang:
Thanks @msmeissn ! 

@hogo6002 can you see if we can start ingesting this into our test instance?

hogo6002:
Hey @msmeissn, I have a question about the data prefixes and would like some clarification.

The ossf schema lists the valid prefixes for [SUSE as only `SUSE-SU-`](https://ossf.github.io/osv-schema/#id-modified-fields), but I noticed that `all.json` also contains entries other than `SUSE-SU-` (security updates), such as `SUSE-RU-` (recommended updates) and `SUSE-FU-` (feature updates). 

I just want to confirm if we only want to ingest data with the `SU-` prefix into OSV or if we want to ingest all the data from `all.json` into OSV.

Also, I'm wondering if entries with other prefixes, such as `SUSE-RU-` and `SUSE-FU-`, are actually security-related data (they do have related CVEs on their records)?

msmeissn:
ok, problem is that occasionaly "recommended (bugfix)" or feature updates also include CVEs and I generate entries for those. Note that only bugfix or freature updates that have CVEs will be reported for OSV.

I sent a PR to osv-schema to allow them, so we can consume those.

hogo6002:
> I sent a PR to osv-schema to allow them, so we can consume those.

Thanks! I will start to ingest data into our test instance.

hogo6002:
SUSE/openSUSE data is available on **test.osv.dev** now!
[...]

PR #2570, opened 2024-09-05
feat: add SUSE/openSUSE source to test instance

hogo6002:
Add SUSE and openSUSE source (around 20k records), details: https://github.com/google/osv.dev/issues/2543
SUSE prefixes: `SUSE-SU-` (most records), `SUSE-OU-`, `SUSE-FU-` and `SUSE-RU-`
openSUSE prefixes: `openSUSE-SU-`

Merge after https://github.com/google/osv.dev/pull/2571

### DIFF
--- a/source_test.yaml
+++ b/source_test.yaml
@@ -204,19 +204,6 @@
   link: 'https://github.com/ossf/malicious-packages/blob/main/'
   editable: False
 
-  - name: 'open-suse'
-  versions_from_repo: False
-  rest_api_url: 'https://ftp.suse.com/pub/projects/security/osv/all.json'
-  type: 2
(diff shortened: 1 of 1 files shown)


michaelkedar (commented):
I think you might need to add SUSE/openSUSE  to `_ecosystems.py` (at least as `OrderingUnsupportedEcosystem`) before we start ingesting the records. I'm pretty sure there's a check somewhere to prevent ecosystems we're unaware of from being added.

Ideally, add the ecosystems first in a separate PR, because `sync-source-repo` will finish before the changes are deployed to the importer/workers

michaelkedar (approved):
Just let #2571 finish `build-and-stage` before merging

PR #2571, opened 2024-09-05
feat: add SUSE/openSUSE ecosystem

hogo6002:
add SUSE/openSUSE ecosystem.

### DIFF
--- a/osv/ecosystems/_ecosystems.py
+++ b/osv/ecosystems/_ecosystems.py
@@ -65,8 +65,10 @@ _ecosystems = {
     'ConanCenter': OrderingUnsupportedEcosystem(),
     'GitHub Actions': OrderingUnsupportedEcosystem(),
     'Linux': OrderingUnsupportedEcosystem(),
+    'openSUSE': OrderingUnsupportedEcosystem(),
     'OSS-Fuzz': OrderingUnsupportedEcosystem(),
     'Photon OS': OrderingUnsupportedEcosystem(),
+    'SUSE': OrderingUnsupportedEcosystem(),
 }
 
 # Semver-based ecosystems, should correspond to _ecosystems above.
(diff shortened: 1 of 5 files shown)


another-rex (approved):

osv/ecosystems/_ecosystems.py
```diff
@@ -65,8 +65,10 @@
     'ConanCenter': OrderingUnsupportedEcosystem(),
     'GitHub Actions': OrderingUnsupportedEcosystem(),
     'Linux': OrderingUnsupportedEcosystem(),
+    'openSUSE': OrderingUnsupportedEcosystem(),
     'OSS-Fuzz': OrderingUnsupportedEcosystem(),
     'Photon OS': OrderingUnsupportedEcosystem(),
+    'SUSE': OrderingUnsupportedEcosystem(),
 }
 
 # Semver-based ecosystems, should correspond to _ecosystems above.
```
oliverchang, line 74: Do we need to modify get() as well?

https://github.com/google/osv.dev/blob/7b948b1d877be849dfb65626caf85e59ac964b28/osv/ecosystems/_ecosystems.py#L111

Also this map looks out of date. E.g. Rocky Linux should be returning `RockyLinux` instead? 

For the actual SuSE ecosystem, I think we can just go ahead and add it properly now since it just uses RPM ordering under the hood.

hogo6002: Updated the map and added sorting for SUSE.

oliverchang (approved):

osv/ecosystems/_ecosystems.py
```diff
@@ -130,6 +127,12 @@ def get(name: str) -> Ecosystem:
   if name.startswith('Ubuntu'):
     return Ubuntu()
 
+  if name.startswith('openSUSE'):
+    return OpenSUSE()
+
+  if name.startswith('SUSE'):
```
michaelkedar, line 133: nit: I think these checks should include the `:`, since it seems likes it's required.

michaelkedar (approved):

osv/ecosystems/_ecosystems.py
hogo6002: If we include `:`, then all the API quires will have to include the release name such as "SUSE:Linux Enterprise Software Development Kit 11 SP4". I think it's not ideal.

PR #2574, opened 2024-09-06
fix(importer): update log message with error details

hogo6002:
Add error message into the log for better debugging


another-rex (approved):

another-rex:
You might also want to print out the type of the exception. E.g. `e.__class__.__name__`. See https://stackoverflow.com/questions/9823936/how-do-i-determine-what-type-of-exception-occurred

docker/importer/importer.py
```diff
@@ -700,8 +700,9 @@ def _process_updates_rest(self, source_repo: osv.SourceRepository):
         # No need to log a full error, as this is expected result.
         logging.info('Entry does not have an OSV entry: %s', vuln.id)
         continue
-      except Exception:
-        logging.error('Failed to parse %s', vuln.id)
+      except Exception as e:
+        logging.error('Failed to parse %s: error type: %s, details: %s',
```
andrewpollock, line 704: https://docs.python.org/3/library/logging.html#logging.Logger.exception might be more appropriate here?

```suggestion
        logging.excecption('Failed to parse %s: error type: %s, details: %s',
```

andrewpollock (approved):
LGTM with a potential alternative...

PR #2579, opened 2024-09-09
chore(vulnfeeds): pin to Cloud SDK 485.0.0

andrewpollock:
Harmonise explicitly on 485.0.0 until GCS performance regression is resolved


go.work
hogo6002, line 1: I'm curious why would we have this `go.work` file at the first place if it does nothing.

hogo6002 (approved):
LGTM!

go.work
andrewpollock: https://www.reddit.com/r/golang/comments/15v8i06/why_gowork_file_should_not_be_committed/ talks about it in more detail

another-rex: Do you want to add go.work file to the gitignore?

PR #2605, opened 2024-09-11
fix(sources): switch oss-fuzz to https

andrewpollock:
The staging instance has no SSH credentials and we aren't writing anything back so don't need to authenticate

### DIFF
--- a/source_test.yaml
+++ b/source_test.yaml
@@ -209,6 +209,7 @@
   type: 0
   ignore_patterns: ['^(?!OSV-).*$']
   directory_path: 'vulns'
+  # deliberately HTTPS due to lack of SSH credentials in Staging.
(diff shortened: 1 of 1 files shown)


source_test.yaml
```diff
@@ -209,7 +209,7 @@
   type: 0
   ignore_patterns: ['^(?!OSV-).*$']
   directory_path: 'vulns'
-  repo_url: 'ssh://github.com/google/oss-fuzz-vulns'
+  repo_url: 'https://github.com/google/oss-fuzz-vulns'
```
michaelkedar, line 212: nit: this is technically more correct
```suggestion
  repo_url: 'https://github.com/google/oss-fuzz-vulns.git'
```

michaelkedar (approved):

oliverchang (approved):

PR #2614, opened 2024-09-12
fix: Special case paging for ubuntu entries

another-rex:
Add a special case for ubuntu entries to reduce the page size by 10x. Because Ubuntu entries are incredibly large.

### DIFF
--- a/gcp/api/server.py
+++ b/gcp/api/server.py
@@ -63,6 +63,8 @@ _MAX_VULN_LISTED_POST_EXCEEDED = 5
 # Max responses before MAX_VULN_RESP_THRESH has been exceeded
 _MAX_VULN_LISTED_PRE_EXCEEDED = 1000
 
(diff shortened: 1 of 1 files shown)


gcp/api/server.py
```diff
@@ -733,6 +741,12 @@ def do_query(query: osv_service_v1_pb2.Query,
         'version specified in params and purl query',
     )
 
+  # Hack to work around ubuntu having extremely large individual entries
+  if (ecosystem.startswith("Ubuntu") or
```
oliverchang, line 745: nit: single quotes for strings.

oliverchang (approved):

gcp/api/server.py
hogo6002, line 745: I think only `linux` related packages would reach the limit. Should we only apply this to packages start with `linux`?

hogo6002 (approved):
Thanks Rex!

gcp/api/server.py
another-rex: Applied to only get linux packages.

PR #2615, opened 2024-09-12
test(ecosystems): add tests for ecosystem consistency

andrewpollock:
Test that ecosystem name hardcoding is internally consistent and consistent with the OSV Schema

Requires OSV Schema submodule to be at or past https://github.com/ossf/osv-schema/commit/84969b48584c72424595f1d918865da865e03a45 (currently unreleased) to pass.


osv/ecosystems/ecosystems_test.py
```diff
@@ -0,0 +1,58 @@
...
+      schema = json.load(schema_f)
+    self.schema_ecosystems = re.compile(schema['$defs']['ecosystem']['pattern'])
+    self.canonical_ecosystems = _ecosystems._ecosystems.keys()  # pylint: disable=protected-access
+
+  def test_ecosystem_supported_by_schema(self):
+    """Test ecosystems referenced exist in schema definition"""
+    for ecosystem in self.canonical_ecosystems:
+      self.assertTrue(
+          self.schema_ecosystems.match(ecosystem),
+          msg=f'"{ecosystem}" not defined in schema')
```
michaelkedar, line 39: Wondering if there might be a way to also check the inverse of this.
i.e. check if there's any ecosystems that are allowed in the schema that `ecosystems.get()` doesn't have.

Though, I can't really think of a way to do this given it's a regex pattern (unless we want to fuzz all possible strings)

osv/ecosystems/ecosystems_test.py
```diff
@@ -0,0 +1,58 @@
...
+from . import _ecosystems
+
+
+class EcosystemsTest(unittest.TestCase):
+  """Ecosystem helpers tests."""
+
+  def setUp(self):
+    with open('osv/osv-schema/validation/schema.json') as schema_f:
+      schema = json.load(schema_f)
+    self.schema_ecosystems = re.compile(schema['$defs']['ecosystem']['pattern'])
```
michaelkedar, line 31: I think we might have to be careful with potential differences between python and json schema regex syntax.
Notably, this string in the schema uses `[[:digit:]]` which doesn't parse as intended in python.

I wonder if the jsonschema python package has a way to check individual fields?

osv/ecosystems/ecosystems_test.py
```diff
@@ -0,0 +1,58 @@
...
+    self.canonical_ecosystems = _ecosystems._ecosystems.keys()  # pylint: disable=protected-access
+
+  def test_ecosystem_supported_by_schema(self):
+    """Test ecosystems referenced exist in schema definition"""
+    for ecosystem in self.canonical_ecosystems:
+      self.assertTrue(
+          self.schema_ecosystems.match(ecosystem),
```
michaelkedar, line 38: nit: this should probably be using `self.assertIsNotNone`

osv/ecosystems/ecosystems_test.py
andrewpollock: https://github.com/ossf/osv-schema/pull/280 is the companion piece to this, and yes, https://json-schema.org/draft/2020-12/json-schema-core#name-regular-expressions implies this is somewhat brittle.

osv/ecosystems/ecosystems_test.py
andrewpollock: > Wondering if there might be a way to also check the inverse of this.
> i.e. check if there's any ecosystems that are allowed in the schema that `ecosystems.get()` doesn't have.

I had thought of that, but then it's somewhat to be expected as new ecosystems onboard, that they get defined in the schema first and then in OSV.dev code, so we wouldn't want this test to start failing when that happens, right?

osv/ecosystems/ecosystems_test.py
andrewpollock: Good nit. Fixed in b06c127e

osv/ecosystems/ecosystems_test.py
michaelkedar: I suppose that makes sense.
On the other hand, the tests failing would be good to signal that we need to add the ecosystem before we start attempting to ingest them into the database.

michaelkedar (approved):

osv/ecosystems/ecosystems_test.py
andrewpollock: > On the other hand, the tests failing would be good to signal that we need to add the ecosystem before we start attempting to ingest them into the database.

That makes me think there might be an opportunity to flag a premature addition to sources{_test}.yaml 🤔

andrewpollock: No, that isn't going to work because there's no direct linkage between a source and the ecosystems it provides...

Issue #2402, opened 2024-07-18
Missing PURL converters


michaelkedar:
We've added a few [OSV ecosystems](https://github.com/google/osv.dev/blob/260a49adcac4d7d06d75b2f53c0915225e15518e/osv/ecosystems/_ecosystems.py#L33) without adding [PURL conversions](https://github.com/google/osv.dev/blob/260a49adcac4d7d06d75b2f53c0915225e15518e/osv/purl_helpers.py#L18) for them.

Ecosystems that don't have corresponding PURL entries (And their [known purl type](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst)):

- [x] `AlmaLinux`
- [ ] `Android`
- [ ] `Bioconductor`
- [x] `Bitnami` (`bitnami`)
- [x] `Chainguard` (`apk/chainguard`)
- [x] `ConanCenter` (`conan`)
- [x] `CRAN` (`cran`)
- [ ] `GHC`
- [ ] `GitHub Actions` (`github`?)
- [ ] `Linux`
- [ ] `Photon OS`
- [x] `Rocky Linux`
- [x] `SwiftURL` (`swift`) (thanks @sschuberth!)
- [x] `Ubuntu` (`deb/ubuntu`)
- [x] `Wolfi` (`apk/wolfi`)

Some of the OSV records have PURLs already because the data source provides them (e.g. Bitnami, Chainguard & Wolfi)

I think we need to come up with some way to make sure we always have these, and explicitly define ecosystems with no PURL equivalents.

andrewpollock:
> I think we need to come up with some way to make sure we always have these, and explicitly define ecosystems with no PURL equivalents.

I'm trying to pull together a checklist for onboarding a new data source, the closest thing to date is https://github.com/google/osv.dev/issues/2252#issuecomment-2138754180 but I was completely unaware of this functionality being a thing until now, so this also needs to be added.

sschuberth:
Is [this](https://github.com/google/osv.dev/pull/2788) all that needs to be done for the case of Swift?

hogo6002:
related discussion: https://github.com/google/osv.dev/pull/2900#discussion_r1861438920

PR #2635, opened 2024-09-18
feat(data source): re-enable UBUNTU-CVE- into prod instance

hogo6002:
Add [`UBUNTU-CVE-` records](https://github.com/canonical/ubuntu-security-notices/tree/main/osv/cve) into prod instance

Test instance PR: https://github.com/google/osv.dev/pull/2607

### DIFF
--- a/source.yaml
+++ b/source.yaml
@@ -300,12 +300,12 @@
 - name: 'ubuntu'
   versions_from_repo: False
   type: 0
-  ignore_patterns: ['^(?!USN-).*$']
+  ignore_patterns: ['^(?!(USN|UBUNTU)-).*$']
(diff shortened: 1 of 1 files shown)


oliverchang (approved):
Did ingesting this into the test instance all look OK ?

hogo6002:
> Did ingesting this into the test instance all look OK ?

The only issue I found is that some UBUNTU-CVEs are missing the `purl` field. But I think this is our problem, as we are missing some purl converters (https://github.com/google/osv.dev/issues/2402). Other issues, such as the Linux package query one, have all been fixed. So, I think the records are good to merge for now, and we should probably implement a purl converter for all newly added Linux distros later.

PR #2636, opened 2024-09-19
fix(aliaslookup): do the right thing with multiple inputs

andrewpollock:
I think there was some copy pasta here from another tool, this wasn't doing the right thing when invoked with a list of CVEs.

### DIFF
--- a/tools/aliaslookup/aliaslookup.py
+++ b/tools/aliaslookup/aliaslookup.py
@@ -22,20 +22,21 @@ def lookup_by_aliases(client: datastore.Client,
   Returns:
     a datastore.query.Iterator
   """
-  query = client.query(kind='Bug')
-  # the "IN" operator only supports 30 identifiers as a maximum, so we need to
-  # batch into multiple a composite OR query
-  # This still caps out at 100 total filters, sadly
-  query.add_filter(
-      filter=datastore.query.PropertyFilter('aliases', 'IN', identifiers))
-  if verbose:
-    print(f'Running query {query.filters[0]} '
-          f'on {query.kind} (in {query.project})...')
-  result = list(query.fetch())
-  if verbose:
-    print(f'Retrieved {len(result)} bugs')
+  for identifier in identifiers:
+    query = client.query(kind='Bug')
+    query.add_filter(
+        filter=datastore.query.PropertyFilter('aliases', 'IN', [identifier]))
+    if verbose:
+      print(f'Running query {query.filters[0]} '
+            f'on {query.kind} (in {query.project})...')
+    result = list(query.fetch())
+    if verbose:
+      print(f'Retrieved {len(result)} bugs')
 
-  return result
+    if len(result) > 0:
+      yield result[0]
+    else:
+      continue
 
 
 def main() -> None:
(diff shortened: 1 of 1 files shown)


tools/aliaslookup/aliaslookup.py
```diff
@@ -22,20 +22,21 @@ def lookup_by_aliases(client: datastore.Client,
...
-          f'on {query.kind} (in {query.project})...')
-  result = list(query.fetch())
-  if verbose:
-    print(f'Retrieved {len(result)} bugs')
-
-  return result
+  for identifier in identifiers:
+    query = client.query(kind='Bug')
+    query.add_filter(
+        filter=datastore.query.PropertyFilter('aliases', 'IN', [identifier]))
```
michaelkedar, line 28: I think this *should* be doing a query on the [`AliasGroup`](https://github.com/google/osv.dev/blob/d51109b8198cc49cc7d47aefd29aed05902483ba/osv/models.py#L878-L881) kind, then getting the Bugs from the `bug_ids`

IIRC some gotchas: if something has 0 aliases, it won't be in any `AliasGroup`. Also, a `bug_id` in an `AliasGroup` may not exist in our db.

andrewpollock: > I think this _should_ be doing a query on the [`AliasGroup`](https://github.com/google/osv.dev/blob/d51109b8198cc49cc7d47aefd29aed05902483ba/osv/models.py#L878-L881) kind, then getting the Bugs from the `bug_ids`

I looked at that, then realized past me had written this, and used it for the task. I think I'll write a separate script to query `AliasGroup` for _computed_ aliases (as opposed to aliases-on-records-as-imported).

andrewpollock:
At least I managed to fix the submodule mess... 🤬

tools/aliaslookup/aliaslookup.py
```diff
@@ -22,20 +22,21 @@ def lookup_by_aliases(client: datastore.Client,
   Returns:
     a datastore.query.Iterator
   """
```
michaelkedar, line 24: nit: This doesn't seem right. I think it yields entities?
I think it was also wrong before the change?

tools/aliaslookup/aliaslookup.py
```diff
@@ -22,20 +22,21 @@ def lookup_by_aliases(client: datastore.Client,
...
+      print(f'Running query {query.filters[0]} '
+            f'on {query.kind} (in {query.project})...')
+    result = list(query.fetch())
+    if verbose:
+      print(f'Retrieved {len(result)} bugs')
+
+    if len(result) > 0:
+      yield result[0]
+    else:
+      continue
```
michaelkedar, line 39: *Should* this only be yielding the first result per identifier? Before it was returning all of them.

michaelkedar (approved):

tools/aliaslookup/aliaslookup.py
andrewpollock: Agreed, thanks. Fixed in 65e80f09

tools/aliaslookup/aliaslookup.py
andrewpollock: Umm, yeah, I don't know what I was thinking earlier in the week (or when I first wrote this). Tidied things up significantly in 65e80f09

michaelkedar (approved):

PR #2650, opened 2024-09-24
fix(ecosystems): treat Red Hat as supported

andrewpollock:
Correct a typo and also a thinko with normalizing fall-through ecosystem names

We'll need to reimport the Red Hat source in Staging once this change is live there.

### DIFF
--- a/osv/ecosystems/_ecosystems.py
+++ b/osv/ecosystems/_ecosystems.py
@@ -64,7 +64,7 @@ _ecosystems = {
     'Linux': OrderingUnsupportedEcosystem(),
     'OSS-Fuzz': OrderingUnsupportedEcosystem(),
     'Photon OS': OrderingUnsupportedEcosystem(),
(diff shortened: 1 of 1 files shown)


oliverchang (approved):

osv/ecosystems/_ecosystems.py
```diff
@@ -64,7 +64,7 @@
     'Linux': OrderingUnsupportedEcosystem(),
     'OSS-Fuzz': OrderingUnsupportedEcosystem(),
     'Photon OS': OrderingUnsupportedEcosystem(),
-    'Red Hat': OrderingUnsupportedEcosystem,
+    'Red Hat': OrderingUnsupportedEcosystem(),
```
oliverchang, line 67: We should create a proper RedHat ecosystem in a followup PR, and we should have RPM ordering already so API queries will work out of the box.

PR #2668, opened 2024-09-25
fix(renovate): try to improve lockfile PR titles

andrewpollock:
Make the per-group lockfile maintenance PRs more obviously per-group

This is something of an experiment, based on:
[`prTitle`](https://docs.renovatebot.com/configuration-options/#prtitle) -> [`commitMessage`](https://docs.renovatebot.com/configuration-options/#commitmessage) and the defaults for
[`lockFileMaintenance`](https://docs.renovatebot.com/configuration-options/#lockfilemaintenance)


andrewpollock:
Actually, looking at the lockfile PRs currently getting spat out, some of them already have group-identifying information in their titles and some don't... 🤔

renovate.json
```diff
@@ -12,7 +12,8 @@
   "osvVulnerabilityAlerts": true,
   "lockFileMaintenance": {
     "enabled": true,
-    "schedule": ["before 6am on wednesday"]
+    "schedule": ["before 6am on wednesday"],
+    "commitMessageTopic": "group {{groupName}}"
```
cuixq, line 16: I am not sure if `groupName` will work here since `lockFileMaintenance` is separate from `packageRules`
consider using `packageFile`? https://docs.renovatebot.com/templates/#other-available-fields

andrewpollock: It's so strange. Looking at the console:

image

(which is also how they render in https://github.com/google/osv.dev/issues/714)

the two without titles correlate with
- https://github.com/google/osv.dev/pull/2743
- https://github.com/google/osv.dev/pull/2741

The ones that have something useful in the title look like it is maybe `parentDir`? The fact that it's already including something sometimes makes me think that further adjustment of the string isn't the right answer, it's failing to set a variable already...

cuixq: after taking a detailed look on the PRs generated this morning:
 - the two lockfile maintenance PRs without group name did not list any update explicitly, the update is `lockFileMaintenance`
 - other PRs with group name also listed package updates besides `lockFileMaintenance`

so my guess is for purely lock file maintenance PR, the group name is not added (which we actually want and this PR should help). For PRs with both lockfile maintenance and updates, the group names are added because this is supported to be a general update PR with grouping rules, but lockfile maintenance shouldn't be the PR main body.

another-rex:
We probably want to merge this anyway, since this also fixes a problem currently where if only one thing gets updated the PR name is that dependency, which usually very unhelpful.

cuixq (approved):

PR #2669, opened 2024-09-25
fix(importer): retry oss-fuzz GCS exports

andrewpollock:
I noticed evidence of GCS upload failures from this while I was scrutinizing an importer run's logs.


andrewpollock:
@oliverchang is there a way to cross-reference things to see what past failed exports may exist?

oliverchang (approved):

oliverchang:
> @oliverchang is there a way to cross-reference things to see what past failed exports may exist?

Hmm, nothing out of the box unfortunately. We'd have to write up a script of some sort to do that (which may be useful generally for monitoring purposes if we run it regularly?)

Issue #2670, opened 2024-09-25
Vulns from REST sources may be missed due to timing


michaelkedar:
The importer uses `last_update_date` to compare both [the Last-Modified of the HEAD request](https://github.com/google/osv.dev/blob/d6ca60c8bdf9313f976075a0aee3657d2b9e5857/docker/importer/importer.py#L718-L725) and [the `modified` time of the individual records](https://github.com/google/osv.dev/blob/d6ca60c8bdf9313f976075a0aee3657d2b9e5857/docker/importer/importer.py#L734-L736). `last_update_date` gets set to the time when the [importer started processing the source](https://github.com/google/osv.dev/blob/d6ca60c8bdf9313f976075a0aee3657d2b9e5857/docker/importer/importer.py#L764-L765).

This can cause an edge case where we miss a few updates:
```
00:10 - REST data source updates.
00:15 - Importer runs. Sets last_update_date to 00:15
00:20 - REST data source updates. Adds records with modified dates between 00:10 - 00:20.
00:30 - Importer runs. Processes the vulnerabilities with modified dates > 00:15 (last_update_date).
```
Here, the importer won't detect the vulnerabilities from the REST source that were modified between `00:10 - 00:15`.

andrewpollock:
> last_update_date gets set to the time when the [importer started processing the source](https://github.com/google/osv.dev/blob/d6ca60c8bdf9313f976075a0aee3657d2b9e5857/docker/importer/importer.py#L764-L765)

That's at the end of the run, not the start. Does that change your hypothesis at all?

michaelkedar:
> > last_update_date gets set to the time when the [importer started processing the source](https://github.com/google/osv.dev/blob/d6ca60c8bdf9313f976075a0aee3657d2b9e5857/docker/importer/importer.py#L764-L765)
> 
> That's at the end of the run, not the start. Does that change your hypothesis at all?

The value is [set at the start](https://github.com/google/osv.dev/blob/d6ca60c8bdf9313f976075a0aee3657d2b9e5857/docker/importer/importer.py#L712), but either way, no - there'd be an issue as long as `last_import_date` is set anywhere between the two REST updates.

michaelkedar:
I've got #2672 on production now, which should help prevent this if the source is well-behaved.

I think we also should be doing some periodic complete reimports of all our sources to catch anything that might be missed for whatever reason.

PR #2672, opened 2024-09-26
fix: Set `last_update_date` to the time reported by the REST endpoint

michaelkedar:
Use the `Last-Modified` header (or the latest `modified` vuln date if the header is missing) to set the REST SourceRepository's `last_update_date`, instead of using the time when the importer ran.

Should alleviate #2670 for well-behaved REST sources, though it still doesn't handle cases where the source itself adds a record with date before the previous updated date.

While I was here, I also tidied up the `ignore_last_import_time` logic a bit (because the long line splitting was annoying).

### DIFF
--- a/docker/importer/importer.py
+++ b/docker/importer/importer.py
@@ -704,35 +704,44 @@ class Importer:
     """Process updates from REST API."""
     logging.info('Begin processing REST: %s', source_repo.name)
 
-    ignore_last_import_time = (
-        source_repo.ignore_last_import_time or not source_repo.last_update_date)
-    if ignore_last_import_time:
+    last_update_date = source_repo.last_update_date or datetime.datetime.min
+    if source_repo.ignore_last_import_time:
+      last_update_date = datetime.datetime.min
       source_repo.ignore_last_import_time = False
       source_repo.put()
-    import_time_now = utcnow()
+
     request = requests.head(source_repo.rest_api_url, timeout=_TIMEOUT_SECONDS)
     if request.status_code != 200:
       logging.error('Failed to fetch REST API: %s', request.status_code)
       return
 
-    if 'Last-Modified' in request.headers:
-      last_modified = datetime.datetime.strptime(
-          request.headers['Last-Modified'], _HTTP_LAST_MODIFIED_FORMAT)
-      # Check whether endpoint has been modified since last update
-      if not ignore_last_import_time and (last_modified
-                                          < source_repo.last_update_date):
-        logging.info('No changes since last update.')
-        return
+    request_last_modified = None
+    if last_modified := request.headers.get('Last-Modified'):
+      try:
+        request_last_modified = datetime.datetime.strptime(
+            last_modified, _HTTP_LAST_MODIFIED_FORMAT)
+        # Check whether endpoint has been modified since last update
+        if request_last_modified <= last_update_date:
+          logging.info('No changes since last update.')
+          return
+      except ValueError:
+        logging.error('Invalid Last-Modified header: "%s"', last_modified)
 
     request = requests.get(source_repo.rest_api_url, timeout=_TIMEOUT_SECONDS)
     # Parse vulns into Vulnerability objects from the REST API request.
     vulns = osv.parse_vulnerabilities_from_data(
         request.text, source_repo.extension, strict=self._strict_validation)
+
+    vulns_last_modified = datetime.datetime.min
     # Create tasks for changed files.
     for vuln in vulns:
       import_failure_logs = []
-      if not ignore_last_import_time and vuln.modified.ToDatetime(
-      ) < source_repo.last_update_date:
+      vuln_modified = vuln.modified.ToDatetime()
+      if request_last_modified and vuln_modified > request_last_modified:
+        logging.warning('%s was modified (%s) after Last-Modified header (%s)',
+                        vuln.id, vuln_modified, request_last_modified)
+      vulns_last_modified = max(vulns_last_modified, vuln_modified)
+      if vuln_modified <= last_update_date:
         continue
       try:
         # TODO(jesslowe): Use a ThreadPoolExecutor to parallelize this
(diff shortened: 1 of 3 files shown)


docker/importer/importer.py
```diff
@@ -704,35 +704,44 @@ def _process_updates_rest(self, source_repo: osv.SourceRepository):
     """Process updates from REST API."""
```
andrewpollock, line 704: Could you please add a prose description of the intended logic to the docstring?

andrewpollock (changes requested):

docker/importer/importer_test.py
```diff
@@ -849,36 +849,40 @@ def test_all_updated(self, unused_mock_time: mock.MagicMock,
...
     thread.start()
     self.source_repo.last_update_date = datetime.datetime(2020, 1, 1)
-    self.source_repo.put()
+    repo = self.source_repo.put()
     imp = importer.Importer('fake_public_key', 'fake_private_key', self.tmp_dir,
                             importer.DEFAULT_PUBLIC_LOGGING_BUCKET, 'bucket',
                             False, False)
     imp.run()
     self.assertEqual(mock_publish.call_count, data_handler.cve_count)
+    self.assertEqual(repo.get().last_update_date, datetime.datetime(2024, 1, 1))
```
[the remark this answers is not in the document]
michaelkedar: Added in 9ce7a86

docker/importer/importer.py
michaelkedar: Added in [2841c89](https://github.com/google/osv.dev/pull/2672/commits/2841c894b9213bdada5469c8ca1349aefbf547f5)

docker/importer/importer.py
```diff
@@ -704,35 +704,44 @@ def _process_updates_rest(self, source_repo: osv.SourceRepository):
...
+        request_last_modified = datetime.datetime.strptime(
+            last_modified, _HTTP_LAST_MODIFIED_FORMAT)
+        # Check whether endpoint has been modified since last update
+        if request_last_modified <= last_update_date:
+          logging.info('No changes since last update.')
+          return
+      except ValueError:
+        logging.error('Invalid Last-Modified header: "%s"', last_modified)
 
     request = requests.get(source_repo.rest_api_url, timeout=_TIMEOUT_SECONDS)
```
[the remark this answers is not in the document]
michaelkedar: I've tried to do this in 085e161 using [requests / urllib3](https://requests.readthedocs.io/en/latest/user/advanced/#example-automatic-retries), to avoid having to add a new dependency.

docker/importer/importer.py
```diff
@@ -701,42 +703,77 @@ def _process_deletions_bucket(self,
...
       source_repo.ignore_last_import_time = False
       source_repo.put()
-    import_time_now = utcnow()
-    request = requests.head(source_repo.rest_api_url, timeout=_TIMEOUT_SECONDS)
+
+    s = requests.Session()
+    adapter = HTTPAdapter(
+        max_retries=Retry(
+            total=3, status_forcelist=[502, 503, 504], backoff_factor=1))
+    s.mount(source_repo.rest_api_url, adapter)
```
oliverchang, line 730: Rather than mounting these explicitly, would it make sense (and is it possible to) just do a catchall here to apply this to all URLs?

oliverchang (approved):

docker/importer/importer.py
michaelkedar: Worked out how to catchall by mounting on `http://` and `https://`

docker/importer/importer.py
```diff
@@ -701,42 +703,77 @@ def _process_deletions_bucket(self,
...
       source_repo.ignore_last_import_time = False
       source_repo.put()
-    import_time_now = utcnow()
-    request = requests.head(source_repo.rest_api_url, timeout=_TIMEOUT_SECONDS)
+
+    s = requests.Session()
+    adapter = HTTPAdapter(
+        max_retries=Retry(
+            total=3, status_forcelist=[502, 503, 504], backoff_factor=1))
+    s.mount('http://', adapter)
```
another-rex, line 730: Should we be supporting http (not s) endpoints?

docker/importer/importer.py
```diff
@@ -754,14 +791,13 @@ def _process_updates_rest(self, source_repo: osv.SourceRepository):
...
-        import_failure_logs.append('Failed to parse vulnerability "' + vuln.id +
-                                   '"')
+        import_failure_logs.append(f'Failed to parse vulnerability "{vuln.id}"')
         continue
 
     replace_importer_log(storage.Client(), source_repo.name,
                          self._public_log_bucket, import_failure_logs)
 
-    source_repo.last_update_date = import_time_now
+    source_repo.last_update_date = request_last_modified or vulns_last_modified
```
another-rex, line 800: Maybe use `max` here?

michaelkedar: You can't do a `max` comparison with a `None` and a `datetime` if `request_last_modified` isn't set.

But also, I want it to be consistent with which time it uses, preferring the server's `Last-Modifed` time since it's check first for early exit.

docker/importer/importer.py
michaelkedar: The tests use http (although I haven't actually added a test for the retries). I don't think there's any harm in supporting it here regardless.

docker/importer/importer_test.py
```diff
@@ -821,7 +825,7 @@ def setUp(self):
     self.tmp_dir = tempfile.mkdtemp()
 
     tests.mock_datetime(self)
-    warnings.filterwarnings("ignore", "unclosed", ResourceWarning)
+    warnings.filterwarnings('ignore', category=SystemTimeWarning)
```
andrewpollock, line 828: I didn't realize there was a way to squelch these, this is awesome! Thank you!!!

andrewpollock (approved):
LGTM

another-rex (approved):

PR #2674, opened 2024-09-27
fix(debian): make fault-tolerant HTTP downloads

andrewpollock:
I noticed a recent spate of 503s in the error logging that may have benefited from this


hogo6002 (approved):
Thanks Andrew!

PR #2683, opened 2024-10-01
refactor(importer): reintroduce #2644 minus performance regression

andrewpollock:
This reverts commit https://github.com/google/osv.dev/commit/165b2fe90d273fcef3bc7a396d5f4ec41f48c9ff.

Fix performance regression introduced:

- don't create a new storage client and Data Store client for each
  thread processing a GCS blob.


oliverchang (approved):

another-rex (approved):

docker/importer/importer.py
```diff
@@ -315,11 +315,107 @@ def _vuln_ids_from_gcs_blob(self, client: storage.Client,
...
+        strict=self._strict_validation)
+
+    # TODO(andrewpollock): integrate with linter here.
+
+    # This is the atypical execution path (when reimporting is triggered)
+    if ignore_last_import_time:
+      return blob_hash, blob.name
+
+    # If being run under test, reuse existing NDB client.
+    ctx = ndb.context.get_context(False)
```
oliverchang, line 395: nit: s/ctx/ndb_ctx/ to be clearer.

docker/importer/importer.py
```diff
@@ -315,11 +315,107 @@ def _vuln_ids_from_gcs_blob(self, client: storage.Client,
...
+    if ctx is None:
+      # Production. Use the NDB client passed in.
+      ctx = ndb_client.context()
+    else:
+      # Unit testing. Reuse the unit test's existing NDB client to avoid
+      # "RuntimeError: Context is already created for this thread."
+      ctx.set_memcache_policy(False)
```
oliverchang, line 402: Are these two lines necessary, given that we expect these to be set by the test framework? 

It's also better to exclude these generally to keep this cleaner and avoid tying this to test-only code to begin with. I.e. it makes more sense to have this code be to generally support the case where there is an existing context to use, as opposed to something that's specific to testing.

docker/importer/importer.py
andrewpollock: Renamed in [81e9dc5](https://github.com/google/osv.dev/pull/2683/commits/81e9dc537353d9b36fce32216abe995f4e62b15c)

docker/importer/importer.py
andrewpollock: Yeah, good catch. Initially I hadn't been directly reusing the test framework's client. Removed in [81e9dc5](https://github.com/google/osv.dev/pull/2683/commits/81e9dc537353d9b36fce32216abe995f4e62b15c)

another-rex (approved):

oliverchang (approved):

Issue #1404, opened 2023-06-20
Support RedHat vulnerabilities


fingeromer:
Currently OSV supports a few operating system ecosystems like Debian & Alpine. 
We would like to open a feature request for supporting RedHat ecosystem vulnerabilities. 

Thanks, have a nice day.

mprpic:
The Red Hat ecosystem is large and varied so we're still working out the kinks on how to best structure the data in the OSV schema, but it's in progress! Since there is interest in this data, I'll ask here @fingeromer, are you mostly interested in data on vulnerabilities affecting RPMs shipped on RHEL? Or other Red Hat products as well?

andrewpollock:
The onboarding process is a little bit bespoke and toilsome at the moment, but something we're continuously improving on and streamlining with each new data source onboarded. I would like to get it to the point of being much more checklist/cookbook driven than it currently is. My detailed response here is an (ongoing) experiment at further process improvement and seeks to address some [recent actionable feedback received by another data source onboarding](https://github.com/google/osv.dev/issues/2089#issuecomment-2071549142). Your actionable feedback is also very welcome.

In a nutshell:

- [ ] Decide if you're going to publish records via a Git repository, GCS bucket or [REST endpoint](https://google.github.io/osv.dev/rest-api-contribution/) (I'm going to assume a Git repository?)
- [ ] Create a PR to [reserve a prefix in the OSV-Schema](https://ossf.github.io/osv-schema/#id-modified-fields) (worked examples: https://github.com/ossf/osv-schema/pull/235 https://github.com/ossf/osv-schema/pull/223 https://github.com/ossf/osv-schema/pull/219)
  - [ ] We review the records you start publishing for OSV Schema [correctness](https://github.com/ossf/osv-schema/tree/main/validation) and [quality](https://google.github.io/osv.dev/data_quality.html) (the work happening under the [OSV Data Quality Program](https://github.com/orgs/google/projects/62) is also relevant here, as an FYI) as part of reviewing and merging that PR
- [ ] Create a PR to extend [purl_helpers.py](https://github.com/google/osv.dev/blob/master/osv/purl_helpers.py) (if appropriate)
- [ ] Create a PR to start [importing the records you are publishing into our test instance of OSV.dev](https://github.com/google/osv.dev/blob/master/source_test.yaml) and validate everything is working as intended there (worked example: https://github.com/google/osv.dev/pull/2086)
[...]

PR #2699, opened 2024-10-02
feat(source): update redhat-stage temp source to production source

jasinner:
Testing of Red Hat RPM security data.

See https://github.com/google/osv.dev/issues/1404

### DIFF
--- a/source_test.yaml
+++ b/source_test.yaml
@@ -256,15 +256,18 @@
   link: 'https://github.com/RConsortium/r-advisory-database/blob/main/'
   editable: False
 
-- name: 'redhat-stage'
+- name: 'redhat'
   versions_from_repo: False
-  type: 1
-  bucket: 'osv-test-cve-osv-conversion'
-  directory_path: 'redhat-temporary'
+  rest_api_url: 'https://security.access.redhat.com/data/osv/all.json'
+  type: 2
+  directory_path: 'data/osv'
+  detect_cherrypicks: False
   ignore_patterns: ['^(?!RH[BES]{1}A-).*$']
   extension: '.json'
   db_prefix: ['RHBA-', 'RHEA-', 'RHSA-']
-  link: 'https://storage.googleapis.com/osv-test-cve-osv-conversion/'
+  ignore_git: False
+  human_link: 'https://access.redhat.com/errata/{{ BUG_ID }}'
+  link: 'https://security.access.redhat.com/data/osv/'
   editable: False
 
 - name: 'rockylinux'


source_test.yaml
```diff
@@ -256,15 +256,18 @@
...
-- name: 'redhat-stage'
+- name: 'redhat'
   versions_from_repo: False
-  type: 1
-  bucket: 'osv-test-cve-osv-conversion'
-  directory_path: 'redhat-temporary'
+  rest_api_url: 'https://security.access.redhat.com/data/osv/all.json'
+  type: 2
+  directory_path: 'data/osv'
+  detect_cherrypicks: False
```
jasinner, line 264: I don't know what this setting does.

source_test.yaml
```diff
@@ -256,15 +256,18 @@
...
   ignore_patterns: ['^(?!RH[BES]{1}A-).*$']
   extension: '.json'
   db_prefix: ['RHBA-', 'RHEA-', 'RHSA-']
-  link: 'https://storage.googleapis.com/osv-test-cve-osv-conversion/'
+  ignore_git: False
```
jasinner, line 268: It's not a git repo so probably irrelevant?

source_test.yaml
```diff
@@ -256,15 +256,18 @@
...
+  human_link: 'https://access.redhat.com/errata/{{ BUG_ID }}'
+  link: 'https://security.access.redhat.com/data/osv/'
```
jasinner, line 270: This link returns 404 because we don't have a index.html setup. Does it matter?

andrewpollock (approved):

source_test.yaml
andrewpollock: It is, but it's also harmless.

source_test.yaml
andrewpollock: I was just [sanity checking that](https://github.com/google/osv.dev/blob/bb583e85b054e3afec31fc52b5f7ab57cf2cf6f0/gcp/appengine/frontend3/src/templates/vulnerability.html#L35). I don't expect this to be a problem, but it's something to review post-import.

source_test.yaml
andrewpollock: It's only relevant for [analysing Git repositories](https://github.com/google/osv.dev/blob/fe6155f7cfa0e5df0ae1ef20c7b16f5c20bebed1/osv/impact.py#L120), and I don't believe the records being published have any Git commit ranges on them, and this is generally the correct approach as far as I understand cherry picking and the analysis of them...

PR #2701, opened 2024-10-04
chore(frontend): improve frontend vuln debuggability

andrewpollock:
I hit weirdness with an incomplete vulnerability page, which took longer than I'd have liked to get to the bottom of. It was because the source `redhat-stage` was renamed to `redhat` AND the particular record I was looking at had failed to import successfully from the newly named `redhat` source.

This scenario is unexpected enough that it shouldn't be something that happens silently.

### DIFF
--- a/gcp/appengine/frontend_handlers.py
+++ b/gcp/appengine/frontend_handlers.py
@@ -380,13 +380,16 @@ def add_links(bug):
 
 
 def add_source_info(bug, response):
-  """Add source information to `response`."""
+  """Add upstream provenance information to `response`."""
   if bug.source_of_truth == osv.SourceOfTruth.INTERNAL:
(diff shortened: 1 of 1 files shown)


hogo6002 (approved):
Interesting, I also renamed ubuntu to `ubuntu-cve` and `ubuntu-usn`, I wonder if the same thing happened.

andrewpollock:
> Interesting, I also renamed ubuntu to `ubuntu-cve` and `ubuntu-usn`, I wonder if the same thing happened.

That will be a little harder to determine, given that rename essentially bifurcated the pool of record IDs.

Once I realized what the problem was, I could then use `tools/datafix/list_ids_for_source.py` to query for IDs with the old source ID. Anything still existing from that source ID didn't get overwritten by the new import from the new source ID.

PR #2705, opened 2024-10-08
feat: add an EMPTY ecosystem for withdrawn records

hogo6002:
Some records will be withdrawn from upstream and may also have their affected package information deleted. These records currently cannot be exported and may cause confusion for our customers due to missing records.  Adding an "EMPTY" ecosystem will represent these records for exporting.

### DIFF
--- a/osv/models.py
+++ b/osv/models.py
@@ -397,9 +397,9 @@ class Bug(ndb.Model):
           break
 
     # If a withdrawn record has no affected package,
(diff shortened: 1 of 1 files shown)


another-rex:
It's maybe even worth adding square brackets around `[Empty]` just to make sure that it will never be a real ecosystem in the future.

another-rex (approved):
LGTM, though let's wait till after this week's release before merging this in.

cuixq (approved):

PR #2707, opened 2024-10-08
feat: skip exporting data into subdirectories

hogo6002:
Linux distributions have a large number of releases and products. Exporting each release into a subdirectory wastes a lot of memory and reduces exporter performance. Skipping the exporting of these subdirectories and keeping one directory for each ecosystem will improve exporter performance.

This PR will be merged after the weekly release, and performance will be tested on the test instance. If all looks good, an announcement email will be sent before moving this into production.

### DIFF
--- a/docker/exporter/export_runner.py
+++ b/docker/exporter/export_runner.py
@@ -54,6 +54,9 @@ def main():
   with concurrent.futures.ThreadPoolExecutor(
       max_workers=args.processes) as executor:
     for eco in ecosystems:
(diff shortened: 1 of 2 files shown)


another-rex (commented):
LGTM, can you also update the documentation of our exports?

oliverchang (approved):

hogo6002:
> LGTM, can you also update the documentation of our exports?

I will test the performance on a test instance first and update it in a follow-up PR after verifying the results.

PR #2726, opened 2024-10-09
fix: catch exception parsing CVSS scores on frontend

michaelkedar:
The CVSS calculators throw exceptions when parsing fails, which were uncaught and causing 500's on our website.


another-rex (approved):

hogo6002 (approved):
Thanks Michael!

cuixq (approved):

PR #2727, opened 2024-10-09
fix: CVSS severity rating on vulnerability.html

another-rex:
Looks something like this:
image

Fixes the "AttributeError'NoneType' object has no attribute 'lower'" error

### DIFF
--- a/gcp/appengine/frontend_handlers.py
+++ b/gcp/appengine/frontend_handlers.py
@@ -749,7 +749,7 @@ def display_severity_rating(severity: dict) -> str:
   """Return base score and rating of the severity."""
   severity_base_score, severity_rating = calculate_severity_details(severity)
   if severity_base_score is None:
-    return "Invalid Severity Rating"
+    return 'Invalid Severity Rating'
 
   return f"{severity_base_score} ({severity_rating})"
 
--- a/gcp/appengine/frontend_handlers.py
+++ b/gcp/appengine/frontend_handlers.py
@@ -758,7 +758,7 @@ def display_severity_rating(severity: dict) -> str:
 def severity_level(severity: dict) -> str:
   """Return rating of the severity."""
   _, rating = calculate_severity_details(severity)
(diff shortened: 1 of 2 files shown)


hogo6002 (approved):

gcp/appengine/frontend_handlers.py
```diff
@@ -748,14 +748,17 @@ def link_to_deps_dev(package, ecosystem):
 def display_severity_rating(severity: dict) -> str:
   """Return base score and rating of the severity."""
   severity_base_score, severity_rating = calculate_severity_details(severity)
+  if severity_base_score is None:
+    return "Invalid Severity Rating"
```
michaelkedar, line 752: nit: use single quotes
```suggestion
    return 'Invalid Severity Rating'
```

gcp/appengine/frontend_handlers.py
```diff
@@ -748,14 +748,17 @@ def link_to_deps_dev(package, ecosystem):
...
+
   return f"{severity_base_score} ({severity_rating})"
 
 
 @blueprint.app_template_filter('severity_level')
 def severity_level(severity: dict) -> str:
   """Return rating of the severity."""
   _, rating = calculate_severity_details(severity)
-  return rating.lower()
+  return (rating and rating.lower()) or "invalid"
```
michaelkedar, line 761: nit: double quotes

Also, imo this might be a bit clearer:
```suggestion
  return 'invalid' if rating is None else rating.lower()
```

another-rex: Yeah good idea!

michaelkedar (approved):

oliverchang (commented):
thanks! should we cut another website release @hogo6002  ?

hogo6002:
> thanks! should we cut another website release @hogo6002 ?

Done, https://osv.dev/vulnerability/GHSA-f3cx-396f-7jqp works now

Issue #2423, opened 2024-07-30
Export an `all.zip` containing every vulnerability.


michaelkedar:
We currently do not have an easy way to download every single OSV record - we only have the zip files for individual ecosystems, which a) duplicates vulnerabilities that are in multiple ecosystems and b) misses vulnerabilities that aren't in any ecosystems (e.g. Android seems to be removing all details from records that they withdraw, meaning they will no longer show up in any zip file)

We should export an `all.zip` containing every vulnerability for consumers who want to download the entire database (e.g. deps.dev).

PR #2731, opened 2024-10-11
feat(exporter): add a unified all.zip

hogo6002:
Add a ZIP file containing records from all ecosystems to solve https://github.com/google/osv.dev/issues/2423.  Write everything to the default work directory for reuse by the new unified all.zip (instead of using temporary directories). Clean up the work directory afterward.

### DIFF
--- a/docker/exporter/export_runner.py
+++ b/docker/exporter/export_runner.py
@@ -15,10 +15,11 @@
 """OSV Exporter."""
 import argparse
 import concurrent.futures
+import glob
 import logging
 import os
 import subprocess
-import shutil
+import tempfile
 import zipfile as z
 
 from google.cloud import ndb, storage
@@ -47,21 +48,24 @@ def main():
       default=os.cpu_count() or DEFAULT_EXPORT_PROCESSES)
   args = parser.parse_args()
 
-  # Clean up the work directory, in case other job didn't clean it properly.
-  clean_work_dir(args.work_dir)
   query = osv.Bug.query(projection=[osv.Bug.ecosystem], distinct=True)
   ecosystems = [bug.ecosystem[0] for bug in query if bug.ecosystem] + ['list']
 
-  with concurrent.futures.ThreadPoolExecutor(
-      max_workers=args.processes) as executor:
-    for eco in ecosystems:
-      # Skip exporting data for child ecosystems (e.g., 'Debian:11').
-      if ':' in eco:
-        continue
-      executor.submit(spawn_ecosystem_exporter, args.work_dir, args.bucket, eco)
-  # Upload a ZIP file containing records from all ecosystems.
-  aggregate_all_vulnerabilities(args.work_dir, args.bucket)
-  clean_work_dir(args.work_dir)
+  # Set TMPDIR to change the tempfile default directory
+  tmp_dir = os.path.join(args.work_dir, 'tmp')
+  os.makedirs(tmp_dir, exist_ok=True)
+  os.environ['TMPDIR'] = tmp_dir
+
+  with tempfile.TemporaryDirectory() as export_dir:
+    with concurrent.futures.ThreadPoolExecutor(
+        max_workers=args.processes) as executor:
+      for eco in ecosystems:
+        # Skip exporting data for child ecosystems (e.g., 'Debian:11').
+        if ':' in eco:
+          continue
+        executor.submit(spawn_ecosystem_exporter, export_dir, args.bucket, eco)
+    # Upload a ZIP file containing records from all ecosystems.
+    aggregate_all_vulnerabilities(export_dir, args.bucket)
 
 
 def spawn_ecosystem_exporter(work_dir: str, bucket: str, eco: str):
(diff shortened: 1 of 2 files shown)


docker/exporter/export_runner.py
```diff
@@ -74,6 +76,39 @@ def spawn_ecosystem_exporter(work_dir: str, bucket: str, eco: str):
...
+  for ecosystem_dir, _, files in os.walk(work_dir):
+    if ecosystem_dir == work_dir:
+      continue
+    for vuln_filename in files:
+      if not vuln_filename.endswith('.json'):
+        continue
+      all_vulns[vuln_filename] = os.path.join(ecosystem_dir, vuln_filename)
+
+  with z.ZipFile(output_zip, 'a') as all_zip:
+    for vuln_filename in sorted(all_vulns):
```
michaelkedar, line 97: Curious if there's any particular reason to sort these?

docker/exporter/export_runner.py
```diff
@@ -58,6 +57,9 @@ def main():
       if ':' in eco:
         continue
       executor.submit(spawn_ecosystem_exporter, args.work_dir, args.bucket, eco)
+  # Upload a ZIP file containing records from all ecosystems.
+  aggregate_all_vulnerabilities(args.work_dir, args.bucket)
+  clean_work_dir(args.work_dir)
```
michaelkedar, line 62: Maybe also call `clean_work_dir` at the beginning of main, in case something doesn't get clean up properly from a different job

michaelkedar (approved):
Nice work!

docker/exporter/export_runner.py
hogo6002: I remember we encountered some all.zip issue [because it was not sorted](https://github.com/google/osv.dev/pull/2311). So I think it would be better to sort this all.zip as well. (But I guess it's not necessary, as I got all the file names through the disk directory, which should already be sorted. But I still feel better to sort it just in case some race conditions.)

docker/exporter/export_runner.py
hogo6002: good point, updated.

docker/exporter/export_runner.py
```diff
@@ -74,6 +78,39 @@ def spawn_ecosystem_exporter(work_dir: str, bucket: str, eco: str):
...
+    for vuln_filename in sorted(all_vulns):
+      file_path = all_vulns[vuln_filename]
+      all_zip.write(file_path, os.path.basename(file_path))
+
+  storage_client = storage.Client()
+  bucket = storage_client.get_bucket(export_bucket)
+  upload_single(bucket, output_zip, zip_file_name)
+
+
+def clean_work_dir(work_dir: str):
```
another-rex, line 108: Rather than using clean_work_dir, let's just create a temporary directory similar to how it was done in the individual exporter runners, since this is what it's used for. (just make sure to set the environment variable for the temporary directory to the working directory so it always writes to the ssd)

And pass the temporary directory path into the individual exporter runs as the work dir.

docker/exporter/export_runner.py
```diff
@@ -74,6 +78,39 @@ def spawn_ecosystem_exporter(work_dir: str, bucket: str, eco: str):
...
+def aggregate_all_vulnerabilities(work_dir: str, export_bucket: str):
+  """
+  Aggregates vulnerability records from each ecosystem into a single zip
+  file and uploads it to the export bucket.
+  """
+  zip_file_name = 'all.zip'
+  output_zip = os.path.join(work_dir, zip_file_name)
+  all_vulns = {}
+
+  for ecosystem_dir, _, files in os.walk(work_dir):
```
another-rex, line 90: A cleaner way to do this could be with python globs. https://docs.python.org/3/library/glob.html (See examples at the bottom). a glob of `**/*.json` should select all the files you need without additional if statements.

another-rex: (Make sure to set recursive to true) `glob.glob("**/*.json", recursive=True)`

docker/exporter/export_runner.py
```diff
@@ -74,6 +78,39 @@ def spawn_ecosystem_exporter(work_dir: str, bucket: str, eco: str):
...
+    if ecosystem_dir == work_dir:
+      continue
+    for vuln_filename in files:
+      if not vuln_filename.endswith('.json'):
+        continue
+      all_vulns[vuln_filename] = os.path.join(ecosystem_dir, vuln_filename)
+
+  with z.ZipFile(output_zip, 'a') as all_zip:
```
another-rex, line 98: Apparently python's zip library uses a lot of memory (which might be why we run out of memory in our current exporter), so this might be a problem. No action required for now, though if this ends up using too much memory we can replace this zip library with a streaming one.

docker/exporter/exporter.py
```diff
@@ -130,12 +120,22 @@ def _export_to_file_and_zipfile(bug):
     with concurrent.futures.ThreadPoolExecutor(
         max_workers=_EXPORT_WORKERS) as executor:
       # Note: all.zip is included here
```
another-rex, line 122:
```suggestion
      # Note: the individual ecosystem all.zip is included here
```

Just clarify a bit since we now have a complete all.zip as well

another-rex (commented):
Looking good! Added some comments

docker/exporter/exporter.py
```diff
@@ -130,12 +120,22 @@ def _export_to_file_and_zipfile(bug):
...
-      for filename in os.listdir(tmp_dir):
-        executor.submit(self.upload_single, bucket,
-                        os.path.join(tmp_dir, filename),
+      for filename in os.listdir(ecosystem_dir):
+        executor.submit(upload_single, bucket,
+                        os.path.join(ecosystem_dir, filename),
                         f'{ecosystem}/{filename}')
 
 
+def upload_single(bucket, source_path, target_path):
```
another-rex, line 129: Please add type annotations here

docker/exporter/export_runner.py
hogo6002: updated

docker/exporter/export_runner.py
hogo6002: updated

another-rex (approved):

PR #2732, opened 2024-10-11
fix(nvd-cve-osv): add a 2% deletion threshold check

andrewpollock:
After the events of #2678, add in a threshold check of the number of GCS objects that would get deleted and abort if more than 10% would be.

Follow
https://google.github.io/styleguide/shellguide.html#which-shell-to-use for setting shell options.


vulnfeeds/cmd/nvd-cve-osv/run_cve_to_osv_generation.sh
```diff
@@ -56,6 +60,14 @@ for (( YEAR = $(date +%Y) ; YEAR >= ${FIRST_INSCOPE_YEAR} ; YEAR-- )); do
 done
 
 # Copy results to GCS bucket.
+objs_present=$(gsutil ls "${OSV_OUTPUT_GCS_PATH}" | wc -l)
+objs_deleted=$(gsutil -m rsync -n -c -d "${WORK_DIR}/nvd2osv/gcs_stage" "${OSV_OUTPUT_GCS_PATH}" 2>&1 | grep -c "Would delete")
+
+if (( objs_deleted < objs_present * 0.1 )); then
```
oliverchang, line 66: should this be > instead?

andrewpollock: Yes. Fixed this, and many other things surfaced from rigorous manual testing in [bb87c97](https://github.com/google/osv.dev/pull/2732/commits/bb87c979c8c58eba16e56c223f263076c542266c)

another-rex:
Given #2730 would it make sense to also make this 2%? 10% of all CVE records seem like a lot of entries.

andrewpollock:
> Given #2730 would it make sense to also make this 2%? 10% of all CVE records seem like a lot of entries.

Fair call. I've defaulted it to this and made it able to be overridden in case this value turns out to be troublesome.

oliverchang (approved):

PR #2762, opened 2024-10-21
fix(staging-api-test): increase memory limit

hogo6002:
We recently added more Linux ecosystem records, and the original memory limit is not enough to fetch all vulnerabilities. Increase the memory limit and the runtime."

### DIFF
--- a/deployment/clouddeploy/gke-workers/environments/oss-vdb-test/staging-api-test.yaml
+++ b/deployment/clouddeploy/gke-workers/environments/oss-vdb-test/staging-api-test.yaml
@@ -3,7 +3,7 @@ kind: CronJob
 metadata:
   name: staging-api-test
 spec:
-  schedule: "50 9 * * *"
+  schedule: "50 21 * * *"
   concurrencyPolicy: Forbid
   jobTemplate:
     spec:
@@ -19,8 +19,8 @@ spec:
             resources:
               requests:
                 cpu: 1.5
-                memory: "4G"
+                memory: "10G"
               limits:
                 cpu: 2
-                memory: "10G"
+                memory: "30G"
           restartPolicy: Never
--- a/deployment/clouddeploy/gke-workers/environments/oss-vdb-test/staging-api-test.yaml
+++ b/deployment/clouddeploy/gke-workers/environments/oss-vdb-test/staging-api-test.yaml
@@ -3,7 +3,8 @@ kind: CronJob
 metadata:
   name: staging-api-test
 spec:
-  schedule: "50 21 * * *"
+  # Runs at 10:00 PM UTC (9:00 AM AEST) every day.
+  schedule: "0 22 * * *"
   concurrencyPolicy: Forbid
   jobTemplate:
     spec:
--- a/deployment/clouddeploy/gke-workers/environments/oss-vdb-test/staging-api-test.yaml
+++ b/deployment/clouddeploy/gke-workers/environments/oss-vdb-test/staging-api-test.yaml
@@ -3,8 +3,8 @@ kind: CronJob
 metadata:
   name: staging-api-test
 spec:
-  # Runs at 10:00 PM UTC (9:00 AM AEST) every day.
-  schedule: "0 22 * * *"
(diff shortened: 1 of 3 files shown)


deployment/clouddeploy/gke-workers/environments/oss-vdb-test/staging-api-test.yaml
```diff
@@ -3,7 +3,7 @@ kind: CronJob
 metadata:
   name: staging-api-test
 spec:
-  schedule: "50 9 * * *"
+  schedule: "50 21 * * *"
```
another-rex, line 6: What timezone is this time in?

hogo6002: I think all cron jobs are using UTC time?

another-rex: Also, I think it'll be better to have this run on the hour rather than at 50 minutes, since it runs for over an hour anyway it doesn't when it starts, so let's just simplify it.

another-rex: > I think all cron jobs are using UTC time? 

Can we have this run at 9:00 AEST? This way we can see the results live throughout the day.

another-rex: Oh that's what this is already doing!, can you just add a quick comment here about that.

hogo6002: Yeah, that is why I set it to this time. I will update it to 10pm UTC

hogo6002: Updated (may vary due to daylight saving time)

andrewpollock: FYI, you can [set a timezone explicitly](https://github.com/google/osv.dev/blob/e96a5770781054ab72fbc2ddd3d9cb92e187fce7/deployment/clouddeploy/gke-workers/base/nvd-cve-osv.yaml#L6), so you don't have to do mental arithmetic to try and get it into local time. It even displays something useful when you're trying to reason about them:

```
$ kubectl get cronjobs
NAME                      SCHEDULE        TIMEZONE           SUSPEND   ACTIVE   LAST SCHEDULE   AGE
alias-computation         10/15 * * * *   <none>             False     0        5m36s           355d
alpine-cve-convert        0 */1 * * *     <none>             False     0        30m             642d
backup                    0 18 * * *      <none>             False     0        13h             329d
combine-to-osv            30 */1 * * *    <none>             False     1        36s             642d
cpe-repo-gen              0 6 */1 * *     <none>             False     0        90m             265d
debian-convert            0 */1 * * *     <none>             False     0        30m             615d
debian-copyright-mirror   0 6 */1 * *     <none>             False     1        90m             573d
debian-cve-convert        0 */1 * * *     <none>             False     0        30m             196d
debian-first-version      0 1 * * *       <none>             False     0        6h30m           615d
exporter                  */30 * * * *    <none>             False     1        30m             677d
generate-sitemap          30 8 * * *      <none>             False     0        23h             131d
importer                  */15 * * * *    <none>             False     1        36s             679d
importer-deleter          * */3 * * *     <none>             False     1        39m             166d
indexer-controller        0 0 * * *       <none>             False     0        7h30m           570d
nvd-cve-osv               0 6,13 * * *    Australia/Sydney   False     0        5h30m           517d
nvd-mirror                0 */2 * * *     <none>             False     0        90m             327d
staging-api-test          50 9 * * *      <none>             False     0        21h             42d
```

hogo6002: Thanks Andrew, updated to Sydney time

andrewpollock (approved):

PR #2766, opened 2024-10-22
feat(importer): enable per-source strict mode and record per-record schema validation failures

andrewpollock:
This commit adds support for enabling strict mode on a per SourceRepository basis, per-record quality findings in a new Data Store Kind, and reporting JSON Schema validation failures as record quality findings.


docker/importer/importer.py
```diff
@@ -182,6 +182,67 @@ def _request_internal_analysis(self, bug):
...
+
+    Sets the finding's last_attempt to now, and adds the finding to the list of
+    findings for the record (if any already exist)
+    """
+
+    # Get any current findings for this record.
+    findingtimenow = utcnow()
+    existing_findings = osv.ImportFinding.query(
+        osv.ImportFinding.bug_id == bug_id)
+    if list(existing_findings):
```
oliverchang, line 233: nit: do this `list()` once and store it in a variable. Otherwise I think this means we'll be running the query twice when we do both line 233 and line 234.

oliverchang: Also: do we expect only a single `ImportFinding` per ID? 

In that case, I'd suggest keying `ImportFinding` on the bug ID, so that there is a guaranteed one-to-one mapping.

andrewpollock: > nit: do this `list()` once and store it in a variable. Otherwise I think this means we'll be running the query twice when we do both line 233 and line 234.

Addressed in [bc85d0f](https://github.com/google/osv.dev/pull/2766/commits/bc85d0f0cd0c58356b70af2a64b4c813dc56d088)

> Also: do we expect only a single ImportFinding per ID?

Yes.

> In that case, I'd suggest keying ImportFinding on the bug ID, so that there is a guaranteed one-to-one mapping.

Can you elaborate on this? Is this feedback on https://github.com/google/osv.dev/blob/281ed7e0606cc6572b93e1dae581e1535f333963/osv/models.py#L915-L920 ?

oliverchang: > > nit: do this `list()` once and store it in a variable. Otherwise I think this means we'll be running the query twice when we do both line 233 and line 234.
> 
> Addressed in [bc85d0f](https://github.com/google/osv.dev/pull/2766/commits/bc85d0f0cd0c58356b70af2a64b4c813dc56d088)
> 
> > Also: do we expect only a single ImportFinding per ID?
> 
> Yes.
> 
> > In that case, I'd suggest keying ImportFinding on the bug ID, so that there is a guaranteed one-to-one mapping.
> 
> Can you elaborate on this? Is this feedback on
> 
> https://github.com/google/osv.dev/blob/281ed7e0606cc6572b93e1dae581e1535f333963/osv/models.py#L915-L920
> 
> ?

Kind of -- it's feedback for how we are creating and retrieving this, which is relevant to this PR. See https://googleapis.dev/python/python-ndb/latest/key.html and https://github.com/google/osv.dev/blob/281ed7e0606cc6572b93e1dae581e1535f333963/osv/models.py#L513 for an existing example.

We can control the key of this Datastore entity, and set it to `bug_id`. This has the benefit of:

1) Faster lookup, because we can just do a Datastore lookup by key (instead of a Datastore query). NDB let's us do this by just doing `Model.get_by_id(...)`. 
2) Ensuring this 1:1 mapping.

andrewpollock: Thank you, is e75ea315 what's needed?

oliverchang: Nice! Yes, but with that we can now also simplify 

```python
    existing_findings = list(
        osv.ImportFinding.query(osv.ImportFinding.bug_id == bug_id))
   for existing_finding in existing_findings: 
    ...
```

to 

```python
existing_finding = osv.Importfinding.get_by_id(bug_id)
```

andrewpollock: Got it, thanks. Refactored accordingly in aabdaa18

oliverchang (approved):

PR #2782, opened 2024-10-23
feat(staging-api-test): replace datastore query with bucket access

hogo6002:
Replace Datastore bug query with `all.zip`


docker/staging_api_test/retrieve_bugs_from_bucket.py
```diff
@@ -0,0 +1,143 @@
...
+
+def main() -> None:
+  osv.logs.setup_gcp_logging('staging-test')
+
+  if not os.path.exists(BUG_DIR):
+    # This will take around 10 mins
+    seed = random.randrange(100)
+    # The seed value can be replaced for debugging
+    random.seed(seed)
+    logging.info('Random seed %d', seed)
```
michaelkedar, line 135: Could you add a comment as to why you generate the seed like this?

docker/staging_api_test/retrieve_bugs_from_bucket.py
```diff
@@ -0,0 +1,143 @@
...
+  logging.info('Start to download %s.', ZIP_FILE_PATH)
+  storage_client = storage.Client()
+  bucket = storage_client.get_bucket(VULN_BUCKET)
+  try:
+    blob = bucket.blob(ZIP_FILE_PATH)
+    file_path = os.path.join(tmp_dir, ZIP_FILE_PATH)
+    blob.download_to_filename(file_path)
+  except Exception as e:
+    logging.exception('Failed to download all.zip: %s', e)
+  logging.info('Downloaded %s.', ZIP_FILE_PATH)
```
michaelkedar, line 85: I'm not sure, but I think this will continue to run after the exception happens.
You might want to `raise e` or something after logging the exception?

docker/staging_api_test/retrieve_bugs_from_bucket.py
```diff
@@ -0,0 +1,143 @@
...
+  os.makedirs(BUG_DIR, exist_ok=True)
+  tmp_dir = os.path.join(BUG_DIR, 'tmp')
+  os.makedirs(tmp_dir, exist_ok=True)
+  os.environ['TMPDIR'] = tmp_dir
+  logging.info('Start to process %s.', ZIP_FILE_PATH)
+
+  def write_to_json():
+    """Writes to a new JSON file."""
+    file_name = f'{BUG_DIR}/all_bugs_{file_counter}.json'
+    with open(file_name, 'w+') as f:
```
michaelkedar, line 102: nit: I think this can just be `'w'`

docker/staging_api_test/retrieve_bugs_from_bucket.py
```diff
@@ -0,0 +1,143 @@
...
+
+  entries_per_file = 10000  # amount of bugs per file
+  file_counter = 0
+  os.makedirs(BUG_DIR, exist_ok=True)
+  tmp_dir = os.path.join(BUG_DIR, 'tmp')
+  os.makedirs(tmp_dir, exist_ok=True)
+  os.environ['TMPDIR'] = tmp_dir
+  logging.info('Start to process %s.', ZIP_FILE_PATH)
+
+  def write_to_json():
```
another-rex, line 99: Can you move this function out to the top level and pass in the required variables as arguments? Currently it's hard to reason about especially with bug_info_list being defined in what looks like a different scope that the function.

docker/staging_api_test/retrieve_bugs_from_bucket.py
```diff
@@ -0,0 +1,143 @@
...
+      'db_id': bug['id'],
+      'project': affected_package.get('name', None),
+      'ecosystem': affected_package.get('ecosystem', None),
+      'affected_fuzzy': affected_fuzzy
+  }
+
+
+def download_vuln_zip(tmp_dir: str) -> None:
+  """Downloads all.zip file from bucket."""
+  logging.info('Start to download %s.', ZIP_FILE_PATH)
```
another-rex, line 76: Maybe add `VULN_BUCKET/ZIP_FILE_PATH`?

docker/staging_api_test/retrieve_bugs_from_bucket.py
```diff
@@ -0,0 +1,143 @@
...
+          continue
+
+  logging.info('All results saved to %s.', BUG_DIR)
+
+
+def main() -> None:
+  osv.logs.setup_gcp_logging('staging-test')
+
+  if not os.path.exists(BUG_DIR):
+    # This will take around 10 mins
```
another-rex, line 131: Hmm do you know what most of the time is spent on?

docker/staging_api_test/retrieve_bugs_from_bucket.py
```diff
@@ -0,0 +1,143 @@
...
+    bug_info_list = []
+    with zipfile.ZipFile(all_zip, 'r') as vuln_zip:
+      for filename in vuln_zip.namelist():
+        try:
+          with vuln_zip.open(filename) as file:
+            bug = json.load(file)
+            bug_info_list.append(format_bug_for_output(bug))
+            if len(bug_info_list) == entries_per_file:
+              write_to_json()
+              file_counter += 1
```
another-rex, line 118: Rather than maintaining a mutable state, we can use `itertools.batched` (https://docs.python.org/3/library/itertools.html#itertools.batched) to split the loop out into individual batches.

This is easier to reason about, and makes it easy if we want to parallelise it later.

docker/staging_api_test/retrieve_bugs_from_bucket.py
```diff
@@ -0,0 +1,143 @@
...
+    logging.exception('Failed to download all.zip: %s', e)
+  logging.info('Downloaded %s.', ZIP_FILE_PATH)
+
+
+def get_bugs_from_export() -> None:
+  """Gets all bugs from the exported all.zip and writes to `BUG_DIR`."""
+
+  entries_per_file = 10000  # amount of bugs per file
+  file_counter = 0
+  os.makedirs(BUG_DIR, exist_ok=True)
```
another-rex, line 93: I don't think this line is needed, the next makedirs will make both directories.

docker/staging_api_test/retrieve_bugs_from_bucket.py
```diff
@@ -0,0 +1,143 @@
...
+    blob.download_to_filename(file_path)
+  except Exception as e:
+    logging.exception('Failed to download all.zip: %s', e)
+  logging.info('Downloaded %s.', ZIP_FILE_PATH)
+
+
+def get_bugs_from_export() -> None:
+  """Gets all bugs from the exported all.zip and writes to `BUG_DIR`."""
+
+  entries_per_file = 10000  # amount of bugs per file
```
another-rex, line 91: Can you move this to a constant at the top of the file.

another-rex (commented):
Nice, added some comments

another-rex (approved):

docker/staging_api_test/retrieve_bugs_from_bucket.py
hogo6002: added comments

docker/staging_api_test/retrieve_bugs_from_bucket.py
hogo6002: added sys.exit

docker/staging_api_test/retrieve_bugs_from_bucket.py
hogo6002: Removed the comment. It was a legacy one for Datastore query.

docker/staging_api_test/retrieve_bugs_from_bucket.py
hogo6002: Yeah, it is better to do that, but it's available for Python 3.12, and we are currently still using 3.11.
I refactored the code a bit, and it will be easy to convert to `itertools.batched` later on when we switch to 3.12.

docker/staging_api_test/retrieve_bugs_from_bucket.py
hogo6002: removed

docker/staging_api_test/retrieve_bugs_from_bucket.py
```diff
@@ -0,0 +1,150 @@
...
+    blob = bucket.blob(ZIP_FILE_PATH)
+    file_path = os.path.join(tmp_dir, ZIP_FILE_PATH)
+    blob.download_to_filename(file_path)
+  except Exception as e:
+    logging.exception('Failed to download all.zip: %s', e)
+    sys.exit(1)
+  logging.info('Downloaded %s.', zip_full_path)
+
+
+def write_to_json(bug_info_list: list) -> None:
```
another-rex, line 92: nit: annotate the content of the list (I think it is `dict[str, any]`?)

docker/staging_api_test/retrieve_bugs_from_bucket.py
another-rex: sgtm

another-rex (approved):
LGTM

docker/staging_api_test/retrieve_bugs_from_bucket.py
hogo6002: It's actually a list of dict, updated it to `list[dict[str, any]]`

PR #2783, opened 2024-10-24
fix: use projection for related queries

michaelkedar:
The addition queries to find the related bugs was querying for the whole record, even though it only needs the id.
Changed the queries to grab only the db_id.

### DIFF
--- a/osv/models.py
+++ b/osv/models.py
@@ -965,6 +965,7 @@ def get_aliases_async(bug_id) -> ndb.Future:
 @ndb.tasklet
 def get_related_async(bug_id) -> ndb.Future:
   """Gets related bugs asynchronously."""
-  related_bugs = yield Bug.query(Bug.related == bug_id).fetch_async(keys_only=True)
+  related_bugs = yield Bug.query(Bug.related == bug_id).fetch_async(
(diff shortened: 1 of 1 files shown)


hogo6002 (approved):
LGTM!

osv/models.py
```diff
@@ -722,8 +722,8 @@ def to_vulnerability(self, include_source=False, include_alias=True):
     aliases = []
 
     if include_alias:
-      related_bugs = Bug.query(Bug.related == self.db_id).fetch()
-      related_bug_ids = [bug.db_id for bug in related_bugs]
+      related_bugs = Bug.query(Bug.related == self.db_id).fetch(keys_only=True)
+      related_bug_ids = [bug.id() for bug in related_bugs]
```
oliverchang, line 726: hmm, is this going to work? `bug.id()` may not be the same as `bug.db_id` for historical reasons (though we can probably clean that up via a database migration)

oliverchang: ref: https://github.com/google/osv.dev/issues/2786

oliverchang (approved):

another-rex (approved):

PR #2796, opened 2024-10-28
feat(api): log queries for test instance

hogo6002:
We're seeing some 502 and 504 errors reported from API queries. Without the details of each query, it's hard to reproduce and find the root causes. 
Adds logs for the test instance. It does not record actual user queries, only queries from our load tests.

### DIFF
--- a/gcp/api/server.py
+++ b/gcp/api/server.py
@@ -903,6 +903,7 @@ def query_by_commit(
   Returns:
     list of responses (return values from to_response)
   """
+
   query = osv.AffectedCommits.query(osv.AffectedCommits.commits == commit)
 
   context.query_counter += 1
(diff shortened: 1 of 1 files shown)


oliverchang (approved):
thanks!

another-rex (commented):
I think it might be better to print the entire query in one log in the QueryAffected and QueryAffectedBatch. Otherwise for batch queries we will be printing out a lot of lines, which might make debugging harder.

another-rex:
Our log storage might explode though haha, we might want to tweak the retention again after this is merged.

hogo6002:
I'll merge this first to allow for some bake time, which will give us time to fix any errors before tomorrow's release. If any issues arise, I will address them in a separate PR.

PR #2802, opened 2024-10-29
fix: Upgrade the website's memory limit

another-rex:
Some big records were still causing a memory outage for osv.dev, this increases it up to 4Gb's, the maximum memory limit on one core.

The osv.dev website only uses around ~1-2 instances, compared to our API which uses 20 instances, and is also using 8Gb per instance, so this is pretty small increase in relative cost. 

(The gitignore is for the hurl tool, which allows me to send http requests for testing)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ build/lib
 .history/*
 **/.history/*
 */datastore-helper/
+hurl-scripts/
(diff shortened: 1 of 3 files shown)


hogo6002 (approved):
Would 2Gi be enough?

another-rex:
Maybe, but I don't think it is worth optimising for because API uses 8GB with 20 instances.

oliverchang (approved):

PR #2818, opened 2024-10-30
feat: Make it easier to launch the api server locally without the backend.

another-rex:
Commonly when testing changes for the backend, we want to keep the ESP translation layer up continuously, while rapidly restarting the python code with changes.

This makes it easier to launch just the translation layer.

### DIFF
--- a/gcp/api/test_server.py
+++ b/gcp/api/test_server.py
@@ -183,11 +183,11 @@ if __name__ == '__main__':
     print(f'Usage: {sys.argv[0]} path/to/credential.json')
     sys.exit(1)
 
-  no_backend = False
+  no_backend_flag = False
   if len(sys.argv) == 3:
-    no_backend = sys.argv[2] == '--no-backend'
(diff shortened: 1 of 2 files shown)


hogo6002 (approved):

gcp/api/test_server.py
```diff
@@ -178,7 +183,11 @@ def start(credential_path, port=_ESP_PORT, backend_port=_BACKEND_PORT):
     print(f'Usage: {sys.argv[0]} path/to/credential.json')
     sys.exit(1)
 
-  server = start(sys.argv[1])
+  no_backend = False
+  if len(sys.argv) == 3:
+    no_backend = sys.argv[2] == '--no-backend'
+
+  server = start(sys.argv[1], no_backend=no_backend)
```
another-rex, line 190:
```suggestion
  no_backend_flag = False
  if len(sys.argv) == 3:
    no_backend_flag = sys.argv[2] == '--no-backend'

  server = start(sys.argv[1], no_backend=no_backend_flag)
```

Issue #2188, opened 2024-05-09
The minimum quality bar is sustainably enforced by OSV.dev


andrewpollock:
Records below the quality bar are not imported into or exported by OSV.dev

andrewpollock:
#2582 and #2766 are the first parts of this work

andrewpollock:
# Status Update

Piloting JSON Schema compliance enforcement in Staging for all sources (except RustSec Advisory DB until https://github.com/rustsec/advisory-db/issues/2135 gets addressed)

# Next Steps

- Assess effectiveness and enable for Production
- Integrate OSV Linter with Importer

PR #2837, opened 2024-11-07
feat(importer): prepare for selective strict mode enablement (staging)

andrewpollock:
This commit explicitly sets strict mode on a per-repository basis. This change is a no-op until the importer's `--strict_validation` flag is also set to True, which will still then be a no-op for all sources other than `cve-osv` where we can trial having it enabled.

Part of #2188


oliverchang (approved):

hogo6002 (commented):
Why not set the default value of strict_validation to `False` when it's not defined in the source, and only enable it when explicitly set to `True`?

andrewpollock:
> Why not set the default value of strict_validation to `False` when it's not defined in the source, and only enable it when explicitly set to `True`?

I'm opting to make things very explicit to avoid any nasty surprises, especially when the flag flip for the importer happens.

PR #2886, opened 2024-11-21
feat(workers): Enable kube-state-metrics with managed prometheus

rjerrems:
This is basically the configuration https://cloud.google.com/stackdriver/docs/managed-prometheus/exporters/kube_state_metrics here with:

- Split into a file per resource
- Updated version to kube-state-metrics 2.14.0
- Changed scrape interval from 30s to 5s
- Updated to only look at cronjob and pod, we can expand this later if we wish


andrewpollock (approved):
Very cool, so this enables the metric collection with alerting on those metrics in a future PR...

michaelkedar (approved):
LGTM

Issue #2786, opened 2024-10-24
Migrate Datastore Bug keys after removing logic to create source-specific keys


oliverchang:
In https://github.com/google/osv.dev/pull/2435, we stopped creating `Bug` ndb Keys with a `source_id:bug ID` format. 

However, existing entities might still have this, and should be migrated so we can delete `db_id` completely and just rely on the NDB key.

andrewpollock:
FYI https://github.com/search?q=repo%3Agoogle%2Fosv.dev+path%3A%2F%5Etools%5C%2Fdatafix%5C%2F%2F+db_id&type=code is relying on `db_id` so will require reworking to accommodate.

PR #2896, opened 2024-11-25
feat: Add script to refresh IDs to the new format.

another-rex:
fix: #2786

This script can also be used if we ever change the key format in the future, we just need to update the check for whether a key needs to be updated or not. 

I'll run this on a machine in the US when merged.

### DIFF
--- a/tools/datafix/refresh_ids.py
+++ b/tools/datafix/refresh_ids.py
@@ -11,7 +11,6 @@ import argparse
 import json
 import functools
 import time
-import os
 import typing
 
 MAX_BATCH_SIZE = 500
--- a/tools/datafix/refresh_ids.py
+++ b/tools/datafix/refresh_ids.py
@@ -47,8 +47,8 @@ def get_relevant_ids(verbose: bool) -> list[str]:
   return relevant_ids
 
 
-def reput_bugs(dryrun: bool, verbose: bool, loadcache: str) -> None:
-  """ Reput all bugs from a given source."""
+def refresh_ids(dryrun: bool, verbose: bool, loadcache: str) -> None:
+  """Update bugs IDs to the new format"""
 
   relevant_ids = []
   if loadcache:
@@ -67,7 +67,7 @@ def reput_bugs(dryrun: bool, verbose: bool, loadcache: str) -> None:
 
   # This handles the actual transaction of reputting
   # the bugs with ndb
-  def _reput_ndb(batch: int):
+  def _refresh_ids(batch: int):
     buf: list[osv.Bug] = [
         osv.Bug.get_by_id(r) for r in relevant_ids[batch:batch + MAX_BATCH_SIZE]
     ]
@@ -95,7 +95,7 @@ def reput_bugs(dryrun: bool, verbose: bool, loadcache: str) -> None:
       num_reputted += len(relevant_ids[batch:batch + MAX_BATCH_SIZE])
       print(f"Reput {num_reputted} bugs... - "
             f"{num_reputted/len(relevant_ids)*100:.2f}%")
-      ndb.transaction(functools.partial(_reput_ndb, batch))
+      ndb.transaction(functools.partial(_refresh_ids, batch))
     except DryRunException:
       # Don't have the first batch's transaction-aborting exception stop
       # subsequent batches from being attempted.
@@ -138,7 +138,7 @@ def main() -> None:
   client = ndb.Client(project=args.project)
   print(f"Running on project {args.project}.")
   with client.context():
-    reput_bugs(args.dryrun, args.verbose, args.loadcache)
+    refresh_ids(args.dryrun, args.verbose, args.loadcache)
 
 
 if __name__ == "__main__":
--- a/tools/datafix/refresh_ids.py
+++ b/tools/datafix/refresh_ids.py
@@ -1,0 +1,125 @@
+#!/usr/bin/env python3
+""" Utility to update the datastore key of each Bug to the new format determined by the pre put hook.
+
+    Does this by deleting and reputting each Bug entry.
+"""
+from google.cloud import ndb
+import osv
+
+import argparse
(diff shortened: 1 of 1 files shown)


tools/datafix/refresh_ids.py
```diff
@@ -0,0 +1,125 @@
...
+
+def reput_bugs(dryrun: bool, verbose: bool) -> None:
+  """ Reput all bugs from a given source."""
+
+  # Uncomment below to load the state and skip the get_relevant_ids func
+  # relevant_ids = json.load(open('relevant_ids.json', 'r'))
+  relevant_ids = get_relevant_ids(verbose)
+
+  # Store the state incase we cancel halfway to avoid having to do the initial query again
+  json.dump(relevant_ids, open('relevant_ids.json', 'w'))
```
michaelkedar, line 48: nit: the file doesn't get (explicitly) closed.

tools/datafix/refresh_ids.py
```diff
@@ -0,0 +1,125 @@
...
+      num_reputted += len(relevant_ids[batch:batch + MAX_BATCH_SIZE])
+      print(
+          f"Reput {num_reputted} bugs... - {num_reputted/len(relevant_ids)*100:.2f}%"
+      )
+      ndb.transaction(functools.partial(_reput_ndb, batch))
+    except Exception as e:
+      # Don't have the first batch's transaction-aborting exception stop
+      # subsequent batches from being attempted.
+      if dryrun and e.args[0].startswith("Dry run mode"):
+        print("Dry run mode. Preventing transaction from committing")
```
michaelkedar, line 87: nit: probably better to create a `DryRunException` to raise/catch , instead of doing this check with the exception text.

tools/datafix/refresh_ids.py
another-rex: Fixed

tools/datafix/refresh_ids.py
another-rex: Done

tools/datafix/refresh_ids.py
```diff
@@ -0,0 +1,131 @@
...
+      default=True,
+      help="Abort before making changes")
+  parser.add_argument(
+      "--verbose",
+      action=argparse.BooleanOptionalAction,
+      dest="verbose",
+      default=False,
+      help="Print each ID that needs to be processed")
+  parser.add_argument(
+      "--project",
```
hogo6002, line 117: Can we add a `--source` to allow only refresh IDs from a certain source repository.

tools/datafix/refresh_ids.py
```diff
@@ -0,0 +1,131 @@
...
+    if res.key.id() != res.db_id:  # type: ignore
+      relevant_ids.append(res.db_id)
+      if verbose:
+        print(res.db_id + ' - ' + res.key.id())  # type: ignore
+
+  print(f"Found {len(relevant_ids)} / {counter} relevant bugs to refresh.")
+  return relevant_ids
+
+
+def reput_bugs(dryrun: bool, verbose: bool) -> None:
```
hogo6002, line 44: We already have a `reput_bugs.py` file. Could we simply modify it to add the new function?

tools/datafix/refresh_ids.py
```diff
@@ -0,0 +1,131 @@
...
+import typing
+
+MAX_BATCH_SIZE = 500
+
+
+class DryRunException(Exception):
+  """This exception is raised to cancel a transaction during dry runs"""
+
+
+def get_relevant_ids(verbose: bool) -> list[str]:
```
andrewpollock, line 22: Please add a docstring that captures for the reader what the relevant IDs are that this returns.

tools/datafix/refresh_ids.py
```diff
@@ -0,0 +1,131 @@
...
+
+  print(f"Found {len(relevant_ids)} / {counter} relevant bugs to refresh.")
+  return relevant_ids
+
+
+def reput_bugs(dryrun: bool, verbose: bool) -> None:
+  """ Reput all bugs from a given source."""
+
+  relevant_ids = []
+  # Uncomment below to load the state and skip the get_relevant_ids func
```
andrewpollock, line 48: Rather than relying on doing this, either:
- add a flag, or
- make this behaviour (noisily) contingent on the existence of the file

tools/datafix/refresh_ids.py
another-rex: This would require adding a new index which is why it is not an option here.

another-rex: (db_id, source)

tools/datafix/refresh_ids.py
another-rex: I copied this file from reput_bugs.py, I think the functionality is different enough I didn't want to over complicate the existing file, since that file already works and does what it intends to do. This is mostly for a one off use-case to update the Bug IDs anyway.

another-rex: I've updated the function names to better clarify the difference.

tools/datafix/refresh_ids.py
another-rex: Done

tools/datafix/refresh_ids.py
another-rex: Made a flag

andrewpollock (approved):
Will this get run against staging first to catch any runtime corner cases not easily picked up by code review?

another-rex:
Yep, I've tested this with a few records in staging already and it seems to work. Will do a full pass for staging before applying it to production.

another-rex:
A few issues I can see is all of those old oss-fuzz records that's hidden but still exists in our database, which might fail when we try to reput it.

hogo6002 (approved):

oliverchang (approved):
very nice!

PR #2898, opened 2024-11-26
fix(website): replace confusing utils.is_prod() function

michaelkedar:
`utils.is_prod()` was actually checking if we were running on Cloud Run (returning True if running on 'oss-vdb-test').
Renamed it to `is_cloud_run()` to be less confusing.
Also created `utils.api_url()` to use where `is_prod` was being used incorrectly.


gcp/website/utils.py
```diff
@@ -15,13 +15,34 @@
...
+  Returns the URL of the OSV API for the current GCP project.
+
+  Falls back to 'api.osv.dev' if URL cannot be determined.
+  """
+  try:
+    _, project = google.auth.default()
+  except DefaultCredentialsError:
+    project = '<unknown>'
+
+  # TODO(michaelkedar): Is there a way to avoid hard-coding this?
```
andrewpollock, line 36: Might it be easier (and potentially safer) to hard-code the production project ID (which is somewhat inescapable, and also more durable) and fail open to the non-production hostname for everything else?

andrewpollock (approved):

gcp/website/utils.py
michaelkedar: Changed it to fall back to `api.test.osv.dev`.

another-rex (approved):

Issue #2842, opened 2024-11-08
Fetching vulnerabilities via api for ubuntu packages by purl with version does not work


Andre-85:
Hello everyone,
I tried to fetch vulnerabilities via api with the purl (Package URL). The SBOM from which I am extracting the packages, the purl has following format: pkg:deb/ubuntu/PACKAGE_NAME@VERSION. 

What I've tested:
1. the api works with purls like this for e.g. maven packages. Test:
`curl -d '{"package": {"purl": "pkg:maven/org.apache.struts/struts2-core@2.0.14"}}' https://api.osv.dev/v1/query`
4. Ubuntu purls work if i know the arch and the distribution. Test: 
`curl -d '{"package": {"purl": "pkg:deb/ubuntu/ruby-sinatra?arch=src?distro=jammy"}}' https://api.osv.dev/v1/query`
6. It works if I split the purl in package-name, version, namespace(==ecosystem). Test: 
`curl -d '{"package": {"name": "atftp"}, "version": "0.7.git20120829-3.1~0.18.04.1", ecosystem: "ubuntu"}' https://api.osv.dev/v1/query`
8. But giving the purl only for the same package and version like in 3., returns nothing. Test: 
`curl -d '{"package": {"purl": "pkg:deb/ubuntu/atftp@0.7.git20120829-3.1~0.18.04.1"}}' https://api.osv.dev/v1/query`
10. Additionally i found out that this works for debian packages coming directly from Debian. Test:  
`curl -d '{"package": {"purl": "pkg:deb/debian/nginx@1.14.2-2+deb10u3"}}' https://api.osv.dev/v1/query`

Right now I use (3.) as a workaround, but i would be nice if it could work like (1.).Thank you very much for your help in advance.Greetings,
André

cuixq:
@hogo6002 can you help to answer this?

Andre-85:
If you could help this would be great. Meanwhile it tried to understand how the osv.dev service works. My plan was to run a api-server in a Docker/Podman container and test patches on it and connect it to osv.dev database in the internet. Setting up with Podman works:
`
podman build --file gcp/api/Dockerfile -t osv-api .
podman run -p 8000:8000 osv-api:latest

`
But because of missing/wrong ndb credentials I cannot connect to osv.dev database. Did I do a mistake or is such a test setup not possible at all because the access to the database itself is private to google's osv.dev service?

hogo6002:
Hi @Andre-85, thanks for reporting this issue!

[...]

PR #2900, opened 2024-11-26
refactor(api): refactor purl API query code

hogo6002:
Converted PURL queries to package queries in `do_query()` to remove unnecessary code. Querying the Datastore directly with PURLs gives the same results as querying by package name, so we can simplify things by treating PURLs like regular package queries after extracting the `package/ecosystem/version`.

This will also resolve issue #2842 by rewriting the PURL-to-ecosystem logic. For most ecosystems, we can get the name from `purl.type`, but for Linux distributions, we need to use `purl.namespace`.


osv/purl_helpers.py
```diff
@@ -63,7 +107,8 @@ def package_to_purl(ecosystem: str, package_name: str) -> str | None:
...
 
-def purl_to_ecosystem(purl_type: str) -> str | None:
-  """Convert purl to a specific ecosystem string"""
-  ecosystem_purl = {v: k for k, v in PURL_ECOSYSTEMS.items()}
-  return ecosystem_purl.get(purl_type)
+def parse_purl_ecosystem(purl: PackageURL) -> str | None:
+  """Extracts the ecosystem name from a PackageURL by checking
+  its `type` and `namespace`."""
+  return PURL_TYPE_ECOSYSTEMS.get(
+      purl.type, PURL_NAMESPACE_ECOSYSTEMS.get(purl.namespace, None))
```
andrewpollock, line 114: Could you add some tests for this as well please?

andrewpollock (commented):
Nice simplification!

osv/purl_helpers.py
```diff
@@ -63,7 +107,8 @@ def package_to_purl(ecosystem: str, package_name: str) -> str | None:
   return f'pkg:{purl_type}/{_url_encode(package_name)}{suffix}'
 
 
-def purl_to_ecosystem(purl_type: str) -> str | None:
-  """Convert purl to a specific ecosystem string"""
-  ecosystem_purl = {v: k for k, v in PURL_ECOSYSTEMS.items()}
-  return ecosystem_purl.get(purl_type)
+def parse_purl_ecosystem(purl: PackageURL) -> str | None:
```
michaelkedar, line 110: Maybe this one function can parse out the ecosystem, name and version, just so we don't have to extract it ourselves later.

osv/purl_helpers.py
michaelkedar, line 114: fwiw, I think this will convert invalid purls e.g. `pkg:badtype/ubuntu/curl` -> `Ubuntu`

This function actually seems like a good candidate to use Python's new-ish [`match` statements](https://docs.python.org/3/tutorial/controlflow.html#match-statements):
```python
def parse_purl_ecosystem(purl: PackageURL) -> str | None:
  match purl:
    case PackageURL(type='cargo'):
      return 'crates.io'
    case PackageURL(type='deb', namespace='debian'):
      return 'Debian'
    case PackageURL(type='deb', namespace='ubuntu'):
      return 'Ubuntu'
    case _:
      return None
```

You can also do something like this to grab fields into variables:
```python
match purl:
  case PackageURL(type='cargo', name=pkg, version=ver):
    return 'crates.io', pkg, ver
```

michaelkedar (commented):
Nice!

osv/purl_helpers.py
hogo6002: updated, thanks!

hogo6002: added, also checked integration tests

osv/purl_helpers.py
```diff
@@ -63,7 +66,84 @@ def package_to_purl(ecosystem: str, package_name: str) -> str | None:
   return f'pkg:{purl_type}/{_url_encode(package_name)}{suffix}'
 
 
-def purl_to_ecosystem(purl_type: str) -> str | None:
-  """Convert purl to a specific ecosystem string"""
-  ecosystem_purl = {v: k for k, v in PURL_ECOSYSTEMS.items()}
-  return ecosystem_purl.get(purl_type)
+def parse_purl(purl_str: str) -> tuple[str, str, str] | None:
```
oliverchang, line 69: nit: can we used a namedtuple here instead? 

https://docs.python.org/3/library/collections.html#collections.namedtuple

oliverchang (commented):
very nice! Is there a way we can do some kind of testing to make sure we're not introducing any regressions on matching behaviour? 

i.e. some kind of comprehensive comparison between test (with this deployed) and prod (without this deployed) for specific ecosystems?

hogo6002:
> very nice! Is there a way we can do some kind of testing to make sure we're not introducing any regressions on matching behaviour?

We have integration tests for the API (LONG_TESTS), which I have manually checked. These tests do not produce any 500 errors. Also, after checking in the test instance, it  will be tested by performance testing, and I will monitor for any new 500 errors. I think this change should be safe.

osv/purl_helpers.py
hogo6002: Yeah, namedtuple is more suitable for this use case. Updated, thanks!

gcp/api/server.py
```diff
@@ -710,52 +711,59 @@ def do_query(query: osv_service_v1_pb2.Query,
...
-  purl_version = None
-  if purl_str:
-    try:
-      purl = PackageURL.from_string(purl_str)
-      purl_version = purl.version
-    except ValueError:
-      context.service_context.abort(grpc.StatusCode.INVALID_ARGUMENT,
-                                    'Invalid Package URL.')
+  # convert purl to package names
+  purl = purl_helpers.parse_purl(purl_str)
```
another-rex, line 728: nit: only pass in purl_str if it is not empty.

gcp/api/fixtures/IntegrationTests_api_query_response.txt
```diff
@@ -1,120 +1,244 @@
...
-    '200:{"commit": "d374094d8c49b6b7d288f307e11217ec5a502391", "package": {"purl": "pkg:cargo/crossbeam-utils", "version": "0.8.5"}}\n',
-    '200:{"commit": "d374094d8c49b6b7d288f307e11217ec5a502391", "package": {"purl": "pkg:cargo/crossbeam-utils"}}\n',
-    '200:{"commit": "d374094d8c49b6b7d288f307e11217ec5a502391", "package": {"purl": "pkg:cargo/crossbeam-utils@0.8.5", "version": "0.8.5"}}\n',
-    '200:{"commit": "d374094d8c49b6b7d288f307e11217ec5a502391", "package": {"purl": "pkg:cargo/crossbeam-utils@0.8.5"}}\n',
-    '200:{"commit": "d374094d8c49b6b7d288f307e11217ec5a502391", "package": {"purl": "pkg:pypi/numpy", "version": "0.8.5"}}\n',
-    '200:{"commit": "d374094d8c49b6b7d288f307e11217ec5a502391", "package": {"purl": "pkg:pypi/numpy"}}\n',
-    '200:{"commit": "d374094d8c49b6b7d288f307e11217ec5a502391", "package": {"purl": "pkg:pypi/numpy@8.24.0", "version": "0.8.5"}}\n',
-    '200:{"commit": "d374094d8c49b6b7d288f307e11217ec5a502391", "package": {"purl": "pkg:pypi/numpy@8.24.0"}}\n',
-    '200:{"commit": "d374094d8c49b6b7d288f307e11217ec5a502391", "package": {"version": "0.8.5"}}\n',
+[   '400:{"commit": "d374094d8c49b6b7d288f307e11217ec5a502391", "package": '
```
hogo6002, line 1: This long test has been skipped and hasn't been updated in a long time (it is also failing for the master branch). The script is generated using the command `TESTS_GENERATE=1`. Most of the 400 failures are caused by querying both the `PURL` and `package` name at the same time, which is invalid.

We could also re-enable long tests on the pipeline (run after merge), it doesn't run too long (maybe 1 min)

osv/purl_helpers.py
```diff
@@ -63,7 +66,84 @@ def package_to_purl(ecosystem: str, package_name: str) -> str | None:
...
+    purl_str: The Package URL string to parse.
+
+  Returns:
+    A tuple containing the ecosystem, package,
+    and version, or None if parsing fails.
+  """
+
+  try:
+    purl = PackageURL.from_string(purl_str)
+  except ValueError:  # Catch potential parsing errors
```
another-rex, line 86: I think you should keep the behavior of raising on error. This should be clearer to read than returning None and checking whether purl_str is not empty, and purl is None.

gcp/api/server.py
```diff
@@ -382,14 +381,16 @@ def query_info(query) -> tuple[str, str | None, str | None]:
   version = query.version
   if query.package.purl:
     try:
-      purl = PackageURL.from_string(query.package.purl)  # can raise ValueError
+      purl = purl_helpers.parse_purl(query.package.purl)  # can raise ValueError
```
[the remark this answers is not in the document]
another-rex: Ah if you are using named tuples then this is not necessary.

osv/purl_helpers.py
```diff
@@ -63,7 +66,84 @@ def package_to_purl(ecosystem: str, package_name: str) -> str | None:
...
+    return None
+
+  match purl:
+    case PackageURL(type='bitnami'):
+      ecosystem = 'Bitnami'
+    case PackageURL(type='cargo'):
+      ecosystem = 'crates.io'
+    case PackageURL(
+        type='golang', namespace=namespace, name=name, version=version):
+      # OSV uses the combined purl.namespace and purl.name for Go package names
```
andrewpollock, line 96: Nit: I think it's not so much OSV as it is Go itself?

```suggestion
      # Go uses the combined purl.namespace and purl.name for Go package names
```

osv/purl_helpers.py
```diff
@@ -63,7 +66,84 @@ def package_to_purl(ecosystem: str, package_name: str) -> str | None:
...
+      # Example:
+      #   pkg:golang/github.com/cri-o/cri-o
+      #   -> namespace: github.com/cri-o
+      #   -> name: cri-o
+      #   -> package name in OSV: github.com/cri-o/cri-o
+      return 'Go', namespace + '/' + name, version
```
andrewpollock: This being the odd-one-out return path I feel is likely to bite in the future, could you make this one like the rest (or the rest all early return as well) to avoid any potential reading comprehension failure induced bugs (or just cognitive load) in the future?

andrewpollock (approved):

osv/purl_helpers.py
hogo6002: updated

osv/purl_helpers.py
hogo6002: removed early return here

gcp/api/server.py
```diff
@@ -710,52 +709,59 @@ def do_query(query: osv_service_v1_pb2.Query,
...
 
-  purl = None
-  purl_version = None
+  # convert purl to package names
   if purl_str:
     try:
-      purl = PackageURL.from_string(purl_str)
-      purl_version = purl.version
+      purl = purl_helpers.parse_purl(purl_str)
     except ValueError:
```
another-rex, line 729: We probably want to let them know the exact reason their PURL is invalid.

osv/purl_helpers.py
```diff
@@ -63,7 +69,84 @@ def package_to_purl(ecosystem: str, package_name: str) -> str | None:
...
+
+  Returns:
+    A tuple containing the ecosystem, package,
+    and version, or None if parsing fails.
+  """
+
+  try:
+    purl = PackageURL.from_string(purl_str)
+  except ValueError as e:  # Catch potential parsing errors
+    raise e
```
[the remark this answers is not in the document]
hogo6002: removed catching and raising, added comment

gcp/api/server.py
hogo6002: Discussed offline, we'll keep the current invalid PURL error message for now. Our integration tests rely on error messages, and providing detailed PURL parsing failure reasons doesn't adding too much benefits.

osv/purl_helpers.py
```diff
@@ -63,7 +69,82 @@ def package_to_purl(ecosystem: str, package_name: str) -> str | None:
...
+  """
+
+  purl = PackageURL.from_string(
+      purl_str)  # May raise ValueError for invalid PURL strings
+
+  package = purl.name
+  version = purl.version
+
+  match purl:
+    case PackageURL(type='bitnami'):
```
oliverchang, line 91: using match/case here is very neat! But, it does introduce 2 sources of truths for the purl type -> osv ecosystem mapping. 

would it be better to create a `PURL_TYPE_TO_OSV_ECOSYSTEM` dict automatically from  `PURL_ECOSYSTEMS` above? 

And then we can use it here just by doing a dictionary lookup on the parsed `purl`s type value?

hogo6002: The existing `PURL_ECOSYSTEMS` is outdated and insufficient to support all Linux distros. My previous implementation used dictionaries, but required three separate mappings:
- PURL type to ecosystem for language ecosystems.
- PURL namespace to ecosystem for Linux distros.
- PURL namespace to type for Linux distros. (in case queries like `pkg:badtype/ubuntu/package`)
Also we have some special handling for Go, and it's best keep using a match/case statement.

I can update the `package_to_purl` function in a followup PR to try using common dictionaries or keep the same structures  with the new `parse_purl` function.

oliverchang: ah, I see, thanks! if there was a way to consolidate the source of truth a bit more in another PR, that would be ideal.

hogo6002: Yep, I'll do that in a separate PR. I wanted to avoid any unrelated changes in this one (in case anything breaks).

oliverchang (approved):

hogo6002:
Merge it now to allow for some bake time on the test instance.

Issue #1489, opened 2023-07-24
Validate Renovate configuration changes


andrewpollock:
- https://docs.renovatebot.com/config-validation/
- https://github.com/marketplace/actions/validate-renovate-configuration-with-renovate-config-validator

The most appropriate thing to do is probably add a GitHub Action to run the validator on PRs that modify it.

PR #2917, opened 2024-11-26
ci: add a workflow to validate renovate config

cuixq:
https://github.com/google/osv.dev/issues/1489

This PR adds a workflow to validate renovate configuration as specified by [renovatebot docs](https://docs.renovatebot.com/config-validation/#default-behavior).

I have tested the workflow by modifying the `renovate.json` with invalid format.


another-rex (approved):

hogo6002 (approved):

andrewpollock (approved):
Awesome!

There may be some wall clock benefit to only running these on PRs where the config has changed. See https://github.com/ossf/osv-schema/blob/main/.github/workflows/validate-schema.yml for inspiration.

cuixq:
PTAL - I updated the `yml` to only trigger on changes with `renovate.json`, as well as pinning both the actions.

PR #2918, opened 2024-11-27
fix: Attempt at grouping lockfile maintenance

another-rex:
Not sure if this will work, but it will be quite nice if all the lockfile maintenance can be done in 1 PR to reduce weekly chore.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -12,7 +12,8 @@
   "osvVulnerabilityAlerts": true,
   "lockFileMaintenance": {
     "enabled": true,
-    "schedule": ["before 6am on wednesday"]
+    "schedule": ["before 6am on wednesday"],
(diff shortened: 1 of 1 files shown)


hogo6002 (approved):

andrewpollock:
What was the reasoning behind the groupings in the first place?

andrewpollock (approved):

another-rex:
Grouping was for updating specific packages, it makes sense to review one project at a time (so you can approve the docs update, but change/fix/hold off the worker dependency update.).

With lockfile maintenance, the workflow (at least for me) is literally just approving it as long as everything looks roughly fine and only the lockfile is changed. There's no need for this to be separate. (Apparently the default is supposed to be just 1 group? not sure how it got split up)

PR #2920, opened 2024-11-27
chore: Remove deprecated setup.py and fix license

another-rex:
Remove setup.py as it is unnecessary, as pyproject.toml is a superset of it.

Also fix the license, not sure why it was labelled MIT, when all our license files and headers are Apache-2.0.


andrewpollock (approved):
/cc @calebbrown woohoo!

calebbrown:
Great work! 😁

oliverchang (approved):
nice! are we able to check if https://github.com/google/osv.dev/blob/master/.github/workflows/publish-to-pypi.yaml still works with this?

another-rex:
I manually ran the command (build) with setup.py deleted and it created the same files as when setup.py were present.

PR #2925, opened 2024-11-27
fix: oss-fuzz + exporter error

another-rex:
For some reason, recently the [EMPTY] ecosystem exporter started failing. The failure is because oss-fuzz special case in the to_vulnerability() code always assumes that there is a project name available. 

This is not the case for withdrawn, public, INVALID status entries, throwing an exception.

I'm not clear on why suddenly it started happening today, when the record has not been changed since 2021. It could have something to do with the Key change done recently, perhaps it was failing at an earlier error because the key was invalid. 

Added some additional logging as well to make it easier to debug the exact record that caused this error next time.

### DIFF
--- a/docker/exporter/exporter.py
+++ b/docker/exporter/exporter.py
@@ -102,16 +102,20 @@ class Exporter:
       files_to_zip = []
 
       @ndb.tasklet
-      def _export_to_file_and_zipfile(bug):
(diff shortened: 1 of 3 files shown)


hogo6002 (approved):
Thanks!

oliverchang (approved):

PR #2928, opened 2024-11-27
fix(worker): do not update source if it is unknown

cuixq:
Currently, `get_source_repository` returns `None` for unknown sources, which causes `AttributeError`.
To prevent this, this PR logs the source and then returns.


docker/worker/worker.py
```diff
@@ -333,6 +333,10 @@ def _source_update(self, message):
     deleted = message.attributes['deleted'] == 'true'
 
     source_repo = osv.get_source_repository(source)
+    if source_repo is None:
+      logging.info('Failed to get source repository %s', source)
```
another-rex, line 337: Should this be logging.error instead? I don't think we should be expecting invalid sources.

cuixq: yes!

another-rex (approved):

hogo6002 (approved):
Thanks!

docker/worker/worker.py
```diff
@@ -333,6 +333,10 @@ def _source_update(self, message):
     deleted = message.attributes['deleted'] == 'true'
 
     source_repo = osv.get_source_repository(source)
+    if source_repo is None:
+      logging.error('Failed to get source repository %s', source)
```
andrewpollock, line 337: To make this logged error more debuggable and actionable, it may be better to raise an exception to the caller, which will then be able to add the offending record's ID to the exception.

cuixq: yes - I also thought of adding ID to the error message so raising an exception is a good idea! I have made https://github.com/google/osv.dev/pull/2933 to address this.

PR #2933, opened 2024-11-28
fix(worker): raise an exception for invalid source

cuixq:
Follow up on https://github.com/google/osv.dev/pull/2928

Instead of logging the error, raise an exception so that caller is able to add record ID to the error.


andrewpollock (approved):
Thanks!

PR #2934, opened 2024-11-29
fix(faq): Revert custom header color and text settings for better contrast

hogo6002:
Our FAQ uses [Just the Docs](https://just-the-docs.com/docs/configuration/) dark mode, but we overwrote the default text and header colour, changing two slightly different colours to all white. This makes our `h3` tags difficult to find. This PR reverts the color scheme settings for headers and body text to improve visibility (slightly different colours between headers and body text).
image


andrewpollock (approved):
Thanks! Let's see how this looks...

another-rex (approved):

Issue #2793, opened 2024-10-28
fixit: reduce memory usage


oliverchang:
Profile and reduce memory usage across API, frontend.

cuixq:
Cloud Profile is set for test instance: https://github.com/google/osv.dev/pull/2941
And `build-essential` is required as per [documentation](https://cloud.google.com/profiler/docs/profiling-python): https://github.com/google/osv.dev/pull/2944

michaelkedar:
Cloud Profiler had to be disabled in #3827
Looking at the profiler, most of the problems were to do with the database structure, which is being worked on in #3850

PR #2936, opened 2024-11-29
feat(api): set up google cloud profiler

cuixq:
https://github.com/google/osv.dev/issues/2793

This PR attempts to set up Google Cloud Profiler following this [guide](https://cloud.google.com/profiler/docs/profiling-python).


another-rex:
Related https://github.com/google/osv.dev/pull/1305 . I've added google cloud profiler in that PR before, it didn't work that well when used with appengine, though maybe now that it is running in cloud run it will work better.

Looking at the commit history, it looks like you should use the wheel version: https://github.com/google/osv.dev/pull/1305/commits/5ef22aaab708df7fc91f0472c603caf3c7ea1018

cuixq:
since we are with Cloud Run instead of App Engine now, maybe let's try with the default version first?

michaelkedar (approved):

another-rex (approved):

gcp/api/server.py
```diff
@@ -1395,6 +1397,13 @@ def main():
     setup_gcp_logging('api-backend')
     logging.getLogger().addFilter(trace_filter)
 
+    # Profiler initialization. It starts a daemon thread which continuously
+    # collects and uploads profiles. Best done as early as possible.
+    try:
+      googlecloudprofiler.start(service="osv-api-profiler")
```
hogo6002, line 1403: Do you want to enable this only on the test instance first and then enable it on prod after verification? `if get_gcp_project() == _TEST_INSTANCE:`

cuixq: ah! thanks!

Issue #1714, opened 2023-10-10
Improve the osv.dev repo's folder structure


michaelkedar:
The repo is quite messy and it's hard to know where things are.
The current state of things (from my understanding):
| folder | what |
|--------|--------|
| `deployment/`| Terraform, Cloud Deploy & App Engine config files, a few Cloud Build yamls, old (no longer used?) `api-staging` and `api-test` Cloud Run configs  |
| `docker/` | CI docker files (`ci`, `deployment`, `terraform`), `indexer`, and `worker`/`importer`/`exporter` code & docker files, `worker-base` docker files |
| `docs/` | Jekyll files, [`build_swagger.py`](https://github.com/google/osv.dev/blob/2f2ed2cf9b46dc8a31eb71ebb84d5f55d915d1fa/docs/build_swagger.py), [`tools.go`](https://github.com/google/osv.dev/blob/2f2ed2cf9b46dc8a31eb71ebb84d5f55d915d1fa/docs/tools.go) |
| `gcp/api` | OSV API server files (including files for the local ESP), protobuf files in `/v1`|
| `gcp/appengine` | Mostly the backend of the osv.dev website, with the frontend in `frontend3`. Blog entires. App Engine Cron Handlers (to be removed). Also has the datastore index.yaml | 
| `gcp/functions` | The Cloud Function for publishing PyPI vulnerabilities, unmaintained |
| `osv/` | Main Python library, used in basically all Python services. Include ecosystem helpers and Datastore model definitions |
| `tools/` | Misc scripts/tools, mostly intended for development (datastore stuff, linting), but also the `indexer-api-caller` for indexer calling |
| `vulnfeeds/` | Go module for (mostly) the NVD CVE conversion, also the Alpine feed converter, and the Debian feed converter in `tools/debian` (which is written in Python) |

It'd be nice to untangle this to have a clearer structure for each component of OSV (especially the `docker/` stuff).

PR #2938, opened 2024-11-29
refactor: move workers and indexer out of `docker/` into `gcp/`

michaelkedar:
#1714 
Move the indexer to `gcp/indexer` and the python workers into `gcp/workers/`, so our deployed code is more together and separated from our ci stuff.

I also want to deal with the shared worker Dockerfile & poetry environment, but this PR already involved a lot of potentially risky renaming so I've started with this.


another-rex (approved):
LGTM, though it's a bit hard to tell whether things will break or not.

hogo6002 (approved):
LGTM. Maybe we can merge this after next week's release to allow for more bake time?

PR #2941, opened 2024-11-29
fix(api): only start cloud profiler for test instance

cuixq:
Only start the cloud profiler for test instance so that we can continue our investigation on this without interrupting production release.


hogo6002 (approved):
Thanks!

PR #2945, opened 2024-12-02
feat(worker): remove stale import findings for successful record

andrewpollock:
This adds functionality to the worker to clear any existing import findings for a record that previously failed to import, but now does. This avoids confusing historical findings existing for records that are successfully importing now.

Part of #2189 and #2188


docker/worker/worker.py
```diff
@@ -562,6 +563,14 @@ def _notify_ecosystem_bridge(self, vulnerability):
             push_topic,
             data=json.dumps(osv.vulnerability_to_dict(vulnerability)).encode())
 
+  def _maybe_remove_import_findings(self, vulnerability: osv.Bug):
+    """Remove any stale import findings for a successfully processed Bug,"""
+
+    finding = osv.ImportFinding.get_by_id(vulnerability.id())
+    if finding:
+      logging.info('Removing stale import finding for %s', vulnerability.id())
+      finding.key.delete()
```
hogo6002, line 572: Just a random thought, will we ever need to query historical finding data? For example, to evaluate how many records the linter helped fix?

andrewpollock: > will we ever need to query historical finding data?

In my opinion, it's more of a "nice to have" than a strong "need" in terms of requirement. Yes, it's helpful to be able to quantify impact. We can determine that from the logs for as long as we have log retention for...

oliverchang (approved):

hogo6002 (approved):

another-rex (approved):

PR #2967, opened 2024-12-09
fix(api): handle golang subpath

hogo6002:
Related issue: https://github.com/google/osv-scanner/issues/1428
Added support for Go module PURL strings containing `subpath`

### DIFF
--- a/gcp/api/server.py
+++ b/gcp/api/server.py
@@ -734,6 +734,19 @@ def do_query(query: osv_service_v1_pb2.Query,
           'Invalid PURL.',
       )
 
+    if package_name:  # Purls already include the package name
+      context.service_context.abort(
+          grpc.StatusCode.INVALID_ARGUMENT,
+          'name specified in a PURL query',
+      )
+
+    if ecosystem:
+      # Purls already include the ecosystem inside
+      context.service_context.abort(
+          grpc.StatusCode.INVALID_ARGUMENT,
+          'ecosystem specified in a PURL query',
+      )
+
     if purl is None:
       # TODO(gongh@): Previously, we didn't perform any PURL validation.
       # All unsupported PURL queries would simply return a 200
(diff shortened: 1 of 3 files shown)


michaelkedar (approved):

gcp/api/server.py
```diff
@@ -734,26 +734,28 @@ def do_query(query: osv_service_v1_pb2.Query,
...
           'name specified in a PURL query',
       )
+
     if ecosystem:
       # Purls already include the ecosystem inside
       context.service_context.abort(
           grpc.StatusCode.INVALID_ARGUMENT,
           'ecosystem specified in a PURL query',
       )
+
```
another-rex, line 749: What was the reason this is moved down?

another-rex (commented):
Hmm I wonder if the other ecosystems has subpaths, we might not need to special case go here.

hogo6002:
> Hmm I wonder if the other ecosystems has subpaths, we might not need to special case go here.

From the [PURL spec](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#golang), subpaths are only for golang and cocoapods

hogo6002:
> > Hmm I wonder if the other ecosystems has subpaths, we might not need to special case go here.
> 
> From the [PURL spec](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#golang), subpaths are only for golang and cocoapods

But one thing I'm concerned about is that each ecosystem might have special cases, like how Go has the `subpath` component. Due to the number of ecosystems we support and the differences in PURL formats between them, it's hard to guarantee that we're handling all the special cases correctly. To catch potential issues from API changes in the future, I think we need more integration tests for each ecosystem (we can put them in long tests and only enable on post-checkin pipeline). I created an [issue](https://github.com/google/osv.dev/issues/2968) to track this

another-rex (approved):

PR #2971, opened 2024-12-10
feat(API): update version matching for NuGet, PyPI and RubyGems

hogo6002:
[Malicious package](https://github.com/ossf/malicious-packages/tree/main/osv/malicious) publishes OSV records for crates.io, npm, NuGet, PyPI, and RubyGems. Queries for NuGet, PyPI, and RubyGems (npm and crates.io use semantic versioning, so the matching process is different) only match vulnerabilities against specific `affected versions`. However, malicious package records may only provide `affected ranges` instead of individual versions in some cases (e.g. https://api.osv.dev/v1/vulns/MAL-2022-7426). OSV also can't enumerate affected versions for a malicious package as those versions have been deleted. This causes issues like https://github.com/google/osv.dev/issues/2407

Switching the API query version matching from `_query_by_generic_version()` to [`_query_by_comparing_versions()`](https://github.com/google/osv.dev/blob/4d981692feb5e088d12177d97a44fe30701b3854/gcp/api/server.py#L1186) can address this issue. The `_query_by_comparing_versions()` function matches [both affected versions and affected ranges](https://github.com/google/osv.dev/blob/4d981692feb5e088d12177d97a44fe30701b3854/gcp/api/server.py#L1381), but might slow down the performance for a bit.

Adding this PR after the end-of-year release to give more time to verify performance on the test instance before rolling out to prod.

### DIFF
--- a/gcp/api/integration_tests.py
+++ b/gcp/api/integration_tests.py
@@ -378,6 +378,52 @@ class IntegrationTests(unittest.TestCase,
         timeout=_TIMEOUT)
     self.assertEqual(0, len(response.json()))
 
+  def test_malicious_package_matching(self):
+    """"Test malicious package query"""
+    # Test matching by affected ranges
+    mal_2022_7426 = self._get('MAL-2022-7426')
+
+    expected_vulns = [
+        mal_2022_7426,
+    ]
+
+    package = 'pymocks'
+    ecosystem = 'PyPI'
+
+    response = requests.post(
+        _api() + _BASE_QUERY,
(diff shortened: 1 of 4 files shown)


another-rex:
Can you update the title to be more about switching the 3 ecosystems to the new matching method? While this change is motivated by improving Malicious package matching, the actual change will primarily affects normal matching as well (if I understand correctly), would be good to clarify that.

another-rex (approved):

Issue #2963, opened 2024-12-05
Update Ubuntu package matching behavior


another-rex:
OSV-Scanner currently can only reliably get Ubuntu ecosystems with their release number (e.g. Ubuntu:22.04), but cannot get whether it is of the :LTS or :Pro variant without some vague string parsing.

Is there a difference in packages between Ubuntu:22.04 and Ubuntu:22.04:LTS at the same version number? E.g. is `socat` at version `1.7.4.1-3ubuntu1` the same between Ubuntu:22.04 and the LTS? 

If they are the same, for our API, it should be fine to strip :LTS when matching, so a query with ecosystem of Ubuntu:22.04 and Ubuntu:22.04:LTS will return the same results.

Also does a similar thing apply the :Pro variant as well.

@dodys might you be able to provide some answers here? :)

dodys:
@another-rex what you are referring to Ubuntu:22.04 is actually Ubuntu:22.04:LTS, we have no data that would have only Ubuntu:22.04, so I don't understand the difference you are seeing.


The :Pro variant is indeed an extra, as it is packages that don't go to the normal archive.

another-rex:
Thanks for the quick reply! It seems I misunderstood how Ubuntu releases work. I was assuming that a release starts off not marked as LTS and then becomes LTS, but it looks like LTS releases are started as LTS releases from the beginning, so this makes sense now.

For the :Pro variants, if we found a package matching the version specified on a advisory (e.g. `matrix-synapse@0.19.2+dfsg-6` from https://osv.dev/vulnerability/UBUNTU-CVE-2024-53863), on an Ubuntu:18.04 machine/image (unknown whether it is Pro or not), can we assume that the advisory applies to that package?

dodys:
@another-rex sorry for the late reply, I was out the previous days.

Pro is a bit trickier, therefore I will try to list every case below:
1. Pro only exists for LTS releases, therefore Interim releases like 25.10, there's no Pro variant.
2. For LTS releases that have already passed their 5 years of support, like 14.04, 16.04 and 18.04, you might see either the LTS or the Pro variant:
2.1. The LTS variant if the fix was published publicly;
2.2 The Pro variant if the fix was published under Pro (only for universe packages) or if there's no fix available yet (that's the case with the matrix-synapse CVE you mentioned for 18.04).
[...]

PR #2984, opened 2024-12-12
feat: expand Ubuntu ecosystem to include more variants

hogo6002:
Related issue https://github.com/google/osv.dev/issues/2963

OSV-Scanner mismatches Ubuntu vulnerabilities due to variations in ecosystem naming (e.g., :LTS, :Pro). 
Adds a function to expand the Ubuntu ecosystem list with all possible variants, ensuring accurate vulnerability matching.

TODO:

- [ ] reput all Ubuntu vulnerabilities
- [ ] support naming variations in API query

### DIFF
--- a/osv/ecosystems/_ecosystems.py
+++ b/osv/ecosystems/_ecosystems.py
@@ -144,6 +144,31 @@ def normalize(ecosystem_name: str):
   return ecosystem_name.split(':')[0]
 
 
+def add_matching_ecosystems(original_set: set[str]) -> set[str]:
+  """
+  For Linux distributions, some release versions may have different variants.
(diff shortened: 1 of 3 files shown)


another-rex (approved):
LGTM, though you might also need to update the API. This only allows the initial query to match successfully, but the API have additional checks within python to see if an affected field matches, and I'm not sure if that only checks version, or also checks ecosystem as well.

hogo6002:
> LGTM, though you might also need to update the API. This only allows the initial query to match successfully, but the API have additional checks within python to see if an affected field matches, and I'm not sure if that only checks version, or also checks ecosystem as well.

Good catch, I will fix the API part in a followup PR after I re-put test instance Ubuntu records

PR #2986, opened 2024-12-13
feat(api): match ecosystem names without variants

hogo6002:
Resolves https://github.com/google/osv.dev/issues/2963

Updates the API query to support variant-insensitive matching of affected package ecosystem names. A match is found if the package ecosystem name, after removing all variants, is the same as the queried ecosystem name.

TODO:
- [ ] Add integration tests after reput Ubuntu records on prod (after the end-of-year release freeze)

### DIFF
--- a/gcp/api/server.py
+++ b/gcp/api/server.py
@@ -983,7 +983,7 @@ def _is_version_affected(affected_packages,
 
     if ecosystem:
       # If package ecosystem has a :, also try ignoring parts after it.
-      if is_matching_package_ecosystem(affected_package.package, ecosystem):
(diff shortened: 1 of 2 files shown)


another-rex (commented):
Could we merge normalize and remove_variants? Or are they for two different purposes.

hogo6002:
> Could we merge normalize and remove_variants? Or are they for two different purposes.

I personally want to keep them separate for now. normalize is also used in the front end. I think it's good to have a separate function to identify the base ecosystem name (e.g., Debian) of each release (e.g., Debian:11).

another-rex (approved):

PR #2987, opened 2024-12-16
feat(api): add `/v1experimental/importfindings/{source}` API endpoint for linter findings

hogo6002:
Add new API endpoint `/v1experimental/importfindings/{source}` to return invalid records per source that were found by the linter in the importer.

Sample query (test instance data):
image

Part of https://github.com/google/osv.dev/issues/2189


gcp/api/server.py
```diff
@@ -345,6 +346,32 @@ def DetermineVersion(self, request, context: grpc.ServicerContext):
...
+      context.abort(grpc.StatusCode.INVALID_ARGUMENT, 'Source is required.')
+    if get_gcp_project() == _TEST_INSTANCE:
+      logging.info('Checking import finding for %s\n', source)
+
+    query = osv.ImportFinding.query(osv.ImportFinding.source == source)
+    import_findings: list[osv.ImportFinding] = query.fetch()
+    invalid_records = []
+    for finding in import_findings:
+      invalid_records.append(
+          importfinding_pb2.ImportFinding(
```
oliverchang, line 365: nit: can we add a `to_proto` method on `models.ImportFinding` instead? This puts the logic in a place closer to the actual model and is more consistent with how we do convert Bug to Vulnerability.

gcp/api/server.py
```diff
@@ -345,6 +346,32 @@ def DetermineVersion(self, request, context: grpc.ServicerContext):
     res = determine_version(request.query, context).result()
     return res
 
+  @ndb_context
+  @trace_filter.log_trace
+  def ImportFindings(self, request, context: grpc.ServicerContext):
+    """Return a list of `ImportFinding` for a given source."""
+    source = request.source
+    # TODO(gongh@): add source check
```
oliverchang, line 354: what source check do you mean? i.e. a source name that doesn't exist? 

It would be good to make this clearer in the comment.

gcp/api/v1/osv_service_v1.proto
```diff
@@ -166,4 +177,11 @@ service OSV {
       body: "query"
     };
   }
+
+  // Get import findings per source.
+  rpc ImportFindings(ImportFindingsParameters) returns (ImportFindingList) {
+    option (google.api.http) = {
+      get: "/v1/importfindings/{source}"
```
oliverchang, line 184: should make this v1experimental for now? 

In case we want to change/break this?

oliverchang (commented):
awesome! 

Do we want to add documentation now?

gcp/api/server.py
hogo6002: added

andrewpollock (approved):
LGTM. I'll see if I can add some basic documentation to this PR today or tomorrow.

oliverchang (approved):

PR #3003, opened 2024-12-19
fix: pin poetry requirements by hash

michaelkedar:
Added a requirements.txt file (with hashes) to use to install poetry in our Docker containers.


another-rex (approved):
Huh, I guess this also fixed the issue of centralising the poetry version definition as well.

PR #3047, opened 2025-01-10
feat(api): switch Debian version query to range matching

hogo6002:
The current Debian's vulnerability matching only checking the affected version. This is a problem for container scanning, as we only enumerate versions during the import from source. So, many vulnerabilities are mismatched if the package is slightly newer than the import time. For example, the `linux` package at version 6.1.112-1 shows about 490 vulnerabilities for Debian 12. but at version 6.1.119-1, it only finds 14 vulnerabilities.

Switching to range matching to check both the affected range and affected versions.

### DIFF
--- a/gcp/api/integration_tests.py
+++ b/gcp/api/integration_tests.py
@@ -575,13 +575,18 @@ class IntegrationTests(unittest.TestCase,
 
     self.assert_results_equal({'vulns': another_expected}, response.json())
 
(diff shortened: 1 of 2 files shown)


oliverchang (approved):

another-rex (approved):

hogo6002:
I will merge this PR in after this week's release

PR #3053, opened 2025-01-15
fix: packages now show in list when there isn't an affected version.

### LINKED ISSUE #3006: Packages with no affected range, but only affected versions, show up as Not Specified on osv.dev/list
**Describe the bug**
Packages with no affected range, but only affected versions, show up as Not Specified on osv.dev/list. This is most easily seen with MAL advisories.

**To Reproduce**
https://osv.dev/list?ecosystem=npm

**Expected behaviour**
It should show the package

jess-lowe:
Closes #3006 

Before:

image

After:

image

### DIFF
--- a/gcp/website/frontend_handlers.py
+++ b/gcp/website/frontend_handlers.py
@@ -693,27 +693,16 @@ def list_packages(vuln_affected: list[dict]):
   packages = []
 
   for affected in vuln_affected:
-    ranges = affected.get('ranges', [])
-    if not ranges:
-      if 'package' not in affected:
-        continue
+    if 'package' in affected:
       package_entry = affected['package']['ecosystem'] + '/' + affected[
           'package']['name']
       if package_entry not in packages:
         packages.append(package_entry)
-    else:
-      for affected_range in ranges:
-        if affected_range['type'] in ['ECOSYSTEM', 'SEMVER']:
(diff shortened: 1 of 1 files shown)


gcp/website/frontend_handlers.py
```diff
@@ -693,18 +693,27 @@ def list_packages(vuln_affected: list[dict]):
...
-          continue
-        package_entry = affected['package']['ecosystem'] + '/' + affected[
-            'package']['name']
-        if package_entry not in packages:
-          packages.append(package_entry)
-      elif affected_range['type'] == 'GIT':
-        parsed_scheme = strip_scheme(affected_range['repo'])
-        if parsed_scheme not in packages:
-          packages.append(parsed_scheme)
+    ranges = affected.get('ranges', [])
```
hogo6002, line 696: Thanks! Looks really good, just one thing to mention here: maybe we can combine the logic a bit to check if the `package` field exists first. For all non-Git packages, the record should provide the `affected.package` (For invalid records, we can't show anything here either. ). So, we only need to handle cases where `affected.package` is provided or the `affected.ranges` contains Git information.

jess-lowe: Done :)

hogo6002 (approved):
LGTM!

PR #3057, opened 2025-01-16
Added support for underscores in headers

### LINKED ISSUE #3055: Support Salesforce Integration by Allowing SFDC_STACK_DEPTH Header
**Is your feature request related to a problem? Please describe.**

Salesforce automatically appends the SFDC_STACK_DEPTH header to all outgoing HTTP requests. This header cannot be removed and causes 400 errors in scenarios where the receiving endpoint does not accept unexpected headers, even when the request is otherwise valid.

**Describe the solution you'd like**

Ignore/Whitelist SFDC_STACK_DEPTH header

jess-lowe:
Calls with headers including underscores should no longer be rejected. 

Fix: added the [--underscores_in_headers](https://cloud.google.com/endpoints/docs/openapi/specify-esp-v2-startup-options#security) flag in both prod and test instance. 

Closes #3055


oliverchang (approved):
thanks! were you able to test that this works locally?

jess-lowe:
> thanks! were you able to test that this works locally?

Yes, it works with the test server :)

PR #3059, opened 2025-01-17
fix: add error handling for invalid version query

hogo6002:
- Handle cases where version parsing fails.
- Show a more descriptive error message for invalid PURL.

### DIFF
--- a/gcp/api/server.py
+++ b/gcp/api/server.py
@@ -749,10 +749,10 @@ def do_query(query: osv_service_v1_pb2.Query,
   if purl_str:
     try:
       purl = purl_helpers.parse_purl(purl_str)
-    except ValueError:
+    except ValueError as e:
       context.service_context.abort(
           grpc.StatusCode.INVALID_ARGUMENT,
-          'Invalid PURL.',
+          f'{e}',
       )
 
     if package_name:  # Purls already include the package name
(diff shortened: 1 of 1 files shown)


gcp/api/server.py
```diff
@@ -1257,36 +1257,41 @@ def _query_by_comparing_versions(context: QueryContext, query: ndb.Query,
...
-      # Skips if the affected package name does not match
-      # the queried package name.
-      if package_name != package.name:
-        continue
-
-      if _is_affected(ecosystem, version, affected_package):
-        bugs.append(bug)
-        context.total_responses.add(1)
+  try:
+    while (yield it.has_next_async()):
```
michaelkedar, line 1261: Maybe put the `try` inside the `while` loop, so that an invalid record doesn't stop processing of all the following bugs

gcp/api/server.py
```diff
@@ -1257,36 +1257,41 @@ def _query_by_comparing_versions(context: QueryContext, query: ndb.Query,
...
+          continue
+
+        if _is_affected(ecosystem, version, affected_package):
+          bugs.append(bug)
+          context.total_responses.add(1)
+          break
+  except TypeError as e:
+    logging.warning(e)
+  except ValueError as e:
+    logging.warning(e)
```
michaelkedar, line 1293: [`logging.exception`](https://docs.python.org/3/library/logging.html#logging.Logger.exception) would be more informative to use here

e.g.
```python
except ValueError:
  logging.exception('failed to compare versions')
  ```
 logs the error message and the stack trace

michaelkedar, line 1293: nit: this can be combined as `except (TypeError, ValueError) as e:`

Though, maybe this should just catch every exception `except Exception as e:`

gcp/api/server.py
hogo6002: The failures are most likely due to the queried version. To reduce duplicate warning logs, I initially wanted to end the current version query after finding one error. But sure, I can keep the error checks within the while loop for now in case the error stems from a specific related bug.

michaelkedar (approved):

PR #3083, opened 2025-01-22
fix: manually install poetry in debian-convert

michaelkedar:
I missed this in #3003, and looks like poetry on apk is version 1, which broke when we updated to v2

Had to change the debian-convert image to build from the root of the repo to be able to grab the poetry requirements.txt from the docker directory.


jess-lowe (approved):
LGTM!

Issue #3082, opened 2025-01-22
Normalize package names for certain ecosystems


hogo6002:
We have a [`maybe_normalize_package_names`](https://github.com/google/osv.dev/blob/2267278b87bdec799505eb3b9e1e716dfd75554b/gcp/workers/worker/worker.py#L281) function in the worker that normalizes PyPI package names. To ensure accuracy, we should also add this function to the API server to normalize queried package names. This normalization issue may also occur in other ecosystems, and we should investigate further.

PR #3088, opened 2025-01-24
fix(api): normalize PyPI package names

hogo6002:
Partially resolves https://github.com/google/osv.dev/issues/3082

Normalize PyPI package names in API queries: Refactor `worker.maybe_normalize_package_names` to use shared code.


another-rex (commented):
Can you also replace the existing instance of this function as well?

oliverchang:
> Can you also replace the existing instance of this function as well?

+1 we should refactor https://github.com/google/osv.dev/blob/dca42d3ad4507ddac2b0103307805f5783927293/gcp/workers/worker/worker.py#L286 to use this.

osv/ecosystems/_ecosystems.py
```diff
@@ -184,3 +186,12 @@ def is_supported_in_deps_dev(ecosystem_name: str) -> bool:
 
 def map_ecosystem_to_deps_dev(ecosystem_name: str) -> str:
   return _OSV_TO_DEPS_ECOSYSTEMS_MAP.get(ecosystem_name)
+
+
+def maybe_normalize_package_names(package_name: str, ecosystem: str):
+  """Normalize package names as necessary."""
+  if ecosystem == "PyPI":
```
oliverchang, line 193: but: single quotes

oliverchang (approved):

another-rex (approved):

osv/ecosystems/_ecosystems.py
```diff
@@ -184,3 +186,12 @@ def is_supported_in_deps_dev(ecosystem_name: str) -> bool:
 
 def map_ecosystem_to_deps_dev(ecosystem_name: str) -> str:
   return _OSV_TO_DEPS_ECOSYSTEMS_MAP.get(ecosystem_name)
+
+
+def maybe_normalize_package_names(package_name: str, ecosystem: str):
```
oliverchang, line 191: late comment: also `-> str` ?

hogo6002: added

PR #3150, opened 2025-02-12
fix(debian-copyright-mirror): overhaul for more efficient run length

andrewpollock:
The k8s CronJob monitoring has highlighted that the debian-copyright-mirror job appears to be overrunning. The wget mirroring approach is grossly inefficient.

This replaces that with a more efficient parallel curl execution, using a YAML file conveniently available to construct the URLs to download.

It sacrifices the attempt at only updating changed files, as this isn't easily possible with curl.

The runtime is now < 10 minutes in local testing!

### DIFF
--- a/vulnfeeds/cmd/debian-copyright-mirror/debian-copyright-mirror.py
+++ b/vulnfeeds/cmd/debian-copyright-mirror/debian-copyright-mirror.py
@@ -1,0 +1,163 @@
+"""
+Download all of the copyright files for packages in main in Debian unstable.
+
+This:
+
+Parses https://metadata.ftp-master.debian.org/changelogs/filelist.yaml.xz 
+    to identify the files to retrieve
+Generates a curl configuration to download the URLs
+Executes curl to download the URLs in parallel
+"""
+
+import os
+import argparse
+import yaml
+import lzma
+import urllib.request
+import subprocess
+from typing import List, Optional, Dict
+
+
+class Error(Exception):
+  """General purpose error class."""
+
+
+def download_url_to_directory(url: str,
+                              directory: str,
+                              filename: Optional[str] = None) -> Optional[str]:
+  """
+    Downloads a URL to a specified directory.
+
+    Args:
+        url (str): The URL to download.
+        directory (str): The directory to save the file to.
+        filename (str, optional): The filename to use. If None, the filename
+                                   is extracted from the URL. Defaults to None.
+
+    Returns:
+        str: The full path to the downloaded file, or None on error.
+  """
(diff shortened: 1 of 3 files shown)


vulnfeeds/cmd/debian-copyright-mirror/debian-copyright-mirror.py
```diff
@@ -0,0 +1,163 @@
...
+
+Parses https://metadata.ftp-master.debian.org/changelogs/filelist.yaml.xz 
+    to identify the files to retrieve
+Generates a curl configuration to download the URLs
+Executes curl to download the URLs in parallel
+"""
+
+import os
+import argparse
+import yaml
```
michaelkedar, line 14: We need to install [PyYAML](https://pypi.org/project/PyYAML/) in the docker container for this.

It's probably overkill to make a whole poetry environment for this & install poetry in the Dockerfile just for one dependency. Maybe we could just make a requirements file (with hashes) like [we have to install poetry](https://github.com/google/osv.dev/blob/911a80f78eaa496c62a1eda08a76d78521c65c08/docker/poetry/requirements.in).

And hm, `gcr.io/google.com/cloudsdktool/google-cloud-cli:485.0.0-alpine` doesn't seem to have `pip` installed, (but it does have `venv`). Not really sure what the best course of action there is.

vulnfeeds/cmd/debian-copyright-mirror/debian-copyright-mirror.py
```diff
@@ -0,0 +1,163 @@
...
+  """
+    Execute curl with the supplied configuration in the specified directory.
+
+    Args:
+        configuration (str): path to configuration file.
+        directory (str): path to set current working directory to.
+ """
+
+  os.makedirs(directory)
+  _ = subprocess.run(
```
michaelkedar, line 135: nit: is there a reason to do `_ = subprocess.run(...)` instead of just `subprocess.run(...)`?

michaelkedar (changes requested):
Amazing improvement!

vulnfeeds/cmd/debian-copyright-mirror/debian-copyright-mirror.py
andrewpollock: I guess not, probably habit from writing so much Go. Deloused in [ce7bad7](https://github.com/google/osv.dev/pull/3150/commits/ce7bad7ee7ee8f670835a38b5e06aca774bfc0a3)

michaelkedar (approved):

PR #3154, opened 2025-02-13
test(vulnfeeds): make flaky tests skippable on Cloud Build

andrewpollock:
`cves.TestExtractGitCommit` and `cves.ValidateAndCanonicalizeLink` are consistently failing on Cloud Build for git.kernel.org, which makes me believe Cloud Build's egress IPs may have been blocked.

Rather than remove these tests entirely, this makes them skippable on Cloud Build, and skips them, to unblock PRs that are currently unmergeable due to these test failures.

### DIFF
--- a/vulnfeeds/cves/versions_test.go
+++ b/vulnfeeds/cves/versions_test.go
@@ -631,8 +631,7 @@ func TestExtractGitCommit(t *testing.T) {
 
 	for _, tc := range tests {
 		if _, ok := os.LookupEnv("BUILD_ID"); ok && tc.skipOnCloudBuild {
-			t.Logf("test %q: skipping on Cloud Build", tc.description)
-			continue
+			t.Skipf("test %q: running on Cloud Build", tc.description)
 		}
 		got, err := extractGitCommit(tc.inputLink, tc.inputCommitType)
 		if err != nil && !tc.expectFailure {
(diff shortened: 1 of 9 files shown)


vulnfeeds/cves/versions_test.go
```diff
@@ -1078,9 +1085,14 @@ func TestValidateAndCanonicalizeLink(t *testing.T) {
...
 			wantErr:           false,
+			skipOnCloudBuild:  true, // observing indications of IP denylisting as at 2025-02-13
 		},
 	}
 	for _, tt := range tests {
+		if _, ok := os.LookupEnv("BUILD_ID"); ok && tt.skipOnCloudBuild {
+			t.Logf("test %q: skipping on Cloud Build", tt.name)
+			continue
+		}
 		t.Run(tt.name, func(t *testing.T) {
```
michaelkedar, line 1096: This logic could probably be moved inside of `t.Run` and use [`t.Skipf`](https://pkg.go.dev/testing#T.SkipNow)

michaelkedar (approved):

vulnfeeds/cves/versions_test.go
andrewpollock: Thanks for the tip! Adjusted in https://github.com/google/osv.dev/pull/3154/commits/938da4e7c0265bea343ddb75065bf4619dc647e1

jess-lowe (approved):

Issue #734, opened 2022-09-21
Investigate if https://staticcheck.io/ can and should be added to the Go code checks


andrewpollock:
https://staticcheck.io/ is an additional option for Go code quality checking. Investigate if it's appropriate to add along with `go vet`.

andrewpollock:
@another-rex also mentioned https://golangci-lint.run/usage/linters/

andrewpollock:
Also related: https://github.com/golang/vuln/blob/master/checks.bash

another-rex:
In osv-scanner we use https://golangci-lint.run/ which includes staticcheck as one of the options. We might want to just copy over osv-scanner's config file and apply it to all the go code in this repo as well.

PR #3164, opened 2025-02-18
ci(golint): add golint-ci GitHub Action

andrewpollock:
This commit replicates OSV-Scanner's golangci-lint action configuration

Fixes: #734


another-rex (commented):
You also should copy over the golangci config yaml file. `.golangci.yaml`

Issue #1871, opened 2023-12-12
Don't log import latency every time the task times out


michaelkedar:
The workers log latency regardless of whether the task finished or not:
https://github.com/google/osv.dev/blob/01173bfd380841914bda13453d3a1853395924de/docker/worker/worker.py#L623-L628

This messes with our import latency SLO monitoring, since jobs repeatedly get processed & timed out until it's processed correctly or dead-lettered.

We should also add a subscription to the dead letter topic to monitor failed imports

PR #3170, opened 2025-02-19
fix(worker): only log import latency for non-time outs

andrewpollock:
This adjusts the latency logging behaviour so that it only happens when
a task completes successfully (as opposed to times out, and is subject to being retried).

Fixes: #1871


another-rex:
Hmm this PR has no files?

andrewpollock:
Oh I seem to have made this PR twice, this is dupe of #3167

Issue #3052, opened 2025-01-14
Support `upstream` field in OSV schema


oliverchang:
The next OSV schema release will add an `upstream` field: https://github.com/ossf/osv-schema/pull/312

We need to support this on OSV.dev's side.

oliverchang:
Putting this in your queue also @jess-lowe as another starter project :)

@hogo6002 FYI!

PR #3172, opened 2025-02-21
feat: add support for an 'upstream' field in the models

jess-lowe:
This PR aims to prepare the models.py file to support the 'upstream' field to be introduced by https://github.com/ossf/osv-schema/pull/312.  Part of addressing #3052

This will calculate all of the transitive upstream dependencies of a given CVE. Hierarchy will be calculated and determined by the frontend in a later PR.

### DIFF
--- a/gcp/workers/alias/upstream_computation.py
+++ b/gcp/workers/alias/upstream_computation.py
@@ -1,0 +1,104 @@
+#!/usr/bin/env python3
+# Copyright 2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""OSV Upstream relation computation."""
+
+import datetime
+from google.cloud import ndb
+
+import osv
+import osv.logs
+
+import logging
+
+
+def _compute_upstream(target_bug_id, bugs):
+  """Computes all upstream vulnerabilities for the given bug ID.
+  The returned list contains all of the bug IDs that are upstream of the
+  target bug ID, including transitive upstreams."""
+  visited = set()
+  target_bug_upstream = bugs[target_bug_id]
+  if not target_bug_upstream:
+    return []
+  to_visit = set(target_bug_upstream)
+  bug_ids = []
+  while to_visit:
+    bug_id = to_visit.pop()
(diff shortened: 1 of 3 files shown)


gcp/workers/alias/upstream_computation.py
```diff
@@ -0,0 +1,104 @@
...
+  # Use (> '' OR < '') instead of (!= '') / (> '') to de-duplicate results
+  # and avoid datastore emulator problems, see issue #2093
+  bugs = osv.Bug.query(ndb.OR(osv.Bug.upstream > '', osv.Bug.upstream < ''))
+
+  all_upstream_group = osv.UpstreamGroup.query()
+
+  # for every bug, check if it has an UpstreamGroup.
+  for bug in bugs:
+    # check if the db key is also a db_id in all_upstream_group
+    b = all_upstream_group.filter(osv.UpstreamGroup.db_id == bug.db_id)
```
hogo6002, line 90: what does the b stand for? can we have a better naming?

another-rex (commented):
Some minor code comments here, will discuss the rest in the design doc.

gcp/workers/alias/upstream_computation.py
```diff
@@ -0,0 +1,104 @@
...
+
+  new_group = osv.UpstreamGroup(db_id=bug_id)
+  new_group.upstream_ids = upstream_ids
+  new_group.last_modified = datetime.datetime.now()
+  new_group.put()
+
+
+def _update_group(upstream_group, upstream_ids: list):
+  """Updates the alias group in the datastore."""
+  if len(upstream_ids) < 1:
```
hogo6002, line 63: Does the upstream_ids include the bug itself? why delete the upstream group when only 1 upstream?

hogo6002 (approved):
LGTM!!!

osv/models.py
```diff
@@ -259,6 +259,8 @@ class Bug(ndb.Model):
   aliases: list[str] = ndb.StringProperty(repeated=True)
   # Related IDs.
   related: list[str] = ndb.StringProperty(repeated=True)
+  # Upstream IDs.
+  upstream: list[str] = ndb.StringProperty(repeated=True)
```
another-rex, line 263: Please rename this to upstream_raw as described in the doc.

another-rex: Can you also expand the docstring above this to explain that this is upstream ids directly from the import and does not include any transitive upstreams.

osv/models.py
```diff
@@ -905,6 +934,16 @@ class AliasDenyListEntry(ndb.Model):
   bug_id: str = ndb.StringProperty()
 
 
+class UpstreamGroup(ndb.Model):
+  """Upstream group for storing transitive upstreams of a Bug"""
```
another-rex, line 938: Can you expand the description here to explain why this is in a separate table rather than directly on the bug entry itself? 

I.e. to prevent additional race conditions, by having it in a separate table only worker will modify Bug's directly.

another-rex (approved):
LGTM, thanks!

PR #3193, opened 2025-02-27
feat(cron): Adding upstream calculation to the alias cron job

jess-lowe:
Change: moved alias_computation and upstream_computation into bash script to be run in the same cron job. 

Chose not to run in parallel out of concern for memory usage.


hogo6002 (approved):
LGTM

PR #3196, opened 2025-02-28
fix: importer not importing upstream values from sources 

jess-lowe:
As the internal data store name for upstream from external sources is 'upstream_raw', it needs to be set properly when received from the sources.


hogo6002 (approved):
LGTM!

PR #3197, opened 2025-02-28
fix(exporter): accept the current reality of run length

andrewpollock:
The exporter is reliably taking >50m to complete, based on recent runs.

This adjusts the monitoring accordingly.

### DIFF
--- a/deployment/clouddeploy/gke-workers/base/exporter.yaml
+++ b/deployment/clouddeploy/gke-workers/base/exporter.yaml
@@ -3,9 +3,9 @@ kind: CronJob
 metadata:
   name: exporter
   labels:
-    cronLastSuccessfulTimeMins: "90"
+    cronLastSuccessfulTimeMins: "180"
 spec:
-  schedule: "*/30 * * * *"
+  schedule: "*/60 * * * *"
   concurrencyPolicy: Forbid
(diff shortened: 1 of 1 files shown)


oliverchang (approved):

deployment/clouddeploy/gke-workers/base/exporter.yaml
```diff
@@ -3,9 +3,9 @@ kind: CronJob
 metadata:
   name: exporter
   labels:
-    cronLastSuccessfulTimeMins: "90"
+    cronLastSuccessfulTimeMins: "180"
 spec:
-  schedule: "*/30 * * * *"
```
hogo6002, line 8: I think it's fine to keep it every `*/30`, as we've set the concurrency policy to 'forbid'. Setting `*/30` gives us more frequent runs than `*/60` if the cron job takes just slightly over 60 minutes.

andrewpollock: Okay, d17ae0f7 leaves the run frequency at 30 minutes.

jess-lowe (approved):
😢

hogo6002 (approved):

another-rex (approved):

PR #3199, opened 2025-02-28
feat(exporter): add a safe_upload_single function

andrewpollock:
This adds a file size delta function to prevent the unintended overwriting of a file with a significantly smaller one.

It only uses it for the unified all.zip upload at present, as I'm not sure how to identify when to use it for the ecosystem-specific zip files as they seem to be bundled in with the individual record files.

### DIFF
--- a/gcp/workers/exporter/exporter.py
+++ b/gcp/workers/exporter/exporter.py
@@ -31,6 +31,7 @@ import osv.logs
 DEFAULT_WORK_DIR = '/work'
 
 DEFAULT_EXPORT_BUCKET = 'osv-vulnerabilities'
+DEFAULT_SAFE_DELTA_PCT = 10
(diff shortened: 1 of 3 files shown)


gcp/workers/exporter/exporter.py
```diff
@@ -146,6 +151,41 @@ def upload_single(bucket: Bucket, source_path: str, target_path: str):
     logging.exception('Failed to export: %s', e)
 
 
+def safe_upload_single(bucket: Bucket,
+                       source_path: str,
+                       target_path: str,
+                       safe_delta_pct: int = 10):
```
another-rex, line 157: nit: Maybe move the default up to a const at the top of the file.

another-rex (approved):

PR #3202, opened 2025-02-28
fix(vulnfeeds): be stricter about acceptable commits from URLs

andrewpollock:
This tightens up what `cves.Commit()` considers acceptable by validating the assumed commit hash from GitHub and GitLab URLs as matching a SHA1 regex.

This prevents tags and tags with ancestry specification from being erroneously treated as a commit hash.

e.g.
- https://github.com/curl/curl/commit/curl-7_50_2
- https://github.com/curl/curl/commit/curl-7_50_2~32

as seen in the likes of CVE-2016-7141

### DIFF
--- a/vulnfeeds/cves/versions.go
+++ b/vulnfeeds/cves/versions.go
@@ -549,6 +549,8 @@ func Commit(u string) (string, error) {
 		return "", err
 	}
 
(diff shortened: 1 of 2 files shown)


oliverchang (approved):

jess-lowe (approved):
Very interesting! LGTM

PR #3205, opened 2025-03-03
feat(test): add V8 advisory to test environment.

vsutedjo:
Add V8 advisory to the test environment.

As per internal discussions, we are adding a V8 advisory to OSV.
The Chromium team has a [policy](https://chromium.googlesource.com/chromium/src/+/refs/heads/main/docs/security/updates.md) that states that direct embedders of Chromium packages such as V8 are responsible for using an up-to-date version of V8 packages. The definition of that is that the V8 package is at most one week old.

Since these types of packages do not have their own identifiers and therefore no advisory could be created, enforcement of such policies was not possible. We now created a synthetic advisory generator that takes such time-based policies and converts them into regularily updated advisories that OSV can read:
https://github.com/google/chromium-policy-vulnfeed/blob/main/advisories/V8-advisory.json

### DIFF
--- a/source_test.yaml
+++ b/source_test.yaml
@@ -425,6 +425,5 @@
   human_link: 'https://github.com/google/chromium-policy-vulnfeed/blob/main/advisories/V8-advisory.json'
   link: 'https://github.com/google/chromium-policy-vulnfeed/blob/main/advisories/V8-advisory.json'
   editable: False
-  key_path: 'OSV'
   repo_username: 'git'
   strict_validation: True
--- a/source_test.yaml
+++ b/source_test.yaml
@@ -418,12 +418,13 @@
   type: 0
   ignore_patterns: [ '^(?!V8-advisory).*$' ]
   repo_url: 'https://github.com/google/chromium-policy-vulnfeed.git'
-  detect_cherrypicks: False
+  detect_cherrypicks: True
   extension: '.json'
   db_prefix: ['V8-']
   ignore_git: False
-  human_link: 'https://github.com/google/chromium-policy-vulnfeed/blob/main/advisories/V8-advisory.json'
-  link: 'https://github.com/google/chromium-policy-vulnfeed/blob/main/advisories/V8-advisory.json'
+  human_link: 'https://github.com/google/chromium-policy-vulnfeed/blob/main/advisories/'
+  
+  link: 'https://github.com/google/chromium-policy-vulnfeed/blob/main/advisories/'
   editable: False
   repo_username: 'git'
   strict_validation: True
--- a/source_test.yaml
+++ b/source_test.yaml
@@ -413,3 +413,18 @@
   repo_username: 'git'
   strict_validation: True
 
+- name: 'V8'
+  versions_from_repo: True
+  type: 0
+  ignore_patterns: [ '^(?!V8-advisory).*$' ]
+  repo_url: 'https://github.com/google/chromium-policy-vulnfeed.git'
+  detect_cherrypicks: False
+  extension: '.json'
+  db_prefix: ['V8-']
+  ignore_git: False
+  human_link: 'https://github.com/google/chromium-policy-vulnfeed/blob/main/advisories/V8-advisory.json'
+  link: 'https://github.com/google/chromium-policy-vulnfeed/blob/main/advisories/V8-advisory.json'
+  editable: False
+  key_path: 'OSV'
+  repo_username: 'git'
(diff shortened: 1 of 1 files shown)


source_test.yaml
```diff
@@ -413,3 +413,18 @@
...
+  ignore_patterns: [ '^(?!V8-advisory).*$' ]
+  repo_url: 'https://github.com/google/chromium-policy-vulnfeed.git'
+  detect_cherrypicks: False
+  extension: '.json'
+  db_prefix: ['V8-']
+  ignore_git: False
+  human_link: 'https://github.com/google/chromium-policy-vulnfeed/blob/main/advisories/V8-advisory.json'
+  link: 'https://github.com/google/chromium-policy-vulnfeed/blob/main/advisories/V8-advisory.json'
+  editable: False
+  key_path: 'OSV'
```
oliverchang, line 428:
```suggestion
```

This isn't needed for this source, and will break the ingestion.

oliverchang (commented):
awesome!

source_test.yaml
```diff
@@ -413,3 +413,17 @@
...
+- name: 'V8'
+  versions_from_repo: True
+  type: 0
+  ignore_patterns: [ '^(?!V8-advisory).*$' ]
+  repo_url: 'https://github.com/google/chromium-policy-vulnfeed.git'
+  detect_cherrypicks: False
+  extension: '.json'
+  db_prefix: ['V8-']
+  ignore_git: False
+  human_link: 'https://github.com/google/chromium-policy-vulnfeed/blob/main/advisories/V8-advisory.json'
```
jess-lowe, line 425:
```suggestion
  human_link: 'https://github.com/google/chromium-policy-vulnfeed/blob/main/advisories/'
  
```
To align with our validator for sourcerepositories, this link should point at the collection of advisories rather than a single one (regardless if there is only 1 advisory) :)

source_test.yaml
```diff
@@ -413,3 +413,17 @@
...
+  link: 'https://github.com/google/chromium-policy-vulnfeed/blob/main/advisories/V8-advisory.json'
```
jess-lowe, line 426:
```suggestion
  link: 'https://github.com/google/chromium-policy-vulnfeed/blob/main/advisories/'
```
Same as above.

source_test.yaml
```diff
@@ -413,3 +413,17 @@
   repo_username: 'git'
   strict_validation: True
 
+- name: 'V8'
+  versions_from_repo: True
+  type: 0
+  ignore_patterns: [ '^(?!V8-advisory).*$' ]
+  repo_url: 'https://github.com/google/chromium-policy-vulnfeed.git'
+  detect_cherrypicks: False
```
oliverchang, line 421:
```suggestion
  detect_cherrypicks: True
```

We need this to ensure we properly treat all branches outside of the specified fix commits as vulnerable.

oliverchang (approved):

oliverchang:
@vsutedjo there's some small issues to resolve:

```
Failed to validate loaded OSV entry: '2024-11-22T16:36:05+01:00' does not match '[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}(\\.[0-9]+)?Z'
```

from https://github.com/google/chromium-policy-vulnfeed/blob/760a43d6e72401432d9ef9f84ed0048457461b87/advisories/V8-advisory.json#L5

This needs to end in "Z" (i.e. be UTC).

Also, re the vulnerability ID, would it make sense to make it `"id": "V8-FRESHNESS"` instead of just `"id": "V8"` ?

vsutedjo:
> @vsutedjo there's some small issues to resolve:
> 
> ```
> Failed to validate loaded OSV entry: '2024-11-22T16:36:05+01:00' does not match '[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}(\\.[0-9]+)?Z'
> ```
> 
> from https://github.com/google/chromium-policy-vulnfeed/blob/760a43d6e72401432d9ef9f84ed0048457461b87/advisories/V8-advisory.json#L5
> 
> This needs to end in "Z" (i.e. be UTC).
> 
> Also, re the vulnerability ID, would it make sense to make it `"id": "V8-FRESHNESS"` instead of just `"id": "V8"` ?

Changed:

- published is now also UTC
- vuln-ID is V8-FRESHNESS

Issue #3161, opened 2025-02-16
Missing version in affected version list of the GIT section of CVE-2024-50382 and https://osv.dev/vulnerability/CVE-2024-50383


sha-cy:
**Describe the bug**
In the affected version range of the GIT section, there are a few missing versions of Botan 2.*, specifically 2.18 and 2.19 and their subversions.

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to ['CVE-2024-50382'](https://osv.dev/vulnerability/CVE-2024-50382) or ['CVE-2024-50383'](https://osv.dev/vulnerability/CVE-2024-50383)
2. Click on the GIT section of the affected libraries
3. Click on 2.*

**Expected behaviour**
a full list of the library 2.* versions

**Screenshots**

<img width="1254" alt="Image" src="https://github.com/user-attachments/assets/23d49686-e1bc-4b69-b0f0-b63eedb65461" />

<img width="1222" alt="Image" src="https://github.com/user-attachments/assets/0e3c3aab-7922-4290-8c49-bab0e24c7018" />

**Additional context**
I am reporting it here and not in NVD as the vulnerabilities are still under analysis there, so there are no assigned CPEs yet, and looking through the documentation, I saw no reason for those versions to not be included as well.

andrewpollock:
Confirmed:

```
$ comm -23 <(git ls-remote -t https://github.com/randombit/botan | awk '{ print $2 }' | fgrep tags/2. | sed 's!refs/tags/!!' | sort --version-sort) <(curl -s https://api.osv.dev/v1/vulns/CVE-2024-50382 | jq -r 
'.affected[] | select(.ranges[].type == "GIT") | .versions[]' | grep '^2.' | sort --version-sort)
2.17.1
2.17.2
2.17.3
2.18.0
2.18.1
2.18.2
2.19.0
2.19.1
2.19.2
2.19.3
2.19.4
2.19.5
```

```
$ comm -23 <(git ls-remote -t https://github.com/randombit/botan | awk '{ print $2 }' | fgrep tags/2. | sed 's!refs/tags/!!' | sort --version-sort) <(curl -s https://api.osv.dev/v1/vulns/CVE-2024-50383 | jq -r 
'.affected[] | select(.ranges[].type == "GIT") | .versions[]' | grep '^2.' | sort --version-sort)
2.17.1
2.17.2
2.17.3
2.18.0
2.18.1
2.18.2
2.19.0
2.19.1
2.19.2
[...]

PR #3207, opened 2025-03-04
feat: Add `consider_all_branches` to SourceRepository.

oliverchang:
This decouples the behaviour of considering all branches from `detect_cherrypicks`.

Context: We need to analyze all branches for the newly added source in #3205, but V8 is a huge repo and `detect_cherrypicks` is too expensive (taking over 2 hours).

Also add some optimisations for dealing with large repos:

- Pre-compute branches that contain relevant commits when not detecting cherrypicks to avoid walking those branches.
- Short circuit branch walking when we've previously seen a commit before (and there was no start commit to bound it).


another-rex (approved):

andrewpollock:
Can you add support for this to analyzetool while you're at it? Will this be of any benefit to that situation we were talking about the other week with https://github.com/google/osv.dev/issues/3161 ?

osv/impact.py
```diff
@@ -120,8 +120,9 @@ def ranges(self):
 class RepoAnalyzer:
   """Repository analyzer."""
```
andrewpollock, line 121: There's quite a bit of nuance going on in `get_affected`, potentially antithetical to the constructor. I think it would be worthwhile for readability/comprehensibility to surface this either in the class docstring or the constructor's docstring

oliverchang: done!

oliverchang:
> Can you add support for this to analyzetool while you're at it? Will this be of any benefit to that situation we were talking about the other week with #3161 ?

Thanks for the callout. I've added support for this to the analyze tool. 

Re #3161, partially yes. We can't enable this at scale though, until we are confident we're not going to report a lot of false positives in branches where we don't have the fix recorded.

jess-lowe (approved):

PR #3208, opened 2025-03-04
feat(frontend): Upstream and Downstream hierarchy display on frontend

jess-lowe:
Added the logic to display the hierarchy of the upstream and downstream vuln entries. 

Example of top, middle, and bottom levels of vulns:

image

image

image


gcp/website/frontend_handlers.py
```diff
@@ -786,3 +803,201 @@ def cvss_calculator_url(severity):
...
+        output_lines.append("<li><a href=\"/vulnerability/" + vuln_id + "\">" +
+                            vuln_id + " </a></li>")
+      else:
+        output_lines.append("<li>" + vuln_id + "</li>")
+
+    if vuln_id in graph:
+      for child in graph[vuln_id]:
+        if child != target_bug_id:
+          output_lines.append("<ul>")
+          print_subtree(child)
```
oliverchang, line 840: does this handle cycles?

jess-lowe: Just added a change that will prevent the cycle from displaying, and will log an error, but won't prevent the whole vuln from being displayed.

gcp/website/frontend_handlers.py
```diff
@@ -302,9 +303,25 @@ def bug_to_response(bug, detailed=True):
...
+    add_stream_info(bug, response)
+
   return response
 
 
+def add_stream_info(bug, response):
+  """Add upstream hierarchy information to `response`."""
+  # Check whether there are upstreams
+  if bug.upstream_raw:
+    upstream_hierarchy_string = get_upstreams_of_vulnerability(bug.db_id)
```
hogo6002, line 315: Maybe we can use response['upstream'], which already contains upstream information from the upstream_group. Passing response['upstream'] into the function can save a query.

jess-lowe: As far as I can tell, response['upstream'] doesn't exist. and bug only has bug.upstream_raw, which we can't rely on to be the full set of transitive upstreams to feed to the hierarchy string generator.

hogo6002: Yeah `bug` itself doesn't contain the final upstream, but `vulnerability` does. The [response](https://github.com/google/osv.dev/blob/cbde8081cd2f5aa544bc7f820a459845947509d3/gcp/website/frontend_handlers.py#L293) here is converted from `vulnerability`. We did queries for upstream and aliases inside the [to_vulnerability()](https://github.com/google/osv.dev/blob/cbde8081cd2f5aa544bc7f820a459845947509d3/osv/models.py#L754) function.

jess-lowe: True! Optimized using this - thanks!

gcp/website/frontend_handlers.py
```diff
@@ -302,9 +303,25 @@ def bug_to_response(bug, detailed=True):
   if detailed:
     add_links(response)
     add_source_info(bug, response)
+    add_stream_info(bug, response)
```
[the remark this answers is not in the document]
jess-lowe: This depends on the number of hierarchical upstreams primarily.
A SUSE entry with 3 immediate upstreams takes about 1.8 seconds, out of the 3.5 total in the `bug_to_response` section. 
A USN entry with a bunch of hierarchical upstreams take 10+ seconds - which isn't ideal. 
CVE entries with no upstreams - only downstream - take ~ 1.2 seconds. 

It's really the `get_multi` query that's hitting it hard. As discussed offline, I'll look into moving this computation into the cron job, and we'll see in staging how querying from a closer data center might affect query speeds.

hogo6002 (approved):
LGTM for now. I also agree with adding the hierarchy calculation to the cron job itself. The code logic is a bit complicated and mixing HTML with Python code makes it hard to maintain. But considering we only apply this to the staging instance, I think it's okay to merge for performance monitoring. We can then probably separate the hierarchy calculation part into the cron job if we see performance issues.

jess-lowe:
Will merge this PR to test performance in staging after this week's release. Changes should only affect staging. Follow up PRs will address performance if necessary.

PR #3209, opened 2025-03-04
feat(source): Add V8 to prod.

vsutedjo:
Following up on #3205, we are adding V8 to the prod environment.

As per internal discussions, we are adding a V8 advisory to OSV.
The Chromium team has a [policy](https://chromium.googlesource.com/chromium/src/+/refs/heads/main/docs/security/updates.md) that states that direct embedders of Chromium packages such as V8 are responsible for using an up-to-date version of V8 packages. The definition of that is that the V8 package is at most one week old.

Since these types of packages do not have their own identifiers and therefore no advisory could be created, enforcement of such policies was not possible. We now created a synthetic advisory generator that takes such time-based policies and converts them into regularily updated advisories that OSV can read:
https://github.com/google/chromium-policy-vulnfeed/blob/main/advisories/V8-advisory.json


oliverchang:
> I wonder if we should have been able to see the V8 record here? https://test.osv.dev/vulnerability/V8-FRESHNESS

This is pending some optimisations on our end: https://github.com/google/osv.dev/pull/3207. 

Also, let's hold off on merging this until we validate that the test instance behaves as expected with the V8 record.

oliverchang:
This is now successfully imported into the test instance: https://test.osv.dev/vulnerability/V8-FRESHNESS

vsutedjo:
Agree. #3224 fixes the source link.

oliverchang:
We'll merge this Monday AU time, to avoid causing potential issues over the weekend.

PR #3229, opened 2025-03-07
feat: added a way to locally test frontend changes without connecting to GCP

jess-lowe:
@michaelkedar and I created figured out how to utilize the datastore emulator to mock the data from locally rather than connecting to staging, so that small frontend tests, like checking behaviour consequences of bad data, can be mocked without introducing bad data in staging. 

Future PRs could introduce a better data set and method: for example, a public bucket or being able to upload a zip of OSV vulns to test on. 

For now this suits the purpose of testing behaviour when there are cycles in the graph.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,11 +52,10 @@ run-website-staging:
 	cd gcp/website/blog && hugo --buildFuture -d ../dist/static/blog
 	cd gcp/website && $(install-cmd) && GOOGLE_CLOUD_PROJECT=oss-vdb-test $(run-cmd) python main.py
 
-
 run-website-emulator:
 	cd gcp/website/frontend3 && npm install && npm run build
 	cd gcp/website/blog && hugo --buildFuture -d ../dist/static/blog
-	cd gcp/website && $(install-cmd) && GOOGLE_CLOUD_PROJECT=test-osv DATASTORE_EMULATOR_PORT=5002 $(run-cmd) python frontend_emulator.py
+	cd gcp/website && $(install-cmd) && DATASTORE_EMULATOR_PORT=5002 $(run-cmd) python frontend_emulator.py
 
 # Run with `make run-api-server ARGS=--no-backend` to launch esp without backend.
 run-api-server:
--- a/Makefile
+++ b/Makefile
@@ -52,6 +52,12 @@ run-website-staging:
 	cd gcp/website/blog && hugo --buildFuture -d ../dist/static/blog
 	cd gcp/website && $(install-cmd) && GOOGLE_CLOUD_PROJECT=oss-vdb-test $(run-cmd) python main.py
 
+
+run-website-emulator:
+	cd gcp/website/frontend3 && npm install && npm run build
+	cd gcp/website/blog && hugo --buildFuture -d ../dist/static/blog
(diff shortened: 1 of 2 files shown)


jess-lowe:
> What do you need to do to get the frontend / API to use this datastore emulator when you start it up? Will it automatically use the emulator if it is running?

Yeah, so if you just run `make run-website-emulator` instead of `run-website-staging` or `run-website` it will use the datastore emulator context instead for the frontend :)

gcp/website/frontend_emulator.py
```diff
@@ -0,0 +1,358 @@
...
+if __name__ == '__main__':
+  ds_emulator = tests.start_datastore_emulator()
+  import main
+  try:
+    with ndb.Client().context() as context:
+      context.set_memcache_policy(False)
+      context.set_cache_policy(False)
+      setUp()
+      main.app.run(host='127.0.0.1', port=8000, debug=False)
+    # main.app.run()
```
michaelkedar, line 356: nit: remove commented line
I also think `app.run` can go outside the `with ... as context`

gcp/website/frontend_emulator.py
```diff
@@ -0,0 +1,358 @@
...
+  osv.UpstreamGroup(
+      id='CYCLE-ROOT-2',
+      db_id='CYCLE-ROOT-2',
+      upstream_ids=['CYCLE-ROOT-1'],
+      last_modified=datetime.datetime(2023, 8, 14)).put()
+
+
+if __name__ == '__main__':
+  ds_emulator = tests.start_datastore_emulator()
+  import main
```
michaelkedar, line 349: Maybe add a comment to explain we need to start the emulator before importing main to make the global ndb client use the emulator.

michaelkedar (approved):

michaelkedar:
> > What do you need to do to get the frontend / API to use this datastore emulator when you start it up? Will it automatically use the emulator if it is running?
> 
> Yeah, so if you just run `make run-website-emulator` instead of `run-website-staging` or `run-website` it will use the datastore emulator context instead for the frontend (and launch the frontend, like how `run-website` does) :)

For reference, `start_datastore_emulator` sets [some environment variables](https://github.com/google/osv.dev/blob/cbde8081cd2f5aa544bc7f820a459845947509d3/osv/tests.py#L110-L111) that makes ndb talk to the emulator

Makefile
```diff
@@ -52,6 +52,12 @@ run-website-staging:
 	cd gcp/website/blog && hugo --buildFuture -d ../dist/static/blog
 	cd gcp/website && $(install-cmd) && GOOGLE_CLOUD_PROJECT=oss-vdb-test $(run-cmd) python main.py
 
+
+run-website-emulator:
+	cd gcp/website/frontend3 && npm install && npm run build
+	cd gcp/website/blog && hugo --buildFuture -d ../dist/static/blog
+	cd gcp/website && $(install-cmd) && GOOGLE_CLOUD_PROJECT=test-osv DATASTORE_EMULATOR_PORT=5002 $(run-cmd) python frontend_emulator.py
```
michaelkedar, line 59: nit: setting `GOOGLE_CLOUD_PROJECT` is probably unnecessary, since `start_datastore_emulator` should do this.

hogo6002 (approved):
Thanks!

another-rex (approved):
Thanks! This is great

PR #3247, opened 2025-03-12
fix: Work around markdown2 incorrect sanitisation.

another-rex:
Added a fragile regex to fix V8's git links from being broken by the markdown2 sanitiser. Created https://github.com/trentm/python-markdown2/issues/621 upstream to hopefully address the issue.


oliverchang (approved):
thanks for the workaround and the investigation!

hogo6002 (approved):

michaelkedar (approved):

PR #3252, opened 2025-03-13
fix: upgrade `packaging` version

### LINKED ISSUE #1271: Use packaging-legacy
Context: https://github.com/google/osv.dev/pull/1270#issuecomment-1534040942

If we use this, we can remove the <22.0 pin for our "packaging" package.

michaelkedar:
Upgrading to Python 3.12 on my computer seems to have broken the API - some problem with `packaging.utils.canonicalize_version` using something that was deprecated.

OSV.dev is still using Python 3.11, so this hasn't been a problem (yet).

Upgrading `packaging` to the newest version seems to fix this. The main osv library was constraining it to the older version because of #1270 - I've just changed it to use `packaging-legacy`, which fixes #1271 (and required re-locking every poetry.lock file depending on it).


oliverchang (approved):
thanks!

PR #3280, opened 2025-03-21
refactor: Move upstream calculation from frontend to cron job

jess-lowe:
We had some concerns about performance of the hierarchy calculation being in the frontend so this should improve the frontend performance by precalculating the hierarchy in the cron job instead. I would include some metrics on how much performance is improved here, but the data needs to be in staging to get this data. 

Upstream hierarchy is stored in JSON format in 'reverse' order e.g. vuln points at all its upstreams. 

Upstream hierarchy must be calculated **AFTER** all transitive upstream vulns in order to have the complete updated set of transitive upstreams to calculate from.

Downstream hierarchy calculation will remain in the frontend as the number of queries and time taken to calculate the hierarchy is negligible, compared to the impact it would have on the cron job - there's no way of checking for all of the vulns that exist as downstreams so **every** `Bug` would need to be checked each time the cron job was run, as opposed to the upstreams that we can check for the presence of. 

Other minor improvements include adding the 'expandible list' class to the upstream and downstream frontend elements so that the list doesn't overwhelm the page.

Relates to frontend for #3052


gcp/workers/alias/upstream_hierarchy_computation.py
```diff
@@ -0,0 +1,95 @@
...
+  raise TypeError
+
+
+def main():
+  """ Updates all upstream groups in the datastore with 
+   their upstream hierarchy relationships """
+  # Find all of the upstream groups
+  all_upstream_group = osv.UpstreamGroup.query()
+
+  for group in all_upstream_group:
```
hogo6002, line 86: this recomputes all upstream hierarchy every run, right? Maybe we can use the last modified time to check if the upstream group was updated? If not, we don't need to compute its hierarchy.

gcp/workers/alias/upstream_hierarchy_computation.py
```diff
@@ -0,0 +1,95 @@
...
+      continue
+    upstream_map[target_bug.db_id] = upstream_map[target_bug.db_id] - v
+  if target_bug.upstream_hierarchy == upstream_map:
+    return upstream_map
+  if upstream_map:
+    target_bug.upstream_hierarchy = json.dumps(
+        upstream_map, default=set_default)
+    target_bug.last_modified = datetime.datetime.now()
+    target_bug.put()
+  return upstream_map
```
hogo6002, line 71: do we still need this return statement? it seems we are not using it from the caller

gcp/workers/alias/upstream_hierarchy_computation.py
```diff
@@ -0,0 +1,95 @@
...
+import datetime
+from google.cloud import ndb
+
+import osv
+import osv.logs
+import json
+
+
+def compute_upstream_hierarchy(target_bug: osv.UpstreamGroup,
+                               bug_groups) -> dict[str, set[str]]:
```
another-rex, line 26: should this be named all_upstream_group and target_upstream_group? bug_group is a bit confusing imo.

another-rex: nit: Can you also add `ndb.Query` type to bug_groups?

The documentation for the type in the docstring is really nice though!

gcp/workers/alias/upstream_hierarchy_computation.py
```diff
@@ -0,0 +1,95 @@
...
+  """
+  visited = set()
+  upstream_map = {}
+  to_visit = set([target_bug.db_id])
+  while to_visit:
+    bug_id = to_visit.pop()
+    if bug_id in visited:
+      continue
+    visited.add(bug_id)
+    bug = bug_groups.filter(osv.UpstreamGroup.db_id == bug_id).get()
```
another-rex, line 43: What is the type of this variable, and is bug the correct name or should this be upstream_group?

gcp/workers/alias/upstream_hierarchy_computation.py
```diff
@@ -0,0 +1,95 @@
...
+          upstream_ids: str[bug_ids]
+          last_modified_date}
+  """
+  visited = set()
+  upstream_map = {}
+  to_visit = set([target_bug.db_id])
+  while to_visit:
+    bug_id = to_visit.pop()
+    if bug_id in visited:
+      continue
```
another-rex, line 41: nit: for the end of blocks, can you add a new line after it to make it easier to read?

gcp/website/frontend_emulator.py
```diff
@@ -362,6 +364,9 @@ def setUp():
       upstream_ids=['CYCLE-ROOT-1'],
       last_modified=datetime.datetime(2023, 8, 14)).put()
 
+  upstream_computation.main()
+  upstream_hierarchy_computation.main()
```
another-rex, line 368: Really cool idea just calling the computation in the emulator!

gcp/workers/alias/upstream_hierarchy_computation.py
```diff
@@ -0,0 +1,95 @@
...
+    for upstream in upstreams:
+      if upstream not in visited and upstream not in to_visit:
+        to_visit.add(upstream)
+      else:
+        if bug_id not in upstream_map:
+          upstream_map[bug_id] = set([upstream])
+        else:
+          upstream_map[bug_id].add(upstream)
+      upstream_map[bug_id] = upstreams
+      to_visit.update(upstreams - visited)
```
another-rex, line 59: Is this line in the right indentation? It looks like you are adding the entire "upstreams" to to_visit every loop of upstreams, which doesn't quite look right...

jess-lowe: Good catch! Probably an artifact from moving everything over 😅 Fixed :)

gcp/workers/alias/upstream_hierarchy_computation.py
jess-lowe: As discussed offline, moved this logic back into `upstream_computation.py` and now it's only computed if the group is updated when the transitive upstream is calculated :)

hogo6002: Thanks!

another-rex (approved):
Suggest adding some additional comments to compute_upstream_hierarchy, specifically for each for loop to explain what it is doing to make it more readable when looking at this in the future. Otherwise LGTM, thanks!

gcp/website/frontend_handlers.py
```diff
@@ -869,8 +867,22 @@ def print_subtree(vuln_id: str) -> None:
...
+  for node, children in graph.items():
+    for child in children:
+      if child not in reversed_graph:
+        reversed_graph[child] = set()
+      reversed_graph[child].add(node)
+
+  return reversed_graph
+
+
+def get_upstreams_of_vulnerability(bug) -> str:
```
hogo6002, line 885: nit: should the return type be `-> str | None`? I see it returns None when cycle detected

gcp/workers/alias/upstream_computation.py
```diff
@@ -76,30 +79,101 @@ def _update_group(upstream_group, upstream_ids: list):
...
+                               all_upstream_groups: ndb.Query):
+  """Computes all upstream vulnerabilities for the given bug ID.
+  The returned list contains all of the bug IDs that are upstream of the
+  target bug ID, including transitive upstreams in a map hierarchy.
+  upstream_group:
+        { db_id: bug id
+          upstream_ids: str[bug_ids]
+          last_modified_date: date
+          upstream_hierarchy: JSON str
+          }
```
hogo6002, line 92: nit: can we format it properly?

gcp/workers/alias/upstream_computation.py
```diff
@@ -76,30 +79,101 @@ def _update_group(upstream_group, upstream_ids: list):
...
+    upstream_map[target_upstream_group
+                 .db_id] = upstream_map[target_upstream_group.db_id] - v
+
+  if target_upstream_group.upstream_hierarchy == upstream_map:
+    return
+
+  if upstream_map:
+    target_upstream_group.upstream_hierarchy = json.dumps(
+        upstream_map, default=set_default)
+    target_upstream_group.last_modified = datetime.datetime.now()
```
hogo6002, line 134: I don't think we'll see this happen, but in theory, there could be a race condition where `bug.upstream_raw` updates between when we update `upstream_group` and `upstream_hierarchy`. I think it's better to not update the upstream's `last_modified` time here, since we already updated it when we updated `upstream_group`. And that modified time is only used to see if upstream_group needs to change.

gcp/workers/alias/upstream_computation.py
```diff
@@ -76,30 +79,101 @@ def _update_group(upstream_group, upstream_ids: list):
...
+    target_upstream_group.put()
+
+
+def set_default(obj):
```
hogo6002, line 138: maybe add a doc string? or move it inside of the compute_upstream_hierarchy function?

hogo6002 (approved):
LGTM with some minor style comments. Thanks Jess! hope this can improve our performance

PR #3282, opened 2025-03-24
chore(vulnfeeds): resume tracking the latest google-cloud-cli image

andrewpollock:
Returning to the generic `alpine` tag will enable Renovate to be able to upgrade this image periodically.

This pins to the current latest Alpine google-cloud-cli Docker image (515.0.0, from 2025-03-18).

We had previously pinned to 485.0.0 (from 2024-07-23), due to performance regressions between `gcloud storage rsync` versus `gsutil rsync`.

That version is:

- getting very old
- showing signs of bitrotting with newer Python runtimes
- potentially hiding deprecations that may surprise us

I believe we have re-converged back on `gsutil` for everything currently, and with the additional k8s CronJob monitoring now in place, we should pick up any future upgrades that introduce any new performance regressions.

### DIFF
--- a/vulnfeeds/cmd/alpine/Dockerfile
+++ b/vulnfeeds/cmd/alpine/Dockerfile
@@ -25,7 +25,7 @@ COPY ./ /src/
 RUN go build -o alpine-osv ./cmd/alpine/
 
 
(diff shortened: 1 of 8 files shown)


another-rex (approved):
Let's wait till after this week's release before merging, and see if we run into problems in staging?

hogo6002 (approved):

jess-lowe (approved):

andrewpollock:
> Let's wait till after this week's release before merging, and see if we run into problems in staging?

Yep, that was my intention

another-rex:
Should we merge this now?

PR #3284, opened 2025-03-24
fix(exporter): increase CPU limit for exporter

andrewpollock:
I *think* the exporter is running into its CPU limit and getting evicted as a result, and the CPU limit seems to be unnecessarily low, given it's running on a highend node with 32 CPUs?

### DIFF
--- a/deployment/clouddeploy/gke-workers/base/exporter.yaml
+++ b/deployment/clouddeploy/gke-workers/base/exporter.yaml
@@ -32,7 +32,7 @@ spec:
                 cpu: "6"
                 memory: "48G"
               limits:
-                cpu: "6"
+                cpu: "32"
                 memory: "128G"
           restartPolicy: OnFailure
           volumes:
--- a/deployment/clouddeploy/gke-workers/base/exporter.yaml
+++ b/deployment/clouddeploy/gke-workers/base/exporter.yaml
@@ -29,10 +29,10 @@ spec:
                 name: "ssd"
             resources:
               requests:
-                cpu: "6"
+                cpu: "12"
                 memory: "48G"
               limits:
-                cpu: "32"
+                cpu: "24"
                 memory: "128G"
           restartPolicy: OnFailure
           volumes:
--- a/.github/workflows/staleness.yml
+++ b/.github/workflows/staleness.yml
@@ -20,6 +20,7 @@ jobs:
           operations-per-run: 100
           remove-stale-when-updated: true
           exempt-issue-labels: "good first issue,backlog"
+          exempt-pr-labels: "backlog"
           exempt-all-assignees: true
           ignore-updates: false
           stale-issue-label: stale


deployment/clouddeploy/gke-workers/base/exporter.yaml
```diff
@@ -32,7 +32,7 @@ spec:
                 cpu: "6"
```
oliverchang, line 32: should we also bump requests to something like 25? to avoid starving the node of resources. 

IIRC we can't put 32 here because there are required pods/overhead on these nodes which make the total available CPUs always less than what the machine actualy has.

another-rex:
Hmm I picked 6 because that is how many ecosystems the exporter_runner is running in parallel. Since each exporter can only run on a single thread I'm not sure increasing the CPU count directly will make that much of an impact. We should still try though.

 > IIRC we can't put 32 here because there are required pods/overhead on these nodes which make the total available CPUs always less than what the machine actualy has.
 
+1, let's pick a limit of 24 or something, and request maybe 12?

andrewpollock:
> Hmm I picked 6 because that is how many ecosystems the exporter_runner is running in parallel.

I'm not following the 6? Am I misinterpreting https://github.com/google/osv.dev/blob/6d7c3dc69bc05e7e1a83f4389e61d430db288416/gcp/workers/exporter/export_runner.py#L44-L48

and

https://github.com/google/osv.dev/blob/6d7c3dc69bc05e7e1a83f4389e61d430db288416/gcp/workers/exporter/export_runner.py#L33

deployment/clouddeploy/gke-workers/base/exporter.yaml
andrewpollock: Agreed, 32 will never be feasible (it;'s 31.something based on what is being reported in the dashboard), but it's the limit, not the request.

andrewpollock:
> Ah, no you are not misinterpreting it, I think it's suppose to say 6 in export_runner, I think I just forgot to change that number when adjusting the number of cores: [#2167 (comment)](https://github.com/google/osv.dev/pull/2167#issuecomment-2097502570)

Right, but unless I'm missing something, that constant isn't factoring into it?

```
$ kubectl --context production exec -it exporter-29047680-vl5qj -- /usr/bin/python3
Python 3.8.10 (default, Mar 11 2025, 17:45:31) 
[GCC 9.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> DEFAULT_EXPORT_PROCESSES=7
>>> os.cpu_count() or DEFAULT_EXPORT_PROCESSES
32
```

another-rex:
🤦 Of course, yes this will just use the given number of CPUs. Huh I was intending for it to use either the request amount or the limit amount, but it looks like `os.cpu_count()` just gets the number of cores on the host machine...

oliverchang (approved):

PR #3298, opened 2025-03-26
fix(upstream): Improve upstream performance with improved query strategy

jess-lowe:
Kudos to @michaelkedar and @hogo6002 for this fix :)

Upstream cron job was taking way too long to run, so reduced querying by using the proper get/fetch method. Will observe in staging.

### DIFF
--- a/gcp/workers/alias/alias_computation.py
+++ b/gcp/workers/alias/alias_computation.py
@@ -56,7 +56,7 @@ def _create_alias_group(bug_ids):
     return
 
   new_group = osv.AliasGroup(bug_ids=bug_ids)
-  new_group.last_modified = datetime.datetime.utcnow()
+  new_group.last_modified = datetime.datetime.now()
   new_group.put()
 
 
--- a/gcp/workers/alias/alias_computation.py
+++ b/gcp/workers/alias/alias_computation.py
@@ -56,7 +56,7 @@ def _create_alias_group(bug_ids):
     return
 
   new_group = osv.AliasGroup(bug_ids=bug_ids)
-  new_group.last_modified = datetime.datetime.now()
+  new_group.last_modified = datetime.datetime.utcnow()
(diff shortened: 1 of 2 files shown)


gcp/workers/alias/upstream_computation.py
```diff
@@ -156,12 +156,12 @@ def main():
   bugs = osv.Bug.query(
       ndb.OR(osv.Bug.upstream_raw > '', osv.Bug.upstream_raw < ''))
   bugs = {bug.db_id: bug for bug in bugs.iter()}
-  all_upstream_groups = osv.UpstreamGroup.query()
+  all_upstream_groups = osv.UpstreamGroup.query().fetch()
+  upstream_groups = {group.db_id: group for group in all_upstream_groups}
```
michaelkedar, line 160: nit: to avoid slight confusion with having two upstream group variables:
```suggestion
  upstream_groups = {group.db_id: group for group in osv.UpstreamGroup.query().iter()}
```

gcp/workers/alias/alias_computation.py
```diff
@@ -56,7 +56,7 @@ def _create_alias_group(bug_ids):
     return
 
   new_group = osv.AliasGroup(bug_ids=bug_ids)
-  new_group.last_modified = datetime.datetime.utcnow()
+  new_group.last_modified = datetime.datetime.now()
```
hogo6002, line 59: Why the change to alias_computation? I suggest we keep everything using `utcnow() `and update upstream_computation to use it as well.

hogo6002 (approved):

gcp/workers/alias/alias_computation.py
jess-lowe: upstream already uses it

hogo6002 (approved):

PR #3299, opened 2025-03-26
feat: Add osvdev pkg

another-rex:
Add the osvdev package as a public package maanged in osvdev's repository. 

This also makes public the batch query helper to make it easier for people to handle results with multiple pages.

While looking though this, I realized there's actually a bug with how paging is being handled currently in osv-scanner, a typo here: 

https://github.com/google/osv-scanner/blob/06755300671eaa9072327b505615462f29e2a879/internal/clients/clientimpl/osvmatcher/osvmatcher.go#L123-L127

means we only actually query two pages deep, and not any further. This also fixes that issue. I'll make a separate PR in osv-scanner to fix it there as well.


apibindings/go/go.mod
```diff
@@ -0,0 +1,9 @@
+module github.com/google/osv.dev/apibindings/go
```
michaelkedar, line 1: Do we want to set up the `osv.dev` vanity URL? And thoughts on this just being `bindings` instead of `apibindings`?

```golang
import "osv.dev/bindings/go/osvdev"
```

apibindings/go/pkg/osvdev/osvdev.go
```diff
@@ -0,0 +1,295 @@
...
+	"golang.org/x/sync/errgroup"
+)
+
+const (
+	QueryBatchEndpoint = "/v1/querybatch"
+	QueryEndpoint      = "/v1/query"
+	GetEndpoint        = "/v1/vulns"
+
+	// DetermineVersionEndpoint is the URL for posting determineversion queries to OSV.
+	DetermineVersionEndpoint = "/v1experimental/determineversion"
```
michaelkedar, line 25: Should DetermineVersion stuff belong in `osvdevexperimental`?

apibindings/go/go.sum
```diff
@@ -0,0 +1,8 @@
+github.com/another-rex/osv.dev/go v0.4.0 h1:1UNsHVhM4KH5faHTXRHF4g1KyDDrrN477w+SxPKoKrM=
+github.com/another-rex/osv.dev/go v0.4.0/go.mod h1:Gb3fa+j3r4ewPodfnLlJRlWWKlML49rJmjLugRv/Wc8=
```
michaelkedar, line 2: Is this up to date? Should this be pointing to your fork?

michaelkedar (commented):
a few small bikeshedding thoughts

apibindings/go/go.mod
another-rex: I guess I can imagine other bindings? e.g. instead of a python library we might have a go library as well, but I think that's thinking too far ahead and you're right we should just stick to bindings.

another-rex: Sure we can do a vanity URL, I'll make a separate PR for it.

apibindings/go/pkg/osvdev/osvdev.go
another-rex: I'm thinking probably not? If we do it has to be detached from the current OSVClient, and I want OSVClient to match the OSV API 1:1.

apibindings/go/go.sum
another-rex: Ah needed to run go mod tidy to delete the outdated go sum lines

apibindings/go/pkg/osvdev/osvdev.go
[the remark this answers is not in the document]
another-rex: Moved out of pkg directory.

oliverchang (approved):

michaelkedar (approved):

PR #3300, opened 2025-03-26
feat: Add go vanity metadata

another-rex:
Add vanity URL metadata to return when go gets a package from osv.dev.

I think this is set up in a way where in the future we can also add osv-scanner to the vanity URL by putting it on a sub path.

Needs to be merged before #3299 for it to work.

### DIFF
--- a/gcp/website/frontend_handlers.py
+++ b/gcp/website/frontend_handlers.py
@@ -55,6 +55,8 @@ _VALID_VULN_ID = _WORD_CHARACTERS_OR_DASH_OR_COLON
 _BLOG_CONTENTS_DIR = 'blog'
 _DEPS_BASE_URL = 'https://deps.dev'
 _FIRST_CVSS_CALCULATOR_BASE_URL = 'https://www.first.org/cvss/calculator'
+_GO_VANITY_METADATA = '<meta name="go-import" content="osv.dev/bindings/go git https://github.com/google/osv.dev/bindings/go">'
+
 _ndb_client = ndb.Client()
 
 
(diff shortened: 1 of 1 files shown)


gcp/website/frontend_handlers.py
```diff
@@ -124,8 +128,21 @@ def index_v2_with_subpath(subpath):
   return redirect('/' + subpath)
 
 
+@blueprint.route('/bindings/go')
+def go_bindings_vanity():
+  if request.args.get('go-get', 0) == 1:
```
michaelkedar, line 133: This should be something like
```python
if request.args.get('go-get') == '1':
```
or
```python
if request.args.get('go-get', 0, type=int) == 1:
```

gcp/website/frontend_handlers.py
```diff
@@ -55,6 +55,10 @@
 _BLOG_CONTENTS_DIR = 'blog'
 _DEPS_BASE_URL = 'https://deps.dev'
 _FIRST_CVSS_CALCULATOR_BASE_URL = 'https://www.first.org/cvss/calculator'
+_GO_VANITY_METADATA = \
+  ('<meta name="go-import" '
+   'content="osv.dev git https://github.com/google/osv.dev">')
```
michaelkedar, line 60: Can we add a check so this is only on prod?
I don't know what would happen if `test.osv.dev` would return this.

michaelkedar (commented):
Hmm I wonder if we could achieve this with the load balancer

gcp/website/frontend_handlers.py
another-rex: I think it's fine for the test to return this? it will just fail to install, since package name would not match, and this way we can test the responses while in the test instance.

another-rex:
Yes I think this can be done in the load balancer, we just upload the response to a GCP bucket. But I think it'll be clearer and more maintainable directly in python here.

gcp/website/frontend_handlers.py
another-rex: Ah good catch, updated this.

michaelkedar (approved):

PR #3302, opened 2025-03-27
chore(k8s): don't restart failing high-frequency CronJobs

andrewpollock:
This adjusts the restart policy for high(ish) frequency CronJobs from `OnFailure` to `Never`

The thinking here is that restarting a failing pod with the exact same container image (and therefore build of the binary) is unlikely to be helpful in the cases where the problem is a runtime bug in that binary, and it just delays the next scheduled run of what is potentially a newer container image with a fix for the aforementioned bug (or possibly a different k8s specification).

If the problem is transient, the next scheduled run will be "soon enough" to be morally equivalent to just restarting the same pod.

And we have monitoring to still catch any longer-term crash looping that prevents a successful completion.

It also fixes a string quoting issue that the linter was complaining about

### DIFF
--- a/deployment/clouddeploy/gke-workers/base/alias-computation.yaml
+++ b/deployment/clouddeploy/gke-workers/base/alias-computation.yaml
@@ -17,9 +17,9 @@ spec:
             imagePullPolicy: Always
             resources:
               requests:
-                cpu: 1
+                cpu: "1"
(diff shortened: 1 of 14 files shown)


hogo6002 (approved):
LGTM!

jess-lowe (approved):

PR #3306, opened 2025-03-28
feat(vulnfeeds): Add test expiry date to vulnfeeds

jess-lowe:
In the case of a known outage, we can enable a time for the tests to be ignored until.

### DIFF
--- a/vulnfeeds/cves/versions_test.go
+++ b/vulnfeeds/cves/versions_test.go
@@ -724,7 +724,6 @@ func TestExtractGitCommit(t *testing.T) {
 			t.Skipf("test %q: running on Cloud Build", tc.description)
 		}
 		if time.Now().Before(tc.disableExpiryDate) {
-			println(tc.description)
 			continue
 		}
 		got, err := extractGitCommit(tc.inputLink, tc.inputCommitType)
--- a/vulnfeeds/cves/versions_test.go
+++ b/vulnfeeds/cves/versions_test.go
@@ -150,384 +150,383 @@ func TestParseCPE(t *testing.T) {
 
 func TestRepo(t *testing.T) {
 	tests := []struct {
-		description     string // human-readable description of test case
-		inputLink       string // a possible repository URL to call Repo() with
-		expectedRepoURL string // The expected  repository URL to get back from Repo()
-		expectedOk      bool   // If an error is expected
-		disableExpiryDate time.Time
+		description       string    // human-readable description of test case
+		inputLink         string    // a possible repository URL to call Repo() with
+		expectedRepoURL   string    // The expected  repository URL to get back from Repo()
+		expectedOk        bool      // If an error is expected
+		disableExpiryDate time.Time // If test needs to be disabled due to known outage.
 	}{
 		{
-			description:     "GitHub compare URL",
-			inputLink:       "https://github.com/kovidgoyal/kitty/compare/v0.26.1...v0.26.2",
-			expectedRepoURL: "https://github.com/kovidgoyal/kitty",
-			expectedOk:      true,
+			description:       "GitHub compare URL",
+			inputLink:         "https://github.com/kovidgoyal/kitty/compare/v0.26.1...v0.26.2",
+			expectedRepoURL:   "https://github.com/kovidgoyal/kitty",
+			expectedOk:        true,
 			disableExpiryDate: NO_EXPIRY,
 		},
 		{
-			description:     "GitLab compare URL",
-			inputLink:       "https://gitlab.com/mayan-edms/mayan-edms/-/compare/development...master?from_project_id=396557&straight=false",
-			expectedRepoURL: "https://gitlab.com/mayan-edms/mayan-edms",
-			expectedOk:      true,
+			description:       "GitLab compare URL",
+			inputLink:         "https://gitlab.com/mayan-edms/mayan-edms/-/compare/development...master?from_project_id=396557&straight=false",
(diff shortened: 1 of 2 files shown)


vulnfeeds/cves/versions_test.go
```diff
@@ -903,37 +991,45 @@ func TestExtractVersionInfo(t *testing.T) {
...
 			expectedVersionInfo: VersionInfo{
 				AffectedVersions: []AffectedVersion{{Introduced: "4.0.0", LastAffected: "4.2"}},
 			},
+
+			disableExpiryDate: NO_EXPIRY,
 		},
 	}
 
 	for _, tc := range tests {
+		if time.Now().Before(tc.disableExpiryDate) {
```
hogo6002, line 1030: should we specify timezone here?

hogo6002 (approved):

vulnfeeds/cves/versions_test.go
```diff
@@ -631,20 +707,26 @@ func TestExtractGitCommit(t *testing.T) {
...
+			disableExpiryDate:      NO_EXPIRY,
 		},
 	}
 
 	for _, tc := range tests {
 		if _, ok := os.LookupEnv("BUILD_ID"); ok && tc.skipOnCloudBuild {
 			t.Skipf("test %q: running on Cloud Build", tc.description)
 		}
+		if time.Now().Before(tc.disableExpiryDate) {
+			println(tc.description)
```
another-rex, line 727: You want the `t.Logf` function here, and probably a more descriptive log, e.g. `Skipping test: %s`

vulnfeeds/cves/versions_test.go
another-rex: Probably not? I think the usecase here is for disabling for multiple days, so a few hours probably doesn't matter I think.

andrewpollock (changes requested):
Please use some variation of https://pkg.go.dev/testing#T.Skip so they don't get silently skipped, to help with visibility.

For extra awesome, highlighting when a test case has recently _started_ running again may help uphold the principle of least surprise.

There's a few other tests with external dependencies as well, want to add them to this PR?

another-rex:
> Please use some variation of https://pkg.go.dev/testing#T.Skip so they don't get silently skipped, to help with visibility.
> 
> For extra awesome, highlighting when a test case has recently _started_ running again may help uphold the principle of least surprise.
> 
> There's a few other tests with external dependencies as well, want to add them to this PR?

+1, you will need to wrap each test case in it's own separate t.Run(). This could actually speed up the tests by a lot because it means it will run the tests in parallel rather than sequentially like it is currently doing.  (Please also use t.Parallel() in each t.Run())

jess-lowe:
> > Please use some variation of https://pkg.go.dev/testing#T.Skip so they don't get silently skipped, to help with visibility.
> > For extra awesome, highlighting when a test case has recently _started_ running again may help uphold the principle of least surprise.
> > There's a few other tests with external dependencies as well, want to add them to this PR?
> 
> +1, you will need to wrap each test case in it's own separate t.Run(). This could actually speed up the tests by a lot because it means it will run the tests in parallel rather than sequentially like it is currently doing. (Please also use t.Parallel() in each t.Run())

You were right:
Before:
image
After:
image

vulnfeeds/cves/versions_test.go
```diff
@@ -135,339 +138,402 @@ func TestParseCPE(t *testing.T) {
...
 
 	for _, tc := range tests {
-		got, err := Repo(tc.inputLink)
-		if err != nil && tc.expectedOk {
-			t.Errorf("test %q: Repo(%q) unexpectedly failed: %+v", tc.description, tc.inputLink, err)
-		}
-		if !reflect.DeepEqual(got, tc.expectedRepoURL) {
-			t.Errorf("test %q: Repo(%q) was incorrect, got: %#v, expected: %#v", tc.description, tc.inputLink, got, tc.expectedRepoURL)
-		}
+		tc := tc // Capture range variable
```
another-rex, line 523: Not necessary since go1.22, as they make a new variable each time now. 

Same for all the other tc := tc lines.

vulnfeeds/cves/versions_test.go
```diff
@@ -135,339 +138,402 @@ func TestParseCPE(t *testing.T) {
...
 		{
-			description:     "GitHub compare URL",
-			inputLink:       "https://github.com/kovidgoyal/kitty/compare/v0.26.1...v0.26.2",
-			expectedRepoURL: "https://github.com/kovidgoyal/kitty",
-			expectedOk:      true,
+			description:       "GitHub compare URL",
+			inputLink:         "https://github.com/kovidgoyal/kitty/compare/v0.26.1...v0.26.2",
+			expectedRepoURL:   "https://github.com/kovidgoyal/kitty",
+			expectedOk:        true,
+			disableExpiryDate: NO_EXPIRY,
```
another-rex, line 168: Thinking about this, I think by default, time.Time{} is a 0 value, which will always be in the past. So there's actually no need for disableExpiryDate on any of the cases where we are not disabling.

another-rex (commented):
Woo nice speedup.

vulnfeeds/cves/versions_test.go
jess-lowe: Fixed :)

another-rex (dismissed):
LGTM, thanks!

andrewpollock (dismissed):
Awesome improvements!

hogo6002 (approved):

PR #3329, opened 2025-04-04
docs: Update osv-scanner install instructions on osv.dev to V2

another-rex:
Turns out we never updated the website to have v2 install instructions.


jess-lowe (approved):

oliverchang:
We probably also want to update the example invocations and add container scanner results.

Maybe also fit in a screenshot of the HTML output as well?

PR #3349, opened 2025-04-08
fix(importer): Add extra logging to track deletions

jess-lowe:
With the importer-deleter cron job acting up occasionally, extra logging is valuable here.

### DIFF
--- a/gcp/workers/importer/importer.py
+++ b/gcp/workers/importer/importer.py
@@ -827,9 +827,10 @@ class Importer:
     if (len(vulns_to_delete) / len(vuln_ids_for_source) * 100) >= threshold:
       logging.error(
           'Cowardly refusing to delete %d missing records from '
-          'GCS for: %s', len(vulns_to_delete), source_repo.name)
+          'GCS for: %s', len(vulns_to_delete), source_repo.name, extra={
+          })
       vulns = [v.id for v in vulns_to_delete]
-      logging.error('Vulnerabilities to delete: %s', vulns)
+      logging.info('Vulnerabilities to delete: %s', vulns)
       return
 
     # Request deletion.
--- a/gcp/workers/importer/importer.py
+++ b/gcp/workers/importer/importer.py
@@ -826,8 +826,10 @@ class Importer:
     if (len(vulns_to_delete) / len(vuln_ids_for_source) * 100) >= threshold:
       logging.error(
           'Cowardly refusing to delete %d missing records from '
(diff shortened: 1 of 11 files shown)


gcp/workers/importer/importer.py
```diff
@@ -816,6 +828,8 @@ def _process_deletions_bucket(self,
       logging.error(
           'Cowardly refusing to delete %d missing records from '
           'GCS for: %s', len(vulns_to_delete), source_repo.name)
+      vulns = [v.id for v in vulns_to_delete]
+      logging.error('Vulnerabilities to delete: %s', vulns)
```
michaelkedar, line 832: Make this a `logging.info` so we don't trigger alerts for both logs.
Or, just add it to the `json_fields` of the previous log.

gcp/workers/importer/importer_test.py
```diff
@@ -715,7 +715,7 @@ def test_bucket_deletion(self, unused_mock_time: mock.MagicMock,
       imp.run()
     # The schema validation of failures of the files in GCS by
     # _process_deletions_bucket() causes, plus an extra one from the safeguard.
-    self.assertEqual(4, len(logs.output))
+    self.assertEqual(5, len(logs.output))
     self.assertEqual(
         "ERROR:root:Cowardly refusing to delete 2 missing records from GCS for: test",  # pylint: disable=line-too-long
         logs.output[-1])
```
michaelkedar, line 721: I assume this line has to change, since `Cowardly refusing to delete...` is no longer the last log

gcp/workers/importer/importer.py
```diff
@@ -123,7 +123,8 @@ def replace_importer_log(client: storage.Client, source_name: str,
                          bucket_name: str, import_failure_logs: List[str]):
   """Replace the public importer logs with the new one."""
   bucket: storage.Bucket = client.bucket(bucket_name)
-  upload_string = '--- ' + datetime.datetime.utcnow().isoformat() + ' ---\n'
+  upload_string = '--- ' + datetime.datetime.now(
+      datetime.timezone.utc).isoformat() + ' ---\n'
```
michaelkedar, line 127: nit: I think
```suggestion
  upload_string = f'--- {datetime.datetime.now(datetime.UTC).isoformat()} ---\n'
```
will just barely fit on one line and be easier to read

michaelkedar (approved):

michaelkedar (approved):

Issue #3353, opened 2025-04-10
GIT ranges from `curl` source not being analyzed


michaelkedar:
> Not sure if this is a separate bug or a misunderstanding of the commit based queries. Looking at https://api.osv.dev/v1/vulns/CURL-CVE-2025-0167 you can see that vulnerability was introduced at `46620b97431e19c53ce82e55055c85830f088cf4`. 
> 
> When I query with 
> `curl -s -d \
>   '{"commit": "46620b97431e19c53ce82e55055c85830f088cf4"}' 
>   "https://api.osv.dev/v1/query"`
> 
> I do not get CURL-CVE-2025-0167 reported in the result. 

 _Originally posted by @douglasclarke in [#3331](https://github.com/google/osv.dev/issues/3331#issuecomment-2790474955)_

This is because we have [`ignore_git: True`](https://github.com/google/osv.dev/blob/c0da8edbb7a38d383c042228e92e93b8d9b41b2b/source.yaml#L98) on the curl source.

I don't think if anyone remembers _why_ it's set to be ignored 🙃.
We should try turning this on to enable commit queries on CURL records.

PR #3354, opened 2025-04-10
fix: analyze git commits from curl on test instance

michaelkedar:
Looking into #3353, testing analyzing these on the test instance.
I believe this requires a re-import of all the CURL records.


jess-lowe (approved):

another-rex:
Likely late next week if everything works as expected, there'll be another PR to land this in production.

PR #3356, opened 2025-04-10
feat(frontend): Added weighted search results

jess-lowe:
When a user searches for a specific vulnerability ID, that should be the first record that shows up, not just the most recently updated record. 

As such we added some extra ordering on record results: the ID given > alias of the ID > in the ID's downstream > related to the ID.

May need to add functionality in the future for also including the ID's upstreams. Unfortunately, this only works 100% as intended with one page of results.

### DIFF
--- a/gcp/website/frontend_handlers.py
+++ b/gcp/website/frontend_handlers.py
@@ -535,17 +535,21 @@ def osv_query(search_string, page, affected_only, ecosystem):
   bugs, _, _ = query.fetch_page(
       page_size=_PAGE_SIZE, offset=(page - 1) * _PAGE_SIZE)
 
+  # The following only works 100% as intended if all of these results appear in
+  # the first page
   if _VALID_VULN_ID.match(search_string):
     sorting_bugs = list(bugs)
+    # yapf: disable
     bugs = sorted(
         sorting_bugs,
         key=lambda bug: (
-            0 if bug.db_id == search_string else 1
-            if search_string in bug.aliases else 2
-            if search_string in bug.upstream_raw else 3
-            if search_string in bug.related else
+            0 if bug.db_id == search_string else
+            1 if search_string in bug.aliases else
+            2 if search_string in bug.upstream_raw else
+            3 if search_string in bug.related else
             4,  # Default priority for items not matching specific criteria
             -bug.timestamp.timestamp()))
+    # yapf: enable
   result_items = [bug_to_response(bug, detailed=False) for bug in bugs]
 
   results = {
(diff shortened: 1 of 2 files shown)


gcp/website/frontend_handlers.py
```diff
@@ -534,8 +534,19 @@ def osv_query(search_string, page, affected_only, ecosystem):
...
       page_size=_PAGE_SIZE, offset=(page - 1) * _PAGE_SIZE)
-  for bug in bugs:
-    result_items.append(bug_to_response(bug, detailed=False))
+
+  if _VALID_VULN_ID.match(search_string):
+    sorting_bugs = list(bugs)
+    bugs = sorted(
+        sorting_bugs,
+        key=lambda bug: (
+            0 if bug.db_id == search_string else 1
```
another-rex, line 543: Does the python formatter allow you to rearrange this, so the number always appears before the if, which I think makes more sense.

gcp/website/frontend_handlers.py
```diff
@@ -534,8 +534,19 @@ def osv_query(search_string, page, affected_only, ecosystem):
 
   bugs, _, _ = query.fetch_page(
       page_size=_PAGE_SIZE, offset=(page - 1) * _PAGE_SIZE)
-  for bug in bugs:
-    result_items.append(bug_to_response(bug, detailed=False))
+
```
another-rex, line 537: Add a comment/TODO for this check to say that it doesn't work properly if there are multiple pages.

another-rex (dismissed):

gcp/website/frontend_handlers.py
```diff
@@ -534,8 +534,23 @@ def osv_query(search_string, page, affected_only, ecosystem):
...
+    # yapf: disable
+    bugs = sorted(
+        sorting_bugs,
+        key=lambda bug: (
+            0 if bug.db_id == search_string else
+            1 if search_string in bug.aliases else
+            2 if search_string in bug.upstream_raw else
+            3 if search_string in bug.related else
+            4,  # Default priority for items not matching specific criteria
+            -bug.timestamp.timestamp()))
```
michaelkedar, line 551: nit: I think this would be easier to understand if this function wasn't defined as inline with this nested ternary. (I particularly find the fact that this is a tuple hard to notice)

 e.g.
```python
def search_sort_key(bug: osv.Bug):
  key = 4
  if bug.db_id == search_string:
    key = 0
  elif ...
  ...
  return (key, -bug.timestamp.timestamp())

...

bug = sorted(sorting_bugs, key=search_sort_key)
```

(I don't have a strong opinion on this - feel free to ignore if you don't feel this is worthwhile)

michaelkedar (approved):

PR #3358, opened 2025-04-11
fix(tests): Update integration tests due to breakage

jess-lowe:
Something updated one of our live data testcases so have to update the testcase as well

### DIFF
--- a/gcp/api/integration_tests.py
+++ b/gcp/api/integration_tests.py
@@ -47,195 +47,190 @@ def _api():
 class IntegrationTests(unittest.TestCase,
                        tests.ExpectationTest(_TEST_DATA_DIR)):
   """Server integration tests."""
-
-  _VULN_890 = {
-      'vulns': [{
-          'id':
-              'CVE-2024-2002',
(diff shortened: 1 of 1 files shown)


gcp/api/integration_tests.py
```diff
@@ -49,50 +49,193 @@ class IntegrationTests(unittest.TestCase,
...
-              'ecosystem': 'OSS-Fuzz',
-              'name': 'libdwarf',
-              'purl': 'pkg:generic/libdwarf'
-          },
-          'ranges': [{
-              'events': [{
-                  'introduced': 'b55ce0185528bf0a99e375cf8f3c84b76b6881a3'
-              }, {
-                  'fixed': 'cd741379bd0203a0875b413542d5f982606ae637'
+      'vulns': [{
```
another-rex, line 52: Let's save CVE-2024-2002 in a separate variable, and just mention both of them in the assert.

another-rex (dismissed):

another-rex (approved):

PR #3359, opened 2025-04-11
fix: analyze git commits from curl on prod

jess-lowe:
Test instance worked as intended so pushing change to prod as well. 

Closes: #3353


another-rex (approved):
FYI we'll try to merge this in on Monday to see if anything comes up over the weekend on staging.

michaelkedar (approved):

PR #3360, opened 2025-04-11
fix: cap the length of string values in query

cuixq:
There are bad value errors reported for some queries:
```
BadValueError: Indexed value must be at most 1500 bytes
```

To fix this, this PR caps the length of the string value reported to a reasonable length.


another-rex (approved):
Can you add the limit to the actual frontend as well? i.e. the textbox

cuixq:
done

another-rex (approved):

PR #3361, opened 2025-04-11
feat(search): improve search through adding more sections of the id to search indices

jess-lowe:
Mostly to address 'cve-xxx' not showing up in 'ubuntu-cve-xxx' search indices, and thus ubuntu not showing when searching for the cve.


osv/models.py
```diff
@@ -364,7 +364,18 @@ def _tokenize(self, value):
       return []
 
     value_lower = value.lower()
-    return re.split(r'\W+', value_lower) + [value_lower]
+    tokens = {token for token in re.split(r'\W+', value_lower) if token}
+    tokens.add(value_lower)
+
+    parts = value_lower.split('-')
```
oliverchang, line 370: Can you add some code comments here to explain what this is doing?

osv/models.py
```diff
@@ -364,7 +364,25 @@ def _tokenize(self, value):
...
+    # Add subsection combinations from id (split at '-') in the search indices
+    # Specifically addresses situation in which UBUNTU-CVE-XXXs don't show up
+    # when searching for the CVE-XXX.
+    # Does not account for combinations with the suffix sections ':' like SUSE
+    parts = value_lower.split('-')
+    num_parts = len(parts)
+    if num_parts > 1:
+      for length in range(2, num_parts + 1):
+        for i in range(num_parts - length + 1):
+          sub_parts = parts[i:i + length]
```
michaelkedar, line 382: nit: personally, I would find this easier to follow:
```python
for i in range(num_parts-1):
  for j in range(i+1, num_parts):
    sub_parts = parts[i:j+1]
```

(I think that's equivalent?)

osv/models.py
```diff
@@ -364,7 +364,25 @@ def _tokenize(self, value):
...
+    tokens.add(value_lower)
+
+    # Add subsection combinations from id (split at '-') in the search indices
+    # Specifically addresses situation in which UBUNTU-CVE-XXXs don't show up
+    # when searching for the CVE-XXX.
+    # Does not account for combinations with the suffix sections ':' like SUSE
+    parts = value_lower.split('-')
+    num_parts = len(parts)
+    if num_parts > 1:
+      for length in range(2, num_parts + 1):
```
michaelkedar, line 380: nit: you can save an indent if you get rid of this length check, since `range(2, 2)` will just do 0 iterations.

osv/models.py
michaelkedar: An example might also be helpful
```
e.g. `a-b-c-d' becomes ['a-b', 'b-c', 'c-d', 'a-b-c', 'b-c-d', 'a-b-c-d']
```

michaelkedar (approved):

osv/models.py
another-rex: I think there's an existing functools function to do this.

osv/models.py
```diff
@@ -364,7 +364,25 @@ def _tokenize(self, value):
       return []
 
     value_lower = value.lower()
-    return re.split(r'\W+', value_lower) + [value_lower]
+
+    # Deconstructs the id given into parts by retrieving parts that are
+    # alphanumeric.
+    tokens = {token for token in re.split(r'\W+', value_lower) if token}
```
another-rex, line 370: You can skip this step, and just fold this into the the subsections function below by starting at length 1 in the loop.

another-rex: Unless there's a special case where things are not separated by -?, If so do add a comment here explaining that.

jess-lowe: SUSE use the ':' separator for their suffixes, so this is a catch all case, and combining them again would be difficult 

- added a comment.

michaelkedar (approved):

Issue #3400, opened 2025-04-29
Support MinimOS datasource


bearjah:
Hi,
I’m part of the [Minimus](https://minimus.io/) team. Minimus delivers secure, minimal container images with auto-generated SBOMs and real-time vulnerability threat intelligence to help reduce vulnerability risk.
We also maintain a minimal operating system called MinimOS. We’ve been publishing our security advisories using the [OSV schema](https://ossf.github.io/osv-schema/) and would like to contribute them as a new data source.

Details:

Vulnerabilities index:

- https://packages.mini.dev/advisories/osv/all.json

Individual advisories:

- https://packages.mini.dev/advisories/osv/<ADVISORY-ID\>.json

All files adhere to the OSV schema.
We’re ready to proceed with the remaining onboarding steps for MinimOS.
Thanks for your consideration!

/label datasource

michaelkedar:
Thanks for doing this!
I've merged #3402, so we should be seeing the records populating on https://test.osv.dev soon (and queryable on the https://api.test.osv.dev/v1 API)

If all seems good, we can merge #3403 next week to get this on production.

bearjah:
Thanks! @michaelkedar

another-rex:
@bearjah We noticed a bunch of records (1051 to be exact) have been deleted from the all.json, but still accessible if you go to the exact ID. Is this intentional? 

Example being: MINI-2226-54r3-mf8j , it does not exist in the all.json, but is still accessible here:  https://packages.mini.dev/advisories/osv/MINI-2226-54r3-mf8j.json

Full list of removed IDs here:

```
[...]

PR #3403, opened 2025-04-29
Add MinimOS feed to production

bearjah:
ref: https://github.com/google/osv.dev/issues/3400

### DIFF
--- a/source.yaml
+++ b/source.yaml
@@ -235,6 +235,20 @@
   editable: False
   strict_validation: True
 
+- name: 'minimos'
+  versions_from_repo: False
+  rest_api_url: 'https://packages.mini.dev/advisories/osv/all.json'
+  type: 2
+  ignore_patterns: ['^(?!MINI-).*$']
+  directory_path: 'advisories/osv'
+  detect_cherrypicks: False
+  extension: '.json'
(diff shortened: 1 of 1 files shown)


jess-lowe (approved):

hogo6002:
Hi, our osv-linter found around 147 records with invalid version range. For example: https://test.osv.dev/vulnerability/MINI-27hm-pcm3-8hc5 has both version `0` as introduced and fixed version, which means the package is not really affected by this vulnerability. Can you take a look at it?

OSV linter API: `http://api.test.osv.dev/v1experimental/importfindings/minimos`

hogo6002:
> Is there another method to mark unaffected CVEs, or should they be dropped?

Thanks for the clarification. I think for now we can keep them. In the long term, OSV will have a more detailed plan on how to handle those records.

hogo6002 (approved):

another-rex:
@bearjah There's some issues with how minimOS OSV records are being published. It seems like you guys are updating the all.json before adding the records, meaning when we see your all.json updated, we can't fetch the actual records it points to.

Is it possible to make sure the records are in place before all.json is updated?

PR #3424, opened 2025-05-07
fix(workers): bump memory limits on alias/upstream cron

michaelkedar:
The upstream computation was being killed for using too much memory.

(For some reason, GKE does not consider this a failed cron...)


jess-lowe (dismissed):
We'll look into why this is happening but good fix in the meantime. 😬

jess-lowe (approved):

PR #3425, opened 2025-05-07
fix(workers): Added index projection to reduce memory + added logging

jess-lowe:
Drastically reduced memory consumption of upstream computation by 15x, with a side of reduced time and CPU usage.

Done by added index projection to the bug entity query.

Also added logging for future debugging.


michaelkedar (dismissed):
Just confirming for next release person that the index update is already on prod.

another-rex (dismissed):

gcp/workers/alias/upstream_computation.py
```diff
@@ -155,12 +156,12 @@ def main():
   # and avoid datastore emulator problems, see issue #2093
```
michaelkedar, line 156: nit: this comment is no longer accurate

gcp/workers/alias/upstream_computation.py
```diff
@@ -155,12 +156,12 @@ def main():
...
-  bugs = {
-      bug.db_id: bug
-      for bug in bugs.iter(projection=[osv.Bug.db_id, osv.Bug.upstream_raw])
-  }
+  bugs_query = osv.Bug.query(osv.Bug.upstream_raw > '')
+
+  bugs = defaultdict(list)
+  for bug in bugs_query.iter(projection=[osv.Bug.db_id, osv.Bug.upstream_raw]):
+    bugs[bug.db_id].append(bug.upstream_raw[0])
+  bugs = {k: list(set(v)) for k, v in bugs.items()}
```
michaelkedar: There's actually quite a bit of conversion between `list` and `set` going on here and in `compute_upstream`

I think you might be able to get away with using only sets throughout (use `defaultdict(set)` instead)

gcp/workers/alias/upstream_computation.py
```diff
@@ -21,15 +21,16 @@
 import osv.logs
 import json
 import logging
+from collections import defaultdict
 
 
-def compute_upstream(target_bug, bugs: dict[str, osv.Bug]) -> list[str]:
+def compute_upstream(target_bug, bugs) -> list[str]:
```
michaelkedar, line 27: nit: maybe keep the type annotation `bugs: dict[str, list[str]]` (or `dict[str, set[str]]` from my other comment)

gcp/workers/alias/upstream_computation.py
michaelkedar: Oh this has changed since I wrote the comment?

michaelkedar (approved):

PR #3427, opened 2025-05-08
fix(vulnfeed-tests): Disable ffmpeg tests temporarily

jess-lowe:
ffmpeg 'average load is too high' currently, so disabling tests so PRs build


vulnfeeds/cves/versions_test.go
```diff
@@ -631,6 +631,7 @@ func TestExtractGitCommit(t *testing.T) {
 				Repo:  "https://git.ffmpeg.org/ffmpeg.git",
 				Fixed: "c94875471e3ba3dc396c6919ff3ec9b14539cd71",
 			},
+			disableExpiryDate: time.Date(2025, 3, 31, 12, 30, 0, 0, time.Local),
```
michaelkedar, line 634: This is for March 31?

jess-lowe: Good catch!

michaelkedar (approved):

PR #3430, opened 2025-05-13
feat: add MinimOS ecosystem helper

michaelkedar:
ref: #3400 
Adding a ecosystem helper for the new `MinimOS` ecosystem, which appears to use standard APK-based versioning.
This is identical to the `Wolfi`/`Chainguard` helpers.

Also updated the osvschema submodule to pick up the `MinimOS` ecosystem.


michaelkedar:
@bearjah I've assumed that minimus doesn't have package versions that don't conform to the standard APK rules.

Issue #760, opened 2022-09-30
Support direct linking by alias


andrewpollock:
> [redacted] linking to https://osv.dev/vulnerability/GHSA-jfh8-c2jp-5v3q for Log4Shell.

> I wish it was possible to link to https://osv.dev/vulnerability/CVE-2021-44228

oliverchang:
I wonder what the behaviour should be if there are multiple that link to the same alias.

oliverchang:
> I wonder what the behaviour should be if there are multiple that link to the same alias.

Oh, maybe we can redirect to the list page instead with the filter set to the alias in that case.

andrewpollock:
I see in the case of https://osv.dev/vulnerability/GHSA-jfh8-c2jp-5v3q, it has three aliases, two of which link to other entries that cite each other. It's basically the (current) lack of an entry for the CVE that makes this not work.

I'm thinking we'd call this "done" once we're successfully importing the CVE record from the NVD?

andrewpollock:
> I'm thinking we'd call this "done" once we're successfully importing the CVE record from the NVD?

Are there other data sources than CVE that can get referenced as an alias without currently being imported?

oliverchang:
> I see in the case of https://osv.dev/vulnerability/GHSA-jfh8-c2jp-5v3q, it has three aliases, two of which link to other entries that cite each other. It's basically the (current) lack of an entry for the CVE that makes this not work.
> 
> I'm thinking we'd call this "done" once we're successfully importing the CVE record from the NVD?

Yes, that's one approach, but it does introduce questions around duplicates even if we can link them via aliases. 

> Are there other data sources than CVE that can get referenced as an alias without currently being imported?

Yes, there is no requirement that anything in `aliases` are OSV formatted. There are going to be things like SNYK-, RHSA- in there.

zahraaalizadeh:
Is this issue still applicable or a matter of concern? @oliverchang

another-rex:
I believe we do still want this, though the need for it is greatly reduced now that we are importing git CVE entries directly. E.g. The example given by Andrew is no longer an issue, since it now links to the GIT entry.

andrewpollock:
This is essentially search-by-alias functionality, when the aliased record doesn't actually exist in OSV.dev.

e.g. when https://osv.dev/GHSA-9p26-698r-w4hx aliases CVE-2024-23650 and CVE-2024-23650 doesn't exist in OSV.dev.

[...]

PR #3455, opened 2025-05-19
feat(frontend): Added rerouting for aliases and queries 

jess-lowe:
Closes https://github.com/google/osv.dev/issues/760

Intended behaviour:

For an ID that we do not have in our datastore, that is part of an alias group with 1 other vuln, using /{vuln-id} will redirect to the other alias vuln. If there are more aliases in the group, or no alias at all, the user will be rerouted to the search query page, with the chance of finding a relevant downstream vuln. 

Current best example is CVE-2024-40120. this may change if we import the record at some point.

Also updated the emulator with some examples to test.

### DIFF
--- a/gcp/website/frontend_handlers.py
+++ b/gcp/website/frontend_handlers.py
@@ -269,8 +269,11 @@ def vulnerability(vuln_id: str):
   """Vulnerability page."""
   try:
     vuln = osv_get_by_id(vuln_id)
+  except VulnerabilityNotImported:
+    # TODO: handle showing import findings
+    abort(404)
   except VulnerabilityNotFound as e:
-    return redirect(f'/list?q={e.vuln_id}')
+    return redirect(url_for('list_vulnerabilities', q=e.vuln_id))
 
   api_url = utils.api_url()
 
@@ -291,9 +294,9 @@ def vulnerability_redirector(potential_vuln_id: str):
   try:
     bug = osv_get_by_id(potential_vuln_id)
   except VulnerabilityNotFound as e:
-    return redirect(f'/list?q={e.vuln_id}')
+    return redirect(url_for('list_vulnerabilities', q=e.vuln_id))
 
-  path = safe_join('/vulnerability/', bug.get('id', ''))
+  path = safe_join('/vulnerability', bug.get('id', ''))
   return redirect(path or '/vulnerability/')
 
 
@@ -590,6 +593,8 @@ def osv_get_by_id(vuln_id: str) -> dict:
   if not bug:
     alias = check_for_aliases(vuln_id)
     if not alias:
+      if osv.ImportFinding.get_by_id(vuln_id):
+        raise VulnerabilityNotImported(vuln_id)
       raise VulnerabilityNotFound(vuln_id)
     bug = alias
 
--- a/gcp/website/frontend_handlers.py
+++ b/gcp/website/frontend_handlers.py
@@ -267,9 +267,11 @@ def list_vulnerabilities():
 @blueprint.route('/vulnerability/<vuln_id>')
 def vulnerability(vuln_id):
   """Vulnerability page."""
-  vuln = osv_get_by_id(vuln_id)
+  vuln, route = osv_get_by_id(vuln_id)
 
   api_url = utils.api_url()
+  if route:
+    return redirect(route)
 
   return render_template(
       'vulnerability.html', vulnerability=vuln, api_url=api_url)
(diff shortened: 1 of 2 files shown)


gcp/website/frontend_handlers.py
```diff
@@ -571,29 +574,36 @@ def get_vuln_count_for_ecosystem(ecosystem: str) -> int:
   return ecosystem_counts.get(ecosystem, 0)
 
 
-def osv_get_by_id(vuln_id):
+def osv_get_by_id(vuln_id) -> tuple[osv.Bug | None, str | None]:
```
oliverchang, line 577: since we're here, can you add `:str` annotation to the argument?

gcp/website/frontend_handlers.py
```diff
@@ -571,29 +574,36 @@ def get_vuln_count_for_ecosystem(ecosystem: str) -> int:
...
-    return None
 
   bug = osv.Bug.get_by_id(vuln_id)
   if not bug:
-    if osv.ImportFinding.get_by_id(vuln_id):
-      # abort(404) is too simplistic for this.
-      raise VulnerabilityNotImported(vuln_id)
-    # abort(404) is too simplistic for this.
-    raise VulnerabilityNotFound(vuln_id)
+    alias_group = osv.AliasGroup.query(osv.AliasGroup.bug_ids == vuln_id).get()
```
oliverchang, line 584: might be worth putting this in its own function to make this function easier to understand.

gcp/website/frontend_handlers.py
```diff
@@ -285,9 +287,10 @@
     return None
 
   # This may raise an exception directly or via abort() for failed retrievals.
-  bug = osv_get_by_id(potential_vuln_id)
-
-  return redirect(f'/vulnerability/{bug["id"]}')
+  bug, route = osv_get_by_id(potential_vuln_id)
+  if bug:
+    return redirect(f'/vulnerability/{bug["id"]}')
```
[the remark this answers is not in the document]
oliverchang: Please fix.

gcp/website/frontend_handlers.py
```diff
@@ -285,9 +287,10 @@
...
+  return redirect(route)
```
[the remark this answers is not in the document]
oliverchang: Please fix.

gcp/website/frontend_handlers.py
```diff
@@ -265,9 +265,12 @@ def list_vulnerabilities():
...
 
 @blueprint.route('/vulnerability/<vuln_id>')
-def vulnerability(vuln_id):
+def vulnerability(vuln_id: str):
   """Vulnerability page."""
-  vuln = osv_get_by_id(vuln_id)
+  try:
+    vuln = osv_get_by_id(vuln_id)
+  except VulnerabilityNotFound as e:
+    return redirect(f'/list?q={e.vuln_id}')
```
oliverchang, line 273: can we use `url_for` here instead? and set `q` via that function?

oliverchang (commented):
nice! one more comment to see if we can make things a bit cleaner (and improve the current status quo too)

gcp/website/frontend_handlers.py
oliverchang: (and same for all other instances)

gcp/website/frontend_handlers.py
```diff
@@ -285,9 +288,13 @@ def vulnerability_redirector(potential_vuln_id):
...
 
   # This may raise an exception directly or via abort() for failed retrievals.
-  bug = osv_get_by_id(potential_vuln_id)
+  try:
+    bug = osv_get_by_id(potential_vuln_id)
+  except VulnerabilityNotFound as e:
+    return redirect(f'/list?q={e.vuln_id}')
 
-  return redirect(f'/vulnerability/{bug["id"]}')
+  path = safe_join('/vulnerability/', bug.get('id', ''))
```
hogo6002, line 296: do we still need the second forward slash if we use safe_join()

gcp/website/frontend_handlers.py
```diff
@@ -571,31 +581,43 @@ def get_vuln_count_for_ecosystem(ecosystem: str) -> int:
...
-def osv_get_by_id(vuln_id):
+def osv_get_by_id(vuln_id: str) -> dict:
   """Gets bug details from its id. If invalid, aborts the request."""
   if not vuln_id:
     abort(400)
-    return None
 
   bug = osv.Bug.get_by_id(vuln_id)
   if not bug:
-    if osv.ImportFinding.get_by_id(vuln_id):
```
hogo6002, line 582: This is used to display records found by the linter that we rejected right? Maybe we could move this check after the alias verification instead of delete it.

hogo6002 (commented):
Nice feature!

gcp/website/frontend_handlers.py
jess-lowe: Added a TODO, as this isn't really implemented properly yet

oliverchang (approved):

PR #3456, opened 2025-05-19
feat: add a cron job to run linter and update datastore regularly

hogo6002:
Resolves https://github.com/google/osv.dev/issues/2189

We have added a linter API to surface all findings. This PR adds a cron job to run the linter (offline checks only) daily against all existing OSV records and adds all findings into our datastore (test instance only). This allows users to actually query the invalid historical records. Example:

Next steps:
- Integrate the linter into the importer, rejecting records at creation time.
- Add more necessary checks, such as ensuring that bugs cannot exist in both the related and upstream fields.
- Dashboard or other ways to better visualize findings, and changes of invalid records.
- Runs on production.


gcp/workers/linter/linter.py
```diff
@@ -0,0 +1,203 @@
...
+
+def download_zip(all_zip: str):
+  """Downloads all.zip file from bucket."""
+  storage_client = storage.Client()
+  bucket = storage_client.get_bucket(VULN_BUCKET)
+  try:
+    blob = bucket.blob(ZIP_FILE_PATH)
+    blob.download_to_filename(all_zip)
+  except Exception as e:
+    logging.exception('Failed to download all.zip: %s', e)
```
oliverchang, line 97: nit: remove all.zip reference, since this function is more general and not specific to downloading all.zip

gcp/workers/linter/linter.py
```diff
@@ -0,0 +1,203 @@
...
+    sys.exit(1)
+  logging.info('Downloaded %s.', all_zip)
+
+
+def download_osv_data():
+  """download osv data from all.zip"""
+  tmp_dir = os.path.join(TEST_DATA, 'tmp')
+  os.makedirs(tmp_dir, exist_ok=True)
+  os.environ['TMPDIR'] = tmp_dir
```
oliverchang, line 106: nit: rather than set an env var, which is a permanent side effect, can we pass it to the `dir` argument of https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory ?

gcp/workers/linter/linter.py
```diff
@@ -0,0 +1,203 @@
...
+import tempfile
+import zipfile
+
+from google.cloud import ndb, storage
+
+import osv.logs
+from osv.models import utcnow
+
+# The Go binary is copied to /usr/local/bin in the Dockerfile
+OSV_LINTER_PATH = '/usr/local/bin/osv-linter'
```
oliverchang, line 32: I wonder if we can avoid this coupling somehow. Can we just use `osv-linter`, and assume it can be found in `PATH`?

gcp/workers/linter/linter.py
```diff
@@ -0,0 +1,203 @@
...
+VULN_BUCKET = 'osv-test-vulnerabilities'
+ZIP_FILE_PATH = 'all.zip'
+TEST_DATA = './test_data'
```
oliverchang, line 35: this seems dangerous since we don't know where the CWD is. I would suggest adding a `--work_dir` arg.

oliverchang (commented):
nice! 

re "Integrate the linter into the importer, rejecting records at creation time."

Do we need to start making a distinction between fatal linter errors vs warnings?

gcp/workers/linter/linter.py
hogo6002: updated function for more general file downloading purpose

gcp/workers/linter/linter.py
hogo6002: yep, calling osv-linter should be enough

hogo6002:
> re "Integrate the linter into the importer, rejecting records at creation time."
> 
> Do we need to start making a distinction between fatal linter errors vs warnings?

My idea is that this cron job focus on historical records, which is why I've only enabled offline checks (not including package or version existence checks) . If a record fail within this cron job, it will almost certainly indicate an issue with the record itself. Then we will record this into our datastore without any doubts. 

For the importer, I plan to enable all checks. But I will not reject records that only failed package or version existence checks. So for the first integration, I will hardcode the handling for these two specific checks. We will then monitor the accuracy of our results to determine if they should be marked as warnings.

hogo6002:
This requires OSV linter provides a JSON output format, so pending on https://github.com/ossf/osv-schema/pull/352

michaelkedar (approved):

michaelkedar (approved):

PR #3458, opened 2025-05-20
test: split Cloud Build tests into multiple builds

michaelkedar:
#3452 Step 1 for running tests only on relevant PRs.
I've created a different `cloudbuild.yaml` file for a bunch of our components - basically each lockfile has its own file.

After this is merged, I'll split up the PR triggers to only run on the relevant files (that I've listed in the yaml files).

I've also made a slight change to the datastore emulator to allow the worker/importer tests to be run in parallel.

### DIFF
--- a/gcp/api/cloudbuild.yaml
+++ b/gcp/api/cloudbuild.yaml
@@ -25,6 +25,9 @@ steps:
 - name: 'gcr.io/cloud-builders/git'
   id: 'init'
   args: ['submodule', 'update', '--init']
+  # if this is invoked from another cloud build, this will fail as it is not a git repo
+  # the invoking cloud build file should run this step.
+  allowFailure: true
 - name: 'gcr.io/oss-vdb/ci'
   id: 'sync'
   dir: gcp/api
--- a/gcp/api/cloudbuild.yaml
+++ b/gcp/api/cloudbuild.yaml
@@ -12,25 +12,31 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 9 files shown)


hogo6002 (dismissed):

gcp/api/cloudbuild.yaml
```diff
@@ -0,0 +1,45 @@
...
+  args: ['poetry', 'sync']
+
+- name: 'gcr.io/oss-vdb/ci'
+  id: 'api-tests'
+  dir: gcp/api
+  #TODO: Update test scripts to support not supplying a credential.
+  args: ['bash', '-ex', 'run_tests.sh', '/workspace/dummy.json']
+  env:
+    - CLOUDBUILD=1
+  waitFor: ['init', 'sync']
```
oliverchang, line 43: nit: is this necessary? without this argument we also have the same desired effect right? 

(similar comment for other instances).

oliverchang (dismissed):
nice!

gcp/api/cloudbuild.yaml
michaelkedar: eh - it's necessary for workers & lib-tests, since they have multiple test steps that we want to run concurrently.
I don't think there's much harm in also having it in api/website/vulnfeeds just to be consistent.

michaelkedar: I also meant to put `waitFor: ['-']` in the `sync` steps but I forgot...

hogo6002 (approved):

Issue #1016, opened 2023-02-10
API: query vulnerabilities by cve id


huanceng:
The current api `/v1/vulns/{id}` returns a vulnerability object for a given **OSV ID**. If a cve id is given, the api will return `Bug not found.`

For example, GHSA-h592-38cm-4ggp and its corresponding cve id CVE-2017-15095.

Do you plan to develop such an API?

oliverchang:
This does sound useful thanks, tracking as a FR.

rdeepc:
Hi @oliverchang,

I am interested in working on this.

andrewpollock:
More abstractly, it sounds like you're wishing for the `/v1/vulns{id}` endpoint to support lookups by alias as well as the primary ID?

huanceng:
Hi @andrewpollock , `/v1/vulns/{id}` or a new endpoint are both acceptable.

andrewpollock:
Some interim thoughts to improve this user experience:

```
$ GET -s https://api.osv.dev/v1/vulns/CVE-2025-0001
404 Not Found
{"code":5,"message":"Bug not found."}
```

* include a link to https://osv.dev/faq in the message text
* add a new FAQ entry about bugs not found/404s from the `/v1/vulns/{id}` API endpoint
* add to https://google.github.io/osv.dev/post-v1-query/ about querying by first class vulnerabilities versus aliases

andrewpollock:
Related: #760

jess-lowe:
I believe this was closed by #3462. Feel free to reopen if not solved!

cmoulliard:
I don't think that the issue is fixed as we continue to get a HTTP 404 response with body
```
❯ http https://api.osv.dev/v1/vulns/CVE-2026-40970
HTTP/1.1 404 Not Found
Content-Length: 89
alt-svc: h3=":443"; ma=2592000,h3-29=":443"; ma=2592000
content-type: application/json
date: Fri, 03 Jul 2026 12:22:34 GMT
server: Google Frontend
...

{
    "code": 5,
    "message": "Bug not found, but the following aliases were: GHSA-c96x-rpm4-349p"
}
```
If I extract the GHSA id, then I can requery the API to find the record
```
❯ http https://api.osv.dev/v1/vulns/GHSA-c96x-rpm4-349p
...
{
    "affected": [
        {
            "database_specific": {
[...]

PR #3462, opened 2025-05-20
feat(server): Add a check for aliases if bug not found, and surface them in api

jess-lowe:
If a bug is not found, check if it is a listed alias, and surface that in the API for users. 

Somewhat solves #1016

### DIFF
--- a/gcp/api/server.py
+++ b/gcp/api/server.py
@@ -1,4 +1,4 @@
-# Copyright 2021 Google LLC
+# Copyright 2025 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


gcp/api/server.py
```diff
@@ -166,7 +166,24 @@ class OSVServicer(osv_service_v1_pb2_grpc.OSVServicer,
...
+
+    if not bug:
+      # Check for aliases
+      alias_group = osv.AliasGroup.query(
+          osv.AliasGroup.bug_ids == request.id).get()
+      if alias_group:
+        string_builder = {
+            f'{alias} ' for alias in alias_group.bug_ids if alias != request.id
+        }
+        alias_string = ''.join(string_builder)
```
michaelkedar, line 178: nit: Is there a reason to use a set here (i.e. `{}` vs `[]`)
Also, instead of adding the space to the alias ID, you could do `' '.join(...)`
(and I guess it could also be one line)

michaelkedar (approved):

hogo6002:
Can we add a test for it

hogo6002 (approved):
LGTM!

PR #3487, opened 2025-05-22
feat(vulnfeeds): Support extracting github tag links from CVE references

jess-lowe:
It involved a serious refactor but this should enable support for extracting github tag links from cve references.

comments on improving code quality more than welcome


vulnfeeds/cves/versions.go
```diff
@@ -613,10 +394,45 @@ func Commit(u string) (string, error) {
 
 	// TODO(apollock): add support for resolving a GitHub PR to a commit hash
 
+	// Support for resolving a Github tag to a commit hash
+	// example: https://github.com/redis/redis/releases/tag/6.2.17
+	if parsedURL.Host == "github.com" {
```
jess-lowe, line 399: note: this is the main logic change

vulnfeeds/cves/versions.go
```diff
@@ -613,10 +394,45 @@ func Commit(u string) (string, error) {
...
 }
 
+func resolveGitTag(parsedURL *url.URL, u string, gitSHA1Regex *regexp.Regexp) (string, error) {
+	directory, tag := path.Split(parsedURL.Path)
+	normalizedTag, err := git.NormalizeVersion(tag)
+	if err == nil {
+		tag = normalizedTag
+	}
+
+	if strings.HasSuffix(directory, "tag/") {
```
another-rex, line 416: Invert this if statement

vulnfeeds/cves/versions.go
```diff
@@ -613,10 +394,45 @@ func Commit(u string) (string, error) {
...
+			return "", err
+		}
+		for t, nTag := range normalizedTags {
+			if tag == t && gitSHA1Regex.MatchString(nTag.Commit) {
+				return nTag.Commit, nil
+			}
+		}
+
+	}
+	return "", err
```
another-rex, line 433: If the URL is not a tag url, I think we should return an error rather than just empty string.

vulnfeeds/cves/versions.go
```diff
@@ -613,10 +394,45 @@ func Commit(u string) (string, error) {
...
+			return possibleCommitHash, nil
+		}
+	}
 	// If we get to here, we've encountered an unsupported URL.
 	return "", fmt.Errorf("Commit(): unsupported URL: %s", u)
 }
 
+func resolveGitTag(parsedURL *url.URL, u string, gitSHA1Regex *regexp.Regexp) (string, error) {
+	directory, tag := path.Split(parsedURL.Path)
+	normalizedTag, err := git.NormalizeVersion(tag)
```
another-rex, line 411: Do we need to continue if this fails? Since we are matching against normalized tags further down anyway if this fails we can probably just exit here.

another-rex (commented):
Thanks! Added some comments

another-rex (approved):

PR #3488, opened 2025-05-22
feat(linter): remove finding entries when the linter no longer reports the issues.

hogo6002:
Updates the cron job to also delete stale findings. 

It first queries all existing findings from the database, then updates/creates entries based on new finding results, and finally removes findings no longer reported.


michaelkedar (approved):
lgtm

PR #3489, opened 2025-05-23
feat: Support querying by tags for GIT affected.

### LINKED ISSUE #3331: Query API missing GIT ecosystem
**Is your feature request related to a problem? Please describe.**
The query API doesn't recognize 'GIT' as a supported ecosystem (returns "Invalid ecosystem." error).

**Describe the solution you'd like**
I would like to be able to query GIT vulnerabilities (found [here](https://osv.dev/list?q=&ecosystem=GIT)) via the query API.

another-rex:
Also smoothed out some pain points in testing

Resolves #3331

Changes made:
- Edited preput to also include repo name in project field if it contains a GIT affected field
- Added special matching rules to match package name against repo_urls, and GIT ecosystems
- Removed footgun in the hard coded GOOGLE_CLOUD_PROJECT being oss-vdb, so you can edit it only in the make file to change projects. Added run-api-server-test to run api server in test instance.


tools/datafix/reput_bugs.py
jess-lowe, line 1: it says this is an empty file?

osv/models.py
```diff
@@ -337,7 +337,7 @@ def id(self):
 
   @property
   def repo_url(self):
-    """Repo URL."""
+    """Repo URL. Assumes there is only ever one repo URL per advisory."""
```
michaelkedar, line 340: It's probably extremely unlikely an advisory would have more than one git repo, but I wonder if we should think about storing the repo in the AffectedCommits, and do the filtering logic when querying those.

michaelkedar (approved):

michaelkedar (dismissed):

jess-lowe (approved):

Issue #3365, opened 2025-04-15
Avoid download archive if there is no updates


kalidor:
**Is your feature request related to a problem? Please describe.**
We use to download the archive all.zip (`https://osv-vulnerabilities.storage.googleapis.com/all.zip`), unzip it, and work with data.
Some of our tools can download it few times, but avoiding it will be better for the bandwith.
The tools are not working together, but having a way to detect if an archive has been updated.

**Describe the solution you'd like**
Adding a simple JSON file with the list of files and associated hash (hash of the content of the file, not the archive), will let us detect if something changed, compare to the previous download.
Something like:
```json
{
  "filename1.json": "<computed hash1>",
  "filename2.json": "<computed hash2>",
  ... and so on
}
```

This could also be applied to each ecosystem archive.

**Describe alternatives you've considered**
None. However, i read the other ticket about this (https://github.com/google/osv.dev/issues/1993) which has been considered as completed.

**Additional context**
This could also be useful on your side to not update the `modified` field date if the content of the OSV entry doesn't change at all.

I read the contribution are welcome, that's nice, i didn't find a way to potentially test code before making a PR...

another-rex:
Thanks for the report!

Currently the hash of the all.zip we are exporting changes between each exporter run even though there are no changes (because of the modified time of each record changing). We are looking into options to avoid this from happening, which should allow you to see if all.zip has changed from your current version. (I believe google cloud buckets returns headers with the hashes of the file which can be used to check for differences).

Another thing we are looking into is a better compression format to reduce bandwidth of each download.

> This could also be useful on your side to not update the modified field date if the content of the OSV entry doesn't change at all.

Do you mean the field within the JSON record, or the file's modified date? I believe we don't make changes to the modified field in the record itself, but we do update the file modified date, which we are looking into addressing (it should also solve the inconsistent all.zip hash issue)

> I read the contribution are welcome, that's nice, i didn't find a way to potentially test code before making a PR...

Making changes to the importing and exporting infra is a bit difficult because of the coupling we currently have with our services. We are looking to improve this this year.

trueserjo:
[...]

PR #3491, opened 2025-05-26
fix: Make reproducable zips in exports

another-rex:
Set the modification date on the OSV records written to disk to be of the Modification time of the record.

Partially resolves: #3365


another-rex:
Modified is the last-modified datastore entry.

hogo6002 (approved):
nice!

hogo6002 (approved):

PR #3580, opened 2025-06-17
feat(website): Make determination of alias/related/upstream links faster

michaelkedar:
Instead of sequentially querying the *whole* Bug in datastore to check if it's in the database for rendering links in alias/related/upstream/downstream fields, collect all the possible IDs and query them beforehand. (This involved some slight refactoring to how the up/downstream html is generated)

Locally, this made loading the page for [USN-7449-1](https://osv.dev/vulnerability/USN-7449-1) go from >6 minutes to ~8 seconds (which is like a 50x speedup). 
For reference, loading the live page on osv.dev took me just over 50 seconds without this improvement, so this should hopefully go down to less than 1 second after it's merged.


gcp/website/frontend_handlers.py
```diff
@@ -336,21 +336,62 @@ def bug_to_response(bug, detailed=True):
     add_links(response)
     add_source_info(bug, response)
     add_stream_info(bug, response)
+    add_known_osv_bugs(response)
+    add_stream_strings(bug, response)
   return response
 
 
+def add_known_osv_bugs(response):
```
another-rex: nit: Add type hints

gcp/website/frontend_handlers.py
```diff
@@ -1044,11 +1081,11 @@ def _get_downstreams_of_bug(bug_id: str, bugs: dict[str,
   return downstreams
 
 
-def compute_downstream_hierarchy(target_bug_id: str,
-                                 downstreams: dict[str, list[str]]) -> str:
+def compute_downstream_hierarchy(
+    target_bug_id: str,
+    downstreams: dict[str, list[str]]) -> tuple[set[str], dict[str, set[str]]]:
```
another-rex: We can probably alias this big tuple into its own type so we can reuse it above in add_known_osv_bugs

another-rex (approved):
Nice! Though I think local testing gets a lot more benefit from less sequential datastore queries than production because of latency, so likely will be a lot less than 50x (guess only 5x-10x speedup?)

gcp/website/frontend_handlers.py
michaelkedar: Added hints to the changed functions

gcp/website/frontend_handlers.py
michaelkedar: turned this into a `typing.NamedTuple` of `ComputedHierarchy`

another-rex (approved):

jess-lowe (approved):
Great work!

jess-lowe (approved):

PR #3591, opened 2025-06-20
fix(vulnfeeds): deduplicate commit hashes on AffectedCommits

### LINKED ISSUE #3523: Duplicate fix commit hases
Image

Example: https://osv.dev/vulnerability/CVE-2022-25761

jess-lowe:
fixes #3523 
- added deduplication logic when duplicate commit hashes are found. 
- Added a testcase to test VersionInfo end-to-end to ensure no duplicate hashes.
- Also added a series of unit tests to test deduplication logic

There's probably a less convoluted way to do it, so would appreciate input.

### DIFF
--- a/vulnfeeds/cves/versions.go
+++ b/vulnfeeds/cves/versions.go
@@ -654,9 +654,7 @@ func deduplicateAffectedCommits(commits []models.AffectedCommit) []models.Affect
 		return []models.AffectedCommit{}
 	}
 	sort.Sort(ByAffectedCommit(commits))
-	uniqueCommits := slices.CompactFunc(commits, func(a, b models.AffectedCommit) bool {
-		return a.Repo == b.Repo && a.Introduced == b.Introduced && a.Fixed == b.Fixed && a.LastAffected == b.LastAffected
-	})
+	uniqueCommits := slices.Compact(commits)
 	return uniqueCommits
 }
 


vulnfeeds/cves/versions_test.go
```diff
@@ -970,6 +989,125 @@ func TestVersionInfoDuplicateDetection(t *testing.T) {
...
+
+func (a ByAffectedCommit) Less(i, j int) bool {
+	// Primary sort key: Repo
+	if a[i].Repo != a[j].Repo {
+		return a[i].Repo < a[j].Repo
+	}
+	// Secondary sort key: Fixed
+	return a[i].Fixed < a[j].Fixed
+}
+func TestDeduplicateAffectedCommits(t *testing.T) {
```
another-rex:
```suggestion
}

func TestDeduplicateAffectedCommits(t *testing.T) {
```

vulnfeeds/cves/versions_test.go
```diff
@@ -970,6 +989,125 @@ func TestVersionInfoDuplicateDetection(t *testing.T) {
...
+			// is determined by the order they were first encountered.
+			// However, when comparing slices, the order *must* match exactly.
+			// To make the comparison robust for the map method,
+			// we sort both the 'got' and 'expected' slices before comparison,
+			// ensuring that we are only testing for the presence of unique items,
+			// not their specific output order.
+			sort.Sort(ByAffectedCommit(got))
+			sort.Sort(ByAffectedCommit(tt.expected))
+
+			if !reflect.DeepEqual(got, tt.expected) {
```
another-rex: You can use `let diff := cmp.Diff(a, b, cmpopts.SortSlices(sortFunc)); diff != ""` to save a bunch of code here, and also pretty print the differences in the test output so it's easier to debug when it breaks.

vulnfeeds/cves/versions.go
```diff
@@ -626,13 +626,36 @@ func cleanVersion(version string) string {
...
+	if len(commits) == 0 {
+		return []models.AffectedCommit{}
+	}
+
+	seen := make(map[string]struct{})
+	uniqueCommits := make([]models.AffectedCommit, 0, len(commits))
+
+	for _, commit := range commits {
+		// Create a unique key for each commit based on the fields that define uniqueness.
+		key := commit.Repo + "|" + commit.Introduced + "|" + commit.Fixed + "|" + commit.LastAffected // Using a separator to avoid issues if parts of the string could combine
```
another-rex: I think you can just use commit as the key here, as it only contains strings.

another-rex: If you don't need to maintain the same order, you can also just use slices.Sort(), then slices.Compact(), and skip the set entirely. Though if you do need to maintain the order, please add a comment to the function explaining that.

vulnfeeds/cves/versions.go
```diff
@@ -626,23 +627,36 @@ func cleanVersion(version string) string {
...
-		// Create a unique key for each commit based on the fields that define uniqueness.
-		key := commit.Repo + "|" + commit.Introduced + "|" + commit.Fixed + "|" + commit.LastAffected // Using a separator to avoid issues if parts of the string could combine
-		if _, found := seen[key]; !found {
-			uniqueCommits = append(uniqueCommits, commit)
-			seen[key] = struct{}{}
-		}
-	}
-
+	sort.Sort(ByAffectedCommit(commits))
+	uniqueCommits := slices.CompactFunc(commits, func(a, b models.AffectedCommit) bool {
```
another-rex: Do you need compactfunc here? I believe if a struct is only simple fields, it automatically does field by field comparison.  (might be wrong though)

vulnfeeds/cves/versions.go
```diff
@@ -626,23 +627,36 @@ func cleanVersion(version string) string {
...
-	for _, commit := range commits {
-		// Create a unique key for each commit based on the fields that define uniqueness.
-		key := commit.Repo + "|" + commit.Introduced + "|" + commit.Fixed + "|" + commit.LastAffected // Using a separator to avoid issues if parts of the string could combine
-		if _, found := seen[key]; !found {
-			uniqueCommits = append(uniqueCommits, commit)
-			seen[key] = struct{}{}
-		}
-	}
-
+	sort.Sort(ByAffectedCommit(commits))
```
another-rex, line 656: Same here, you might be able to just use sort without ByAffectedCommit.

jess-lowe: Have to use ByAffectedCommit or it errors that []models.AffectedCommit doesn't implement the sort.Interface (missing the Len method). Also can't use slices.Sort otherwise it erros that models.AffectedCommit does not satisfy cmp.Ordered.

vulnfeeds/cves/versions.go
jess-lowe: I swear when I tried this the other day it gave me errors, but now it's not. Will put up the commit to change that soon.

another-rex (approved):
LGTM, thanks!

hogo6002:
do we need to manually reimport those records?

jess-lowe:
> do we need to manually reimport those records?

Yeah, but let's wait until after #3617 :)

Issue #3263, opened 2025-03-18
new datasource: Alpaquita Linux


i-bs:
Our security team has prepared Alpaquita Linux security data in OSV-schema format. The CLA is submitted. Please help with the futher process. Thanks.

We also have a few questions.
According to the [docs](https://google.github.io/osv.dev/rest-api-contribution/) there are three ways to push data: 1. a public Git repository; 2. a public GCS bucket; 3. to REST API endpoints.

1. Our probably most preferable way is (1) a public Git repository. Is there an example repo for that? I found https://github.com/google/oss-fuzz-vulns/ but really not sure it is that.
2. We'd want our data to be checked for quality beforehand. Where/who should we ask about it?

/label datasource

jess-lowe:
Hi @i-bs, how exciting! Thanks for considering contributing to OSV.dev! 

Credit to @andrewpollock for this:
**OSV.dev's onboarding process in a nutshell (actionable feedback welcome):**

- [ ] Decide if you're going to publish records via a Git repository, GCS bucket or [REST endpoint](https://google.github.io/osv.dev/rest-api-contribution/), which you have - off the top of my head, an example of a public Git repository is AlmaLinux: https://github.com/AlmaLinux/osv-database. 

- [ ] Create a PR to [reserve a prefix in the OSV-Schema](https://ossf.github.io/osv-schema/#id-modified-fields) (worked examples: https://github.com/ossf/osv-schema/pull/235 https://github.com/ossf/osv-schema/pull/223 https://github.com/ossf/osv-schema/pull/219)

   - [ ] We review the records you start publishing for OSV Schema [correctness](https://github.com/ossf/osv-schema/tree/main/validation) and [quality](https://google.github.io/osv.dev/data_quality.html) (the work happening under the [OSV Data Quality Program](https://github.com/orgs/google/projects/62) is also relevant here, as an FYI) as part of reviewing and merging that PR

- [ ] Create a PR to extend [purl_helpers.py](https://github.com/google/osv.dev/blob/master/osv/purl_helpers.py) (if appropriate)

- [ ] Create a PR to start [importing the records you are publishing into our test instance of OSV.dev](https://github.com/google/osv.dev/blob/master/source_test.yaml) and validate everything is working as intended there (worked example: https://github.com/google/osv.dev/pull/2086)

[...]

PR #3608, opened 2025-06-24
feat(test): add BellSoft Security Advisory feed to test environment (…

i-bs:
…#3263)

Ecosystems:
-   Alpaquita Linux
-   BellSoft Hardened Containers

ref: https://github.com/google/osv.dev/issues/3263

### DIFF
--- a/source_test.yaml
+++ b/source_test.yaml
@@ -57,6 +57,21 @@
   editable: False
   strict_validation: True
 
+- name: 'bellsoft'
+  versions_from_repo: False
+  type: 0
+  ignore_patterns: ['^(?!BELL-CVE-).*$']
+  directory_path: 'BELL-CVE'
+  repo_url: 'https://github.com/bell-sw/osv-database.git'
+  detect_cherrypicks: False
+  extension: '.json'
+  db_prefix: ['BELL-']
+  ignore_git: False
+  human_link: 'https://docs.bell-sw.com/security/cves/{{ BUG_ID }}/'
+  link: 'https://github.com/bell-sw/osv-database/blob/master/'
+  editable: False
+  strict_validation: True
+
 - name: 'bitnami'
   versions_from_repo: False
   type: 0


another-rex:
It looks like the records that are in that repository has IDs starting with CVE, but all the records has to being with the prefix defined in the schema: BELL-SA

We would also prefer the filename to also start with BELL-SA, but that's technically optional, but the record ID itself definitely has to start with BELL-SA.

i-bs:
thank you for your guidance.

> It looks like the records that are in that repository has IDs starting with CVE, but all the records has to being with the prefix defined in the schema: BELL-SA

I see. Then e.g. for CVE-2025-48060 we'll have our packages visible at URL https://osv.dev/vulnerability/BELL-SA-2025-48060 , not https://osv.dev/vulnerability/CVE-2025-48060 (with Alpine and Debian packages) ?

If so what can be done for having our packages in https://osv.dev/vulnerability/CVE-2025-48060 ?

> We would also prefer the filename to also start with BELL-SA, but that's technically optional, but the record ID itself definitely has to start with BELL-SA.

we can change it, yes.

Also you didn't notice the wrong `ecosystem` fields. We'll change it too.

jess-lowe:
> I see. Then e.g. for CVE-2025-48060 we'll have our packages visible at URL https://osv.dev/vulnerability/BELL-SA-2025-48060 , not https://osv.dev/vulnerability/CVE-2025-48060 (with Alpine and Debian packages) ?

That is correct. We're looking to actually move the Alpine and Debian packages away from this main CVE ID, and move them to their own records, so the CVE-ID version will only involve data ingested directly from NVD/CVEList. The logic behind this on our end is that there are too many trackers and ecosystems referencing the same vulns to meanfully display them in one record, and not all users will need information from multiple distros. This will also allow better clarity around where we get our data from.

Instead, please put the reference to the CVE-ID you are addressing in the 'upstream' field (this takes an array). On the frontend, all of the 'downstream' vulnerability references will be computed and shown like so on the 'upsteam' CVE ID entry:
image
side note: this section will be moved up higher for better visibility soon (PR is in review).

i-bs:
@jess-lowe , thank you so much for the detailed explanation.

Then before I fix the things you're pointing at, I need this to be checked and merged: https://github.com/ossf/osv-schema/pull/361

Thanks.

jess-lowe:
Awesome, I will nudge the relevant reviewers for you :)

i-bs:
@another-rex , I hope this time everything is right and in place.

One note: please ignore `CVE` folder in the OSV data repo. It'll be gone soon.

Also, if possible, please merge the https://github.com/ossf/osv-schema/pull/364 : it fixes one error in the schema I overlooked.

another-rex (approved):
Thanks!

i-bs:
@another-rex , thanks a lot.

The PR was merged 3 days but I don't see any relevant data at https://test.osv.dev/list . Are there any other steps I need to do?

PR #3616, opened 2025-06-30
chore: Various dependency tidy ups

another-rex:
- Regenerate protos with new protogen version (v6)
- Update poetry in CI to 2.1.3
- Relock all poetry lock files
- Sort keys on write_vulnerabilities to have deterministic tests
- Update snapshots
- Added `.iml` which is intellij config file to the gitignore
- Added a new function to kill datastore emulator if it is still running at the start of the test.
- Update readme to specify we use python 3.12


poetry.lock
```diff
@@ -1,4 +1,4 @@
-# This file is automatically @generated by Poetry 2.0.1 and should not be changed by hand.
+# This file is automatically @generated by Poetry 2.1.2 and should not be changed by hand.
```
cuixq, line 1: 2.1.3?

CONTRIBUTING.md
cuixq, line 1: Poetry version in this doc to be updated too.

cuixq (approved):
Poetry version in https://github.com/google/osv.dev/blob/master/.github/workflows/lint.yaml#L39 as well.

poetry.lock
another-rex: Sticking with 2.1.2 for now since there's no change when relocking with 2.1.3

michaelkedar (approved):

Issue #3633, opened 2025-07-02
Naive CVE5 to OSV Conversion


jess-lowe:
We are looking into implementing using the CVEList as an immediate ingestion pipeline, due to the presence of the versions fields in the CVE 5.x schema format. As such before optimising CVE5 to OSV Conversion, we first need an MVP to build off of.

This converter should only need to convert naively if given data in a common and expected format across the CVE5 JSON.

PR #3590, opened 2025-06-20
feat(vulnfeeds): Update CVE5 struct to be more detailed for eventual CVE5 ingestion

jess-lowe:
This is in preparation for CVE5 ingestion. It's not completed but it's a minimalish set.

PR #3635, opened 2025-07-02
feat(vulnfeeds): Implemented a naiveish CVE5 to OSV converter

jess-lowe:
This change implements the initial naive conversion of the CVE5 to OSV format as discussed in #3633 

Note this does not implement the ingestion pipeline - just the conversion.

Below is an example of a more accurate-than-what-we-have-from-NVD conversion that resulted from this converter. 
image

**Changes**
- New cvelist2osv Converter: Adds a new binary (vulnfeeds/cmd/cvelist2osv/main.go) responsible for processing CVE 5.x JSON files from the CVEList repository and converting them to the OSV schema.
- Refactored Common Logic: To avoid code duplication, common conversion logic has been extracted from the existing nvd-cve-osv converter into a new shared package (vulnfeeds/cmd/common/conversion.go). This includes functions for:
  - Identifying repositories from references.
  - Converting Git versions/tags to commit hashes.
  - Managing a cache of vendor-product-to-repository mappings.
- Updated CVE Structs: The Go structs for CVEs (vulnfeeds/cves/cve.go) have been updated to support the CVE 5.x format, including more detailed fields for affected products, versions, and provider metadata.
- New Test Cases: Added extensive test data and cases to validate the new conversion logic against real-world CVEv5 records.
This work lays the foundation for further enhancements, including more advanced regex-based and AI-assisted data extraction.


vulnfeeds/cmd/common/conversion.go
```diff
@@ -0,0 +1,259 @@
...
+	// If there is only one entry, delete the entry cache entry.
+	if len(cacheEntry) == 1 {
+		delete(cache, *vp)
+		return
+	}
+	cacheEntry = slices.Delete(cacheEntry, i, i+1)
+	cache[*vp] = cacheEntry
+}
+
+// Examines repos and tries to convert versions to commits by treating them as Git tags.
```
hogo6002: Follow the https://tip.golang.org/doc/comment#func to put function name at the beginning of the docstring.

vulnfeeds/cmd/cvelist2osv/main_test.go
```diff
@@ -0,0 +1,359 @@
...
+		{
+			name:  "CVE with lessThan",
+			cveID: "CVE-2025-1110", // GitLab
+			expectedVersionInfo: models.VersionInfo{
+				AffectedVersions: []models.AffectedVersion{
+					{Introduced: "18.0", Fixed: "18.0.1"},
+				},
+			},
+		},
+		// {
```
hogo6002: should we just remove it?

vulnfeeds/cves/cve.go
```diff
@@ -23,49 +23,84 @@ const (
...
+
+type Versions struct {
+	Version         string `json:"version"`
+	Status          string `json:"status"`
+	LessThanOrEqual string `json:"lessThanOrEqual"`
+	LessThan        string `json:"lessThan"`
+	VersionType     string `json:"versionType"`
+}
+
+type CVE5 struct {
```
hogo6002: would anything be empty? do we need to add the `omitempty` tag to any field?

jess-lowe: alright, this is done and updated in the #3590 PR. Please review it there :)

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -0,0 +1,484 @@
...
+)
+
+type ConversionOutcome int
+
+var ErrNoRanges = errors.New("no ranges")
+
+var ErrUnresolvedFix = errors.New("fixes not resolved to commits")
+
+func (c ConversionOutcome) String() string {
+	return [...]string{"ConversionUnknown", "Successful", "Rejected", "NoSoftware", "NoRepos", "NoRanges", "FixUnresolvable"}[c]
```
another-rex: nit:  Format this vertically so it's easier to see if it matches up with the consts.

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -0,0 +1,484 @@
...
+)
+
+var (
+	jsonPath  = flag.String("cve_json", "", "Path to CVEList JSON to examine.")
+	outDir    = flag.String("out_dir", "", "Path to output results.")
+	outFormat = flag.String("out_format", "OSV", "Format to output {OSV,PackageInfo}")
+)
+var Logger utility.LoggerWrapper
+var RepoTagsCache git.RepoTagsCache
+var Metrics struct {
```
another-rex, line 63: Let's define this as an actual type

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -0,0 +1,484 @@
...
+	}
+	var notes []string
+
+	v.Published = formatDateTimeRFC3339(cve.Metadata.DatePublished)
+	v.Modified = formatDateTimeRFC3339(cve.Metadata.DateUpdated)
+	v.References = vulns.ClassifyReferences(cve.Containers.CNA.References)
+	// v.AddSeverity(cve.Metrics)
+	return &v, notes
+}
+func ExtractVersionInfo(cve cves.CVE5, refs []string, validVersions []string, httpClient *http.Client) (v models.VersionInfo, notes []string) {
```
another-rex:
```suggestion
}

func ExtractVersionInfo(cve cves.CVE5, refs []string, validVersions []string, httpClient *http.Client) (v models.VersionInfo, notes []string) {
```

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -0,0 +1,484 @@
...
+	for _, reference := range refs {
+		// (Potentially faulty) Assumption: All viable Git commit reference links are fix commits.
```
another-rex: I would have guessed that some references could be the introduced commit as well?

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -0,0 +1,484 @@
...
+			v.AffectedCommits = append(v.AffectedCommits, commit)
+		}
+	}
+
+	gotVersions := false
+	cna := cve.Containers.CNA
+
+	for _, cveAff := range cna.Affected {
+
+		for _, vInfo := range cveAff.Versions {
```
another-rex:
```suggestion
	for _, cveAff := range cna.Affected {
		for _, vInfo := range cveAff.Versions {
```

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -0,0 +1,484 @@
...
+		Details: cves.EnglishDescription(cve.Containers.CNA.Descriptions),
+		Aliases: aliases,
+		Related: related,
+	}
+	var notes []string
+
+	v.Published = formatDateTimeRFC3339(cve.Metadata.DatePublished)
+	v.Modified = formatDateTimeRFC3339(cve.Metadata.DateUpdated)
+	v.References = vulns.ClassifyReferences(cve.Containers.CNA.References)
+	// v.AddSeverity(cve.Metrics)
```
another-rex: Should this be commented out? If not needed let's just remove it, otherwise add a reason here.

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -0,0 +1,484 @@
...
+	}
+
+	if err != nil {
+		fmt.Printf("Warning: Could not parse date string '%s': %v. Returning original.\n", dateStr, err)
+		return dateStr // Return original if parsing fails
+	}
+	return t.UTC().Format(time.RFC3339) // Format in UTC RFC3339
+}
+
+// BORROWED FROM VULNS.GO
```
another-rex: Can we extract these borrowed functions into a package that can be imported by both? (we can also put stuff like formatDateTime there as well)

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -0,0 +1,484 @@
...
+	gotVersions := false
+	cna := cve.Containers.CNA
+
+	for _, cveAff := range cna.Affected {
+
+		for _, vInfo := range cveAff.Versions {
+			if vInfo.Status != "affected" {
+				continue
+			}
+			introduced := ""
```
another-rex: You can just do `var introduced, fixed, lastAffected string`

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -0,0 +1,484 @@
...
+					notes = append(notes, fmt.Sprintf("Warning: %s is not a valid fixed version", fixed))
+				}
+
+			} else {
+
+				// only version number
+				// naive assumption: it only affects that version. More likely, it affects up to that version
+				// lastaffected = vInfo.Version
+				v.AffectedVersions, _ = cves.ExtractVersionsFromDescription(validVersions, vInfo.Version)
+				//check if it starts with "<" or "before" or "before version"
```
another-rex: I'm  not quite following the comments here, if it is more likely to be affecting up to that version, why are we not putting it in lastaffected?

And is the last comment something that's done in ExtractVersionFromDescription?

vulnfeeds/cmd/cvelist2osv/main.go
another-rex: Please add a docstring for this function, especially for the arguments.

another-rex: It looks like you are only calling this function from one place, but validVersions is nil, so it's a bit unclear what that's suppose to be fore.

vulnfeeds/cmd/cvelist2osv/main.go
another-rex: Also, it looks like this is just overwriting whatever is in v.AffectedVersions every time this runs, since it's in a loop, shouldn't this be appending the versions?

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -0,0 +1,484 @@
...
+			if slices.Contains(v.AffectedVersions, possibleNewAffectedVersion) {
+				// Avoid appending duplicates
+				continue
+			}
+			v.AffectedVersions = append(v.AffectedVersions, possibleNewAffectedVersion)
+		}
+	}
+	if !gotVersions {
+		var extractNotes []string
+		v.AffectedVersions, extractNotes = cves.ExtractVersionsFromDescription(validVersions, cves.EnglishDescription(cve.Containers.CNA.Descriptions))
```
another-rex: This seems to be overwriting v.AffectedVersions again when gotVersion is false, but I think if hasRange is false, you would have filled affectedVersions with something already.

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -0,0 +1,484 @@
...
+	}
+	return v, notes
+}
+
+// Takes a CVE record and outputs an OSV file in the specified directory.
+func CVEToOSV(CVE cves.CVE5, repos []string, cache git.RepoTagsCache, directory string) error {
+	// Create a base OSV record
+	v, notes := FromCVE(cves.CVEID(CVE.Metadata.CVEID), CVE)
+
+	versions, versionNotes := ExtractVersionInfo(CVE, repos, nil, http.DefaultClient)
```
another-rex: I would suggest passing in the http client from the parent here as well. And initialize a client in main with Context.

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -0,0 +1,484 @@
...
+	notes = append(notes, versionNotes...)
+
+	// Attempt to resolve Version Info to commits.
+	cna := CVE.Containers.CNA
+	maybeVendorName := "SOMETHING"
```
another-rex: maybe "Unknown" instead of "Something"?

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -0,0 +1,484 @@
...
+	v.Affected = append(v.Affected, affected)
+
+	// Save OSV record to a directory
+	vulnDir := filepath.Join(directory, maybeVendorName, maybeProductName)
+	err := os.MkdirAll(vulnDir, 0755)
+	if err != nil {
+		Logger.Warnf("Failed to create dir: %v", err)
+		return fmt.Errorf("failed to create dir: %v", err)
+	}
+	notesFile := filepath.Join(vulnDir, v.ID+".notes")
```
another-rex: nit: Move this closer to where the notes is being written

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -0,0 +1,484 @@
...
+		}
+	}
+
+	if gotNoRanges {
+		return ErrNoRanges
+	}
+	if gotUnresolvedFix {
+		return ErrUnresolvedFix
+	}
+	return fileWriteErr
```
another-rex: I think we want to prioritise fileWriteErr over the other errors.

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -0,0 +1,484 @@
...
+	for _, affected := range cve.Containers.CNA.Affected {
+		if affected.CollectionUrl != "" {
+			refs = append(refs, cves.Reference{Url: affected.CollectionUrl})
+		}
+		if affected.Repo != "" {
+			refs = append(refs, cves.Reference{Url: affected.Repo})
+		}
+	}
+
+	slices.SortStableFunc(refs, func(a, b cves.Reference) int {
```
another-rex, line 388: nit: Add a commit here saying this is to remove duplicates URLs

vulnfeeds/cmd/cvelist2osv/main.go
another-rex: Hmm... looking at the actual usages, are any of the fields actually being used except outcomes and OSVRecordsGenerated?

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -0,0 +1,484 @@
...
+	}
+	w.Flush()
+
+	if err = w.Error(); err != nil {
+		return err
+	}
+	return nil
+}
+
+func identifyPossibleURLs(cve cves.CVE5) []cves.Reference {
```
another-rex, line 370: Add a docstring here explaining what this is doing, e.g. collected URLs as cve.Refernces. Btw does this have to be collected as References? it seems to make more sense to just be an array of strings?

another-rex (changes requested):
Thanks! Two more general comments:
- There seems to be a lot of inconsistent whitespace/newlines, can you go over it and just clean that up?
- Copy over the .golangci.yaml file from Andrew's PR here https://github.com/google/osv.dev/pull/3164/files and run it on cvelist2osv. 
```
go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.1.6 run -c "path to config" ./vulnfeeds/cmd/cvelist2osv/...
```

We should do a pass over the existing code and merge that PR, but for now at least we can apply it to new code.

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -0,0 +1,484 @@
...
+var RepoTagsCache git.RepoTagsCache
+var Metrics struct {
+	TotalCVEs           int
+	CVEsForApplications int
+	CVEsForKnownRepos   int
+	OSVRecordsGenerated int
+	Outcomes            map[cves.CVEID]ConversionOutcome // Per-CVE-ID record of conversion result.
+}
+
+// References with these tags have been found to contain completely unrelated
```
hogo6002: nit: starts with "RefTagDenyList ... "

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -0,0 +1,484 @@
...
+var VendorProductDenyList = []cves.VendorProduct{
+	// Three strikes and the entire netapp vendor is out...
+	{Vendor: "netapp", Product: ""},
+	// [CVE-2021-28957]: Incorrectly associates with github.com/lxml/lxml
+	{Vendor: "oracle", Product: "zfs_storage_appliance_kit"},
+	{Vendor: "gradle", Product: "enterprise"}, // The OSS repo gets mis-attributed via CVE-2020-15767
+}
+
+// formatDateTimeRFC3339 parses an ISO-like date string and returns RFC3339 format or original on error.
+func formatDateTimeRFC3339(dateStr string) string {
```
hogo6002: We also have a similar function here https://github.com/google/osv.dev/blob/master/vulnfeeds/vulns/vulns.go#L381
Can we merge two into one?

vulnfeeds/cmd/cvelist2osv/main.go
hogo6002: +1 it will better for us only maintain same function at one place.

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -0,0 +1,484 @@
...
+				fileWriteErr = err
+			} else {
+				Logger.Infof("[%s]: Generated OSV record for %q", CVE.Metadata.CVEID, maybeProductName)
+			}
+			f.Close()
+		}
+	}
+	Logger.Warnf("numNotes %v", len(notes))
+	if len(notes) > 0 {
+
```
hogo6002: remove newline

vulnfeeds/cmd/cvelist2osv/main.go
hogo6002: This function is quite long and doing multiple things. Can we break down into some smaller helper functions to improve readabilty?

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -0,0 +1,484 @@
...
+	for _, cveAff := range cna.Affected {
+		maybeVendorName = cveAff.Vendor
+		maybeProductName = cveAff.Product
+		continue
+	}
+
+	var gotUnresolvedFix, gotNoRanges bool
+
+	if len(versions.AffectedVersions) != 0 {
+		// There are some AffectedVersions to try and resolve to AffectedCommits.
```
hogo6002: This part can be extracted into a separate function to reduce indentation

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -0,0 +1,484 @@
...
+		err = outputOutcomes(Metrics.Outcomes, ReposForCVE, *outDir)
+		if err != nil {
+			// Log entry with size 1.15M exceeds maximum size of 256.0K
+			fmt.Fprintf(os.Stderr, "Failed to write out metrics: %v", err)
+		}
+		// Outcomes is too big to log, so zero it out.
+		Metrics.Outcomes = nil
+		Logger.Infof("%s Metrics: %+v", filepath.Base(*jsonPath), Metrics)
+	}
+
```
hogo6002: remove newline

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -0,0 +1,484 @@
...
+			Logger.Warnf("[%s]: Failed to generate an OSV record: %+v", CVEID, err)
+			if errors.Is(err, ErrNoRanges) {
+				Metrics.Outcomes[CVEID] = NoRanges
+
+			}
+			if errors.Is(err, ErrUnresolvedFix) {
+				Metrics.Outcomes[CVEID] = FixUnresolvable
+
+			}
+			Metrics.Outcomes[CVEID] = ConversionUnknown
```
hogo6002: this will override no matter what error we found. Do you want do:
```
if errors.Is(err, ErrNoRanges) {
				Metrics.Outcomes[CVEID] = NoRanges
			} else if errors.Is(err, ErrUnresolvedFix) {
				Metrics.Outcomes[CVEID] = FixUnresolvable
			} else {
                                 Metrics.Outcomes[CVEID] = ConversionUnknown
                         }
```
instead?

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -0,0 +1,484 @@
...
+		Logger.Fatalf("Failed to open file: %v", err) // double check this is best practice output
+	}
+
+	var cve cves.CVE5
+	err = json.Unmarshal(data, &cve)
+	if err != nil {
+		Logger.Fatalf("Failed to parse CVEList CVE JSON: %v", err)
+	}
+
+	// VPRepoCache := make(VendorProductToRepoMap)
```
hogo6002: should we just remove it?

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -0,0 +1,484 @@
...
+			}
+			if errors.Is(err, ErrUnresolvedFix) {
+				Metrics.Outcomes[CVEID] = FixUnresolvable
+
+			}
+			Metrics.Outcomes[CVEID] = ConversionUnknown
+
+		}
+		Metrics.OSVRecordsGenerated++
+		Metrics.Outcomes[CVEID] = Successful
```
hogo6002: this also overrides the final outcome right?

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -0,0 +1,484 @@
...
+			Logger.Warnf("[%s]: Failed to derive any repos", CVEID)
+		}
+		Logger.Infof("[%s]: Derived %q for CVE", CVEID, repos)
+		ReposForCVE[CVEID] = repos
+	}
+
+	Logger.Infof("[%s]: Repos: %#v", CVEID, ReposForCVE[CVEID])
+
+	switch *outFormat {
+	case "OSV":
```
hogo6002, line 438: do we need switch here? seems only one case.

hogo6002 (commented):
Adding some comments on style. It'll be better to break some long functions into smaller ones. You can do this in a followup PR if that's your plan.

vulnfeeds/cmd/cvelist2osv/main.go
jess-lowe: PackageInfo will be implemented in a future PR :)

vulnfeeds/cmd/cvelist2osv/main.go
jess-lowe: These are brought over from the NVD conversion, and proper metrics and consideration will be determined in a future PR.

vulnfeeds/cmd/cvelist2osv/main.go
jess-lowe: Added docstring, and yeah it does need to be collected as references unfortunately, as it gets passed to FromCVE which calls ClassifyReferences, which uses the tags field of References.

vulnfeeds/cmd/cvelist2osv/main.go
jess-lowe: deleted it entirely to use the alternative

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -82,107 +88,66 @@ var VendorProductDenyList = []cves.VendorProduct{
...
+//   - cna: The CNA container from the CVE5 object.
+//   - gotVersions: A boolean indicating if any versions have already been extracted.
+//   - versionInfo: The models.VersionInfo struct to populate with affected versions.
+//
+// Returns:
+//   - []models.AffectedVersion: The updated slice of affected versions.
+//   - bool: True if any versions were successfully extracted, false otherwise.
+//   - []string: A slice of notes or warnings generated during the extraction process.
+
+func extractVersionsFromAffected(cna cves.CNA, gotVersions bool, versionInfo models.VersionInfo) ([]models.AffectedVersion, bool, []string) {
```
another-rex: If versionInfo is being modified/populated and we need the results, shouldn't it be a pointer?

vulnfeeds/cves/versions.go
```diff
@@ -594,7 +594,7 @@ func processExtractedVersion(version string) string {
 	return version
 }
 
-func ExtractVersionsFromDescription(validVersions []string, description string) ([]models.AffectedVersion, []string) {
+func ExtractVersionsFromText(validVersions []string, description string) ([]models.AffectedVersion, []string) {
```
another-rex: description -> text

vulnfeeds/cmd/cvelist2osv/main.go
another-rex: Also, if it is being populated in place, this should not be returning models.AffectedVersion as well

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -458,18 +444,16 @@ func main() {
...
-			}
-			if errors.Is(err, ErrUnresolvedFix) {
+			} else if errors.Is(err, ErrUnresolvedFix) {
 				Metrics.Outcomes[CVEID] = FixUnresolvable
-
+			} else {
+				Metrics.Outcomes[CVEID] = ConversionUnknown
 			}
-			Metrics.Outcomes[CVEID] = ConversionUnknown
+		} else {
```
another-rex, line 455: Is this correct?

another-rex (changes requested):

vulnfeeds/cmd/cvelist2osv/main.go
jess-lowe: Probably not but this will be reworked in a future PR for metrics anyway.

PR #3637, opened 2025-07-04
fix: 500 error on frontend rendering of some Ubuntu CVEs

another-rex:
Remove assumption on the frontend where we assume that it's only CVSS scores.

This is a bit hacked together and on the frontend it'll look a bit weird because type doesn't actually exist right now:

image


jess-lowe (approved):
LGTM

PR #3638, opened 2025-07-07
perf(api): Improve asynchronous behaviour

michaelkedar:
Change all(?) the ndb calls used throughout the API handlers to use the async versions, and rewrote a few functions to not block on the completion of futures that it doesn't need to (particularly, converting the Bug to a proto, and populating the related/alias/upstream fields of those protos).

Testing on my private instance, this should be a 3-5x speedup for queries with lots of vulnerabilities.

I've also made the batch queries properly set the modified time from the alias/upstream group (which might actually be a tiny performance loss) to ensure the modified time is always the same for the batch query and the get by ID.

I've added `@ndb.synctasklet` to the rpc handlers to let us use `yield` instead of `.result()`, just to discourage usage of it (using `.result()` in a tasklet can cause stack overflows)

### DIFF
--- a/osv/models.py
+++ b/osv/models.py
@@ -665,7 +665,7 @@ class Bug(ndb.Model):
     if self.last_modified:
       modified_times.append(self.last_modified)
 
-    # Correctly fetch the last_modified dates from the upstream/alias groups
+    # Fetch the last_modified dates from the upstream/alias groups.
     alias_future = get_aliases_async(self.id()) if include_alias else None
     upstream_future = (
         get_upstream_async(self.id()) if include_upstream else None)
--- a/osv/models.py
+++ b/osv/models.py
@@ -656,6 +656,37 @@ class Bug(ndb.Model):
 
     return vulnerability_pb2.Vulnerability(id=self.id(), modified=modified)
 
+  @ndb.tasklet
+  def to_vulnerability_minimal_async(self,
+                                     include_alias=True,
+                                     include_upstream=True):
+    """Convert to Vulnerability proto (minimal) asynchronously."""
+    modified_times = []
+    if self.last_modified:
+      modified_times.append(self.last_modified)
+
+    # Correctly fetch the last_modified dates from the upstream/alias groups
+    alias_future = get_aliases_async(self.id()) if include_alias else None
+    upstream_future = (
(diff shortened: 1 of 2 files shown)


osv/models.py
```diff
@@ -656,6 +656,37 @@ def to_vulnerability_minimal(self):
...
+  @ndb.tasklet
+  def to_vulnerability_minimal_async(self,
+                                     include_alias=True,
+                                     include_upstream=True):
+    """Convert to Vulnerability proto (minimal) asynchronously."""
+    modified_times = []
+    if self.last_modified:
+      modified_times.append(self.last_modified)
+
+    # Correctly fetch the last_modified dates from the upstream/alias groups
```
another-rex: nit: "Correctly" probably doesn't make sense as a comment. (Since the code hopefully is always correct)

osv/models.py
```diff
@@ -656,6 +656,37 @@ def to_vulnerability_minimal(self):
...
+
+    modified = None
+    if modified_times:
+      modified = timestamp_pb2.Timestamp()
+      modified.FromDatetime(max(modified_times))
+
+    return vulnerability_pb2.Vulnerability(id=self.id(), modified=modified)
+
   def to_vulnerability(self,
                        include_source=False,
```
another-rex, line 691: Maybe something better suited to a followup PR, but I would suggest removing all these flags and force callers to use the async version if they want to include stuff.

osv/models.py
```diff
@@ -656,6 +656,37 @@ def to_vulnerability_minimal(self):
 
     return vulnerability_pb2.Vulnerability(id=self.id(), modified=modified)
 
+  @ndb.tasklet
+  def to_vulnerability_minimal_async(self,
```
another-rex, line 660: Should we also be converting to_vulnerability_minimal to just call the async version then blocking so they both return the same data?

another-rex (approved):
RIP types, all `ndb.Future`s now.

osv/models.py
michaelkedar: I might just leave it for now, since it's kind of the opposite of your other comment.
(I don't think `to_vulnerability_minimal()` is even used anymore anyway)

hogo6002 (approved):

jess-lowe (approved):

Issue #3350, opened 2025-04-08
OSV.dev needs a navbar search option


jess-lowe:
**Is your feature request related to a problem? Please describe.**
The website doesn't follow normal user flow. We have a lot of recurring users, who just want to search a vuln on our website but have to read a button in the middle of the page instead of clicking a search icon or using a nav search bar

**Describe the solution you'd like**
I want a magnifying glass in the nav bar that either:
- folds out to a full search bar when clicked 
- goes to search vuln database page. 

also add a magnifying glass icon to the 'Search Vulnerability Database' button or remove it entirely. 

**Additional context**

![Image](https://github.com/user-attachments/assets/2857115c-1dd2-4984-a5cd-0eea29205d1e)

ashmod:
i've added search functionality in the navbar + magnifying glass icon to the Search Vulnerability Database button, please review #3657 

<img width="1322" height="539" alt="Image" src="https://github.com/user-attachments/assets/518e4211-3fe9-4be9-90ea-9eb982f4c169" />

Issue #382, opened 2022-04-06
V2 UI: Search suggestions.


oliverchang:
Add some autocompletion/suggestions to the search box

![osv03](https://user-images.githubusercontent.com/759062/200988012-3f061b9c-d3ff-46b7-ae03-2acc1885d180.png)

zahraaalizadeh:
@another-rex Was wondering if you still want to work on this issue? otherwise I'll be happy to work on it.

oliverchang:
This one is probably out of scope -- to do it well it would require a number of complicated backend changes (as well as possibly a DB stack migration).

PR #3657, opened 2025-07-09
feat(website): added search option to navbar

dizzydroid:
# Navbar search option

Addresses #3350

## Description

This PR adds an expandable search bar to the navbar, making it easier for users to search the vulnerability database from any page. The search icon expands into a search field when clicked and collapses when clicked again or when clicking outside the search area. Additionally, I added a search icon to the "Search Vulnerability Database" button for a more intuitive user experience overall.

## Implementation Details

- Created a new `search.js` module with an `ExpandableSearch` class to handle search functionality
- Added search bar HTML structure to `base.html`
- Added SCSS styles for the search bar in `styles.scss`
- Added search suggestions API endpoint in `frontend_handlers.py`
- Ensured proper initialization during page load and Turbo navigation in `index.js`


ashmod:
signed

jess-lowe:
Hey @dizzydroid! Thanks for contributing! I'm so excited to see this issue get picked up - it's been bugging me for months! I've just run the code locally, with `make run-website-emulator` and this is what it looks like for me:

image

The addition of the icon next to the button is awesome, but the expandable search bar doesn't seem to work (or look as intended). I'm not sure how well the suggestions API will work with our current infrastructure - we're using Datastore which doesn't have any good options for fuzzy searching, but kudos if you can make it work nicely! Here's the issue that was previously being tracked for search improvements: #382

ashmod:
> Hey @dizzydroid! Thanks for contributing! I'm so excited to see this issue get picked up - it's been bugging me for months! I've just run the code locally, with `make run-website-emulator` and this is what it looks like for me:
> 
> image
> 
> The addition of the icon next to the button is awesome, but the expandable search bar doesn't seem to work (or look as intended). I'm not sure how well the suggestions API will work with our current infrastructure - we're using Datastore which doesn't have any good options for fuzzy searching, but kudos if you can make it work nicely! Here's the issue that was previously being tracked for search improvements: #382 

Hello, yes that doesn't look right I will investigate this and update in a few hours. Thank you for the info on the suggestions feature, I was not aware there was an open issue :D , I'll look into it and see if it's applicable with the current configuration too.

ashmod:
Hi, @jess-lowe , I have pushed a fix, apparently there was a duplicate search bar I forgot to remove. It should be displaying fine now, let me know if it works now on your end.
(sometimes the cache acts funny so also make sure to do a hard refresh if it doesn't work right away)
image
image

The suggestions are still not functional atm, but the API code is still there. We could probably delegate that to issue #382 as a separate feature since it will probably require some additional refactoring.

jess-lowe:
Works on my end now! - It looks great! Thanks for this!

ashmod:
applied linter fixes

jess-lowe:
Awesome, maybe (if you don't mind), we can extract the suggestions API to a separate PR so that we can merge in the parts that definitely work now, and work on the parts that might be hindered by Datastore and other blockers separately? As you said earlier, the #382 issue would be a great tracker for this.

ashmod:
Yes, I agree! sorry for the late reply (I think I'm on the other side of earth 😄), I've removed the suggestions api from this pr and extracted it to the new PR #3660 as a separate feature.

gcp/website/frontend3/src/index.js
```diff
@@ -49,3 +50,31 @@ export class MdTextFieldWithEnter extends MdFilledTextField {
 }
 customElements.define('md-textfield-with-enter', MdTextFieldWithEnter);
 
+let searchInstance = null;
+
+function initializeSearch() {
+  // Clean up previous instance if it exists
+  if (searchInstance) {
+    // Previous instance will be cleaned up by the new instance
+  }
```
jess-lowe:
```suggestion
```

gcp/website/frontend3/src/search.js
```diff
@@ -0,0 +1,130 @@
...
+      if (isActive) {
+        this.closeSearch(containerInfo);
+      } else {
+        this.openSearch(containerInfo);
+      }
+    };
+    
+    searchToggle.removeEventListener('click', toggleHandler);
+    searchToggle.addEventListener('click', toggleHandler);
+    searchToggle.setAttribute('data-search-initialized', 'true');
```
jess-lowe:
```suggestion
    if (!searchToggle.hasAttribute('data-search-initialized')) {
      searchToggle.addEventListener('click', toggleHandler);
      searchToggle.setAttribute('data-search-initialized', 'true');
    }
```
We can use the data attribute here to prevent attaching the same listener multiple times :) - because toggleHandler is defined inside the `initializeSearchContainer` method, every time  `initializeSearchContainer`  is called, a new `toggleHandler` function is created in memory.

gcp/website/frontend3/src/search.js
```diff
@@ -0,0 +1,130 @@
...
+    const submitHandler = (e) => {
+      const query = searchInput.value.trim();
+      if (!query) {
+        e.preventDefault();
+      }
+    };
+    
+    searchForm.removeEventListener('submit', submitHandler);
+    searchForm.addEventListener('submit', submitHandler);
+    searchForm.setAttribute('data-search-initialized', 'true');
```
jess-lowe:
```suggestion
    if (!searchForm.hasAttribute('data-search-initialized')) {
      searchForm.addEventListener('submit', submitHandler);
      searchForm.setAttribute('data-search-initialized', 'true');
    }
```
Same as above :)

ashmod:
Thanks for the detailed review and the clear explanation! That was really helpful. I've incorporated all the feedback, let me know what you think!

hogo6002 (approved):

PR #3660, opened 2025-07-11
feat(website): Add search suggestions to OSV database

dizzydroid:
This PR introduces a new `/api/search_suggestions` endpoint that aims to provide real-time search suggestions for the OSV vulnerability database.

This PR is part of a larger effort to enhance the search experience across the website (see PR #3657)
Addresses issue #382


ashmod:
image
dropdown suggestions should now work on the Vulnerability Database page.

jess-lowe:
This may be due to the merge conflicts but this is what it currently looks like on my end:
image

ashmod:
I'll resolve the conflicts and look into this, but I have a feeling this could be related to Turbo's cache, did you try a hard reload?

ashmod:
I'm facing the same issue actually but it goes away when I do a hard reload or try in incognito somehow which is strange XD - I will investigate this further and also try to add the dropdown feature to the navbar search too.

ashmod:
I've pushed a fix (hopefully), implemented a few retry mechanisms and fallbacks to ensure the page loads well on the first try. Also restructured the modules by exporting the Search Suggestions class to `search.js` instead of cluttering `index.js`. Let me know if this works now - next step should be implementing the dropdown on the navbar's searchbox.

ashmod:
image
image

Suggestions are currently working well on the website emulator, both frontend and backend seem to be stable on my end - frontend renders well on first load and backend returns the suggestions array in the response body as expected. Let me know if it works on your end too.

(About the styling, I've went with this change because I didn't like the older one especially the white border because it looked strange in the navbar 😅, I've experimented a few styles and this is the one that looked best but feel free to ask for any specific styling changes!)

jess-lowe:
Awesome! My small nit would be - would it be possible to show the ids in the search suggestion in uppercase instead of lower? - most vuln IDs present as uppercase :). I'll review the rest of the code when I get the chance!

gcp/website/frontend3/src/styles.scss
```diff
@@ -1740,56 +1746,110 @@ div.highlight {
   }
 }
 
-// Search suggestions dropdown styles
+/* Search Suggestions Component */
```
another-rex: Can you remove the bottom rounded border on the search box when suggestions is visible to it connects more seamlessly

gcp/website/frontend3/src/index.js
```diff
@@ -53,25 +53,92 @@ customElements.define('md-textfield-with-enter', MdTextFieldWithEnter);
...
 function initializeSearch() {
-  // Clean up previous instance if it exists
   searchInstance = new ExpandableSearch();
 }
 
-if (document.readyState === 'complete' || document.readyState === 'interactive') {
-  setTimeout(initializeSearch, 0);
+// Ensure initialization happens after all dependencies are loaded
+function ensureInitialization() {
+  if (typeof customElements !== 'undefined' && customElements.get('md-filled-text-field')) {
```
another-rex:
```suggestion
  if (customElements && customElements.get('md-filled-text-field')) {
```
I think it's cleaner to just do a truthy check instead of specifically for undefined.

another-rex: Hmm... doing a quick search, it looks like customElements can only be undefined if the browser doesn't support it. In that case we probably shouldn't keep trying to ensureInitialization and just exit early.

gcp/website/frontend3/src/index.js
```diff
@@ -53,25 +53,92 @@ customElements.define('md-textfield-with-enter', MdTextFieldWithEnter);
...
+      this.initializationRetries++;
+      setTimeout(() => this.initializeSuggestions(), 50);
+      return;
+    }
+    
+    try {
+      if (!this.suggestionsManager && this.isConnected) {
+        this.suggestionsManager = new SearchSuggestionsManager(this);
+      }
+    } catch (error) {
```
another-rex, line 123: What types of errors are we expecting here? Having a quick look at the constructor it doesn't look like it's likely to error. 

Can we also just delay the construction till everything is ready rather than retrying?

gcp/website/frontend3/src/search.js
```diff
@@ -127,6 +144,223 @@ export class ExpandableSearch {
...
+    
+    this.input.addEventListener('input', () => {
+      if (this.isDestroyed) return;
+      this.selectedIndex = -1;
+      clearTimeout(this.debounceTimer);
+      this.debounceTimer = setTimeout(() => this.handleInput(), 300);
+    });
+
+    this.input.addEventListener('keydown', (e) => this.handleKeydown(e));
+    this.input.addEventListener('blur', () => setTimeout(() => this.hide(), 200));
```
another-rex: Can you clarify what this delay is for?

gcp/website/frontend3/src/search.js
```diff
@@ -127,6 +144,223 @@ export class ExpandableSearch {
...
+      this.suggestionsElement.style.top = `${formRect.bottom}px`;
+      this.suggestionsElement.style.width = `${formRect.width}px`;
+    } else if (queryContainer) {
+      // Main search
+      const containerRect = queryContainer.getBoundingClientRect();
+      this.suggestionsElement.style.left = `${containerRect.left}px`;
+      this.suggestionsElement.style.top = `${containerRect.bottom}px`;
+      this.suggestionsElement.style.width = `${containerRect.width}px`;
+    } else {
+      // Fallback
```
another-rex: log a warning here, as we shouldn't be relying on the fallback directly.

gcp/website/frontend3/src/search.js
```diff
@@ -127,6 +144,223 @@ export class ExpandableSearch {
...
+    if (this.suggestionsElement) {
+      this.suggestionsElement.style.display = 'none';
+    }
+    this.selectedIndex = -1;
+  }
+
+  updatePosition() {
+    const rect = this.input.getBoundingClientRect();
+    const queryContainer = this.input.closest('.query-container');
+    const isNavbarSearch = this.input.closest('.search-container-nav');
```
another-rex: Can we generalise this, let's add a new class to the container we want the search suggestion to match width to, and use that here.

gcp/website/frontend3/src/search.js
```diff
@@ -127,6 +144,223 @@ export class ExpandableSearch {
...
+  }
+
+  init() {
+    this.createSuggestionsElement();
+    this.setupEventListeners();
+  }
+
+  cleanupOrphanedElements() {
+    const orphanedElements = document.querySelectorAll('.search-suggestions');
+    orphanedElements.forEach(element => {
```
another-rex: Will this cause issues if we have two search boxes on the page? (on the list page we have the main one, and the one in the top left corner that you added earlier.)

gcp/website/frontend3/src/templates/list.html
```diff
@@ -35,10 +35,10 @@ <h1 class="title">Vulnerabilities</h1>
           <form action="{{ url_for('frontend_handlers.list_vulnerabilities') }}" data-turbo-frame="vulnerability-table">
             <div class="mdc-layout-grid__inner">
               <div class="query-container mdc-layout-grid__cell--span-8">
-                <md-textfield-with-enter label="Package or ID search" class="query-field" type="search" name="q"
-                  value="{{ query }}" maxlength="300">
+                <md-textfield-with-suggestions label="Package or ID search" class="query-field" type="search" name="q"
+                  value="{{ query }}">
```
another-rex: Can we keep maxlength?

gcp/website/frontend3/src/search.js
```diff
@@ -127,6 +144,223 @@ export class ExpandableSearch {
...
+    
+    // Add a unique identifier to track this element
+    this.managerId = `suggestions-${Date.now()}-${Math.random().toString(36).substr(2, 9)}`;
+    this.suggestionsElement.dataset.managerId = this.managerId;
+    
+    // Ensure document.body exists before appending
+    if (document.body) {
+      document.body.appendChild(this.suggestionsElement);
+    } else {
+      const checkBody = () => {
```
another-rex: nit: this can just be a top level function, and we don't need the document.body if statement above, just call this function right?

gcp/website/frontend3/src/search.js
```diff
@@ -127,6 +144,223 @@ export class ExpandableSearch {
...
+    this.input.addEventListener('keydown', (e) => this.handleKeydown(e));
+    this.input.addEventListener('blur', () => setTimeout(() => this.hide(), 200));
+  }
+
+  async handleInput() {
+    if (this.isDestroyed) return;
+    
+    const query = this.input.value.trim();
+    
+    if (query.length < 2) {
```
another-rex: nit: make the min length a constant. (And can you bump this up to 3, we can lower it later if it's too high)

gcp/website/frontend_handlers.py
```diff
@@ -1139,3 +1139,29 @@ def compute_downstream_hierarchy(
...
+  db_query = db_query.order(osv.Bug.search_tags)
+  bugs = db_query.fetch(max_suggestions, projection=[osv.Bug.search_tags])
+
+  # Build suggestion list
+  suggestions = sorted(
+      list(
+          set(tag.upper() for bug in bugs for tag in bug.search_tags
+              if tag.lower().startswith(query))))
+
+  return json.dumps({'suggestions': suggestions[:max_suggestions]})
```
another-rex: Can you `update gcp/datastore/index.yml` with a new index with the query parameters here.

gcp/website/frontend_handlers.py
```diff
@@ -1139,3 +1139,29 @@ def compute_downstream_hierarchy(
...
+      osv.Bug.search_tags >= query,
+      osv.Bug.search_tags < query + '\ufffd')
+  db_query = db_query.order(osv.Bug.search_tags)
+  bugs = db_query.fetch(max_suggestions, projection=[osv.Bug.search_tags])
+
+  # Build suggestion list
+  suggestions = sorted(
+      list(
+          set(tag.upper() for bug in bugs for tag in bug.search_tags
+              if tag.lower().startswith(query))))
```
another-rex, line 1168: Let's make this a constant at the top of this file.

gcp/website/frontend3/src/search.js
```diff
@@ -127,6 +144,223 @@ export class ExpandableSearch {
...
+  }
+
+  createSuggestionsElement() {
+    if (this.suggestionsElement) {
+      return;
+    }
+    
+    this.suggestionsElement = document.createElement('div');
+    this.suggestionsElement.classList.add('search-suggestions');
+    this.suggestionsElement.style.display = 'none';
```
another-rex: Rather than editing the style directly, let's just add a class for hiding this. This way we can change the display mode to block or whatever else in the css `.search-suggestions` without having to update the javascript.

gcp/website/frontend3/src/search.js
```diff
@@ -127,6 +144,223 @@ export class ExpandableSearch {
...
+    
+    // Add a unique identifier to track this element
+    this.managerId = `suggestions-${Date.now()}-${Math.random().toString(36).substr(2, 9)}`;
```
another-rex, line 182: Can you edit the  comment to include *why* we need a unique ID here? I.e. it's used for cleanup

gcp/website/frontend3/src/search.js
another-rex: Wait, this is just checking whether managerId exists, wouldn't that always be the case, since it's added every time you create it, and I don't really see anywhere you remove this.

gcp/website/frontend3/src/search.js
```diff
@@ -1,6 +1,9 @@
+import { submitForm } from './index.js';
+
 export class ExpandableSearch {
   constructor() {
     this.containers = [];
+    this.suggestionsManagers = new Map(); // Track suggestions managers per container
```
another-rex: I'm still not quite clear on what the map is for... when will we have multiple suggestionManagers, (since this just exists at the top right corner) especially multiple per instance of ExpandableSearch?

another-rex (changes requested):
Thanks, very nice! Made a bunch of comments. 

I do have a question on why the searchSuggestionsManager need to be cleanedup and recreated so much (at least code length wise), is it just because of the "turbo" load?

ashmod:
@another-rex thanks for the review, I'm going through them now and will update! Yes, I've added those retries because of the Turbo cache bug, it seemed that this was the way I got it to stabilize and not load unstyled without needing a forced reload. Weird bug but that seemed to work :)

gcp/website/frontend3/src/index.js
ashmod: Thanks, fixed in [c25b025](https://github.com/dizzydroid/osv.dev/commit/c25b0250714243ef7b1cc37a34664bc00f750c85)

gcp/website/frontend3/src/index.js
ashmod: The errors could be dependencies not being loaded fully (I initially added this change to try to get to the root of the Turbo cache problem), or maybe just a problem with the DOM's rendering. I've removed the retry logic, should be cleaner now in [c25b025](https://github.com/dizzydroid/osv.dev/commit/c25b0250714243ef7b1cc37a34664bc00f750c85)

gcp/website/frontend3/src/templates/list.html
ashmod: Yes! Added back in [c25b025](https://github.com/dizzydroid/osv.dev/commit/c25b0250714243ef7b1cc37a34664bc00f750c85)

gcp/website/frontend_handlers.py
ashmod: Fixed in [b4accf5](https://github.com/dizzydroid/osv.dev/commit/b4accf58a43a19dceaf8bc9cbc7ea0fda90b416c)

gcp/website/frontend_handlers.py
ashmod: Fixed in [b4accf5](https://github.com/dizzydroid/osv.dev/commit/b4accf58a43a19dceaf8bc9cbc7ea0fda90b416c)

gcp/website/frontend3/src/search.js
ashmod: checkBody is a closure though, wouldn't making it a top-level function mean it would lose access to the current instance?

gcp/website/frontend3/src/search.js
ashmod: This delay should allow time for the click on a suggestion in the dropdown to be processed before the `hide` is triggered on it.

gcp/website/frontend3/src/search.js
ashmod: Should be fixed in [15be169](https://github.com/dizzydroid/osv.dev/commit/15be169bc344d12bd413b9700d20d3438c4621e1)

gcp/website/frontend3/src/search.js
ashmod: Yes! Fixed in [15be169](https://github.com/dizzydroid/osv.dev/commit/15be169bc344d12bd413b9700d20d3438c4621e1)

gcp/website/frontend3/src/search.js
ashmod: You're right! Seems like this was part of a previous attempt and I forgot to remove it, the components already handle their own cleanup so this should be removed entirely. Fixed in [15be169](https://github.com/dizzydroid/osv.dev/commit/15be169bc344d12bd413b9700d20d3438c4621e1)

gcp/website/frontend3/src/search.js
ashmod: Done in [15be169](https://github.com/dizzydroid/osv.dev/commit/15be169bc344d12bd413b9700d20d3438c4621e1)

gcp/website/frontend3/src/search.js
ashmod: Fixed in [15be169](https://github.com/dizzydroid/osv.dev/commit/15be169bc344d12bd413b9700d20d3438c4621e1)

gcp/website/frontend3/src/search.js
ashmod: Yes, sure thing! Added in [15be169](https://github.com/dizzydroid/osv.dev/commit/15be169bc344d12bd413b9700d20d3438c4621e1)

gcp/website/frontend3/src/search.js
ashmod: You're right this seems to be overengineering 😅 I've refactored the logic since we won't be needing multiple suggestionManagers, only the one in the ExpandableSearch. Fixed in [15be169](https://github.com/dizzydroid/osv.dev/commit/15be169bc344d12bd413b9700d20d3438c4621e1)

gcp/website/frontend3/src/styles.scss
ashmod: You mean the main searchbox, the navbar's or both? Also while we're at styling, is the current style/color accent okay or should I change it?

gcp/website/frontend3/src/search.js
another-rex: Ah I see! Can you add a comment here basically describing that? Thanks!

gcp/website/frontend3/src/search.js
another-rex: Sorry, I meant making it a method rather than a closure, then I think you can just call this.checkBody() (well name it something like appendSuggestionsElem), then we can remove the entire outer if statement and just call that.

gcp/website/frontend3/src/styles.scss
another-rex: Probably both, I originally meant the main searchbox, but I think it'll look pretty good for the navbar one as well.

I guess I do like the white border for the main one, and using the same background color as our main bg color, can we have it inherit the border from the component it's dropping down from?

gcp/website/frontend3/src/search.js
ashmod: Yes! It should be already added in [15be169](https://github.com/dizzydroid/osv.dev/commit/15be169bc344d12bd413b9700d20d3438c4621e1)

gcp/website/frontend3/src/search.js
ashmod: Fixed in [8a792e0](https://github.com/dizzydroid/osv.dev/commit/8a792e0cc4b1d53073379e9006f372d9a21e8b6c)

ashmod:
I've modified the styles, this is how the suggestion dropdowns should look like as of now,
image
image

edit: it may not be clearly visible here but the main searchbox (and the navbar's) has a scrollbar when focused on to scroll through suggestions (currently a max of 10 are displayed).

gcp/website/frontend3/src/styles.scss
ashmod: I have updated the styling in [8161371](https://github.com/dizzydroid/osv.dev/commit/8161371338b6414b395e4c91e51b53226bef1d73) , screenshots of their current state are shown in https://github.com/google/osv.dev/pull/3660#issuecomment-3085048975

ashmod:
Applied the required linter formatting. The CI check should now pass.

another-rex:
Looks amazing!

jess-lowe:
LGTM! Great work! We will aim to merge this on Tuesday (Sydney time) after this week's release so we have a week to observe it in the test instance, and make sure the database changes happen smoothly :)

jess-lowe:
Ah it looks like the worker tests are failing because of the added `search_tags` field. Are you able to run `make worker-tests` locally and see the missing tags? If not I can push a commit with the updated testcases.

ashmod:
Thanks for the heads up, I'll run the tests locally and update you shortly!

ashmod:
@jess-lowe I've updated the test data and the tests are passing for me now. Let me know if this works!

another-rex:
It's now live on test.osv.dev! Only some records are showing up as we populate the search_tags.

I did find a bug though, if you scroll down a little bit and try to search, the search dropdown will be offset in the wrong spot. (I think one of the offset calculations in javascript has a sign flipped?)

ashmod:
Awesome, Thanks for the update! Yes it seems like this is happening on both dropdowns, I will look into this and patch it shortly.

ashmod:
Checked the code, calculations look correct, it just didnt account for the relative position of the document.
Should be fixed in #3712

PR #3663, opened 2025-07-14
fix(actions): Update osv-scanner-unified.yml with read permissions

jess-lowe:
OSV-Scanner hasn't run on OSV.dev successfully in ~ 4 months. It seems to be missing read permissions. Hopefully this fixes it?


G-Rath:
No I think this is my fault, and the issue is in `osv-scanner` - https://github.com/google/osv-scanner/pull/2063 should fix this

hogo6002 (approved):

PR #3669, opened 2025-07-14
chore: update go version

hogo6002:


cuixq (approved):

cuixq:
I think https://github.com/google/osv.dev/blob/master/gcp/indexer/go.mod needs update as well.

Issue #1796, opened 2023-11-07
Add "CVE" ecosystem for osv.dev filtering?


oliverchang:
Currently, it's difficult to search for converted CVE records (per https://osv.dev/blog/posts/introducing-broad-c-c++-support/) via the osv.dev UI. 

Some solutions:
- We add a special "CVE" ecosystem to enable filtering by these. 
- We add a "Git" ecosystem to enable filtering by all entries that include Git ranges. This will include more than just CVEs though (e.g. OSS-Fuzz ones). 
- We detect and tag C/C++ relevant advisories and create a "C/C++" ecosystem. 

Adding an ecosystem has the additional benefit of surfacing an additional ecosystem bubble and filter on the home osv.dev UI.

Alternatively:
- We fix search results to show advisories with IDs matching the given search prefix first.

another-rex:
We currently have all the ecosystems roughly mutually exclusive, now the GIT count overlaps with everything else which sort of doubles the advisories in the count on the home page.

another-rex:
An alternative could be to only apply the GIT filter for ecosystems that don't have other ecosystems already present. It is a bit weird to filter by GIT, then see PYSEC advisories.

andrewpollock:
It also makes the `all.zip` in https://storage.googleapis.com/osv-vulnerabilities/index.html?prefix=GIT/ less than ideal.

another-rex:
For future reference, the git meta-ecosystem has now been updated to only be attached to entries without any other ecosystem

oliverchang:
Let's call this done!

PR #3700, opened 2025-07-21
fix: add git ecosystem whether or not another ecosystem already exists

### LINKED ISSUE #3698: Git vulnerabilities are not classified correctly on osv.dev
**Describe the bug**
CVE-2025-31498 on osv.dev has Alpine, Debian and Git vulnerabilities, it correctly shows up for Alpine and Debian for search filters, but not Git. There are a few others with similar issues, I am not sure about the total count without analyzing the entire dataset on osv.dev. 

This is affecting the data dump for GIT https://osv-vulnerabilities.storage.googleapis.com/GIT/all.zip and we are not able to create alerts for these CVEs since they are missing from the zip file.

Image

Image

jess-lowe:
Closes #3698

While this was WAI in original implementation: https://github.com/google/osv.dev/issues/1796, we revised our stance on this.

### DIFF
--- a/osv/models.py
+++ b/osv/models.py
@@ -411,16 +411,15 @@ class Bug(ndb.Model):
         if pkg.package.ecosystem
     }
 
-    # Only attempt to add the Git ecosystem if
-    # there are no existing ecosystems present
-    if not ecosystems_set:
-      for pkg in self.affected_packages:
-        for r in pkg.ranges:
-          if r.type == 'GIT':
-            ecosystems_set.add('GIT')
-            break
-        if 'GIT' in ecosystems_set:
+    # Add the Git ecosystem if it has git ranges
+    
+    for pkg in self.affected_packages:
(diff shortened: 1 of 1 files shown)


osv/models.py
```diff
@@ -411,16 +411,15 @@ def _pre_put_hook(self):  # pylint: disable=arguments-differ
...
-            break
-        if 'GIT' in ecosystems_set:
+    # Add the Git ecosystem if it has git ranges
+    
+    for pkg in self.affected_packages:
+      for r in pkg.ranges:
+        if r.type == 'GIT':
+          ecosystems_set.add('GIT')
           break
+      if 'GIT' in ecosystems_set:
```
another-rex: I just realized this should be moved above the for loop

another-rex (approved):

another-rex (approved):

PR #3701, opened 2025-07-21
feat(vulnfeeds): Add helper to parse GitHub CNA version ranges

### LINKED ISSUE #3699: Create a helper function to extract versions from the Github_M CNA version description
With the [CVE5 to OSV conversion project](#3633), we are currently not able to ingest the CVEs from the Github_M CNA  despite having most/all of the necessary information, as the version is displayed incorrectly:

` "version": ">= 1.32.3, < 1.34.5",`

A helper function that could extract the versions into the[ AffectedVersion](https://github.com/google/osv.dev/blob/f07566a5aea7147155fe5fcb78ee5ab2fda1e04d/vulnfeeds/models/types.go#L71C1-L75C2) format would be immensely helpful.

The above example would give:
`Introduced: 1.32.3`
`Fixed: 1.34.5`

if the second part was "<=", it'd be `LastAffected: 1.34.5` instead

It can probably go here: https://github.com/google/osv.dev/blob/master/vulnfeeds/git/versions.go

From my notes, this regex might be of use:

`^(>=|<=|~|\^|>|<|=)\s*(\d+\.\d+\.\d+)(?:,\s*(>=|<=|~|\^|>|<|=)\s*(\d+\.\d+\.\d+))?$`

dizzydroid:
This PR introduces a helper function, `ParseVersionRange`, and its corresponding unit tests. 
This function is designed to parse the non-standard version range strings found in CVEs from the GitHub CNA.
Resolves #3699


vulnfeeds/git/versions.go
```diff
@@ -137,3 +139,53 @@ func NormalizeVersion(version string) (normalizedVersion string, e error) {
 	normalizedVersion = strings.Join(components, "-")
 	return normalizedVersion, e
 }
+
+// Parse a version range string into an AffectedVersion struct.
```
jess-lowe: The only thing I can see missing is possibly just a comment that this is specifically targeting Github's structure otherwise LGTM.

ashmod: [f9f04b4](https://github.com/dizzydroid/osv.dev/commit/f9f04b471d186f3269a369eca5ffdab734712eb8) 
something like this?

jess-lowe (approved):
LGTM

vulnfeeds/git/versions.go
jess-lowe: Perfect :)

PR #3702, opened 2025-07-22
chore: update schema version in linter

hogo6002:


gcp/workers/linter/Dockerfile
```diff
@@ -20,12 +20,14 @@ RUN apk add --no-cache git
 WORKDIR /src
 RUN git clone https://github.com/ossf/osv-schema.git osv-schema
 
-RUN cd osv-schema && git checkout 21268ba778975b0b6dfd77f4b3e71c1f6449fcdb
+RUN cd osv-schema && git checkout c66187a8068efe94a4a89953699335cbdb9acb9a
```
michaelkedar: To make the docker layer situation a bit better, can you combine these into one RUN command
i.e. `RUN git clone ... && cd osv-schema && git checkout ...`

michaelkedar: Also, could you add a comment to the commit hash to note what version of the schema it's meant to be

hogo6002: updated

michaelkedar (approved):

PR #3704, opened 2025-07-22
fix(source-sync): Update missing source defaults

jess-lowe:
Missing defaults are causing some fields not to be propagated to the SourceRepository Datastore entity.


another-rex (approved):
Is there something we can add to automatically check whether this is missing fields, especially when new fields gets added to  SourceRepository?

jess-lowe:
> Is there something we can add to automatically check whether this is missing fields, especially when new fields gets added to SourceRepository?

We can probably dramatically error the cloud build if it senses a length difference between them, especially if the local repo vers > datastore repo vers.

PR #3708, opened 2025-07-22
feat: Begin populating new database fields

michaelkedar:
This PR begins our database table migrations!
This adds a few new entities into Datastore:
- `Vulnerability`: containing a small amount of information of a vulnerability, with the (overall) modified date, as well as some raw fields that are overwritten by our enrichment.
- `ListedVulnerability`: Minimal vulnerability information that is needed for our website's `/list` page
- `AffectedVersions`: An index for API matching, which should help simplify our API logic and improve performance.
Additionally, full vulnerabilities are stored in the GCS bucket, which will eventually become the new source of truth for the vulnerability records (deprecating the Bug entity)

To populate these I've added a new `_post_put_hook` on the Bug entity, and added some logic to the alias & upstream cron jobs to update the entities/GCS also.

I've only set this to run on the test instance (and when running tests with the emulator) because I don't want this to block releases on prod if it turns out this has significant performance impacts (especially with the blocking writing to the bucket in the alias & upstream crons).

Some other misc things to note:
- I'm writing both `.pb` and `.json` files to to bucket, because the proto files are a bit easier to work with in code, but the JSON is more usable externally. It's probably better to just write the `.pb` and have the exporter create the `.json` files
- I had to remove the transactions from the reput tools, because ndb doesn't like nested transactions
- I have written my own (minimal) mocker for the `google.cloud.storage` operations, and added that to the datastore emulator starter code
- I've made a few changes to some tests that were creating incomplete Bugs (which caused issues when trying to convert them for writing)
- I have a TODO to create another worker/cron job that retries failed GCS writes - that's probably the next step.
- Before we merge this, we should pause the alias/upstream cron jobs, then re-put every Bug (which will take some time) to avoid noisy error logs.


another-rex:
> Before we merge this, we should pause the alias/upstream cron jobs, then re-put every Bug (which will take some time) to avoid noisy error logs.

The reputting is after merging right?

gcp/workers/alias/alias_computation.py
```diff
@@ -15,24 +15,31 @@
...
 import osv
+from osv import gcs
 import osv.logs
 
 ALIAS_GROUP_VULN_LIMIT = 32
 VULN_ALIASES_LIMIT = 5
 
 
-def _update_group(bug_ids, alias_group):
+def _update_group(bug_ids, alias_group, changed_vulns):
```
another-rex: nit: might as well add some types to these

gcp/workers/alias/alias_computation.py
```diff
@@ -42,9 +49,11 @@ def _update_group(bug_ids, alias_group):
   alias_group.bug_ids = bug_ids
   alias_group.last_modified = datetime.datetime.now(datetime.UTC)
   alias_group.put()
+  for vuln_id in bug_ids:
+    changed_vulns[vuln_id] = alias_group
 
 
-def _create_alias_group(bug_ids):
+def _create_alias_group(bug_ids, changed_vulns):
```
another-rex: Same here

gcp/workers/alias/alias_computation.py
```diff
@@ -82,6 +93,73 @@ def _compute_aliases(bug_id, visited, bug_aliases):
...
+  if pb_blob is None:
+    if osv.Vulnerability.get_by_id(vuln_id) is not None:
+      logging.error('vulnerability not in GCS - %s', vuln_id)
+      # TODO(michaelkedar): send pub/sub message to reimport
+    return
+  try:
+    vuln_proto = osv.vulnerability_pb2.Vulnerability.FromString(
+        pb_blob.download_as_bytes())
+  except Exception:
+    logging.exception('failed to download %s protobuf from GCS', vuln_id)
```
another-rex: Can all of this be abstracted to a single get by ID function? I feel like this would be used in quite a few different places, and if we ever decide to put the .pb directly in datastore or another db we can just change the implementation.

gcp/workers/alias/alias_computation.py
```diff
@@ -82,6 +93,73 @@ def _compute_aliases(bug_id, visited, bug_aliases):
...
+
+  def transaction():
+    vuln: osv.Vulnerability = osv.Vulnerability.get_by_id(vuln_id)
+    if vuln is None:
+      logging.error('vulnerability not in Datastore - %s', vuln_id)
+      # TODO: Raise exception
+      return
+    if alias_group is None:
+      modified = datetime.datetime.now(datetime.UTC)
```
another-rex, line 124: Should modified be unconditionally now? If aliases doesn't change we probably don't want to update modified date right?

another-rex: I guess the function doc does have that assumption that something was just deleted, but if  it's easy / there's no additional side effects, we should just only update when there was an alias group in the first place.

gcp/workers/alias/alias_computation.py
```diff
@@ -82,6 +93,73 @@ def _compute_aliases(bug_id, visited, bug_aliases):
...
+
+  try:
+    json_blob = bucket.blob(os.path.join(gcs.VULN_JSON_PATH, vuln_id + '.json'))
+    json_blob.custom_time = modified
+    json_data = json_format.MessageToJson(
+        vuln_proto, preserving_proto_field_name=True, indent=None)
+    json_blob.upload_from_string(json_data, content_type='application/json')
+  except Exception:
+    logging.exception('failed to upload %s json to GCS', vuln_id)
+    # TODO(michaelkedar): send pub/sub message to retry
```
another-rex: Same as above, this is probably a good candidate to tuck into a function to keep things clearer

gcp/workers/alias/alias_computation.py
another-rex: Maybe in the new gcs module

osv/gcs_mock.py
```diff
@@ -0,0 +1,96 @@
...
+  
+  If `directory` is set, blobs will be read from/written to files in the
+  directory, which will remain after the context exits.
+  Otherwise, blobs will be written to a temporary directory, which is deleted
+  when the context exits.
+  """
+  with (tempfile.TemporaryDirectory()
+        if directory is None else contextlib.nullcontext(directory)) as db_dir:
+    pathlib.Path(db_dir, gcs.VULN_JSON_PATH).mkdir(parents=True, exist_ok=True)
+    pathlib.Path(db_dir, gcs.VULN_PB_PATH).mkdir(parents=True, exist_ok=True)
```
another-rex, line 40: Should exist_ok be False here? I'm not sure what situation the dir's would already exist

osv/models.py
```diff
@@ -892,6 +866,407 @@ def to_vulnerability_async(self,
...
+      combo = '-'.join(sub_parts)
+      tokens.add(combo)
+  return tokens
+
+
+# --- Vulnerability Entity ---
+
+
+class Vulnerability(ndb.Model):
+  """Vulnerability entry."""
```
another-rex: Can you describe a bit more what this is for? (e.g. you can just use the PR description)

osv/models.py
```diff
@@ -892,6 +866,407 @@ def to_vulnerability_async(self,
...
+  # The commit hash.
+  commits: list[bytes] = ndb.BlobProperty(repeated=True, indexed=True)
+  # Whether or not the bug is public.
+  public: bool = ndb.BooleanProperty()
+  # The page for this batch of commits.
+  page: int = ndb.IntegerProperty(indexed=False)
+
+
+class AffectedVersions(ndb.Model):
+  """AffectedVersions entry."""
```
another-rex: Same here, please expand this a bit more, like the ListedVulnerability model

osv/models.py
```diff
@@ -892,6 +866,407 @@ def to_vulnerability_async(self,
...
+      if affected.package.ecosystem:
+        all_ecosystems.add(affected.package.ecosystem)
+      for sev in affected.severity:
+        severities.add(
+            (vulnerability_pb2.Severity.Type.Name(sev.type), sev.score))
+      for r in affected.ranges:
+        if r.type == vulnerability_pb2.Range.Type.GIT:
+          all_ecosystems.add('GIT')
+          search_indices.add(r.repo)
+          autocomplete_tags.add(r.repo)
```
another-rex: `.lower()` for this as well

osv/models.py
```diff
@@ -892,6 +866,407 @@ def to_vulnerability_async(self,
...
+    all_packages = set()
+
+    is_fixed = False
+    severities = set()
+    for sev in vulnerability.severity:
+      severities.add(
+          (vulnerability_pb2.Severity.Type.Name(sev.type), sev.score))
+
+    search_indices = set()
+    search_indices.update(_tokenize(vulnerability.id))
```
another-rex, line 1042: Hmm... Since we are doing this big reput anyway, we could just .lower() all of this and make our search case insensitive which we should have done.

But probably a good idea to not change behavior too much for now.

osv/models.py
```diff
@@ -892,6 +866,407 @@ def to_vulnerability_async(self,
...
+      to_delete.extend(r.key for r in removed)
+
+    ndb.put_multi(to_put)
+    ndb.delete_multi(to_delete)
+
+  ndb.transaction(transaction)
+  _write_to_bucket(vuln_pb)
+
+
+def _write_to_bucket(vulnerability: vulnerability_pb2.Vulnerability):
```
another-rex: Could be a good idea to make this public and use it in alias-computation

osv/models.py
```diff
@@ -892,6 +866,407 @@ def to_vulnerability_async(self,
...
+    json_blob.custom_time = modified
+    json_data = json_format.MessageToJson(
+        vulnerability, preserving_proto_field_name=True, indent=None)
+    json_blob.upload_from_string(json_data, content_type='application/json')
+  except Exception:
+    logging.exception('failed to upload %s json to GCS', vuln_id)
+    # TODO(michaelkedar): send pub/sub message to retry
+
+
+_EVENT_ORDER = {
```
another-rex: Let's just put it at the top of the file, since it's already pretty far from where it's used.

another-rex (commented):
Looking great!

I'm not sure how to review models_test.py though...

gcp/workers/alias/alias_computation.py
michaelkedar: I'm pretty sure the only way `_update_vuln_with_group` will be called in main with `alias_group = None` is if the vuln was in an alias group already.

osv/gcs_mock.py
michaelkedar: Though I don't actually use it, the `gcs_mock()` context does allow the directory to be manually specified, which could already have the folders set up from a previous run

gcp/workers/alias/alias_computation.py
michaelkedar: created functions in the gcs module for reading and writing

osv/models.py
michaelkedar: added some description

osv/models.py
michaelkedar: added what it is to be used for to the docstring

osv/models.py
michaelkedar: It is mostly `lower()`ed - `_tokenize()` does lower-case the tokens. The only thing not made lowercase is the repo URLs, but they probably are already lowercase anyway.

PR #3709, opened 2025-07-22
feat(css): Update blogpost styles

another-rex:
Update the blog so that it looks nicer and is easier to read.

New:
image

Existing:

image


jess-lowe (approved):
Much nicer :) LGTM

PR #3712, opened 2025-07-22
fix(website): Correct search dropdown position on scroll

dizzydroid:
Fixes a bug where the search suggestions dropdown was misaligned when the page was scrolled.

The position calculation in `updatePosition()` now includes `window.scrollY` to ensure the dropdown's `top` position is always correct relative to the document, not the viewport.

### DIFF
--- a/gcp/website/frontend3/src/search.js
+++ b/gcp/website/frontend3/src/search.js
@@ -296,13 +296,13 @@ export class SearchSuggestionsManager {
       // Use the designated suggestions container
       const containerRect = suggestionContainer.getBoundingClientRect();
       this.suggestionsElement.style.left = `${containerRect.left}px`;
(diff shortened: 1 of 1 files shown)


gcp/website/frontend3/src/search.js
```diff
@@ -296,13 +296,13 @@ export class SearchSuggestionsManager {
       // Use the designated suggestions container
       const containerRect = suggestionContainer.getBoundingClientRect();
       this.suggestionsElement.style.left = `${containerRect.left}px`;
```
another-rex: Unlikely to happen unless quite zoomed in, but might as well add `window.scrollX` here

ashmod: Yes! Added in [42a624e](https://github.com/google/osv.dev/pull/3712/commits/42a624e0a7acd08547e6ff56f4f7107813463f1b)

PR #3714, opened 2025-07-23
refactor(vulnfeeds): Use OSVSchema Vulnerability structs instead of maintaining a duplicate set

jess-lowe:
Noticed that we were using an outdated Vulnerability struct in vulns.go. This will update vulnfeeds to use the osvschema go bindings instead, which will keep the fields better up to date. 

Also refactored FromCVE to take the specific fields it needs instead of the whole CVE for better modularity.

Sorry for the long PR.


vulnfeeds/vulns/vulns.go
```diff
@@ -168,57 +150,24 @@ type PackageInfo struct {
 	Ecosystem         string             `json:"ecosystem,omitempty" yaml:"ecosystem,omitempty"`
 	PURL              string             `json:"purl,omitempty" yaml:"purl,omitempty"`
 	VersionInfo       models.VersionInfo `json:"fixed_version,omitempty" yaml:"fixed_version,omitempty"`
-	EcosystemSpecific map[string]string  `json:"ecosystem_specific,omitempty" yaml:"ecosystem_specific,omitempty"`
+	EcosystemSpecific map[string]any     `json:"ecosystem_specific,omitempty" yaml:"ecosystem_specific,omitempty"`
```
hogo6002: why change to any?

vulnfeeds/vulns/vulns.go
```diff
@@ -321,37 +270,70 @@ func (v *Vulnerability) AddPkgInfo(pkgInfo PackageInfo) {
...
+func (v *Vulnerability) AddSeverity(metricsData any) {
 	var bestVectorString string
+	var severityType osvschema.SeverityType
 
-	for _, metric := range CVEImpact.CVSSMetricV31 {
-		if bestVectorString != "" {
+	switch md := metricsData.(type) {
+	case *cves.CVEItemMetrics:
+		if md == nil {
 			break
```
hogo6002, line 283: do we still need the break statement here?

vulnfeeds/vulns/vulns.go
```diff
@@ -588,44 +565,47 @@ func unique[T comparable](s []T) []T {
...
-	}
-	var notes []string
-	v.Published = cve.Published.Format(time.RFC3339)
-	v.Modified = cve.LastModified.Format(time.RFC3339)
-	v.References = ClassifyReferences(cve.References)
-	v.AddSeverity(cve.Metrics)
-	return &v, notes
+// There are two id fields passed in as one of the users of this field (PyPi) sometimes has a different id than the CVEID
+// and the ExtractReferencedVulns function uses these in a check to add the other ID as an alias.
+func FromCVE(id cves.CVEID, cveID cves.CVEID, references []cves.Reference, descriptions []cves.LangString, publishedDate time.Time, modifiedDate time.Time, metrics any) *Vulnerability {
```
hogo6002, line 573: can we just pass cve.CVE in?

vulnfeeds/vulns/vulns.go
```diff
@@ -321,37 +270,70 @@ func (v *Vulnerability) AddPkgInfo(pkgInfo PackageInfo) {
...
 			if metric.Type != "Primary" {
 				continue
 			}
 			bestVectorString = metric.CVSSData.VectorString
+			severityType = osvschema.SeverityCVSSV3
+		}
+
+		// No CVSS 3.1, try falling back to CVSS 3.0 if available.
+		if bestVectorString == "" {
+			for _, metric := range md.CVSSMetricV30 {
```
hogo6002: The code here feels quite repeated for V3.0 and V3.1. I guess we can't really extract it into a separate function without using generics, can we?

vulnfeeds/vulns/vulns.go
jess-lowe: no, because we are using both the NVD CVE struct and CVE5 in the future so it's easier to pass in exactly what is needed rather than the whole thing.

vulnfeeds/vulns/vulns.go
jess-lowe: Updated to interface{} to align with go bindings

vulnfeeds/vulns/vulns.go
jess-lowe: Refactored a bit regardless.

hogo6002 (approved):
Thanks!

PR #3723, opened 2025-07-29
feat: add support for the Drupal ecosystem

G-Rath:


G-Rath:
This is no longer needed as we're not going to use a dedicated ecosystem for Drupal

valentijnscholten:
@G-Rath Could you let us know what happened here? I think I am seeing vulnerabilities from the drupal database, i.e. https://osv.dev/vulnerability/DRUPAL-CONTRIB-2023-052. However the Drupal database (or ecosystem) is not mentioned on:
- https://google.github.io/osv.dev/data/#current-data-sources
- https://google.github.io/osv.dev/data/#converted-data
- https://google.github.io/osv.dev/data/#covered-ecosystems

My _feeling_ is that the Drupal database is mirrored/ingested, but it's been made part of the Packagist ecosystem?

I looked for other PRs but I am struggling to find the one that actually adds this.

G-Rath:
@valentijnscholten yes that's correct, we're using the Packagist ecosystem - if you look at the advisory the source is the Drupal advisory database.

We added the db to prod in #4394 and you can read more about it in [our blog post](https://www.ackama.com/articles/drupal-advisories-now-available-in-osv-format/)

Happy to answer any questions either on GH or in the Drupal Slack 🙂

valentijnscholten:
Thanks. The part that I missed is that the conversion to OSV format is actually done in https://github.com/DrupalSecurityTeam/drupal-advisory-database.

PR #3724, opened 2025-07-29
docs(api): Warn users about API response size limit

### LINKED ISSUE #3119: Update documentation for Response Size error
Our API when using HTTP/1.1 has a limit response size of 32MiB, but has no such limit for HTTP2. This limit could be hit in big OS linux queries.

We need to update the documentation to warn of this limit.

dizzydroid:
Updated the API documentation to warn users of API response limit on HTTP/1.1 and recommend using HTTP/2 for large queries.

Fixes #3119


another-rex (approved):
Thanks!

PR #3743, opened 2025-07-29
feat: add Drupal advisory database

G-Rath:
~Pending https://github.com/ossf/osv-schema/pull/372 and friends~


source_test.yaml
```diff
@@ -176,6 +176,20 @@
   editable: False
   strict_validation: True
 
+- name: 'drupal'
+  versions_from_repo: False
+  type: 0
+  ignore_patterns: ['^(?!DRUPAL-).*$']
+  directory_path: 'advisories'
+  repo_url: 'https://github.com/ackama/drupal-advisory-database.git'
```
greggles: Before this gets merged, it would be good to discuss the location of the repo_url, whether it should be on drupal.org or github, if on github where. Thanks!

I'll reach out to discuss.

G-Rath: database repo now lives with the Drupal team

another-rex:
This also needs to update the osv-schema submodule

G-Rath:
~Does it?~ oh right yea cause I changed the linter, even though we didn't end up adding a new ecosystem

PR #3747, opened 2025-07-31
feat(vulnfeeds): CVE5 to OSV conversion w/o version extraction

jess-lowe:
I am splitting out my PRs for the CVE5 to OSV conversion for easier reviewing. 

In this PR, main focus are in the 
cvelist2osv/main.go file
vulns.go file

Version extraction will be in a separate PR. 

This PR also includes some heuristics and metrics stuff, but it's not fully useful or functional yet so don't expect anything there. 

Since #3635 we have adjusted strategies to have an intermediary OSV output that will include the naive versions, to be enriched later by a dedicated enricher:
image


vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -0,0 +1,290 @@
...
+		Metrics.Notes = append(Metrics.Notes, fmt.Sprintf("[%s]: Reference Type %s: %d", cve.Metadata.CVEID, refType, count))
+	}
+}
+
+// FromCVE5 creates a Vulnerability from a CVE5 object.
+func FromCVE5(cve cves.CVE5, refs []cves.Reference) (*vulns.Vulnerability, []string) {
+	aliases, related := vulns.ExtractReferencedVulns(cve.Metadata.CVEID, cve.Metadata.CVEID, refs)
+	var err error
+	var notes []string
+	v := vulns.Vulnerability{}
```
hogo6002: nit: initialize all fields at once? e.g. 
```
v := vulns.Vulnerability{
    SchemaVersion: osvschema.SchemaVersion,
    ID:            string(cve.Metadata.CVEID),
    Summary:       string(cve.Containers.CNA.Title),
}
```

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -0,0 +1,290 @@
...
+	v.SchemaVersion = osvschema.SchemaVersion
+	v.ID = string(cve.Metadata.CVEID)
+	v.Summary = string(cve.Containers.CNA.Title)
+	v.Details = cves.EnglishDescription(cve.Containers.CNA.Descriptions)
+	v.Aliases = aliases
+	v.Related = related
+	v.Published, err = cves.ParseCVE5Timestamp(cve.Metadata.DatePublished)
+	if err != nil {
```
hogo6002, line 83: should we check error first before assigning the value to v.Published/v.Modified? what would the value be when parse failed?

hogo6002 (commented):
Very nice and clean!

vulnfeeds/cmd/cvelist2osv/testdata/TestExtractVersionInfo_CVE_with_lessThan.yaml
```diff
@@ -0,0 +1,3 @@
+---
+version: 2
```
another-rex: Is this suppose to be empty? If so, can you please update the file name

vulnfeeds/cmd/cvelist2osv/main_test.go
```diff
@@ -0,0 +1,74 @@
...
+		log.Fatalf("Failed to decode %q: %+v", fileName, err)
+	}
+	return cve
+}
+
+func TestFromCVE(t *testing.T) {
+	cveData := loadTestData("CVE-2021-44228")
+	refs := identifyPossibleURLs(cveData)
+
+	v, notes := FromCVE5(cveData, refs)
```
another-rex: Let's turn this into a tabular test to make it easier to extend with additional test cases.

vulnfeeds/cves/testdata/Test_MaybeUpdateVPRepoCache_Test_with_no_cache.yaml
```diff
@@ -0,0 +1,3 @@
+---
+version: 2
```
another-rex, line 2: hmm... why are all of these empty?

vulnfeeds/vulns/vulns.go
```diff
@@ -584,6 +617,23 @@ func FromNVDCVE(id cves.CVEID, cve cves.CVE) *Vulnerability {
 	return &v
 }
 
+// GetCPEs extracts CPE strings from a slice of cves.CPE.
+func GetCPEs(cpeApplicability []cves.CPE) []string {
+	var CPEs []string
+	for _, c := range cpeApplicability {
+		for _, node := range c.Nodes {
+			if node.Operator != "OR" {
```
another-rex, line 604: Would this be something unexpected we should log?

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -0,0 +1,290 @@
...
+	"golang.org/x/exp/slices"
+
+	"github.com/google/osv/vulnfeeds/cves"
+	"github.com/google/osv/vulnfeeds/git"
+	"github.com/google/osv/vulnfeeds/utility"
+	"github.com/google/osv/vulnfeeds/vulns"
+	"github.com/ossf/osv-schema/bindings/go/osvschema"
+)
+
+type ConversionOutcome int
```
another-rex: Can you group the ConversionOutcome enum stuff in one place?

vulnfeeds/cmd/cvelist2osv/main.go
jess-lowe: Because we're using the go bindings from `osv-schema` now, so it complains that it doesn't know the struct fields. This was the easier solution than figuring out how to expose the embedded fields.

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -0,0 +1,293 @@
...
+	FixUnresolvable                            // Partial resolution of versions, resulting in a false positive.
+)
+
+var (
+	jsonPath  = flag.String("cve_json", "", "Path to CVEList JSON to examine.")
+	outDir    = flag.String("out_dir", "", "Path to output results.")
+	outFormat = flag.String("out_format", "OSV", "Format to output {OSV}")
+)
+var Logger utility.LoggerWrapper
+var RepoTagsCache git.RepoTagsCache
```
another-rex: Avoid having a global mutable struct, as that'll make parallising it later much harder. Let's just pass this in as a pointer to the functions that need to modify this.

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -0,0 +1,293 @@
...
+	}
+
+	Metrics.Notes = append(Metrics.Notes, notes...)
+	metricsFile := filepath.Join(vulnDir, v.ID+".metrics.json")
+	if marshalledMetrics, err := json.MarshalIndent(Metrics, "", "  "); err != nil {
+		Logger.Warnf("[%s]: Failed to marshal metrics: %v", CVE.Metadata.CVEID, err)
+	} else {
+		if err = os.WriteFile(metricsFile, marshalledMetrics, 0660); err != nil {
+			Logger.Warnf("[%s]: Failed to write %s: %v", CVE.Metadata.CVEID, metricsFile, err)
+		}
```
another-rex, line 154: nit: You can just return fileWriteErr

another-rex: Though what you currently have might be clearer, so up to you.

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -0,0 +1,293 @@
...
+	v.References = vulns.ClassifyReferences(refs)
+	// Add affected version
+	// VERSIONS WILL BE ADDED IN ANOTHER PR
+	v.DatabaseSpecific = make(map[string]interface{})
+	CPEs := vulns.GetCPEs(cve.Containers.CNA.CPEApplicability)
+	if len(CPEs) != 0 {
+		v.DatabaseSpecific["CPE"] = vulns.Unique(CPEs)
+	}
+
+	// TODO: add CWEs
```
another-rex, line 104: Probably need a better name than metrics, since you also use "metrics" for ConversionNotes, where in this case it looks like it's for severity?

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -0,0 +1,293 @@
...
+// addIDToNotes prepends a CVE ID to a slice of notes.
+func addIDToNotes(id cves.CVEID, notes []string) []string {
+	var updatedNotes []string
+	for _, note := range notes {
+		updatedNotes = append(updatedNotes, fmt.Sprintf("[%s] %s", id, note))
+	}
+	return updatedNotes
+}
+
+// determineHeuristics examines a CVE and extracts metrics and heuristics about it.
```
another-rex: Rename determineHeuristics to something more specifically about extracting metrics, and clarify that this does not modify the vulnerability. (I would even avoid taking the vulnerability by pointer, but that might have performance implications)

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -0,0 +1,293 @@
...
+	for _, m := range metrics {
+		v.AddSeverity(m)
+	}
+	return &v, notes
+}
+
+// CVEToOSV converts a CVE into an OSV finding and writes it to a file.
+func CVEToOSV(CVE cves.CVE5, references []cves.Reference, repos []string, directory string) error {
+	// Create a base OSV record
+	v, notes := FromCVE5(CVE, references)
```
another-rex: I think this is just for metrics, so let's group this with metrics writing at the bottom.

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -0,0 +1,293 @@
...
+	if err != nil {
+		Logger.Warnf("Failed to create dir: %v", err)
+		return fmt.Errorf("failed to create dir: %w", err)
+	}
+
+	var fileWriteErr error
+
+	// Only write out the OSV file if we have ranges.
+	outputFile := filepath.Join(vulnDir, v.ID+extension)
+
```
another-rex: I don't think we should be noting write failures, since that's unrelated to the actual record itself, and will be caused by our own infrastructure. Just log these.

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -0,0 +1,293 @@
...
+	cnaAssigner := CVE.Metadata.AssignerShortName
+
+	// Save OSV record to a directory
+	vulnDir := filepath.Join(directory, cnaAssigner)
+	err := os.MkdirAll(vulnDir, 0755)
+	if err != nil {
+		Logger.Warnf("Failed to create dir: %v", err)
+		return fmt.Errorf("failed to create dir: %w", err)
+	}
+
```
another-rex: Hmm... I'm missing where the check for whether we have ranges is?

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -0,0 +1,293 @@
...
+	for _, adp := range cve.Containers.ADP {
+		if len(adp.Metrics) != 0 {
+			metrics = append(metrics, adp.Metrics...)
+		}
+	}
+	for _, m := range metrics {
+		v.AddSeverity(m)
+	}
+	return &v, notes
+}
```
another-rex, line 120: The comments in the functions is really nice! 

I would further refactor it along those commented sections.

Currently this function does quite a few different things as identified by the comments for each section.

You can group the "save OSV record to disk" section into it's own  function, and same for "save metrics to disk"

Then you will essentially have 4 function calls that are really clear:

- Make CVE record
- Save CVE record
- Make Metrics
- Save Metrics

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -0,0 +1,293 @@
...
+	err = json.Unmarshal(data, &cve)
+	if err != nil {
+		Logger.Fatalf("Failed to parse CVEList CVE JSON: %v", err)
+	}
+
+	ReposForCVE := make(map[cves.CVEID][]string)
+	refs := identifyPossibleURLs(cve)
+
+	CVEID := cve.Metadata.CVEID
+
```
another-rex: Hmm... doesn't the notes already have the CVE ID prefixed? (At least looking at the changes you made below it looks like it would already have the CVE ID.

Either way, I think this should be done in ReposFromReferencesCVEList, not something the caller should do.

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -0,0 +1,293 @@
...
+	if len(refs) > 0 {
+		repos, notes := cves.ReposFromReferencesCVEList(string(CVEID), nil, nil, refs, RefTagDenyList, Logger)
```
another-rex: Can you add a comment here to describe what scenerio this would happen in?

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -0,0 +1,293 @@
...
+		addIDToNotes(CVEID, notes)
+		if len(repos) == 0 {
+			Logger.Warnf("[%s]: Failed to derive any repos", CVEID)
+		}
+		Metrics.Notes = append(Metrics.Notes, fmt.Sprintf("[%s]: Derived %q for CVE", CVEID, repos))
+		ReposForCVE[CVEID] = repos
```
another-rex: Repeated note from 3 lines above?

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -0,0 +1,293 @@
...
+	defer logCleanup()
+	data, err := os.ReadFile(*jsonPath)
+	if err != nil {
+		Logger.Fatalf("Failed to open file: %v", err) // double check this is best practice output
+	}
+
+	var cve cves.CVE5
+	err = json.Unmarshal(data, &cve)
+	if err != nil {
+		Logger.Fatalf("Failed to parse CVEList CVE JSON: %v", err)
```
another-rex, line 238: This whole ReposForCVE section seems like it could be a separate function

another-rex (commented):
Nice! Added a bunch of comments :)

vulnfeeds/cmd/cvelist2osv/main.go
jess-lowe: Removed it entirely for now, as it isn't being used properly (will become more useful in the enricher/resolution stage)

vulnfeeds/cmd/cvelist2osv/main.go
jess-lowe: moved

vulnfeeds/cmd/cvelist2osv/main.go
jess-lowe: good catch!

vulnfeeds/cmd/cvelist2osv/main.go
jess-lowe: renamed, and removed the vulnerability reference to only take the references it needs

vulnfeeds/cmd/cvelist2osv/main.go
jess-lowe: refactored the logic out.

vulnfeeds/cmd/cvelist2osv/testdata/TestExtractVersionInfo_CVE_with_lessThan.yaml
jess-lowe: I think these just happen because there are no http calls, so it has nothing to save here - they're auto-generated.

vulnfeeds/vulns/vulns.go
jess-lowe: logged it but not sure it's crazy unexpected - or something we need to handle in it's current state.

vulnfeeds/cmd/cvelist2osv/testdata/TestExtractVersionInfo_CVE_with_lessThan.yaml
another-rex: Ohhh... that makes sense

vulnfeeds/cves/versions.go
```diff
@@ -1026,14 +1026,13 @@ func ReposFromReferences(CVE string, cache VendorProductToRepoMap, vp *VendorPro
 }
 
 // Examines the CVE references for a CVE and derives repos for it, optionally caching it.
-func ReposFromReferencesCVEList(CVE string, cache VendorProductToRepoMap, vp *VendorProduct, refs []Reference, tagDenyList []string, Logger utility.LoggerWrapper) (repos []string, notes []string) {
+func ReposFromReferencesCVEList(CVE string, refs []Reference, tagDenyList []string, Logger utility.LoggerWrapper) (repos []string, notes []string) {
 
 	for _, ref := range refs {
 		// If any of the denylist tags are in the ref's tag set, it's out of consideration.
 		if !RefAcceptable(ref, tagDenyList) {
 			// Also remove it if previously added under an acceptable tag.
```
another-rex: Comment outdated?

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -106,97 +66,161 @@ func FromCVE5(cve cves.CVE5, refs []cves.Reference) (*vulns.Vulnerability, []str
...
-	// Add affected version
-	// VERSIONS WILL BE ADDED IN ANOTHER PR
-	v.DatabaseSpecific = make(map[string]interface{})
-	CPEs := vulns.GetCPEs(cve.Containers.CNA.CPEApplicability)
+	v.Vulnerability.Modified = modified
+
+	// TODO(jesslowe): add logic to also extract cpes from affected field (CVE-2025-1110)
+	CPEs, notes := vulns.GetCPEs(cve.Containers.CNA.CPEApplicability)
 	if len(CPEs) != 0 {
 		v.DatabaseSpecific["CPE"] = vulns.Unique(CPEs)
```
another-rex, line 98: Nit, update vulns.Unique to return empty slice if input is nil / len (0), then  flatten this.

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -106,97 +66,161 @@ func FromCVE5(cve cves.CVE5, refs []cves.Reference) (*vulns.Vulnerability, []str
...
+
+	for _, check := range checks {
+		for _, m := range metricsData {
+			if vectorString := check.getVectorString(m); vectorString != "" {
+				return vectorString, check.severityType
+			}
+		}
+	}
+	return "", ""
+}
```
another-rex: Suggest moving to a separate file (and eventually probably a separate package) to shorten `main.go`

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -106,97 +66,161 @@ func FromCVE5(cve cves.CVE5, refs []cves.Reference) (*vulns.Vulnerability, []str
...
-			Logger.Warnf("[%s]: Failed to write %s: %v", CVE.Metadata.CVEID, metricsFile, err)
+			Logger.Warnf("[%s]: Failed to write %s: %v", id, metricsFile, err)
+			return err
 		}
 	}
+	return nil
+}
+
+// CVEToOSV converts a CVE into an OSV finding and writes it to a file.
+func CVEToOSV(CVE cves.CVE5, directory string) error {
```
another-rex: nit: Could have a clearer name here. Maybe ConvertAndExportCVEToOSV

vulnfeeds/vulns/vulns.go
```diff
@@ -618,19 +595,21 @@ func FromNVDCVE(id cves.CVEID, cve cves.CVE) *Vulnerability {
 }
 
 // GetCPEs extracts CPE strings from a slice of cves.CPE.
```
another-rex, line 597: Add return values to the docstring

vulnfeeds/cmd/cvelist2osv/testdata/TestExtractVersionInfo_CVE_with_lessThan.yaml
another-rex: I guess the part I'm confused about is a lot of those files got "deleted" / moved from another folder, and these specific ones became empty (if you look at the same filename in the original folder that got "deleted", you can see that it had stuff inside before.

another-rex (approved):
LGTM :)

Issue #3087, opened 2025-01-23
CVEs with many affected packages frontend is overwhelming


jess-lowe:
https://github.com/user-attachments/assets/688048cc-8e22-4e9e-9c9b-7d93a78b21a3

Too many affected packages where all of the information is being displayed makes the information overwhelming for users. 

Possible solutions could include:
- Better grouping
- Collapse tabs by default when > ~7 affected packages

hogo6002:
Nice video description!

andrewpollock:
@jess-lowe for the record, could you please state which records this screen cast is showing?

jess-lowe:
> @jess-lowe for the record, could you please state which records this screen cast is showing?

[UBUNTU-CVE-2024-45341](https://osv.dev/vulnerability/UBUNTU-CVE-2024-45341)

Kunal9027:
Hi, @jess-lowe 

I’d like to work on this issue. For auto-collapsing, I believe the following logic can help:

```
{% set collapse_threshold = 7 %}
<spicy-sections
  class="vulnerability-packages{% if vulnerability.affected | length > collapse_threshold %} force-collapse{% endif %}">
```
This would collapse the section when the number of affected packages exceeds 7.


Regarding grouping, I have implemented logic to group affected packages by their ecosystem:

```
{% set grouped_packages = {} %}
{% for affected in vulnerability.affected %}
    {% set ecosystem = affected.package.ecosystem if 'package' in affected else 'Git' %}
    {% if ecosystem not in grouped_packages %}
        {% set _ = grouped_packages.update({ecosystem: []}) %}
    {% endif %}
    {% set _ = grouped_packages[ecosystem].append(affected) %}
{% endfor %}

{% for ecosystem, packages in grouped_packages.items() %}
  <h2>{{ ecosystem }}</h2>
  <div>
    {% for package in packages %}
      <h3>{{ package.package.name }}</h3>
    {% endfor %}
  </div>
{% endfor %}


```


Could you provide any suggestions or feedback on this approach? I’d appreciate any guidance to improve it.

Thanks!
Kunal9027

jess-lowe:
Hey @Kunal9027. Happy for you to work on this - feel free to put the logic into a PR and so we can check out the commit and play around with the frontend that way. 

We have recently added a way of hopefully testing the frontend without GCP access but I haven't yet set up a test case that would help with this specific case, otherwise I'd ask if you'd be able to test it.

Let's move this chat to a PR regardless :)

[...]

PR #3753, opened 2025-08-03
feat(frontend): Prevent horizontal overflow for affected packages

### LINKED ISSUE #3752: Vulnerability page cuts off affected packages on horizontal overflow
On the vulnerability page, when there are more affected packages than can fit on the screen, the display cuts off the remaining packages in some cases. 
For example, on [RHSA-2025:12478](https://osv.dev/vulnerability/RHSA-2025:12478), only 4 of the 5 affected packages are visible.

**Notes:**
- I believe this issue would be particularly noticeable when package names are long.
- This does not affect mobile devices, as they use a vertical, collapsible layout for this section.

dizzydroid:
This PR improves the UX of the "Affected packages" section on the vulnerability page.
Previously, a large number of packages would [in some cases](https://github.com/google/osv.dev/issues/3752) overflow the page and become inaccessible.

This change modifies the rule for displaying the vertical layout to include ecosystem names, ensuring all packages are viewable and accessible.

This change also collapses the vertical tabs when they exceed a certain threshold, to minimize vertical scrolling.

Fixes #3752 , related to #3087


hogo6002:
Hi, thanks for contribution! 

The new added scroll bar applies to both the package selection area and the package detail area. It can add unwanted space to the package detail section:
image

The scroll bar should only be in the package selection area. Let's also only focus on fixing this issue in this PR.

The other issue you are trying to resolve needs more of design. When there are a lot of affected packages, we still like to keep it in vertical layout but with some improvements. We need a better way to group them, such as using drop downs for ecosystem or package names, or collapsing some by default. Let's fix that in a different PR.

ashmod:
> Hi, thanks for contribution!
> 
> The new added scroll bar applies to both the package selection area and the package detail area. It can add unwanted space to the package detail section: image
> 
> The scroll bar should only be in the package selection area. Let's also only focus on fixing this issue in this PR.
> 
> The other issue you are trying to resolve needs more of design. When there are a lot of affected packages, we still like to keep it in vertical layout but with some improvements. We need a better way to group them, such as using drop downs for ecosystem or package names, or collapsing some by default. Let's fix that in a different PR.

Hi, thanks for the feedback!

I initially tried to implement the horizontal scrollbar to the package selection area, however it seems it is not feasible with the current styling, specifically the `<spicy-sections>` web component's inner style rules seem to be interfering with this kind of external modification. (spicy-sections includes the entire container not just the "tabs" part).

I've updated the PR to focus on the root cause of the layout issue.

The problem was that the logic for switching to the vertical view only considered the length of package names, not ecosystem names. I've modified this logic to check the combined length of both, which now correctly triggers the vertical layout for packages with long ecosystem names.

I've also added a conditional to collapse the vertical view by default when there's more than 4 packages on the screen to reduce vertical scrolling and helps pave the way towards targeting the other issue, which will be in a separate PR.

Let me know if any changes are needed!

gcp/website/frontend3/src/templates/vulnerability.html
```diff
@@ -411,7 +411,13 @@ <h3 class="mdc-layout-grid__cell--span-3">
       });
     }
 
-    expandPackageHeaders();
+    const packageHeaders = document.querySelectorAll('.package-header');
+    const EXPAND_PACKAGES_THRESHOLD = 5;
```
hogo6002: Let's change this to something bigger, maybe 10. The reason is we only want to collapse the list when there's an excessive number of packages. For some Linux distros, a record can have over 100 packages, and we certainly don't want all of them expanded by default. For all other ecosystems, it's fine to just show all of them.

gcp/website/frontend_handlers.py
```diff
@@ -719,9 +719,10 @@ def event_value(event):
 @blueprint.app_template_filter('should_collapse')
 def should_collapse(affected):
   """Whether if we should collapse the package tab bar."""
-  total_package_length = sum(
-      len(entry.get('package', {}).get('name', '')) for entry in affected)
-  return total_package_length > 70 or len(affected) > 5
+  total_text_length = sum(
```
hogo6002: Yeah you are right about we didn't consider the length of the ecosystem name. 

But this code isn't quite right here. The length should be max(total_text_length_ecosystem, total_text_length_package). The ecosystem and package names are in different rows, so we shouldn't add their lengths together. I'm also not sure why we're using `> 70` here. A slightly larger number probably work better?
image

So the end condition should be `return len(affected) > 5 or max_total_length > 100` (we can adjust that number after inspecting the data to see what makes the most sense).

gcp/website/frontend3/src/templates/vulnerability.html
ashmod: Fixed in [9acafd9](https://github.com/google/osv.dev/pull/3753/commits/9acafd948e9d8c2cc614e5b04a2d17ce32032782)

gcp/website/frontend_handlers.py
ashmod: I agree, this makes more sense! Should be fixed in [9acafd9](https://github.com/google/osv.dev/pull/3753/commits/9acafd948e9d8c2cc614e5b04a2d17ce32032782)

PR #3755, opened 2025-08-04
feat: Export a modified_id.csv

### LINKED ISSUE #3671: List of published OSV records?
Is there a way to get the list of modified/published records?  The web-ui has it here: https://osv.dev/list?page=2 

But that's not a great way for anyone else to consume this information. Can you make the web-ui list view available as an API endpoint? I want to get a list of the published OSV records, in order of recent publiciation/modification.

I know there is a data dump of all the OSV records. However, if I want to get the latest record set, I have to download the entire ZIP again and work out what records have changed. 

Is there any quick way to work out what records have recently changed so I only have to process a subset of the changed data?

It could be as simple as posting a JSON listing like so: https://packages.cgr.dev/chainguard/osv/all.json

another-rex:
Resolves #3671

This outputs a `modified_id.csv` in this format in each ecosystem directory:
```
<iso modified date>,<id>
```

Then a `modified_id.csv` at the top level containing all vulns:
```
<iso modified date>,<ecosystem_dir>/<id>
```

No deduplication is done between ecosystems.

### DIFF
--- a/gcp/workers/exporter/export_runner.py
+++ b/gcp/workers/exporter/export_runner.py
@@ -122,12 +122,12 @@ def aggregate_all_vulnerabilities(work_dir: str, export_bucket: str):
       reader = csv.reader(infile)
       for line in reader:
         # Create <timestamp>,<dir>/<osv_id>
-        full_modified_list.append(f'{line[0]},{dir_from_work_dir}/{line[1]}')
+        full_modified_list.append((line[0],f'{dir_from_work_dir}/{line[1]}'))
 
   full_modified_list.sort(reverse=True)
 
   with open(output_modified_list, 'w') as outfile:
-    outfile.write('\n'.join(full_modified_list))
+    csv.writer(outfile).writerows(full_modified_list)
 
   storage_client = storage.Client()
   bucket = storage_client.get_bucket(export_bucket)
--- a/gcp/workers/exporter/export_runner.py
+++ b/gcp/workers/exporter/export_runner.py
@@ -122,7 +122,7 @@ def aggregate_all_vulnerabilities(work_dir: str, export_bucket: str):
       reader = csv.reader(infile)
       for line in reader:
         # Create <timestamp>,<dir>/<osv_id>
-        full_modified_list.append((line[0],f'{dir_from_work_dir}/{line[1]}'))
+        full_modified_list.append((line[0], f'{dir_from_work_dir}/{line[1]}'))
 
   full_modified_list.sort(reverse=True)
 
--- a/gcp/workers/exporter/export_runner.py
+++ b/gcp/workers/exporter/export_runner.py
@@ -24,7 +24,7 @@ import zipfile as z
 
 from google.cloud import ndb, storage
 
-from exporter import safe_upload_single
+from exporter import safe_upload_single, LAST_MODIFIED_FILE
 import osv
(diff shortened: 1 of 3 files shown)


gcp/workers/exporter/export_runner.py
```diff
@@ -74,7 +75,9 @@ def spawn_ecosystem_exporter(work_dir: str, bucket: str, eco: str):
   """
   logging.info('Starting export of ecosystem: %s', eco)
   proc = subprocess.Popen([
-      'exporter.py', '--work_dir', work_dir, '--bucket', bucket, '--ecosystem',
+      # Assume exporter.py is in the same directory as this file
+      # This is true for local dev and Docker
+      f'{__file__}/exporter.py', '--work_dir', work_dir, '--bucket', bucket, '--ecosystem',
```
michaelkedar: is `{__file__}/exporter.py` correct here? I thought that would become `path/to/exporter_runner.py/exporter.py`?

gcp/workers/exporter/export_runner.py
```diff
@@ -103,10 +106,31 @@ def aggregate_all_vulnerabilities(work_dir: str, export_bucket: str):
...
+  # Also aggregate modified_list.csv files
+  output_modified_list = os.path.join(work_dir, LAST_MODIFIED_FILE)
+  full_modified_list = []
+  for file_path in glob.glob(
+          os.path.join(work_dir, f'**/{LAST_MODIFIED_FILE}'), recursive=True):
+    dir_from_work_dir = os.path.relpath(os.path.dirname(file_path), work_dir)
+    with open(file_path, 'r') as infile:
+      modified_list = infile.read().splitlines()
+      for modified in modified_list:
+        split_modified = modified.split(',')
```
michaelkedar: It's probably better to use Python's built-in [csv module](https://docs.python.org/3/library/csv.html) instead of manually doing this parsing

gcp/workers/exporter/export_runner.py
another-rex: Ah good point, fixed now and tested locally.

hogo6002:
why not using the JSON format to keep it the same style as our all.json?
```[{
 "id": "OSV-CVE-2020-1111",
 "modified": "2023-12-04T10:12:08.00Z",
},{
 "id": "OSV-CVE-2020-1112",
 "modified": "2023-12-04T10:16:25.00Z"
}]```

gcp/workers/exporter/export_runner.py
another-rex: Good idea, updated to use csv module for reading and writing.

another-rex:
> not de-duplicating these might be a bit odd when we end up with the same vuln multiple times in a row like

Doing this because you can't actually download a JSON without knowing the ecosystem, so if you only want to download the records that have changed, you need the ecosystem prefix. And then it'll be weird if we only pick one prefix by random.

> why not using the JSON format to keep it the same style as our all.json?

CSV has a couple of benefits:
1. Has less redundant characters, reducing file size
2. each line is independent, so you can just stream the first couple of kilobytes of the file, and you can easily parse just that to get the most recent changed files, and it will still be a fully valid CSV.

michaelkedar (approved):

hogo6002 (approved):

PR #3774, opened 2025-08-06
feat(linter): output linter results to a GCS bucket

hogo6002:
The linter cron job now outputs osv-linter results to a public GCS bucket. The results are organized by source, which gives upstream maintainers easy access to the details of each finding.

### DIFF
--- a/gcp/workers/linter/linter.py
+++ b/gcp/workers/linter/linter.py
@@ -37,6 +37,9 @@ ZIP_FILE_PATH = 'all.zip'
 TEST_DATA = '/linter/test_data'
 GCP_PROJECT = 'oss-vdb-test'
 
+LINTER_EXPORT_BUCKET = 'osv-test-public-import-logs'
+LINTER_RESULT_DIR = 'linter-result'
+
 ERROR_CODE_MAPPING = {
     'SCH:001': osv.ImportFindings.INVALID_JSON,
     'REC:001': osv.ImportFindings.INVALID_RECORD,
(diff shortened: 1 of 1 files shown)


gcp/workers/linter/linter.py
```diff
@@ -147,6 +149,38 @@ def record_quality_finding(bug_id: str, source: str,
                   new_findings, source)
 
 
+def upload_record_to_bucket(json_result: Any):
+  """Uploads the linter result to a GCS bucket, creating one file per source."""
+  storage_client = storage.Client()
+  bucket = storage_client.get_bucket(LINTER_EXPORT_BUCKET)
+  prefix_to_source = construct_prefix_to_source_map()
```
another-rex: Out of scope of this PR, but I do wonder why there is this construct function and why it's not just a global variable.

gcp/workers/linter/linter.py
```diff
@@ -147,6 +149,38 @@ def record_quality_finding(bug_id: str, source: str,
...
+    if not results:
+      continue
+
+    target_path = os.path.join(LINTER_RESULT_DIR, source, 'result.json')
+    logging.info('Uploading linter result for source '
+                 '%s'
+                 ' to %s/%s', source, LINTER_EXPORT_BUCKET, target_path)
+
+    blob = bucket.blob(target_path)
+    blob.upload_from_string(
```
another-rex, line 192: Should we be cleaning up previous uploads? E.g. for records that have been fixed.

another-rex: Depends on if the bucket is the source of truth, or you are expecting everyone to go through the API.

another-rex (approved):

michaelkedar (approved):

gcp/workers/linter/linter.py
hogo6002: Good catch! Added cleanup function before writing new ones.

gcp/workers/linter/linter.py
hogo6002: added a global variable

michaelkedar (approved):

PR #3776, opened 2025-08-07
feat(staging): use GCS/new entities for API

michaelkedar:
Trying out the new GCS-backed vulnerabilities for making API queries on the staging instance (only). This should give us a good idea on the performance improvements.
- Created `server_new.py` to (temporarily) keep the new matching logic a bit separate
- Added new required indexes to allow AffectedVersions to be ordered by ID
- Modified the page token to have some metadata that keeps track of the last returned ID (to avoid duplicate vulns across pages)


gcp/api/cursor.py
```diff
@@ -74,13 +76,17 @@ def from_page_token(cls, page_token: str | None) -> Self:
...
+    if len(split_values) >= 2:
       page_token = split_values[1]
       try:
         qc.query_number = int(split_values[0])
       except ValueError as e:
         raise ValueError('Invalid page token.') from e
+      for enc_meta in split_values[2:]:
+        meta = base64.urlsafe_b64decode(enc_meta).decode(errors='ignore')
+        k, _, v = meta.partition('=')
+        qc._metadata[k] = v
```
another-rex: Should this have a check to see if v != ""?

gcp/api/server_new.py
```diff
@@ -0,0 +1,199 @@
...
+    if affected.ecosystem == 'GIT':
+      vers = osv.models.maybe_strip_repo_prefixes(vers, [affected.name])
+    if osv.normalize_tag(version) in osv.normalize_tags(vers):
+      return True
+    if canonicalize_version(version) in (
+        canonicalize_version(v) for v in affected.versions):
+      return True
+    return False
+
+  # Check if this version is affected in the events list
```
another-rex: nit: maybe underline this comment so that it's clear it's the following section that is being described?

gcp/api/server_new.py
```diff
@@ -0,0 +1,199 @@
...
+from packaging.utils import canonicalize_version
+
+from google.cloud import exceptions
+from google.cloud import ndb
+from google.cloud.ndb import tasklets
+from google.protobuf import timestamp_pb2
+
+import osv
+
+CURSOR_LAST_ID = 'last_id'
```
another-rex: Hmm can this be done at the metadata's type level? I.e. make metadata a struct with specific fields, and parse it into those fields, rather than having a dict that anyone can fill with anything.  This way we can also limit the length of the ID to something reasonable as well.

gcp/api/server_new.py
michaelkedar: I've split this up into `_match_versions()` and `_match_events()` instead, which should be clearer

gcp/api/cursor.py
michaelkedar: an empty string doesn't cause problems, and would let us set key-only metadata (though I don't think that we would ever need to)

gcp/api/server_new.py
michaelkedar: I've turned metadata into a dataclass, and made it raise the invalid page token error for unknown keys.
I didn't feel like doing the length check, since it isn't passed directly to datastore anyway.

another-rex (approved):

Issue #3573, opened 2025-06-12
Support openEuler datasource


kirigiricloud:
Hi all,

I'm from openEuler security team. We're an open source community aiming to build a secure and trusted linux distro. We're currently building our security database and exporting security advisories in OSV Schema.

[Edit] Full advisories are now in the openEuler repo: https://repo.openeuler.org/security/data/osv/

We would like OSV to validate and ingest our data, and we're ready to discuss the remaining onboarding steps.

Thanks for your consideration! Any suggestions would be greatly appreciated.

Tony

- [x] Prepare your data \- refer to the [OSV Schema](https://ossf.github.io/osv-schema/) documentation for information on how to properly format the data so it can be accepted.  
        
- [x] Create a PR to [reserve an ID prefix and define a new ecosystem](https://ossf.github.io/osv-schema/#id-modified-fields) ([example](https://github.com/ossf/osv-schema/pull/219)). We review the records you start publishing for OSV Schema [correctness](https://github.com/ossf/osv-schema/tree/main/validation) and [quality](https://google.github.io/osv.dev/data_quality.html) as part of reviewing and merging this PR. [ossf/osv-schema#358]

- [x] Prepare and publish your records via a [Git repository](http:///git-repo-contribution) ([example](https://github.com/AlmaLinux/osv-database/tree/master)). If this method isn’t ideal, we also support publishing records from [REST API endpoints](https://google.github.io/osv.dev/data/new/rest-api) or through a GCS bucket([example](https://storage.googleapis.com/android-osv/)). (https://repo.openeuler.org/security/data/osv/)
        
- [x] To support API querying, please create a PR to extend [purl\_helpers.py](https://github.com/google/osv.dev/blob/master/osv/purl_helpers.py) and create a new ecosystem in [\_ecosystems.py](https://github.com/google/osv.dev/blob/master/osv/ecosystems/_ecosystems.py). You can refer to existing examples showing how to implement support for [Semver](https://github.com/google/osv.dev/blob/139de7b69a2ea39e2113309b3a0a47aab920ddcf/osv/ecosystems/_ecosystems.py#L45) and [non-Semver](https://github.com/google/osv.dev/pull/3430) ecosystems.  [#3777 ] 
        
- [x] Create a PR to start [importing the records you are publishing into our test instance of OSV.dev](https://github.com/google/osv.dev/blob/master/source_test.yaml) and validate everything is working as intended there. [#3779 ]

[...]

PR #3777, opened 2025-08-07
feat: support openeuler ecosystem

kirigiricloud:
ref: #3573 
Extending `purl_helper.py` and adding `openEuler` ecosystem in `_ecosystems.py`; openEuler follows rpm versioning.

PR #3779, opened 2025-08-07
feat: add openEuler to test environment

kirigiricloud:
ref: #3573

### DIFF
--- a/source_test.yaml
+++ b/source_test.yaml
@@ -269,7 +269,7 @@
   rest_api_url: 'https://repo.openeuler.org/security/data/osv/all.json'
   type: 2
   ignore_patterns: ['^(?!OESA-).*$']
-  directory_path: 'data/osv'
+  directory_path: 'security/data/osv'
   detect_cherrypicks: False
   extension: '.json'
   db_prefix: ['OESA-']


source_test.yaml
```diff
@@ -264,6 +264,21 @@
   editable: False
   strict_validation: True
 
+- name: 'openeuler'
+  versions_from_repo: False
+  rest_api_url: 'https://repo.openeuler.org/security/data/osv/all.json'
+  type: 2
+  ignore_patterns: ['^(?!OESA-).*$']
+  directory_path: 'data/osv'
```
another-rex:
```suggestion
  directory_path: 'security/data/osv'
```

(This doesn't actually matter, the rest importer uses link and rest_api_url, but it's nice to be consistent)

another-rex (dismissed):
Let's wait till #3777 is merged in first before merging this.

another-rex (approved):

michaelkedar (approved):

source_test.yaml
kirigiricloud: Thanks for reviewing the code! We will update the prod PR to include this point as well.

PR #3785, opened 2025-08-09
fix(frontend): Escape underscores in Markdown version strings

dizzydroid:
Fixes issue where version strings containing underscores (e.g., [`1381.v2c3a_12074da_b_`](https://osv.dev/vulnerability/GHSA-62jv-j4w7-5hh8#:~:text=Credentials%20Plugin-,1381.v2c3a12074dab_,-redacts%20the%20encrypted)) were incorrectly rendered with italic formatting in vulnerability descriptions due to Markdown processing treating underscores as emphasis markers.

### DIFF
--- a/gcp/website/frontend_handlers.py
+++ b/gcp/website/frontend_handlers.py
@@ -769,6 +769,9 @@ _URL_MARKDOWN_REPLACER = re.compile(r'(<a href=\".*?)(/ /)(.*?\">)')
 def markdown(text):
   """Render markdown."""
   if text:
+    # Escape underscores in version strings.
+    text = text.replace('_', r'\_')
+    
     md = markdown2.markdown(
         text, safe_mode='escape', extras=['fenced-code-blocks'])
     # TODO(michaelkedar): Seems like there's a bug with markdown2 not escaping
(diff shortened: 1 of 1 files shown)


another-rex:
Hmm... this seem more like an upstream data issue, where they are not escaping `_` correctly (or using code blocks). This could break anyone using `_` as an italics marker instead of inside a version string.

another-rex:
Also, this would definitely break people using _ correctly inside code blocks, or already escaping them.

ashmod:
I agree, that's an aggressive fix for what's probably a data upstream issue..
I think one better approach would be only escaping those that appear in version strings as a countermeasure to cases where this may happen. Something like `processed_ver_text = re.sub(r'(\d+\.[a-zA-Z0-9]+)_(\w+)', r'\1\\_\2', ver_text)`, but if this causes overhead then I think the issue should be left as-is for now. The current display issue is minor and doesn't break functionality. LMK what you think @another-rex !

ashmod:
Updated to escape underscores only in version strings where they are not escaped, preserving code blocks and legitimate Markdown usage.

another-rex:
Thanks for the effort! Because the regex is getting complicated, we might be missing cases where it applies when we don't want it to, and currently it's not a big issue, I think let's just leave this for now.

PR #3809, opened 2025-08-15
feat: update osv-schema

michaelkedar:
For the Echo ecosystem

### DIFF
--- a/osv/models.py
+++ b/osv/models.py
@@ -35,7 +35,7 @@ from . import semver_index
 from . import sources
 from . import vulnerability_pb2
 
-SCHEMA_VERSION = '1.7.0'
+SCHEMA_VERSION = '1.7.3'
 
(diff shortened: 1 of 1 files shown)


osv/models.py
```diff
@@ -35,7 +35,7 @@
 from . import sources
 from . import vulnerability_pb2
 
-SCHEMA_VERSION = '1.7.0'
+SCHEMA_VERSION = '1.7.3'
```
michaelkedar, line 38: 1.7.3 hasn't been released yet, but Echo wasn't in 1.7.2

another-rex:
Made https://github.com/ossf/osv-schema/pull/402 to publish 1.7.3

osv/models.py
michaelkedar: Updated everything to point to https://github.com/ossf/osv-schema/tree/v1.7.3

PR #3813, opened 2025-08-15
fix(frontend): fix ecosystem filtering for versioned ecosystem variants

### LINKED ISSUE #3812: Regression: "Fix available" tag shows "No Fix Available" when filtering by ecosystem
After #3784, filtering by ecosystem on the [test instance](https://test.osv.dev) incorrectly shows "No Fix Available" for all vulnerabilities, even when fixes exist in the selected ecosystem.

The feature works correctly in the frontend emulator, suggesting the issue is related to how the live production data is being queried or structured.

**Potential fix:** Remove the default `[]` parameter from the `getattr` call in the `_is_fixed_in_ecosystem` function:
https://github.com/google/osv.dev/blob/3a8174f09206943abe9f8e8958ed3707f274bf8c/gcp/website/frontend_handlers.py#L651
To:
```python
for affected_pkg in bug.affected_packages:
```
since the attribute is guaranteed to exist, the `getattr` with default `[]` may be masking cases where the attribute exists but is still being accessed, causing the function to iterate over an empty list and always return `False` (showing "No Fix Available").

Image

ashmod:
Fixes ecosystem filtering to correctly match versioned ecosystem variants (e.g., filtering by "Alpaquita" now matches "Alpaquita:23", "Alpaquita:stream").

Closes #3812

### DIFF
--- a/gcp/website/frontend_handlers.py
+++ b/gcp/website/frontend_handlers.py
@@ -628,7 +628,7 @@ def osv_query(search_string, page, affected_only, ecosystem):
     eco_variants.add(osv.ecosystems.normalize(ecosystem))
 
     for item, bug_obj in zip(result_items, bugs):
-      item['isFixed'] = _is_fixed_in_ecosystem(bug_obj, eco_variants)
+      item['isFixed'] = _is_fixed_in_ecosystem(bug_obj, eco_variants, ecosystem)
 
   results = {
       'total': total_future.get_result(),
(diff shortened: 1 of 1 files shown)


another-rex:
Hmm, this issue still seems to be showing up for GIT ecosystems. E.g. https://test.osv.dev/list?q=CVE-2025-55285&ecosystem=GIT

another-rex:
The git one is a bit weird, you need to add a special case for it, maybe just try to match all ecosystems if the input ecosystem is GIT.

ashmod:
> Hmm, this issue still seems to be showing up for GIT ecosystems. E.g. https://test.osv.dev/list?q=CVE-2025-55285&ecosystem=GIT

Just noticed that too - this stems from Git ecosystems missing a `package` field in their affected structure. This should be handled now in #3814

PR #3814, opened 2025-08-18
fix(frontend): fix Git ecosystem filtering

ashmod:
Fixes ecosystem filtering for Git vulnerabilities.


ashmod:
Hi @another-rex ,
It seems the recent changes are not yet reflected on the test instance. If it should've propagated by now, would a maintainer with `make run-website` access be able to test it locally using live data to help verify if this works properly? Thanks!

PR #3820, opened 2025-08-19
feat(vulnfeeds): Add version extraction logic for CVE to OSV conversion

jess-lowe:
This PR introduces the core logic for extracting version information during the conversion of CVE JSON 5.0 data into the OSV format. The primary goal is to populate the affected field in the resulting OSV vulnerability file with accurate version ranges.

**Key changes include:**
- Version Range Extraction: Implements functionality to parse version ranges from the affected field in the CVE data, including handling lessThan and lessThanOrEqual properties.
- Inverse Range Calculation: Adds logic to determine affected ranges from unaffected version statuses, which is a common pattern in Linux kernel CVEs
- CPE Version Parsing: Includes support for extracting version information from CPE data when available.

This change focuses exclusively on extracting version and range information as specified in the CVE record. The logic to resolve these versions to specific Git commit hashes is out of scope for this PR and will be addressed in a future change.


vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -59,6 +61,412 @@ func extractConversionMetrics(cve cves.CVE5, refs []osvschema.Reference) {
...
+}
+
+// AddVersionInfo attempts to extract versions into version ranges and events
+// through different methods, sometimes based on specific CNA behaviour.
+func AddVersionInfo(cve cves.CVE5, v *vulns.Vulnerability) []string {
+	var notes []string
+	if cve.Metadata.AssignerShortName == "Linux" {
+		pkg := osvschema.Package{
+			Ecosystem: string(osvschema.EcosystemLinux),
+			Name:      "Kernel"}
```
another-rex:
```suggestion
			Name:      "Kernel"
	}
```

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -59,6 +61,412 @@ func extractConversionMetrics(cve cves.CVE5, refs []osvschema.Reference) {
...
+
+		cpeRanges, cpeStrings, err := findCPEVersionRanges(cve)
+		if err != nil {
+			notes = append(notes, err.Error())
+		}
+		if cpeRanges != nil {
+			nA := osvschema.Affected{
```
another-rex: what is nA? Can you expand it out to a full name?

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -59,6 +61,412 @@ func extractConversionMetrics(cve cves.CVE5, refs []osvschema.Reference) {
...
+				Package: pkg,
+			}
+			for _, r := range cpeRanges {
+				r.Type = osvschema.RangeEcosystem
+				nA.Ranges = append(nA.Ranges, r)
+			}
+			nA.DatabaseSpecific = make(map[string]interface{})
+			nA.DatabaseSpecific["CPEs"] = vulns.Unique(cpeStrings)
+			v.Affected = append(v.Affected, nA)
```
another-rex: This section seems pretty self contained, can you extract that out into a separate file? I imagine in the future we'll have a bunch of these CNA specific logic, so we can have a file containing just that.

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -59,6 +61,412 @@ func extractConversionMetrics(cve cves.CVE5, refs []osvschema.Reference) {
...
+		if adp.Affected != nil {
+			affected = append(affected, adp.Affected...)
+		}
+	}
+
+	// Attempt to extract version ranges from Affected field
+	for _, cveAff := range affected {
+		versionRanges, versionType, extractNotes := ExtractVersionsFromAffectedField(cveAff, cve.Metadata.AssignerShortName)
+		if len(versionRanges) == 0 {
+			notes = append(notes, extractNotes...)
```
another-rex: Notes only need to be appended when no versions are found? Can you add a comment here explaining why the other cases should not append notes.

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -59,6 +61,412 @@ func extractConversionMetrics(cve cves.CVE5, refs []osvschema.Reference) {
...
+			continue
+		} else {
+			gotVersions = true
+		}
+		if versionType == "git" {
+			nA := osvschema.Affected{}
```
another-rex: nit: please expand nA

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -59,6 +61,412 @@ func extractConversionMetrics(cve cves.CVE5, refs []osvschema.Reference) {
...
+
+		// Deal with this later
+		if len(ranges) != 0 {
+			return ranges, string(osvschema.RangeSemVer), notes
+		}
+		// TODO(jesslowe): add more notes here
+		return nil, "", notes
+	}
+
+	versionTypesCount := make(map[string]int)
```
another-rex: Version type really should be an enum. I know that the cve.Affected type itself is not an enum, so we'll need to convert it to a concrete set of values. 

E.g. 

```
Git = "git"
Semver = "Semver"
Ecosystem = "<whatever common ecosystem versioning is.>"
Other = "unexpected other type"
```
If ecosystem one is commonly a bunch of different types, we can just get rid of other and lump any unknown ones into ecosystem

another-rex: Generally when parsing data, as early on in the conversion process we want to get rid of as much uncertainty as possible, so most of the business logic part where we do the interpreting of the data has the most constrained / concrete set of values. 

<sub>*imo idk if this is the best way</sub>

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -59,6 +61,412 @@ func extractConversionMetrics(cve cves.CVE5, refs []osvschema.Reference) {
...
+
+		// Quality check
+		vQuality := vulns.CheckQuality(vers.Version)
+		if vQuality >= vulns.Filler {
+			notes = append(notes, fmt.Sprintf("Version value for %s %s is filler or empty", affected.Vendor, affected.Product))
+		}
+		vLessThanQual := vulns.CheckQuality(vers.LessThan)
+		vLTOEQual := vulns.CheckQuality(vers.LessThanOrEqual)
+
+		hasRange := vLessThanQual <= vulns.Spaces || vLTOEQual <= vulns.Spaces
```
another-rex: hmm vulns.Spaces is kind of confusing, especially <=, which makes me think "worse than or equal". Can you save the quality to a variable somewhere in this func (e.g. `acceptableQuality`) and use that?

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -129,73 +526,79 @@ func writeOSVToFile(id cves.CVEID, cnaAssigner string, vulnDir string, v *vulns.
...
 
-// identifyPossibleURLs extracts all possible URLs from a CVE5 object,
-// including those in CNA and ADP containers, and affected package information.
-// It deduplicates the URLs before returning them.
+// identifyPossibleURLs extracts and deduplicates all URLs from a CVE object.
+// It searches for URLs in the CNA and ADP reference sections, as well as in
+// the 'collectionUrl' and 'repo' fields of the 'affected' entries.
 func identifyPossibleURLs(cve cves.CVE5) []cves.Reference {
-	refs := cve.Containers.CNA.References
+	var refs []cves.Reference
```
another-rex: Huh why this change?

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -59,26 +65,419 @@ func extractConversionMetrics(cve cves.CVE5, refs []osvschema.Reference) {
...
+	// Attempt to extract version ranges from the combined 'affected' fields.
+	for _, cveAff := range affected {
+		versionRanges, versionType, extractNotes := ExtractVersionsFromAffectedField(cveAff, cve.Metadata.AssignerShortName)
+		notes = append(notes, extractNotes...)
+		if len(versionRanges) == 0 {
+			continue
+		}
+		gotVersions = true
+
+		if versionType == "git" {
```
another-rex: 2 things:

- Once versionType is an enum, this could be a specific enum match rather than a string comparison.
- The if statement seems to be very similar between the branches, can the "if" part of the statement just be within the for loop (vr.Type + vr.Repo vs vr.Type)

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -59,26 +65,419 @@ func extractConversionMetrics(cve cves.CVE5, refs []osvschema.Reference) {
...
+	gotVersions := false
+
+	affected := cna.Affected
+	for _, adp := range adps {
+		if adp.Affected != nil {
+			affected = append(affected, adp.Affected...)
+		}
+	}
+
+	// Attempt to extract version ranges from the combined 'affected' fields.
```
another-rex: Just a general observation: 

The comments make it really clear what each section is doing! And the fact most of the logic is nestled away in a function is really nice.

Though there is a surprising amount of code after each extraction function to attach the extracted info back onto the output. I don't think it's something that needs to be fixed, or even how to fix it.

vulnfeeds/cmd/cvelist2osv/main.go
another-rex, line 1: main.go is getting quite big. The new set of functions you added seems self contained, can it be moved to a separate package? (or a separate file if it's too much to move to a different package)

Please do the move in a separate commit from any other changes, so that it's easier to re-review.

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -59,26 +65,419 @@ func extractConversionMetrics(cve cves.CVE5, refs []osvschema.Reference) {
...
+			notes = append(notes, fmt.Sprintf("Extracted versions from description but did not save them: %+v", versions))
+		}
+	}
+
+	return notes
+}
+
+// addToVersionRange is a helper function that adds 'introduced', 'fixed', or 'last_affected'
+// events to an OSV version range. If 'intro' is empty, it defaults to "0".
+func addToVersionRange(versionRange *osvschema.Range, intro string, lastAff string, fixed string) {
```
another-rex: nit: Move addToVersionRange to the bottom. Idea is the functions at the end call no other functions, and all their dependents are above it.

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -59,26 +65,419 @@ func extractConversionMetrics(cve cves.CVE5, refs []osvschema.Reference) {
...
+		}
+	}
+
+	return versionRanges, mostFrequentVersionType, notes
+}
+
+// sortBadSemver provides a custom sorting function for version strings that may not
+// strictly adhere to the SemVer specification. It compares versions numerically,
+// part by part (major, minor, patch).
+func sortBadSemver(a, b string) int {
```
another-rex: nit: put sortBadSemver below findInverseAffectedRanges, since that uses this.

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -59,26 +65,419 @@ func extractConversionMetrics(cve cves.CVE5, refs []osvschema.Reference) {
...
+		ranges, inverseNotes := findInverseAffectedRanges(affected)
+		notes = append(notes, inverseNotes...)
+
+		if len(ranges) != 0 {
+			return ranges, string(osvschema.RangeSemVer), notes
+		}
+		return nil, "", notes
+	}
+
+	versionTypesCount := make(map[string]int)
```
another-rex: Would it be good to do a check here that this is defaultstatus NotAffected, and log / exit early if that's not the case?

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -59,26 +65,419 @@ func extractConversionMetrics(cve cves.CVE5, refs []osvschema.Reference) {
...
+	// Handle cases where a product is marked as "affected" by default, and specific versions are marked "unaffected".
+	if affected.DefaultStatus == "affected" {
+		// For Linux kernel CVEs, this logic is often handled by CPEs, so we skip it here.
+		if cnaAssigner == "Linux" {
+
+			notes = append(notes, "Skipping Linux Affected range versions in favour of CPE versions")
+			return nil, "", notes
+		}
+		// Calculate the affected ranges by finding the inverse of the unaffected ranges.
+		ranges, inverseNotes := findInverseAffectedRanges(affected)
```
another-rex: We have a findInverseAffectedRanges function here, can we also refactor the findNormalAffectedRanges (I'm guessing that's everything after this if block) into its own function as well?  

Ideally we have the same function nest level for different branches.

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -59,26 +65,419 @@ func extractConversionMetrics(cve cves.CVE5, refs []osvschema.Reference) {
...
+	for _, adp := range adps {
+		if adp.Affected != nil {
+			affected = append(affected, adp.Affected...)
+		}
+	}
+
+	// Attempt to extract version ranges from the combined 'affected' fields.
+	for _, cveAff := range affected {
+		versionRanges, versionType, extractNotes := ExtractVersionsFromAffectedField(cveAff, cve.Metadata.AssignerShortName)
+		notes = append(notes, extractNotes...)
```
another-rex: This is probably beyond the scope of this PR, but rather than returning notes, we should just have notes be a struct that is passed around by reference, and directly appended to. Saves all the notes appending that needs to happen with every function call.

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -59,26 +65,419 @@ func extractConversionMetrics(cve cves.CVE5, refs []osvschema.Reference) {
...
+
+	for _, vers := range affected.Versions {
+		if vers.Status != "affected" {
+			continue
+		}
+		if _, ok := versionTypesCount[vers.VersionType]; ok {
+			versionTypesCount[vers.VersionType]++
+		} else {
+			versionTypesCount[vers.VersionType] = 0
+		}
```
another-rex: I don't think this is needed, a map defaults to 0 as the empty value, so you should be able to just have the ++ line by itself.

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -59,26 +65,419 @@ func extractConversionMetrics(cve cves.CVE5, refs []osvschema.Reference) {
...
+						versionRanges = append(versionRanges, versionRange)
+					} else if av.Introduced != "" && av.LastAffected != "" {
+						versionRange := osvschema.Range{}
+						addToVersionRange(&versionRange, av.Introduced, av.LastAffected, "")
+						versionRanges = append(versionRanges, versionRange)
+						notes = append(notes, "vers range updated la")
+					}
+				}
+				continue
+			}
```
another-rex: For these per CNA logic, is it possible to separate out into individual functions for each CNA, and a default  CNA neutral set of logic

vulnfeeds/cmd/cvelist2osv/main.go
another-rex: Hmm, looking at all the usages of addToVersionRange, you seem to be using it with a brand new versionRange every time. Is it possible to just have this return a new versionRange, and rename it buildVersionRange or something. 

I'm guessing this would never add additional events to an existing range, because you always add an introduce every time this function is called, and it's not valid to have multiple introduce in one version range.

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -59,26 +65,419 @@ func extractConversionMetrics(cve cves.CVE5, refs []osvschema.Reference) {
...
+// It handles various scenarios:
+// - If `defaultStatus` is "affected", it may calculate the inverse ranges (common for Linux).
+// - It iterates through `versions` entries with `status: "affected"`.
+// - It constructs ranges from `version`, `lessThan`, and `lessThanOrEqual` fields.
+// - For GitHub CVEs, it uses a special parser for version strings like "< 1.2.3".
+// - For git commits, it creates an introduced event.
+// - As a fallback, it may assume a single version means "fixed at this version, introduced at 0".
+//
+// Returns the extracted OSV ranges, the most frequent version type (e.g., "semver"), and any notes.
+func ExtractVersionsFromAffectedField(affected cves.Affected, cnaAssigner string) (versionRanges []osvschema.Range, rangeType string, notes []string) {
```
another-rex: Should this be a private function? To match the other extractVersions functions. (or make them all public and move it to a separate package).

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -59,26 +65,419 @@ func extractConversionMetrics(cve cves.CVE5, refs []osvschema.Reference) {
...
+			if introduced != "" && fixed != "" {
+				versionRange := osvschema.Range{}
+				addToVersionRange(&versionRange, introduced, "", fixed)
+				versionRanges = append(versionRanges, versionRange)
+			} else if introduced != "" && lastaffected != "" {
+				versionRange := osvschema.Range{}
+				addToVersionRange(&versionRange, introduced, lastaffected, "")
+				versionRanges = append(versionRanges, versionRange)
+			}
+		} else {
```
another-rex: If I'm reading the nesting correctly, you can flatten this else statement and just end the hasRange part with a continue.

vulnfeeds/cmd/cvelist2osv/main.go
another-rex:
```
				versionRange := osvschema.Range{}
				addToVersionRange(&versionRange, introduced, "", fixed)
				versionRanges = append(versionRanges, versionRange)
```
Then you can flatten a lot of these 3 line code into one
```
				versionRanges = append(versionRanges, buildVersionRange(introduced, "", fixed))
```

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -59,26 +65,419 @@ func extractConversionMetrics(cve cves.CVE5, refs []osvschema.Reference) {
...
+			// In this case only vers.Version exists which either means that it is _only_ that version that is
+			// affected, but more likely, it affects up to that version. It could also mean that the range is given
+			// in one line instead - like "< 1.5.3" or "< 2.45.4, >= 2.0 " or just "before 1.4.7", so check for that.
+			notes = append(notes, "Only version exists")
+			// GitHub often encodes the range directly in the version string.
+			if cnaAssigner == "GitHub_M" {
+				av, err := git.ParseVersionRange(vers.Version)
+				if err == nil {
```
another-rex: invert if statement and continue (or return if you extract this section to a func)

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -59,26 +65,419 @@ func extractConversionMetrics(cve cves.CVE5, refs []osvschema.Reference) {
...
+					if av.Introduced != "" && av.Fixed != "" {
```
another-rex: Simplify if statement by extracting and inverting av.Introduced check
```
if av.Introduced == "" { continue }
```

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -59,26 +65,419 @@ func extractConversionMetrics(cve cves.CVE5, refs []osvschema.Reference) {
...
+						addToVersionRange(&versionRange, av.Introduced, av.LastAffected, "")
+						versionRanges = append(versionRanges, versionRange)
+						notes = append(notes, "vers range updated la")
+					}
+				}
+				continue
+			}
+
+			if vers.VersionType == "git" {
+				versionRange := osvschema.Range{}
```
another-rex: You can just specify the Event array inline, or even better just use the existing addToVersionRange / buildVersionRange func here.

vulnfeeds/cmd/cvelist2osv/main.go
another-rex: Even better, you can add a method to the quality type, then you'll have a syntax like so:
```
vLessThanQual.atLeast(acceptableQuality)
```

vulnfeeds/cmd/cvelist2osv/main.go
another-rex: There is probably some advanced interface stuff that we can do here to make it easy to add CNA overrides (similar to scalibr plugins), but let's just keep it simple for now and chuck it into separate functions.

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -59,26 +65,419 @@ func extractConversionMetrics(cve cves.CVE5, refs []osvschema.Reference) {
...
+				addToVersionRange(&versionRange, "0", "", vers.Version)
+				versionRanges = append(versionRanges, versionRange)
+				notes = append(notes, fmt.Sprintf("%s - Single version found %v - Assuming introduced = 0 and Fixed = %v", vQuality, vers.Version, vers.Version))
+			}
+		}
+	}
+	// Determine the most frequent version type to return as the range type.
+	maxCount := 0
+	var mostFrequentVersionType string
```
another-rex: Quick question, what are the different version types we are expecting?

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -59,26 +65,419 @@ func extractConversionMetrics(cve cves.CVE5, refs []osvschema.Reference) {
...
+// considered affected by default, and only unaffected versions are listed.
+// It sorts the introduced and fixed versions to create chronological ranges.
+func findInverseAffectedRanges(cveAff cves.Affected) (ranges []osvschema.Range, notes []string) {
+	var introduced []string
+	var fixed []string
+	for _, vers := range cveAff.Versions {
+		if vers.Status == "affected" {
+			introduced = append(introduced, fmt.Sprintf("%s.0", vers.Version))
+		}
+		if vers.Status == "unaffected" {
```
another-rex: nit: Invert this if statement and continue. Maybe add a note/log an error if that happens (depending on how often status is neither affected or unaffected?)

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -59,26 +65,419 @@ func extractConversionMetrics(cve cves.CVE5, refs []osvschema.Reference) {
...
+// findInverseAffectedRanges calculates the affected version ranges by analyzing a list
+// of 'unaffected' versions. This is common in Linux kernel CVEs where a product is
+// considered affected by default, and only unaffected versions are listed.
+// It sorts the introduced and fixed versions to create chronological ranges.
+func findInverseAffectedRanges(cveAff cves.Affected) (ranges []osvschema.Range, notes []string) {
+	var introduced []string
+	var fixed []string
+	for _, vers := range cveAff.Versions {
+		if vers.Status == "affected" {
+			introduced = append(introduced, fmt.Sprintf("%s.0", vers.Version))
```
another-rex: Why adding .0 to the end?

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -59,26 +65,419 @@ func extractConversionMetrics(cve cves.CVE5, refs []osvschema.Reference) {
...
+		}
+		if vers.Status == "unaffected" {
+			if vers.Version == "0" || vers.VersionType != "semver" {
+				continue
+			}
+			fixed = append(fixed, vers.Version)
+			// Infer the next introduced version from the 'lessThanOrEqual' field.
+			// For example, if "5.10.*" is unaffected, the next introduced version is "5.11.0".
+			minorVers := strings.Split(vers.LessThanOrEqual, ".*")[0]
```
another-rex: What if this version does not have .*?

another-rex: And can you add a test for that?

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -59,26 +65,419 @@ func extractConversionMetrics(cve cves.CVE5, refs []osvschema.Reference) {
...
+	patchA, _ := strconv.Atoi(partsA[2])
+	patchB, _ := strconv.Atoi(partsB[2])
+	return cmp.Compare(patchA, patchB)
+}
+
+// findInverseAffectedRanges calculates the affected version ranges by analyzing a list
+// of 'unaffected' versions. This is common in Linux kernel CVEs where a product is
+// considered affected by default, and only unaffected versions are listed.
+// It sorts the introduced and fixed versions to create chronological ranges.
+func findInverseAffectedRanges(cveAff cves.Affected) (ranges []osvschema.Range, notes []string) {
```
another-rex: This function seems pretty specific to Linux CNA published CVEs, can we add a check for the CNA if we are only currently Linux CNA, and don't convert non linux CNA records here.

another-rex (changes requested):
Overall structure looks pretty good to me!

Added a bunch of comments.

I think if we want to say we are supporting a bunch of different CNAs, we'll need to add a lot more test data in here. Otherwise can also just limit the conversion here to Linux CNA records for now and expand it later.

vulnfeeds/cmd/cvelist2osv/main.go
jess-lowe: Yeah it's definitely something that can be improved - it's somewhat an artifact of moving functions between files and sharing the array became somewhat difficult. I'll look at addressing it in a future pr

vulnfeeds/cmd/cvelist2osv/main.go
jess-lowe: good question. I mostly made this because Linux uses 'semver', and 'original_commit_for_fix' as version types in the same affected collection. maybe it's smarter just to throw out what doesn't match the rest.

vulnfeeds/cmd/cvelist2osv/main.go
jess-lowe: yeah I was considering that but it might get a bit messy. I think that can happen in a later PR

vulnfeeds/cmd/cvelist2osv/main.go
jess-lowe: but otherwise literally everything from 'git', 'semver', 'custom', 'unknown', and nothing at all 'cause it's not a required field.

vulnfeeds/cmd/cvelist2osv/main.go
jess-lowe: nah because it's not a required field, and it's excluded but assumed unaffected a lot of the time.

vulnfeeds/cmd/cvelist2osv/main.go
jess-lowe: moved - hopefully it's alright to review?

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -180,9 +182,10 @@ func AddVersionInfo(cve cves.CVE5, v *vulns.Vulnerability) []string {
 	return notes
 }
 
-// addToVersionRange is a helper function that adds 'introduced', 'fixed', or 'last_affected'
+// buildVersionRange is a helper function that adds 'introduced', 'fixed', or 'last_affected'
 // events to an OSV version range. If 'intro' is empty, it defaults to "0".
-func addToVersionRange(versionRange *osvschema.Range, intro string, lastAff string, fixed string) {
+func buildVersionRange(intro string, lastAff string, fixed string) (versionRange osvschema.Range) {
```
another-rex: nit: no need for a named versionRange return value.

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -269,105 +268,8 @@ func ExtractVersionsFromAffectedField(affected cves.Affected, cnaAssigner string
...
-	// Determine the most frequent version type to return as the range type.
-	maxCount := 0
-	var mostFrequentVersionType string
-	for versionType, count := range versionTypesCount {
-		if count > maxCount {
-			maxCount = count
-			mostFrequentVersionType = versionType
-		}
-	}
+	return findNormalAffectedRanges(affected, cnaAssigner)
```
another-rex: hmm... does this not need to append notes?

vulnfeeds/cmd/cvelist2osv/main.go
another-rex: nit: can you still set vuln.Spaces to a variable at the top of the function?

vulnfeeds/cmd/cvelist2osv/version_extraction.go
```diff
@@ -0,0 +1,413 @@
...
+		if cnaAssigner == "Linux" {
+			notes = append(notes, "Skipping Linux Affected range versions in favour of CPE versions")
+			return nil, VersionRangeTypeUnknown, notes
+		}
+		// Calculate the affected ranges by finding the inverse of the unaffected ranges.
+		ranges, inverseNotes := findInverseAffectedRanges(affected, cnaAssigner)
+		notes = append(notes, inverseNotes...)
+
+		if len(ranges) != 0 {
+			return ranges, VersionRangeTypeSemver, notes
```
another-rex: Is this guarenteed? Or does this also need something similar to what's in findNormalAffectedRanges where we count the most common version type.

It might be clearer to move out the determining version type logic, so that both findInverseAffectedRanges and findNormalAffectedRanges have the same function signature.

jess-lowe: made some changes so they have the same function signature, but from 220-221 in version_extraction.go it pretty much requires that it has the semver type

vulnfeeds/cmd/cvelist2osv/main.go
jess-lowe: it passes the notes through to AddVersionInfo which appends the notes directly

vulnfeeds/cmd/cvelist2osv/version_extraction.go
```diff
@@ -0,0 +1,413 @@
...
+// It handles various scenarios:
+// - If `defaultStatus` is "affected", it may calculate the inverse ranges (common for Linux).
+// - It iterates through `versions` entries with `status: "affected"`.
+// - It constructs ranges from `version`, `lessThan`, and `lessThanOrEqual` fields.
+// - For GitHub CVEs, it uses a special parser for version strings like "< 1.2.3".
+// - For git commits, it creates an introduced event.
+// - As a fallback, it may assume a single version means "fixed at this version, introduced at 0".
+//
+// Returns the extracted OSV ranges, the most frequent version type (e.g., "semver"), and any notes.
+func extractVersionsFromAffectedField(affected cves.Affected, cnaAssigner string) (versionRanges []osvschema.Range, rangeType VersionRangeType, notes []string) {
```
another-rex: you can probably remove all named returns here, and for the linux notes just define the struct inline with the return value. I.e. 
```
return nil, VersionRangeTypeUnknown, []string{"..."}
```

another-rex (approved):
LGTM!

vulnfeeds/cmd/cvelist2osv/version_extraction.go
jess-lowe: done :)

another-rex (approved):

PR #3821, opened 2025-08-20
feat: add `recoverer` for failed tasks

michaelkedar:
Added a new `recoverer` (open to renaming) GKE worker in charge of recovering from database writing errors. It subscribes to the existing `failed-tasks` topic (which is the dead-letter topic for the importer/worker `tasks` topic - I'm pretty sure those were being discarded since it had no subscriptions).
- currently tasks coming from the importer/clusterfuzz are just logged by the recoverer (this will show up in prod)
- Failures from reading/writiing to the GCS bucket backend on staging now publish corresponding messages `failed-tasks`. The recoverer will attempt to rectify these errors.
- Bunch of terraform/cloud deploy/cloud build changes to support this.
- I also did a little refactor - created a utils file with a `get_google_cloud_project()` function that multiple things end up using.

Honestly, I have no idea how much of this code is actually going to run in practice - we haven't really seen any of these GCS errors on staging as of yet. Maybe this will never end up running :upside_down_face: 

In another PR, I'll make a cron job that goes through all the recently-updated `Vulnerability` entities to make sure they have GCS records associated with them (sending problem entities to the `recoverer` to fix)


gcp/workers/recoverer/recoverer.py
```diff
@@ -0,0 +1,230 @@
...
+      response = subscriber.pull(subscription=subscription, max_messages=1)
+      if not response.received_messages:
+        continue
+
+      message = response.received_messages[0].message
+      ack_id = response.received_messages[0].ack_id
+      if handle_task(message):
+        subscriber.acknowledge(subscription=subscription, ack_ids=[ack_id])
+      else:
+        subscriber.modify_ack_deadline(
```
another-rex, line 230: Can you add a comment here to describe what this is doing?

gcp/workers/recoverer/recoverer.py
```diff
@@ -0,0 +1,230 @@
...
+            subscription=subscription, ack_ids=[ack_id], ack_deadline_seconds=0)
+
+
+if __name__ == '__main__':
```
another-rex, line 234: Is this missing setupGcpLogging?

gcp/workers/recoverer/recoverer.py
```diff
@@ -0,0 +1,230 @@
...
+
+HANDLERS = {
+    'gcs_retry': handle_gcs_retry,
+    'gcs_missing': handle_gcs_missing,
+    'gcs_gen_mismatch': handle_gcs_gen_mismatch,
+}
+
+
+def handle_task(message: pubsub_v1.types.PubsubMessage) -> bool:
+  """Handle a task message."""
```
another-rex:
```suggestion
  """Handle a failed task message."""
```
?

gcp/workers/recoverer/recoverer.py
```diff
@@ -0,0 +1,230 @@
...
+  Lazily initialized to allow testing with datastore emulator."""
+  global _ndb_client
+  if _ndb_client is None:
+    _ndb_client = ndb.Client()
+  return _ndb_client
+
+
+def handle_gcs_retry(message: pubsub_v1.types.PubsubMessage) -> bool:
+  """Handle a failed GCS write."""
+  # Check that the record hasn't been written/updated in the meantime.
```
another-rex: Move the comment down to the line where the check actually takes place

gcp/workers/recoverer/recoverer.py
```diff
@@ -0,0 +1,230 @@
...
+  bucket = osv.gcs.get_osv_bucket()
+  path = os.path.join(osv.gcs.VULN_PB_PATH, vuln.id + '.pb')
+  pb_blob = bucket.get_blob(path)
+  if pb_blob and pb_blob.custom_time and pb_blob.custom_time >= modified:
+    logging.warning(
+        'gcs_retry: %s was modified before message was processed: '
+        'message: %s, blob: %s', vuln.id, modified, pb_blob.custom_time)
+    # TODO(michaelkedar): trigger a reimport of the record.
+    return True
+  pb_blob = bucket.blob(path)
```
another-rex, line 69:
```suggestion

  pb_blob = bucket.blob(path)
```

gcp/workers/recoverer/recoverer.py
```diff
@@ -0,0 +1,230 @@
...
+    return False
+
+
+def handle_gcs_missing(message: pubsub_v1.types.PubsubMessage) -> bool:
+  """Handle a failed GCS read."""
+  vuln_id = message.attributes.get('id')
+  if not vuln_id:
+    logging.error('gcs_missing: message missing id attribute: %s', message)
+    return True
+  # Re-put the Bug to regenerate the GCS & Datastore entities
```
another-rex, line 86: Should we check if it has been added to GCS before reputting?

gcp/workers/recoverer/recoverer.py
```diff
@@ -0,0 +1,230 @@
...
+        else:
+          aliases = sorted(set(alias_group.bug_ids) - {vuln_id})
+          aliases_modified = alias_group.last_modified
+        # Only update the modified time if it's actually being modified
+        if vuln_proto.aliases != aliases:
+          vuln_proto.aliases[:] = aliases
+          if aliases_modified > modified:
+            modified = aliases_modified
+          else:
+            modified = datetime.datetime.now(datetime.UTC)
```
another-rex, line 145: Probably fine to leave here for now, but is this code the same as the code in the alias group service? We probably should extract it.

gcp/workers/recoverer/recoverer.py
michaelkedar: Yes - added

gcp/workers/recoverer/recoverer.py
michaelkedar: In this case, it's ok if we don't, since the re-put will completely regenerate the bug with fresh data.

gcp/workers/recoverer/recoverer.py
michaelkedar: Yeah - it's pretty similar to alias/upstream.
It's hard to share code between these two at the moment due to how the dockerfiles are built.

another-rex (approved):

PR #3825, opened 2025-08-22
chore: Upgrade to 3.13

another-rex:
I think I found all the instances of python in dockerfiles.

Not sure if there's any other files that need to be updated


michaelkedar (approved):
Seems like everything I can remember

PR #3826, opened 2025-08-23
docs: Add OSV logo to README.md

ashmod:


ashmod:
On another note, the logo used on the main website is different from the updated one (that is used on the docs site). Should we maybe update it at some point?

jess-lowe (approved):
LGTM, thanks!

PR #3827, opened 2025-08-25
fix: remove `google-cloud-profiler`

michaelkedar:
`google-cloud-profiler` fails to install in Python 3.13 (and was only officially supported up to 3.11).
While useful, we'd prefer to use the newer python version, so removed it.

Also fix the python version in CONTRIBUTING.md that was missed in #3825


cuixq (approved):
`build-essential` probably is not needed neither: https://github.com/google/osv.dev/blob/master/gcp/api/Dockerfile#L17

Issue #3849, opened 2025-08-28
Vanir signature in OSV.dev


cuixq:
[Vanir](https://github.com/google/vanir) is able to identify missing security patches based on the generated signatures for vulnerabilities in OSV format.

We would like to generate Vanir signatures in worker to make the signatures available in OSV API so that Vanir is able to fetch the signatures by query OSV.dev.

PR #3852, opened 2025-08-28
feat(worker): vanir signature generation

cuixq:
https://github.com/google/osv.dev/issues/3849

This PR integrates the vanir tool into `worker` to generate signatures for git commits. If a vulnerability contains at least one GIT type affected range, the worker will invoke vanir to generate signatures. The resulting signatures are stored in the `database_specific` field of an OSV record.

This feature is currently only enabled for test instance to allow for validation before being enabled in production.

Performance: the signature generation slows down the worker test for about 20% (from 280s to 320s) and this may be improved with the TODO below.

TODO: currently the [OSV GIT code extractor](https://github.com/google/vanir/blob/main/vanir/code_extractors/code_extractor_git.py) is used during signature generation, and by default does git clones all the time which takes a lot of time. Considering we already do [clones](https://github.com/google/osv.dev/blob/master/gcp/workers/worker/oss_fuzz.py#L323) in the worker, we probably can implement our own code extractor to improve the performance.


gcp/workers/worker/worker.py
```diff
@@ -498,6 +531,11 @@ def _do_update(self, source_repo, repo, vulnerability, relative_path,
 
     filter_unsupported_ecosystems(vulnerability)
 
+    # Generate Vanir signatures for test instance.
+    project = osv.utils.get_google_cloud_project()
+    if project and project in ('oss-vdb-test', 'test-osv'):
```
michaelkedar: nit:
```suggestion
    if project in ('oss-vdb-test', 'test-osv'):
```
evaluates to False if `project` is `''` (or falsy)

michaelkedar (dismissed):
LGTM

cuixq:
@doryiii FYI

michaelkedar (approved):

PR #3861, opened 2025-09-02
chore: upgrade ubuntu version to 24.04

cuixq:


osv/osv-schema
another-rex, line 1: Did you mean to update osv-schema?

cuixq: I think my local copy was still at v1.7.3 and the submodule has been updated in https://github.com/google/osv.dev/pull/3855

another-rex (approved):

PR #3862, opened 2025-09-02
chore: An attempt at fixing all linting issues in vulnfeeds

another-rex:
Fixes: #734
Closes: #3164

### DIFF
--- a/bindings/go/osvdevexperimental/paging.go
+++ b/bindings/go/osvdevexperimental/paging.go
@@ -1,3 +1,4 @@
+// Package osvdevexperimental contains experimental helpers for querying osv.dev
 package osvdevexperimental
 
 import (
@@ -69,7 +70,9 @@ func BatchQueryPaging(ctx context.Context, c OSVClientInterface, queries []*osvd
 	}
 	// --- Paging logic ---
 	var errToReturn error
+	//nolint:prealloc
 	var nextPageQueries []*osvdev.Query
+	//nolint:prealloc
 	var nextPageIndexMap []int
 	for i, res := range batchResp.Results {
 		if res.NextPageToken == "" {
--- a/bindings/go/osvdevexperimental/paging_test.go
+++ b/bindings/go/osvdevexperimental/paging_test.go
@@ -3,10 +3,10 @@ package osvdevexperimental
 import (
 	"context"
 	"errors"
-	"github.com/ossf/osv-schema/bindings/go/osvschema"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
+	"github.com/ossf/osv-schema/bindings/go/osvschema"
 	"osv.dev/bindings/go/osvdev"
 )
 
@@ -15,7 +15,7 @@ type mockOSVClient struct {
 	batchQueryReponses map[string]*osvdev.BatchedResponse
 }
 
-func (m *mockOSVClient) Query(ctx context.Context, query *osvdev.Query) (*osvdev.Response, error) {
+func (m *mockOSVClient) Query(_ context.Context, query *osvdev.Query) (*osvdev.Response, error) {
 	key := query.PageToken
 	if key == "" {
 		key = "first"
@@ -24,10 +24,11 @@ func (m *mockOSVClient) Query(ctx context.Context, query *osvdev.Query) (*osvdev
 	if !ok {
 		return nil, errors.New("no response for query")
 	}
+
 	return resp, nil
 }
 
-func (m *mockOSVClient) QueryBatch(ctx context.Context, queries []*osvdev.Query) (*osvdev.BatchedResponse, error) {
+func (m *mockOSVClient) QueryBatch(_ context.Context, queries []*osvdev.Query) (*osvdev.BatchedResponse, error) {
 	// For simplicity, we'll just use the first query's page token to determine the response.
 	key := ""
 	if len(queries) > 0 {
@@ -41,6 +42,7 @@ func (m *mockOSVClient) QueryBatch(ctx context.Context, queries []*osvdev.Query)
 	if !ok {
 		return nil, errors.New("no response for batch query")
 	}
+
 	return resp, nil
 }
 
--- a/tools/indexer-api-caller/caller.go
+++ b/tools/indexer-api-caller/caller.go
@@ -18,14 +18,14 @@ var (
 	repoDir   = flag.String("lib", "", "library directory")
 	repoDir2  = flag.String("lib2", "", "specify another directory to compare file hashes to the first")
 	searchDir = flag.String("dir", "", "third party directory containing multiple libraries")
-	fileExts  = []string{
-		".hpp",
-		".h",
-		".hh",
-		".cc",
-		".c",
-		".cpp",
-	}
+	//fileExts  = []string{
+	//	".hpp",
+	//	".h",
+	//	".hh",
+	//	".cc",
+	//	".c",
+	//	".cpp",
+	//}
 )
 
 type Hash = [16]byte
@@ -139,7 +139,7 @@ func buildGit(repoDir string) ([]*FileResult, error) {
 
 	res, err := http.Post("https://api.osv.dev/v1experimental/determineversion", "application/json", strings.NewReader(b.String()))
 	if err != nil {
-		return nil, fmt.Errorf("Failed to make request: %v", err)
+		return nil, fmt.Errorf("failed to make request: %w", err)
 	}
 
 	output, err := io.ReadAll(res.Body)
--- a/vulnfeeds/cmd/combine-to-osv/main.go
+++ b/vulnfeeds/cmd/combine-to-osv/main.go
@@ -1,3 +1,4 @@
+// package main combines CVEs and security advisories into OSV records.
 package main
 
 import (
@@ -96,8 +97,8 @@ func loadInnerParts(innerPartInputPath string, output map[cves.CVEID][]vulns.Pac
 		}
 
 		// Turns CVE-2022-12345.alpine.json into CVE-2022-12345
-		cveId := cves.CVEID(strings.Split(entryInner.Name(), ".")[0])
-		output[cveId] = append(output[cveId], pkgInfos...)
+		cveID := cves.CVEID(strings.Split(entryInner.Name(), ".")[0])
+		output[cveID] = append(output[cveID], pkgInfos...)
 
 		Logger.Infof(
 			"Loaded Item: %s", entryInner.Name())
@@ -108,9 +109,9 @@ func loadInnerParts(innerPartInputPath string, output map[cves.CVEID][]vulns.Pac
 			Logger.Warnf("Failed to get modified time of %s: %s", filePath, err)
 			continue
 		}
-		existingDate, exists := cvePartsModifiedTime[cveId]
+		existingDate, exists := cvePartsModifiedTime[cveID]
 		if !exists || modifiedTime.After(existingDate) {
-			cvePartsModifiedTime[cveId] = modifiedTime
+			cvePartsModifiedTime[cveID] = modifiedTime
 		}
 	}
 }
@@ -152,11 +153,11 @@ func loadParts(partsInputPath string) (map[cves.CVEID][]vulns.PackageInfo, map[c
 func combineIntoOSV(loadedCves map[cves.CVEID]cves.Vulnerability, allParts map[cves.CVEID][]vulns.PackageInfo, cveList string, cvePartsModifiedTime map[cves.CVEID]time.Time) map[cves.CVEID]*vulns.Vulnerability {
 	Logger.Infof("Begin writing OSV files from %d parts", len(allParts))
 	convertedCves := map[cves.CVEID]*vulns.Vulnerability{}
-	for cveId, cve := range loadedCves {
-		if len(allParts[cveId]) == 0 {
+	for cveID, cve := range loadedCves {
+		if len(allParts[cveID]) == 0 {
 			continue
 		}
-		convertedCve := vulns.FromNVDCVE(cveId, cve.CVE)
+		convertedCve := vulns.FromNVDCVE(cveID, cve.CVE)
 		if len(cveList) > 0 {
 			// Best-effort attempt to mark a disputed CVE as withdrawn.
 			modified, err := vulns.CVEIsDisputed(convertedCve, cveList)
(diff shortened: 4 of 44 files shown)


bindings/go/osvdevexperimental/paging.go
```diff
@@ -69,7 +70,9 @@ func BatchQueryPaging(ctx context.Context, c OSVClientInterface, queries []*osvd
 	}
 	// --- Paging logic ---
 	var errToReturn error
+	//nolint:prealloc
 	var nextPageQueries []*osvdev.Query
+	//nolint:prealloc
 	var nextPageIndexMap []int
```
michaelkedar, line 76: I think these could be `:= make([]type, 0, len(batchResp.Results))` instead of using nolint?

vulnfeeds/cmd/cpe-repo-gen/main.go
```diff
@@ -101,7 +100,7 @@ var DebianCopyrightDenylist = []VendorProduct{
 }
 
 // Helper for JSON rendering of a map with a struct key.
-func (vp VendorProduct) MarshalText() (text []byte, err error) {
+func (vp VendorProduct) MarshalText() (text []byte, err error) { //nolint:unparam
 	return []byte(vp.Vendor + ":" + vp.Product), nil
```
michaelkedar: Is this nolint for this?
```suggestion
func (vp VendorProduct) MarshalText() ([]byte, error) {
	return []byte(vp.Vendor + ":" + vp.Product), nil
```

vulnfeeds/cmd/cpe-repo-gen/main.go
```diff
@@ -197,18 +203,19 @@ func outputDescriptionFrequency(df map[string]int, f io.Writer) error {
 		fmt.Printf("Seen %d reference descriptions\n", len(df))
 	}
 	Logger.Infof("Seen %d distinct reference descriptions", len(df))
+
 	return nil
 }
 
 // Checks if a URL relates to the FSF.
-func IsGNUURL(url string) bool {
+func IsGNUURL(u string) bool {
```
michaelkedar, line 211: not sure why this is preferred :shrug:

bindings/go/osvdevexperimental/paging_test.go
```diff
@@ -15,7 +15,7 @@ type mockOSVClient struct {
 	batchQueryReponses map[string]*osvdev.BatchedResponse
 }
 
-func (m *mockOSVClient) Query(ctx context.Context, query *osvdev.Query) (*osvdev.Response, error) {
+func (m *mockOSVClient) Query(_ context.Context, query *osvdev.Query) (*osvdev.Response, error) {
```
jess-lowe, line 18: out of curiosity - why is this replaced with '_'?  Is it because the class demands it, but it isn't used?

vulnfeeds/cmd/cvelist2osv/main.go
```diff
@@ -29,12 +29,12 @@ var Logger utility.LoggerWrapper
...
-	Repos          []string                        // A list of repositories extracted from the CVE's references.
-	RefTypesCount  map[osvschema.ReferenceType]int // A count of each type of reference found.
-	VersionSources []VersionSource                 // A list of the ways the versions were extracted
-	Notes          []string                        // A collection of notes and warnings generated during conversion.
+	CNA            string                          `json:"cna"`             // The CNA that assigned the CVE.
+	Outcome        string                          `json:"outcome"`         // The final outcome of the conversion (e.g., "Successful", "Failed").
+	Repos          []string                        `json:"repos"`           // A list of repositories extracted from the CVE's references.
+	RefTypesCount  map[osvschema.ReferenceType]int `json:"ref_types_count"` // A count of each type of reference found.
+	VersionSources []VersionSource                 `json:"version_sources"` // A list of the ways the versions were extracted
+	Notes          []string                        `json:"notes"`           // A collection of notes and warnings generated during conversion.
```
michaelkedar, line 37: Does this change the output?
Before the change was it e.g.
```json
{
  "CNA": "...",
  "VersionSources": []
}
```
(I don't know the default behaviour of go's JSON marshaller)
Does it matter?

tools/indexer-api-caller/caller.go
```diff
@@ -18,14 +18,14 @@ var (
...
-	fileExts  = []string{
-		".hpp",
-		".h",
-		".hh",
-		".cc",
-		".c",
-		".cpp",
-	}
+	//fileExts  = []string{
+	//	".hpp",
```
jess-lowe, line 22: why not just delete?

bindings/go/osvdevexperimental/paging_test.go
another-rex: Yes, the interface it's implementing requires a context, but since this is a mock, we don't really need to care about the context, so just mark it as unused.

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -164,38 +166,39 @@ func combineIntoOSV(loadedCves map[cves.CVEID]cves.Vulnerability, allParts map[c
 			}
 			if err == nil && !modified.IsZero() {
 				convertedCve.Withdrawn = modified
-				convertedCve.DatabaseSpecific = make(map[string]interface{})
+				convertedCve.DatabaseSpecific = make(map[string]any)
```
jess-lowe, line 168: is 'any' better practise??

vulnfeeds/cmd/cvelist2osv/main_test.go
```diff
@@ -32,8 +33,9 @@ func loadTestCVE(path string) cves.CVE5 {
 	var cve cves.CVE5
 	err = json.NewDecoder(file).Decode(&cve)
 	if err != nil {
-		log.Fatalf("Failed to decode %q: %+v", path, err)
+		panic(fmt.Sprintf("Failed to decode %q: %+v", path, err))
```
michaelkedar: 1. Why is this changed to a panic, but not the `log.Fatalf` in the same function above?
2. Maybe this should be a `t.Helper()` function, and use `t.Fatalf`?

vulnfeeds/cmd/combine-to-osv/main.go
another-rex: Yep, just looks clearer, it was added in 1.18 or something, but it's exactly the same under the hood.

https://cs.opensource.google/go/go/+/refs/tags/go1.23.0:src/builtin/builtin.go;l=97

vulnfeeds/cmd/debian/debian_security_tracker.go
```diff
@@ -1,3 +1,4 @@
+// package main contains the data structures for the Debian Security Tracker.
 package main
```
michaelkedar: :thinking:

tools/indexer-api-caller/caller.go
another-rex: I didn't want to touch the determineversion stuff too much. I think it's still useful to have around to hack on later.

vulnfeeds/cmd/download-cves/main.go
```diff
@@ -39,41 +41,44 @@ func main() {
...
+	if apiKey != "" {
+		// apiKey is the correct header type that NVD expects
+		// https://nvd.nist.gov/developers/start-here
+		//nolint:canonicalheader
+		req.Header.Add("apiKey", apiKey)
 	}
 	backoff := retry.NewExponential(6 * time.Second)
-	if err := retry.Do(context.Background(), retry.WithMaxRetries(3, backoff), func(ctx context.Context) error {
+	if err := retry.Do(context.Background(), retry.WithMaxRetries(3, backoff), func(_ context.Context) error {
 		resp, err := client.Do(req)
```
michaelkedar: Maybe we should be using this context?
```suggestion
	if err := retry.Do(context.Background(), retry.WithMaxRetries(3, backoff), func(ctx context.Context) error {
		req := req.WithContext(ctx)
		resp, err := client.Do(req)
```

vulnfeeds/cves/versions.go
```diff
@@ -139,6 +141,7 @@ func Repo(u string) (string, error) {
 				nil
 		}
 		// GitLab can have a deeper structure to a repo (projects can be within nested groups)
+		//nolint:staticcheck
```
michaelkedar, line 144: Do I want to know what it would take to fix this without the `//nolint`?

vulnfeeds/cves/versions.go
```diff
@@ -478,8 +478,8 @@ func ValidateAndCanonicalizeLink(link string, httpClient *http.Client) (canonica
 		return link, err
 	}
 	backoff := retry.NewExponential(1 * time.Second)
-	if err := retry.Do(context.Background(), retry.WithMaxRetries(3, backoff), func(ctx context.Context) error {
-		req, err := http.NewRequest("HEAD", link, nil)
+	if err := retry.Do(context.Background(), retry.WithMaxRetries(3, backoff), func(_ context.Context) error {
+		req, err := http.NewRequest(http.MethodHead, link, nil)
```
michaelkedar:
```suggestion
	if err := retry.Do(context.Background(), retry.WithMaxRetries(3, backoff), func(ctx context.Context) error {
		req, err := http.NewRequestWithContext(ctx, http.MethodHead, link, nil)
```

vulnfeeds/cves/versions.go
```diff
@@ -1047,14 +1056,15 @@ func ReposFromReferencesCVEList(CVE string, refs []Reference, tagDenyList []stri
 			continue
 		}
 		// If the reference is a commit URL, the repo is inherently useful (but only if the repo still ultimately works).
-		_, err = Commit(ref.Url)
+		_, _ = Commit(ref.URL)
```
michaelkedar: What is this line even doing?

vulnfeeds/faulttolerant/http.go
```diff
@@ -9,40 +10,21 @@ import (
...
-			return retry.RetryableError(fmt.Errorf("bad response: %v", r.StatusCode))
-		default:
-			resp = r
-			return nil
-		}
-	}); err != nil {
-		return nil, fmt.Errorf("fail: %q: %v", url, err)
-	}
-	return resp, err
+	return req(url, "GET")
```
michaelkedar:
```suggestion
	return req(url, http.MethodGet)
```

vulnfeeds/faulttolerant/http.go
```diff
@@ -9,40 +10,21 @@ import (
...
 }
 
-// Make a HTTP HEAD request for url and retry 3 times, with an exponential backoff.
+// Head makes a HTTP HEAD request for url and retry 3 times, with an exponential backoff.
 func Head(url string) (resp *http.Response, err error) {
+	return req(url, "HEAD")
```
michaelkedar:
```suggestion
	return req(url, http.MethodHead)
```

vulnfeeds/faulttolerant/http.go
```diff
@@ -9,40 +10,21 @@ import (
...
+}
+
+// Get makes a HTTP GET request for url and retry 3 times, with an exponential backoff.
+func req(url, method string) (resp *http.Response, err error) {
```
michaelkedar: Something like
```suggestion
// req makes a HTTP request for url and retry 3 times, with an exponential backoff.
func req(url, method string) (resp *http.Response, err error) {
```

vulnfeeds/faulttolerant/http.go
```diff
@@ -51,19 +33,21 @@ func Head(url string) (resp *http.Response, err error) {
 		if err != nil {
 			return err
 		}
-		defer r.Body.Close()
```
michaelkedar, line 54: Do we need to make sure the calling function is calling `r.Body.Close()`?

vulnfeeds/git/versions.go
```diff
@@ -34,7 +34,7 @@ func fuzzyVersionToCommit(normalizedVersion string, repo string, commitType mode
 	// Keep in sync with the regex in models.NormalizeVersion()
 	var validVersionText = regexp.MustCompile(`(?i)(?:rc|alpha|beta|preview)\d*`)
 
-	for _, k := range maps.Keys(normalizedTags) {
+	for k := range maps.Keys(normalizedTags) {
```
michaelkedar: can't this just be 
```suggestion
	for k, _ := range normalizedTags {
```

vulnfeeds/cmd/cpe-repo-gen/main.go
another-rex: url overlaps with a package name.

vulnfeeds/cmd/cpe-repo-gen/main.go
another-rex: Huh, that is weird

vulnfeeds/cmd/cvelist2osv/main_test.go
another-rex: I switched to log.Panicf() now. the reason the linter gave is that Fatalf doesn't run the defer functions, so things that should be closed doesn't get closed.

another-rex: But yeah, definitely should be t.Helper with t.Fatalf, since in tests closing doesn't really matter

vulnfeeds/git/versions.go
another-rex: yeah

vulnfeeds/faulttolerant/http.go
another-rex: Yep, and both calling functions does call close

bindings/go/osvdevexperimental/paging.go
another-rex: Well, you don't know whether the next page token has any next pages for each batch result, and I'm guessing most of the time most queries don't have the next page token, and only a few do, so I changed this to no prealloc. Probably doesn't matter too much either way.

vulnfeeds/cmd/cpe-repo-gen/main.go
another-rex: Nope, turns out the unparam is for the fact that error return value will always be nil.

another-rex: Removed the named return variables anyway.

vulnfeeds/cves/versions.go
another-rex: yeah I think the current version is much clearer. I.e.

Not this or this or this or this...

seems clearer than 

not this, and not this, and not this, and not this

vulnfeeds/cmd/cvelist2osv/main.go
another-rex: "changes it but should be fine" - Jess

vulnfeeds/cves/versions.go
another-rex: No idea... Might just remove it.

another-rex: Actually, I think this should have a err continue here.

michaelkedar (approved):

PR #3863, opened 2025-09-02
fix(vulnfeeds): use new CPE Dictionary JSON Zips instead of old XML dict

### LINKED ISSUE #3847: Update cpe-repo-gen to use new CPE dictionary source
CPE Dictionary from NVD has been moved, and changed from XML format to JSON. 

New link: [ZIP](https://nvd.nist.gov/feeds/json/cpe/2.0/nvdcpe-2.0.zip)

It also seems to be chunked now, rather than one large file. 

Relevant code in https://github.com/google/osv.dev/tree/master/vulnfeeds/cmd/cpe-repo-gen

jess-lowe:
Closes #3847


vulnfeeds/cmd/cpe-repo-gen/cpe-repo-gen_map.sh
```diff
@@ -38,9 +43,10 @@ if [[ -n "${DEBIAN_COPYRIGHT_GCS_PATH}" ]]; then
   MAYBE_USE_DEBIAN_COPYRIGHT_METADATA="--debian_metadata_path ${WORK_DIR}/metadata.ftp-master.debian.org"
 fi
 
-/cpe-repo-gen \
-  --cpe_dictionary "${WORK_DIR}/official-cpe-dictionary_v2.3.xml" \
+./cpe-repo-gen \
```
michaelkedar, line 46: This probably shouldn't have been changed
```suggestion
/cpe-repo-gen \
```

vulnfeeds/cmd/cpe-repo-gen/main.go
```diff
@@ -109,37 +123,58 @@ func (vp VendorProduct) MarshalText() (text []byte, err error) {
...
+			for _, r := range p.CPE.References {
+				refs = append(refs, Reference{URL: r.URL, Description: r.Type})
+			}
 
-	return c, nil
+			cpeItems = append(cpeItems, CPEItem{
+				Name:       p.CPE.CPEName,
+				Deprecated: p.CPE.Deprecated,
+				References: refs,
+			})
```
michaelkedar: `Title` is not being written - I thought it was possible to get this from the JSON (though is it used by us?)

vulnfeeds/cmd/cpe-repo-gen/cpe-repo-gen_map.sh
michaelkedar: actually, I think either works?

vulnfeeds/cmd/cpe-repo-gen/main.go
jess-lowe: I don't think we use it, hence we don't write it

another-rex: If it is not used, should we remove it from `CPEItem`?

Issue #3850, opened 2025-08-28
Improve the OSV.dev database situation


michaelkedar:
Tracking issue for database improvements.
In short, the `Bug` Datastore entity is unwieldy and affecting our ability to make performance improvements. We'd also like to move away from our dependence on Datastore.
We want to store full vulnerabilities in GCS, with smaller query indexes in Datastore to improve things.
Additionally, we want to separate vulnerability ingestion from enrichment (e.g. computing affected versions and commits), to reduce the delay between vulns being published upstream and them being available in OSV.dev.

| Task | Test | Prod |
| - | - | - |
| Create & start populating new Datastore entities and GCS objects | #3708 | TODO |
| Ensure consistency between GCS and Datastore entities | #3821 | TODO |
| Migrate API to new entities | #3776 | TODO |
| Migrate website to new entities | TODO | TODO |
| Change importer to create entities before sending tasks to worker | TODO | TODO |
| Change exporter to read/write from new entities | TODO | TODO |
| Change alias/upstream cron to read from new entities | TODO | TODO |
| Split OSS-Fuzz-specific handling away from main importer/worker | N/A |  TODO |

michaelkedar:
This is largely complete now

PR #3892, opened 2025-09-03
refactor: ecosystem helpers

### LINKED ISSUE #3063: Switch all ecosystem version queries to range match
We currently have three ways of handling API version queries:

- **For semver versions**: We first match against the affected version range ([using normalization](https://github.com/google/osv.dev/blob/22a535c6abbc67d35e6499f845b6152c92629b27/gcp/api/server.py#L1045C1-L1046C59)) and then apply [generic version matching](https://github.com/google/osv.dev/blob/22a535c6abbc67d35e6499f845b6152c92629b27/gcp/api/server.py#L1201C24-L1201C49) (affected versions match) to refine the results.
- For non-semver versions **with comparison support**: We pull all related packages from db, check their affected versions, and then compare them with the affected version ranges.
- For non-semver versions **without comparison support**: We only pull packages that contain the currently queried version within their affected versions (using [`affected_fuzz`](https://github.com/google/osv.dev/blob/22a535c6abbc67d35e6499f845b6152c92629b27/osv/models.py#L283)).

Both semver version query and range comparison query methods match against affected versions and affected ranges, and providing the best coverage. We should use the range comparison method for all non-semver ecosystems, as it offers accurate results (also it doesn't slow down much):
- [ ] Alpine
- [ ] AlmaLinux
- [ ] Bioconductor
- [ ] CRAN
- [ ] Debian
- [ ] Chainguard
- [ ] GHC
- [ ] Hackage
- [ ] Maven
- [ ] NuGet
- [ ] openSUSE
- [ ] Packagist
- [ ] Pub
- [ ] PyPI
- [ ] Red Hat
- [ ] Rocky Linux
- [ ] RubyGems
- [ ] SUSE
- [ ] Ubuntu
- [ ] Wolfi

We could also potentially apply the version comparison method to the semver ecosystem to unify our query methods and simplify our code logic. This needs more testing to assess its impact on performance and accuracy.
- [ ]  semver version (unsure for this one)

### LINKED ISSUE #3431: Unify APK ecosystem helpers
Alpine, Chainguard, Wolfi, and now MinimOS (#3430), all use the Alpine Package Keeper (APK) for their packages.

Currently, Chainguard, Wolfi, and MinimOS all have identical ecosystem helper implementations, since we don't support version enumeration for these. These should be combined into one class.

Alpine implements `enumerate_versions`, and needs to distinguish between Alpine releases, so it probably can stay as its own thing (but maybe could inherit from the others).

### LINKED ISSUE #3745: Refactor Ecosystem helper interface
The [Ecosystem class](https://github.com/google/osv.dev/blob/a179a8e75062733dd4d24d24532548093294684a/osv/ecosystems/helper_base.py#L29) is confusing to use:
- It has abstract methods `sort_key` and `enumerate_versions` that are not required to be implemented (but must be overwritten with a function that raises a `NotImplementedError`)
- To check whether these are implemented, the respective (and poorly named) `supports_comparing` and `supports_ordering` properties must be manually implemented & checked.

In my opinion, it would be better to have a few interfaces:
- `Ecosystem()`, a replacement for the current `OrderingUnsupportedEcosystem`, with whatever base things are needed
- `ComparableEcosystem(Ecosystem, ABC)`, which are ecosystems which we can create orderings of version strings (i.e. which implement `sort_key`)
- `EnumerableEcosystem(ComparableEcosystem)`, which are ecosystems which we can find all concrete versions for (i.e. which implement `enumerate_versions`)

With this, we could replace the `supports_x` properties with `isinstance` calls. It also would make reusing logic for ecosystems which use the same versioning scheme but have different enumerations (e.g. APK/RPM ecosystems #3431)

michaelkedar:
Closes #3745
- Refactored the ecosystem base class to split it into two: `OrderedEcosystem` (for ones that implement sort_key only) and `EnumerableEcosystem`  (which has enumerate versions)
- Removed a few unused / unnecessary methods that were on the old `Ecosystem` class
  - kept `next_version` (but marked it as deprecated) since only malicious packages uses it to convert some ghsa records.
- Got rid of `OrderingUnsupportedEcosystem` since it's no longer particularly useful.
- Consolidated APK-, RPM-, and dpkg-based helpers under a single helper (closes #3431)

I also updated the copyright year on all the ecoystems py files :)

I think this also incidentally closes #3063, since `supports_comparing` is no longer used (it will be superseded by #3850 anyway).

### DIFF
--- a/osv/ecosystems/_ecosystems.py
+++ b/osv/ecosystems/_ecosystems.py
@@ -121,7 +121,10 @@ _OSV_TO_DEPS_ECOSYSTEMS_MAP = {
 def get(name: str) -> OrderedEcosystem | EnumerableEcosystem | None:
   """Get ecosystem helpers for a given ecosystem."""
   name, _, suffix = name.partition(':')
-  return None if (e := _ecosystems.get(name)) is None else e(suffix)
+  ecosys = _ecosystems.get(name)
+  if ecosys is None:
+    return None
+  return ecosys(suffix)
 
 
 def normalize(ecosystem_name: str):


osv/ecosystems/_ecosystems.py
```diff
@@ -117,50 +118,10 @@
...
-
-  if name.startswith('openEuler'):
-    return OpenEuler()
-
-  if name.startswith('SUSE'):
-    return SUSE()
-
-  return _ecosystems.get(normalize(name))
+  name, _, suffix = name.partition(':')
+  return None if (e := _ecosystems.get(name)) is None else e(suffix)
```
another-rex: nit: Let's just expand this out to a full if statement for clarity

osv/ecosystems/_ecosystems.py
```diff
@@ -15,78 +15,79 @@
...
+    'PyPI': PyPI,
+    'Red Hat': RPM,
+    'Rocky Linux': RPM,
+    'RubyGems': RubyGems,
+    'SUSE': RPM,
+    'SwiftURL': SemverEcosystem,
+    'Ubuntu': Ubuntu,
+    'Wolfi': APK,
+
+    # Ecosystems known in the schema, but without implementations.
```
another-rex, line 68: Hmm... is there a test we can add here that reads the ecosystems.json in the osv-schema repo and checks it against this.?

osv/ecosystems/bioconductor_test.py
```diff
@@ -24,7 +25,10 @@ class BioconductorEcosystemTest(vcr.unittest.VCRTestCase):
   def test_next_version(self):
     """Test next_version."""
     ecosystem = ecosystems.get('Bioconductor')
-    self.assertEqual('1.18.0', ecosystem.next_version('a4', '1.16.0'))
-    self.assertEqual('1.20.0', ecosystem.next_version('a4', '1.18.0'))
-    with self.assertRaises(ecosystems.EnumerateError):
-      ecosystem.next_version('doesnotexist123456', '1')
+    with warnings.catch_warnings():
```
another-rex, line 28: did not know warnings was a thing, wow.

another-rex (approved):
This is great!

cuixq (approved):
:tada:

jess-lowe (dismissed):

osv/ecosystems/_ecosystems.py
michaelkedar: I remember bringing this up a while ago - https://github.com/google/osv.dev/pull/2615#discussion_r1756215854

Not sure if you agree/disagree with Andrew's reasoning here. With renovate automatically updating the osv-schema it probably would always fail the tests.

osv/ecosystems/_ecosystems.py
michaelkedar: expanded

osv/ecosystems/_ecosystems.py
another-rex: That should be fine though, with the recent ecosystems, they made both PRs (osv-schema and osv.dev _ecosystem.py) simultaneously. So we just need to merge the ecosystems.py PR before updating the schema to the newest version and it should work right?

michaelkedar: We currently have a test that checks that all the ecosystems in `_ecosystems.py` are valid ecosystems in the schema. Adding the vice versa test (that all schema ecosystems are in `_ecosystems.py`) would mean we'd always have to update both at the same time.

another-rex: Ah I see..., let's leave it as it is then.

jess-lowe (approved):

PR #3897, opened 2025-09-04
refactor(tests): Use context manager, firestore emulator for datastore emulation

michaelkedar:
Does a few things:
1. Changed testing code to use firestore emulator (in datastore mode) instead of the older datastore emulator (in datastore-in-firestore mode), which is newer and seems to be the more preferred emulator to run.
    - **This will require devs to install the gcloud firestore emulator.** I've added a check/runtime error which should make this clearer.
2. Changed the emulator start/stop functions to instead use a context (`with datastore_emulator():`), so stopping the emulator will be handled automatically when the context exits.
3. Made all the unittests use the `enterContext` family of functions to start the emulator, instead of doing it in `__main__` or manually calling `__enter__` / `__exit__` on the context manages in `setUp`/ `tearDown`

One small issue is that the firestore emulator cannot handle writing too many indexed properties (e.g. AffectedCommits, which has up to 10000 indexed commits). I've worked around this limitation in the tests that hit it.


osv/tests.py
```diff
@@ -108,170 +108,108 @@ def commit(self, author_name, author_email, message='Changes'):
...
+    ],
+                                     text=True,
+                                     stderr=subprocess.DEVNULL)
+    return 'Not Installed' not in output
+  except (subprocess.CalledProcessError, FileNotFoundError):
+    return False
+
+
+@contextlib.contextmanager
+def datastore_emulator():
```
another-rex, line 138: should we name this firestore emulator to avoid confusion? Or would that be confusing becuase it's firestore in datastore mode.

another-rex (approved):
Nice!

osv/tests.py
michaelkedar: eh I don't think so - it's still an emulator for datastore, just running on the firestore emulator (just like our database is 'Firestore with Datastore compatibility')

PR #3898, opened 2025-09-04
docs: mention bindings directory in README

cuixq:
The `bindings/` directory was not documented in the main `README.md` file.
  
This change adds an entry for it to the directory table, making the repository overview more complete for anyone navigating the project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,7 @@ consists of:
 
 | directory       | what |
 |-----------------|------|
+| `bindings/`     | Language bindings for the OSV API (currently Go only) |
 | `deployment/`   | Terraform & Cloud Deploy config files <br /> A few Cloud Build config yamls |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -36,6 +36,7 @@ consists of:
 
 | directory       | what |
 |-----------------|------|
+| `bindings/`     | Language bindings for the OSV API (currently Go only) |
```
jess-lowe: Is it worth explicitly stating that these don't include the OSV Schema bindings, or is it clear enough with 'API'?

michaelkedar (dismissed):

README.md
cuixq: I think mentioning the OSV schema bindings may confuse people - to make it clear that these are the bindings for API only, I attach a link to OSV API documentation.

PR #3903, opened 2025-09-06
refactor(api): Normalize Git repo URLs for protocol matching

### LINKED ISSUE #3830: normalize/remove protocols on GIT repos
e.g. [CVE-2017-15650](https://osv.dev/vulnerability/CVE-2017-15650) uses **http:**//git.musl-libc.org/git/musl while [CVE-2025-26519](https://osv.dev/vulnerability/CVE-2025-26519) uses **https:**//git.musl-libc.org/git/musl
(I think we also have records that use git:// or don't have the protocol at all)

Doing an API query by tag will only match if the protocol matches exactly.

We should probably remove the protocol when we're matching against repos to allow us to match these.

ashmod:
Added protocol normalization at the API query level by creating `_normalize_git_repo_url` that removes protocol/scheme for comparison. Updated GIT ecosystem matching logic to use normalized URLs.

Now, for example, both `http://git.musl-libc.org/git/musl` and `https://git.musl-libc.org/git/musl` will normalize to just `git.musl-libc.org/git/musl` before being matched accordingly.

Fixes #3830

### DIFF
--- a/gcp/api/server.py
+++ b/gcp/api/server.py
@@ -113,8 +113,8 @@ _ndb_client = ndb.Client()
 def _normalize_git_repo_url(repo_url: str) -> str:
   """Normalize git repository URL for matching by removing protocol/scheme.
   
-  This enables matching git repositories regardless of whether they use http, https,
-  git, or other protocols. For example:
+  This enables matching git repositories regardless of whether they use
+  http, https, git, or other protocols. For example:
   - http://git.musl-libc.org/git/musl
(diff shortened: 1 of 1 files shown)


michaelkedar:
Hey, thanks for doing this.
I don't think this completely solves the issue - they're still stored & queried from the database with the scheme included. Fixing that is a bit more complicated and overlaps with #3850, so I'll take a look into it.

I'll reopen #3830 just so we don't lose track of this.

PR #3904, opened 2025-09-08
fix(protos): Fixed missing 'GIT' reference type

### LINKED ISSUE #3902: Malformed OSV records
**Describe the bug**
CVE-OSV conversion tooling seems to have an issue where it generates malformed OSV advisories when validated against https://github.com/ossf/osv-schema/blob/main/validation/schema.json

The `type` field of some of the references is missing.

**To Reproduce**
The first reference here shows the issue
https://api.osv.dev/v1/vulns/CVE-2024-57392

it's not missing here though
https://storage.googleapis.com/cve-osv-conversion/osv-output/CVE-2024-57392.json

**Additional context**
This was noticed in CVE-2018-11212 and [CVE-2024-57392](https://osv.dev/vulnerability/CVE-2024-57392) when downloading the Debian ZIP file.

jess-lowe:
'GIT' reference type was missing from the vulnerability protos, causing the reference type to be dropped when ingested by the importer. 

Fixes #3902


michaelkedar (changes requested):
You need to change this in [vulnerability.proto](https://github.com/google/osv.dev/blob/96958296f6f268aa5556b726c27be1a965e36534/osv/vulnerability.proto#L155), then [rebuild the protos](https://github.com/google/osv.dev/blob/96958296f6f268aa5556b726c27be1a965e36534/osv/build_protos.sh)

michaelkedar (approved):
I feel like the GIT reference type was added to the validation schema by accident, since it's not defined in [the schema](https://ossf.github.io/osv-schema/#references-field) (maybe from a misreading of the `FIX` field description, which mentions `GIT` in affected ranges)

I'm okay with adding this, since it seems to have been in the validation for a while. But since it doesn't have a defined meaning, I think we should avoid using it ourselves in the Debian converter code.

jess-lowe:
> I feel like the GIT reference type was added to the validation schema by accident, since it's not defined in [the schema](https://ossf.github.io/osv-schema/#references-field) (maybe from a misreading of the `FIX` field description, which mentions `GIT` in affected ranges)
> 
> I'm okay with adding this, since it seems to have been in the validation for a while. But since it doesn't have a defined meaning, I think we should avoid using it ourselves in the Debian converter code.

I agree - I've updated the PR to return 'Package' instead of 'GIT' in line with how GitHub links are treated

michaelkedar (approved):

PR #3905, opened 2025-09-08
feat(vulnfeeds): CVEList cleanups and batch run scripts

jess-lowe:
Did some cleanup:
- forced semver versions to have a patch version (0)
- Removed unnecessary database specific declarations


vulnfeeds/cmd/cve-bulk-converter/main.go
```diff
@@ -0,0 +1,114 @@
...
+	flag.Parse()
+	var logCleanup = logger.InitGlobalLogger("cvelist-osv", false)
+	defer logCleanup()
+	// --- Main execution ---
+	logger.Infof("Commencing Linux CVE to OSV conversion run")
+
+	repoDir := "cvelistV5"
+
+	// Prepare for processing.
+	localOutputDir := "cvelist2osv"
```
another-rex: These should probably be flags, and the default value can be constants at the top of the file.

vulnfeeds/cmd/cve-bulk-converter/main.go
```diff
@@ -0,0 +1,114 @@
...
+			continue
+		}
+
+		var cve PartialCVE
+		if err := json.Unmarshal(data, &cve); err != nil {
+			logger.Infof("Failed to unmarshal JSON from %s: %v", path, err)
+			continue
+		}
+
+		if cve.Metadata.Assigner == "Linux" && cve.Metadata.State == "PUBLISHED" {
```
another-rex: It's a bit unclear to me that the buck converter only convert Linux records, while the single converter will convert anything. Maybe have this check as a flag to make it explicit.

vulnfeeds/cmd/cve-bulk-converter/main.go
```diff
@@ -0,0 +1,114 @@
...
+	defer wg.Done()
+	for path := range jobs {
+		data, err := os.ReadFile(path)
+		if err != nil {
+			logger.Infof("Failed to read file %s: %v", path, err)
+			continue
+		}
+
+		var cve PartialCVE
+		if err := json.Unmarshal(data, &cve); err != nil {
```
another-rex, line 97: How much time does this actually save? Can you do a quick test with and without the partial JSON to see if this is effective? If it doesn't save too much time maybe just remove the extra step.

vulnfeeds/cmd/cve-bulk-converter/main.go
```diff
@@ -0,0 +1,114 @@
...
+		ID       string `json:"cveId"`
+		Assigner string `json:"assignerShortName"`
+		State    string `json:"state"`
+	} `json:"cveMetadata"`
+}
+
+func main() {
+	// --- Configuration ---
+	years := flag.String("years", "2022,2023,2024,2025", "A comma-separated list of years to process.")
+	workers := flag.Int("workers", 10, "The number of concurrent workers to use for processing CVEs.")
```
another-rex: You might be able to bump this up by quite a bit, depending on how much time is spent reading from disk.

vulnfeeds/cmd/cve-bulk-converter/run-cvelist-converter.sh
```diff
@@ -0,0 +1,74 @@
+#!/bin/bash
+
+# Copyright 2023 Google LLC
```
another-rex: 2023?

vulnfeeds/cmd/cvelist2osv/version_extraction.go
```diff
@@ -242,17 +235,22 @@ func findInverseAffectedRanges(cveAff cves.Affected, cnaAssigner string) (ranges
 	var introduced []string
 	fixed := make([]string, 0, len(cveAff.Versions))
 	for _, vers := range cveAff.Versions {
+		versionValue := vers.Version
 		if vers.Status == "affected" {
-			introduced = append(introduced, vers.Version)
+			if len(strings.Split(versionValue, ".")) != 3 {
```
another-rex: Probably can be more specific to `== 2`. And if it's not == 2 or == 3, then just error out of this record.

vulnfeeds/cmd/cvelist2osv/version_extraction.go
```diff
@@ -428,7 +425,6 @@ func buildVersionRange(intro string, lastAff string, fixed string) osvschema.Ran
 func sortBadSemver(a, b string) int {
```
another-rex: Hmm maybe rename to sortSemverLike? I'm guessing this also sorts good semver versions

vulnfeeds/cmd/cvelist2osv/version_extraction.go
```diff
@@ -439,8 +435,28 @@ func sortBadSemver(a, b string) int {
...
 		}
 	}
+	// If lengths are the same, they're equal.
+	if len(partsA) == len(partsB) {
+		return 0
+	}
+
+	// Determine which version has extra parts and what the result
+	// should be if those parts are non-zero.
+	longerParts := partsB
```
another-rex: Prefer to put this in the else part of the if statement which reads clearer.

PR #3906, opened 2025-09-08
feat(cron): dockerfile and cron job for CVEList conversion

jess-lowe:
merge this after #3905

### DIFF
--- a/deployment/clouddeploy/gke-workers/base/cve5-to-osv.yaml
+++ b/deployment/clouddeploy/gke-workers/base/cve5-to-osv.yaml
@@ -1,0 +1,36 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: cve5-to-osv
+  labels:
+    cronLastSuccessfulTimeMins: "2160"
+spec:
+  timeZone: Australia/Sydney
+  schedule: "0 6,13 * * *"
+  concurrencyPolicy: Forbid
+  jobTemplate:
+    spec:
+      activeDeadlineSeconds: 86400
+      template:
+        spec:
+          containers:
+          - name: cve5-to-osv
+            image: cve5-to-osv
+            imagePullPolicy: Always
+            securityContext:
+              privileged: true
+            resources:
+              requests:
+                cpu: "1"
+                memory: "2G"
+              limits:
+                cpu: "1"
+                memory: "4G"
+            env:
+              - name: WORK_DIR
+                value: /tmp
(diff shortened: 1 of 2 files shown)


vulnfeeds/cmd/cve-converter/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+
+WORKDIR /go/src
+
+COPY go.mod go.sum ./
+RUN go mod download && go mod verify
+
+COPY . .
+RUN CGO_ENABLED=0 go build -v -o /usr/local/bin ./cmd/cvelist2osv ./cmd/cve-converter
+
+FROM gcr.io/google.com/cloudsdktool/google-cloud-cli:alpine@sha256:f2b7bf7556ac1d266c3b4563507d3c7aec1464ec3f6d3cbd114f01011d02f1c3
```
another-rex: I'm not sure if you need to add a comment here about which version this is for renovatebot to pick it up.

vulnfeeds/cmd/cve-converter/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+RUN apk --no-cache add jq
+
+COPY --from=GO_BUILD /usr/local/bin/ ./usr/local/bin/
```
another-rex: This is copying the entire bin dir which could contain a bunch of other unrelated bins. Just select the right bins we need here.

deployment/clouddeploy/gke-workers/base/cve5-to-osv.yaml
```diff
@@ -0,0 +1,36 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: cve5-to-osv
+  labels:
+    cronLastSuccessfulTimeMins: "2160"
+spec:
+  timeZone: Australia/Sydney
+  schedule: "0 6,13 * * *"
```
another-rex: Hmm is this intended to be after 6am and 1pm?

jess-lowe: I think so? this is just copied from the nvd-cve conversion scripts

vulnfeeds/cmd/cve-converter/Dockerfile
jess-lowe: from looking at the other dockerfiles, I'm not sure this is needed?

another-rex (approved):

PR #3941, opened 2025-09-11
fix(cron): Update data dump pulls to 2.0 and revert back to data dump NVD ingestion

### LINKED ISSUE #3933: Consider updating nvd mirror to use NVD data dumps instead of whole API collection
There has occasionally been some hiccups in the nvd-mirror cron job, where the larger nvd-cve JSON file fails to parse properly - leaving a year completely unextracted. 

NVD publishes data dumps by year in the same fashion as we extract at https://nvd.nist.gov/vuln/data-feeds

Maybe we should consider downloading them directly through the data dumps rather that extract them ourselves through the API.

jess-lowe:
Closes #3933

### DIFF
--- a/vulnfeeds/cmd/download-cves/mirror_nvd.sh
+++ b/vulnfeeds/cmd/download-cves/mirror_nvd.sh
@@ -31,7 +31,7 @@ set -e
 
 echo "Downloading the entire NVD"
 mkdir -p "${WORK_DIR}/nvd"
-./download-cves --cvePath "${WORK_DIR}/nvd"
+/usr/local/bin/download-cves --cvePath "${WORK_DIR}/nvd"
 
 echo "Copying files to GCS bucket"
 gcloud config set storage/parallel_composite_upload_enabled True


vulnfeeds/cmd/download-cves/mirror_nvd.sh
```diff
@@ -31,32 +31,7 @@ set -e
...
-        "totalResults": count_matching_cves,
-        "format": .format,
-        "version": .version,
-        "timestamp": .timestamp,
-        "vulnerabilities": .vulnerabilities | map(select(.cve?.id? | startswith("CVE-" + $year + "-")))
-       }' > "${WORK_DIR}/nvd/nvdcve-2.0-${YEAR}.json" &
-done
-
-wait
+./download-cves --cvePath "${WORK_DIR}/nvd"
```
jess-lowe:
```suggestion
/usr/local/bin/download-cves --cvePath "${WORK_DIR}/nvd"
```

PR #3948, opened 2025-09-12
fix(cron): bump memory allocation

jess-lowe:


vulnfeeds/cmd/cve-bulk-converter/run-cvelist-converter.sh
```diff
@@ -50,12 +53,14 @@ fi
...
- --out_dir="${LOCAL_OUT_DIR}"
+ --years="2022,2023,2024,2025" \
+ --out_dir="${LOCAL_OUT_DIR}" \
+ --workers="${NUM_WORKERS}"
 
 # Copy results to staging area.
 echo "Copying CVEList records successfully converted to OSV to aggregated staging"
-find "${LOCAL_OUT_DIR}" -type f -name \*.json -exec cp '{}' "${LOCAL_OUT_DIR}/gcs_stage/"
+find "${LOCAL_OUT_DIR}" -type f -name \*.json \
+  -exec cp '{}' "${LOCAL_OUT_DIR}/gcs_stage/" \;
```
another-rex, line 63: what is this `\;` at the end?

jess-lowe: idk but it was in the cve-to-osv generation bash script and it seems to be broken without it so I'm calling it a lucky semicolon.

another-rex: Oh it marks the end of the -exec part

PR #3951, opened 2025-09-12
feat(vulnfeeds): Enable git resolution for affected versions field CVEList

jess-lowe:
This PR:
- Fixes the metrics notes file concurrency issue
- Enables git resolution for version ranges extracted
- Stores versions converted in the range `database_specific` and stores failed version conversions in the `affected`  `database_specific`
- Enables conversion for GitHub_M scoped CVEs

### DIFF
--- a/vulnfeeds/cmd/cvelist2osv/converter.go
+++ b/vulnfeeds/cmd/cvelist2osv/converter.go
@@ -105,7 +105,7 @@ func FromCVE5(cve cves.CVE5, refs []cves.Reference, metrics *ConversionMetrics) 
 
 	// Try to extract repository URLs from references.
 	repos, repoNotes := cves.ReposFromReferencesCVEList(string(cve.Metadata.CVEID), refs, RefTagDenyList)
-  for _, note := range repoNotes {
+	for _, note := range repoNotes {
 		metrics.AddNote("%s", note)
 	}
 	metrics.Repos = repos
@@ -184,7 +184,7 @@ func ConvertAndExportCVEToOSV(cve cves.CVE5, directory string) error {
 	cveID := cve.Metadata.CVEID
 	cnaAssigner := cve.Metadata.AssignerShortName
 	references := identifyPossibleURLs(cve)
-	metrics := &ConversionMetrics{CVEID: cveID, CNA: cnaAssigner}
+	metrics := ConversionMetrics{CVEID: cveID, CNA: cnaAssigner}
 
 	// Create a base OSV record from the CVE.
 	v := FromCVE5(cve, references, &metrics)


vulnfeeds/cmd/cve-bulk-converter/main.go
```diff
@@ -21,7 +21,7 @@ var (
 	localOutputDir = flag.String("out_dir", "cvelist2osv", "Path to output results.")
 	years          = flag.String("years", "2022,2023,2024,2025", "A comma-separated list of years to process.")
 	workers        = flag.Int("workers", 30, "The number of concurrent workers to use for processing CVEs.")
-	cnas           = flag.String("cnas", "Linux", "A comma-separated list of CNAs to process.")
+	cnas           = flag.String("cnas", "Linux,GitHub_M", "A comma-separated list of CNAs to process.")
```
maennchen, line 24: The EEF CNA is reporting GIT Versions. Can we be added to this list? Example CVE: https://cna.erlef.org/cves/cve-2025-48042.html

jess-lowe: Hey @maennchen, thanks for your enthusiasm with the CVE5 conversion! I tried running the converter with EEF scoped vulns and unfortunately in its current state it's not able to convert EEF vulns effectively, as the data layout differs from the more common expressions I've seen across Linux/GitHub_M/MITRE vulns. Looking a little more into it, it looks like the vulns are generated using Vulnogram, in which case, I'll consider adding an extension to handle Vulnogram generated vulns in a future PR :).

For a more timely ingestion of EEF vulns into OSV, we would highly appreciate Erlang publishing natively in the OSV format for us to ingest (also removes the layer of abstraction from going through the CVE Program to publish changes, and waiting for us to ingest those vulns).

vulnfeeds/cmd/cvelist2osv/converter.go
```diff
@@ -168,18 +173,14 @@ func ConvertAndExportCVEToOSV(cve cves.CVE5, directory string) error {
 	cveID := cve.Metadata.CVEID
 	cnaAssigner := cve.Metadata.AssignerShortName
 	references := identifyPossibleURLs(cve)
+	metrics := &ConversionMetrics{}
 
 	// Create a base OSV record from the CVE.
-	v, notes := FromCVE5(cve, references)
-	Metrics.Notes = append(Metrics.Notes, notes...)
+	v, notes := FromCVE5(cve, references, metrics)
```
another-rex: Since you are  passing metrics in now, this should not be returning notes, just append the notes to the metrics you passed in.

vulnfeeds/cmd/cvelist2osv/converter.go
```diff
@@ -168,18 +173,14 @@ func ConvertAndExportCVEToOSV(cve cves.CVE5, directory string) error {
 	cveID := cve.Metadata.CVEID
 	cnaAssigner := cve.Metadata.AssignerShortName
 	references := identifyPossibleURLs(cve)
+	metrics := &ConversionMetrics{}
```
another-rex: I think this should be the non pointer ver in this variable, and put the & on the places you are passing it in so it's clear it's passing a pointer.

vulnfeeds/cmd/cvelist2osv/version_extraction.go
```diff
@@ -107,31 +109,36 @@ func AddVersionInfo(cve cves.CVE5, v *vulns.Vulnerability) ([]VersionSource, []s
...
-			aff.Ranges = append(aff.Ranges, vr)
-		}
-
+		var aff osvschema.Affected
 		// Special handling for Linux kernel CVEs.
-		if cve.Metadata.AssignerShortName == "Linux" && versionType != VersionRangeTypeGit {
-			aff.Package = osvschema.Package{
-				Ecosystem: string(osvschema.EcosystemLinux),
-				Name:      "Kernel",
+		if cve.Metadata.AssignerShortName == "Linux" {
```
another-rex, line 111: Use a switch / default statement here, as this seems likely that we'll add additional ecosystems for special handling in the future.

vulnfeeds/cmd/cvelist2osv/version_extraction.go
```diff
@@ -164,6 +171,104 @@ func AddVersionInfo(cve cves.CVE5, v *vulns.Vulnerability, metrics *ConversionMe
...
+			logger.Warn("Failed to normalize tags", slog.String("cve", string(cveID)), slog.String("repo", repo), slog.Any("err", err))
+			continue
+		}
+
+		var stillUnresolvedRanges []osvschema.Range
+		for _, vr := range unresolvedRanges {
+			var introducedCommit, fixedCommit, lastAffectedCommit string
+			var resolutionErr error
+
+			for _, ev := range vr.Events {
```
another-rex: For 1 Event will there be more than one entry? I.e. Introduced != "" and Fixed != "" in the same event?

vulnfeeds/cmd/cvelist2osv/version_extraction.go
jess-lowe: this will be addressed in my big refactor :)

vulnfeeds/cmd/cvelist2osv/version_extraction.go
jess-lowe: No, because that's disallowed according to the osvschema, so they aren't written like that :)

vulnfeeds/cmd/cvelist2osv/version_extraction.go
```diff
@@ -164,6 +171,104 @@ func AddVersionInfo(cve cves.CVE5, v *vulns.Vulnerability, metrics *ConversionMe
...
+				logger.Info("Attempting version resolution", slog.String("cve", string(cveID)), slog.Any("event", ev), slog.String("repo", repo))
+				if ev.Introduced != "" {
```
another-rex: We can probably flatten a lot of these if statements by continuing rather than using else. I think it might make it a bit more readable.

vulnfeeds/cmd/cvelist2osv/version_extraction.go
```diff
@@ -164,6 +171,104 @@ func AddVersionInfo(cve cves.CVE5, v *vulns.Vulnerability, metrics *ConversionMe
...
+					lastAffectedCommit, resolutionErr = git.VersionToCommit(ev.LastAffected, normalizedTags)
+					if resolutionErr != nil {
+						logger.Warn("Failed to get Git commit for last affected version", slog.String("cve", string(cveID)), slog.String("version", ev.LastAffected), slog.String("repo", repo), slog.Any("err", resolutionErr))
+					} else {
+						logger.Info("Successfully derived commit for last affected version", slog.String("cve", string(cveID)), slog.String("commit", lastAffectedCommit), slog.String("version", ev.LastAffected))
+					}
+				}
+			}
+
+			resolved := false
```
another-rex: We can use continues here and avoid this variable, and the `else` in the `else if`

vulnfeeds/cmd/cvelist2osv/version_extraction.go
jess-lowe: will fix

PR #3952, opened 2025-09-12
fix: add more information to message of some logs

michaelkedar:
#3944 may have made these logs too short. Added some information back into the message line.

### DIFF
--- a/vulnfeeds/cmd/nvd-cve-osv/main.go
+++ b/vulnfeeds/cmd/nvd-cve-osv/main.go
@@ -449,7 +449,7 @@ func main() {
 			continue
 		}
 
-		logger.Info("Found Repos for CVE"+string(CVEID), slog.String("cve", string(CVEID)), slog.Any("repos", ReposForCVE[CVEID]))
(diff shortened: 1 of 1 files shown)


vulnfeeds/cmd/nvd-cve-osv/main.go
```diff
@@ -445,7 +449,7 @@ func main() {
 			continue
 		}
 
-		logger.Info("Repos", slog.String("cve", string(CVEID)), slog.Any("repos", ReposForCVE[CVEID]))
+		logger.Info("Found Repos for CVE", slog.String("cve", string(CVEID)), slog.Any("repos", ReposForCVE[CVEID]))
```
jess-lowe: I'll probably update this in the future 'cause it's also useful to skim-read the repos sometimes but for now:
```suggestion
		logger.Info("Found Repos for CVE"+string(CVEID), slog.String("cve", string(CVEID)), slog.Any("repos", ReposForCVE[CVEID]))
```

PR #3953, opened 2025-09-14
feat(vulnfeeds): populate package.purl for GIT ranges

### LINKED ISSUE #3807: Create repo purls for GIT ecosystem ranges
**The CVE ID**
https://osv.dev/vulnerability/CVE-2019-25219

**Describe the data quality issue observed**
PURL entries are provided for the Debians. The Git repo could also have a PURL, in this case: `pkg:generic/github.com/chriskohlhoff/asio@asio-x-y-z` where the version comes from the tagging scheme used in that repo.

**Suggested changes to record**
Add a PURL for the Git repo. This supports the use case where the code is being packaged directly by the consumer, as opposed to installed via the Debian.

ashmod:
This change automatically generates `pkg:generic` pURLs from Git repository URLs found in vulnerability data.
A new `BuildGenericRepoPURL()` helper converts URLs from common forges (e.g., GitHub, GitLab, self‑hosted), across the `https://`, `http://`, `git://`, `ssh://`, and `git@host:path` transports, into the `pkg:generic/{host}/{namespace}/{name}` format and enriches records that contain GIT-type ranges. Existing pURLs are not overwritten.

For example,
https://github.com/user/repo would generate the pURL: `pkg:generic/github.com/user/repo`

In addition, when packaging directly from source, versioned repo pURLs derived from tags (e.g., `pkg:generic/github.com/org/repo@repo-x-y-z`) are exposed under `affected.database_specific.repo_purls`; the standard `affected.package.purl` remains the base (unversioned) identifier [as per OSV schema](https://ossf.github.io/osv-schema/#:~:text=The%20purl%20field%20is%20a%20string%20following%20the%20Package%20URL%20specification%20that%20identifies%20the%20package%2C%20without%20the%20%40version%20component). Versioned pURLs are built through `packageurl-go`, so tags containing reserved characters (e.g. `release/1.2.3`, `v1.0 beta`) should be properly encoded.

Unit tests validate the conversion logic, the escape behaviour, and enrichment behavior, and module files are updated accordingly.

Some changes are still needed, especially for frontend display and backfilling historical Git ranges, but I think those can be added in a follow-up pr.

Fixes #3807


jess-lowe:
Hey @ashmod, thanks for this contribution! We're currently working on a refactor behind the scenes that is blocking this being merged - might take a couple of weeks, but we'll get back to it!

PR #3954, opened 2025-09-15
refactor(vulnfeeds): Uncouple Alpine vulns

jess-lowe:


vulnfeeds/vulns/cveworker.go
```diff
@@ -0,0 +1,88 @@
...
+	"path"
+	"time"
+
+	"cloud.google.com/go/storage"
+	"github.com/google/osv/vulnfeeds/utility/logger"
+)
+
+const (
+	// HashMetadataKey is the key for the sha256 hash in the GCS object metadata.
+	HashMetadataKey = "sha256-hash"
```
michaelkedar: nit: does this need to be public?

vulnfeeds/cmd/alpine/main.go
```diff
@@ -31,15 +40,49 @@ func main() {
 		"alpineOutput",
 		alpineOutputPathDefault,
 		"path to output general alpine affected package information")
+	outputBucketName := flag.String("output_bucket", outputBucketDefault, "The GCS bucket to write to.")
+	numWorkersStr := flag.String("num_workers", "64", "Number of workers to process records")
```
another-rex: Why is this a string?

another-rex: Please just change this to an int.

another-rex (dismissed):

PR #3955, opened 2025-09-15
fix(vulnfeeds): make logging a bit better

michaelkedar:
1. Fix the `sourceLocation` log field so that it a) maps to the location where the logger helper function was called and b) only shows the filename, rather than the full path.
2. When not logging in GCP, make the logs :sparkles: colourful :sparkles: 
image


michaelkedar:
Lmk if there's other fields you would like changed for running locally - I can probably remove the time field if it's making the logs too long.

another-rex (approved):

vulnfeeds/utility/logger/logger.go
```diff
@@ -79,10 +80,13 @@ func log(level slog.Level, msg string, a []any) {
 	runtime.Callers(3, pcs[:]) // skip [Callers, log, Info/Warn/etc]
 	r := slog.NewRecord(time.Now(), level, msg, pcs[0])
 	r.Add(a...)
+	//nolint:errcheck
 	slogLogger.Handler().Handle(context.Background(), r)
 }
 
 // Info prints an Info level log.
+//
+//nolint:contextcheck,nolintlint
```
another-rex, line 98: Huh why does this need both ignoring contextcheck, and ignoring nolint at the same time? Seems mutually exclusive

michaelkedar: without any nolints:
```
cmd/download-cves/main.go:105:16: Function `Warn->log` should pass the context parameter (contextcheck)
                                logger.Warn("Failed to decode NVD data", slog.Any("err", err))
```
with just nolint:contextcheck on `Warn`:
 ```
 utility/logger/logger.go:99:1: directive `//nolint:contextcheck` is unused for linter "contextcheck" (nolintlint)
//nolint:contextcheck
 ```
I can't win :pensive: 

 I think it wants the `//nolint:contextcheck` on the calling function, but that's annoying.
 I added it to the other functions to avoid this in the future.

jess-lowe:
YAY colourful logs!

jess-lowe (approved):

PR #3957, opened 2025-09-16
feat(staging): record checker

michaelkedar:
Adds a cron job to validate the consistency of records between GCS and datastore with the new database format (#3850). This sends messages to the recoverer (#3821) to attempt to repair.

The record-checker stores information from its previous runs in datastore in a new JobData entity - where the datastore Key is the name of the metadata, and the value is stored in a `value` property.

I've written this in Go to because it's fairly simple and to set a precedent for writing / migrating other components away from python. I made a copy of  the `logging` submodule from `vulnfeeds` into the new toplevel `go` directory - we'll want to consolidate these soon. There's also room for refactoring the record checker code have some reusable components for e.g. setting up datastore, gcs, or pub/sub.

I added a validation test script to make sure datastore entities written in Python are compatible with the go definitions, and vice versa.


go/cmd/recordchecker/recordchecker.go
```diff
@@ -0,0 +1,313 @@
...
+	return &appEnv{
+		bucket:     bucket,
+		ds:         dsClient,
+		topic:      topic,
+		numWorkers: numWorkers,
+	}, nil
+}
+
+type checkRecordResult struct {
+	ID         string
```
another-rex: Either make the struct public, or it's internal fields private. Probably just make the internal fields private.

go/cmd/recordchecker/recordchecker.go
```diff
@@ -0,0 +1,313 @@
...
+	}
+	if err := writeRecordCheckData(ctx, env.ds, rcData); err != nil {
+		return fmt.Errorf("failed to store run data: %w", err)
+	}
+
+	return nil
+}
+
+func handleResult(ctx context.Context, topic *pubsub.Topic, result checkRecordResult) bool {
+	wasInvalid := false
```
another-rex: Can you add a comment here (or just on the function) about what the return type of wasInvalid is suppose to mean? It's a bit unclear why NeedsRetry makes it invalid.

go/cmd/recordchecker/recordchecker.go
```diff
@@ -0,0 +1,313 @@
...
+		return fmt.Errorf("failed to get prior run data: %w", err)
+	}
+
+	tasksChan := make(chan checkTask)
+	resultsChan := make(chan checkRecordResult)
+
+	// Start the results handler
+	var resultsWg sync.WaitGroup
+	resultsWg.Add(1)
+	var newInvalid []string
```
another-rex: newInvalids?

go/cmd/recordchecker/recordchecker.go
```diff
@@ -0,0 +1,313 @@
...
+	}, nil
+}
+
+type checkRecordResult struct {
+	ID         string
+	NeedsRetry bool
+	Err        error
+}
+
+func checkRecord(ctx context.Context, cl *datastore.Client, bucket *storage.BucketHandle, id string, vuln *models.Vulnerability, out chan<- checkRecordResult) {
```
another-rex: nit: Still feel like it could be clearer to return checkRecordResult, then have it be sent out the channel in the parent.

go/cmd/recordchecker/recordchecker.go
another-rex, line 1: Can you copy in the .golangci.yaml config file from vulnfeeds here for the go linter to the go/ dir.

go/models/internal/validate/validate.go
```diff
@@ -0,0 +1,51 @@
...
+	"os"
+	"time"
+
+	"cloud.google.com/go/datastore"
+	"github.com/google/osv.dev/go/models"
+)
+
+func main() {
+	ctx := context.Background()
+	client, err := datastore.NewClient(ctx, os.Getenv("GOOGLE_CLOUD_PROJECT"))
```
another-rex, line 20: This is communicating with the datastore emulator right? Can you add a comment here clarifying that, and explain how this knows that it should be the datastore emulator it's connecting to.

go/cmd/recordchecker/recordchecker.go
michaelkedar: privatized the internal fields
(except the ones on the datastore struct that needs to be public)

go/cmd/recordchecker/recordchecker.go
michaelkedar: wasInvalid was a bit of a misnomer.
Removed the variable since it was just = needsRetry, and added a function-level comment

go/cmd/recordchecker/recordchecker.go
michaelkedar: done

go/models/internal/validate/validate.go
michaelkedar: added a comment explaining most of that

PR #3978, opened 2025-09-17
fix: alpine versions `<=`/`>=`

michaelkedar:
The alpine versions seem to be falling back to string comparisons for `>=`/`<=`, which is wrong.
This issue is in the univers code we've vendored, which seems to be a bit outdated, but it's [still present on their current version](https://github.com/aboutcode-org/univers/issues/172)

I've added tests to catch this in Alpine and other ecosystems, but APK seems to have been the only ecosystem affected.


osv/ecosystems/alpine_test.py
```diff
@@ -75,6 +75,12 @@ def test_apk(self):
     self.assertLess(
         ecosystem.sort_key('13.0.14.5-r1'), ecosystem.sort_key('16.6-r0'))
 
+    # Check >= / <= methods: https://github.com/google/osv.dev/pull/3978
```
another-rex, line 78: Ah yes, the self referential PR.

PR #3979, opened 2025-09-17
fix(cron): Improve speed of Debian converter

jess-lowe:


deployment/clouddeploy/gke-workers/base/debian-cve-convert.yaml
```diff
@@ -24,13 +24,13 @@ spec:
               privileged: true
             resources:
               requests:
-                cpu: "1"
+                cpu: "8"
```
jess-lowe:
```suggestion
                cpu: "16"
```

deployment/clouddeploy/gke-workers/base/debian-cve-convert.yaml
```diff
@@ -24,13 +24,13 @@ spec:
...
                 memory: "8G"
```
jess-lowe:
```suggestion
                memory: "16G"
```

PR #3982, opened 2025-09-18
fix(ui): Add border to final ecosystem panel

ashmod:
Adds a footer bar so the last ecosystem’s tree ends on a visible horizontal line instead of trailing off.

terminator

### DIFF
--- a/gcp/website/frontend3/src/styles.scss
+++ b/gcp/website/frontend3/src/styles.scss
@@ -1016,10 +1016,11 @@ dl.vulnerability-details,
     }
 
     $last-ecosystem-border-offset: 12px;
+    $last-ecosystem-border-gap: 8px;
     $last-ecosystem-border-thickness: 3px;
 
     .ecosystem-content-panel--last {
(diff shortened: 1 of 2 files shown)


another-rex:
Suggestion is to add back the bottom border, but only for the last package so it ends with a horizontal line. I tried to quickly add it to the original PR, but I think because of some border box issues it gets hidden somehow.

ashmod:
> Suggestion is to add back the bottom border, but only for the last package so it ends with a horizontal line. I tried to quickly add it to the original PR, but I think because of some border box issues it gets hidden somehow. 

I agree, I also tried re-adding the border to the last child but that didn't seem to work. I'll look into this further.

ashmod:
@another-rex lmk what you think!

another-rex:
Nice!, I think let's just make it 1px wide and have the same gap as the gap between ecosystems.

ashmod:
updated

another-rex (approved):

PR #3985, opened 2025-09-18
feat(vulnfeeds): write CVEs that are only Debian without package info for migration

jess-lowe:
So that users aren't shocked by a missing CVE file when the Debian decoupling happens, keep writing out the CVEs but not the pkgInfo information for Debian


another-rex (dismissed):

vulnfeeds/test_data/parts/debian/CVE-2016-1585.debian.json
another-rex, line 1: Can we keep these parts? The tests are still valid, we want to test what happens when parts exist, just that it should be a cve record with empty affected.

another-rex: Well at least keep one.

PR #3987, opened 2025-09-18
fix(vulnfeeds): only upload Debian CVEs if file has changed. 

### LINKED ISSUE #3983: Debian Conversion changes modified date each time
Every time the Debian conversion is run, it updates the modified time. This shouldn't happen if it hasn't changed.

Idea currently:
- csv file to hash updates: CVE_ID, modified date, hash
- read into memory at start, generate file, fill data with modified date, hash it and compare against csv.
- if same, generate file as is, otherwise update modified to time.Now, update dictionary and hash
- regenerate csv file. 

more elegant solution: use metadata

jess-lowe:
Closes #3983

### DIFF
--- a/vulnfeeds/cmd/debian/main.go
+++ b/vulnfeeds/cmd/debian/main.go
@@ -2,20 +2,24 @@
 package main
 
 import (
+	"context"
+	"crypto/sha256"
 	"encoding/csv"
+	"encoding/hex"
 	"encoding/json"
 	"flag"
 	"fmt"
 	"log/slog"
-
 	"net/http"
 	"os"
 	"path"
 	"sort"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
+	"cloud.google.com/go/storage"
 	"github.com/google/osv/vulnfeeds/cves"
 	"github.com/google/osv/vulnfeeds/faulttolerant"
 	"github.com/google/osv/vulnfeeds/models"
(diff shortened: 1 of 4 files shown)


michaelkedar (approved):
lgtm

vulnfeeds/cmd/debian/main.go
```diff
@@ -144,8 +236,13 @@ func generateOSVFromDebianTracker(debianData DebianSecurityTrackerData, debianRe
 			}
 		}
 	}
-
-	return osvCves
+	for _, v := range osvCves {
```
another-rex: Keep the return here and do this outside

vulnfeeds/cmd/debian/main.go
```diff
@@ -29,12 +34,16 @@ const (
 	debianOutputPathDefault  = "debian-cve-osv"
 	debianDistroInfoURL      = "https://debian.pages.debian.net/distro-info-data/debian.csv"
 	debianSecurityTrackerURL = "https://security-tracker.debian.org/tracker/data/json"
+	outputBucketDefault      = "debian-osv"
+	hashMetadataKey          = "sha256-hash"
+	numWorkers               = 128
```
another-rex: Have workers be passed in as a flag. Then set the flag in the script as a env variable.

vulnfeeds/cmd/debian/main.go
another-rex: That way the tests also don't need to have all the channel logic there.

vulnfeeds/cmd/debian/main.go
```diff
@@ -100,7 +200,6 @@ func generateOSVFromDebianTracker(debianData DebianSecurityTrackerData, debianRe
 					Vulnerability: osvschema.Vulnerability{
 						ID:        "DEBIAN-" + cveID,
 						Upstream:  []string{cveID},
-						Modified:  time.Now().UTC(),
```
another-rex, line 103: I am a bit surprised that this is not breaking the tests, can you have a look at what's happening here?

another-rex (approved):
Hopefully 256 isn't too high that we start getting worse performance haha.

michaelkedar:
> Hopefully 256 isn't too high that we start getting worse performance haha.

For reference, the record checker has 50 workers and was using <0.5 CPUs when it was checking the metadata on all of the bucket vulns :slightly_smiling_face:

vulnfeeds/cmd/debian/main.go
jess-lowe: when I wrote the original tests, I made it explicitly ignore modified/published times in its comparisons.

PR #3993, opened 2025-09-18
docs(api): clarify version vs purl usage in query endpoints

### LINKED ISSUE #1900: Document that `version` and purls with versions can't be both specified in a query.
Context: https://github.com/google/osv.dev/issues/1443#issuecomment-1875625789 

We should update https://google.github.io/osv.dev/post-v1-query/#parameters to mention that `version` cannot be specified twice via the `"version"` parameter and the purl.

ashmod:
Documents that `version` and versioned purls are mutually exclusive.
Also fixes a small typo.

Fixes #1900


another-rex (approved):
Really nice!

PR #3996, opened 2025-09-19
feat(tools/datafix):  allow updating all bugs in a source with confirmation

cuixq:
This change enhances the `request_worker_update_record.py` script by adding the ability to process all bug records from a specified source.

Previously, the script required at least one bug ID to be provided, so making it cumbersome to re-process an entire source.

Now, if no bug IDs are supplied:
   1. The script queries all bugs for the given source.
   2. It displays the total number of bugs found.
   3. It prompts the user for confirmation before proceeding to publish the update messages.

 This makes the tool more flexible for bulk updates while maintaining safety with an explicit confirmation step.


tools/datafix/request_worker_update_record.py
```diff
@@ -103,35 +103,55 @@ def main():
...
+    if not source_repo:
+      raise ValueError(f"Source repository '{args.source}' not found.")
+
+    bugs_to_process = []
+    if args.bugs and args.bugs[0]:
+      bugs_to_process = args.bugs[0]
+    else:
+      print(f'No bug IDs provided. Querying all bugs for source {args.source}...')
+      query = osv.Bug.query(osv.Bug.source == args.source)
+      bugs_to_process = [b.key.id() for b in query]
```
michaelkedar: This could probably be a bit more efficient by doing a keys only query. I *think* it's like (but double check this works):
```suggestion
      query = osv.Bug.query(osv.Bug.source == args.source)
      bugs_to_process = [b.id() for b in query.iter(keys_only=True)]
```

michaelkedar (dismissed):

michaelkedar (approved):

PR #4005, opened 2025-09-19
fix(ui): set vuln packages active tab bottom border to red

### LINKED ISSUE #3995: Nit: border-bottom-color is blue instead of red
Image

the bottom border of packages on frontend is blue instead of red because of the `spicy-section` package's default color. 

Ideal fix: remove the spicy-section package entirely
acceptable fix: set the correct bottom color

ashmod:
Fixes #3995


another-rex (approved):
Nice!

PR #4033, opened 2025-09-24
fix: Debian-CVE with no fixed not being converted (and added dryrun mode)

jess-lowe:


vulnfeeds/cmd/debian/main.go
```diff
@@ -44,6 +44,7 @@ func main() {
 	debianOutputPath := flag.String("output_path", debianOutputPathDefault, "Path to output OSV files.")
 	outputBucketName := flag.String("output_bucket", outputBucketDefault, "The GCS bucket to write to.")
 	numWorkers := flag.Int("num_workers", 64, "Number of workers to process records")
+	dryRun := flag.Bool("dry_run", false, "If true, do not write to GCS bucket and instead write to local disk.")
```
another-rex: nit: Let's call this uploadToGCS and default it to false, that way we will default to writing to disk, and only write to GCS when running in cloud.

another-rex (approved):

PR #4034, opened 2025-09-24
fix: api tests

jess-lowe:


gcp/api/integration_tests.py
```diff
@@ -533,8 +533,9 @@ def test_query_purl(self):
 
     expected_deb = [
         self._get('CVE-2018-25047'),
```
cuixq: this can be deleted?

cuixq: let's delete this record

```suggestion
```

jess-lowe: tdlr: importer broken

cuixq (approved):

PR #4035, opened 2025-09-24
feat: Much faster REST API import

another-rex:
21 minutes to 2 seconds. By skipping jsonschema validate for parsing just (modified, id)

### DIFF
--- a/gcp/workers/importer/importer.py
+++ b/gcp/workers/importer/importer.py
@@ -25,6 +25,7 @@ import json
 import logging
 import os
 import requests
+from google.protobuf import json_format
 from requests.adapters import HTTPAdapter
 import shutil
 import threading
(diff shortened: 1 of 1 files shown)


gcp/workers/importer/importer.py
```diff
@@ -903,11 +905,15 @@ def _process_updates_rest(self, source_repo: osv.SourceRepository):
...
-        source_repo.extension,
-        strict=source_repo.strict_validation and self._strict_validation)
+
+    data = json.loads(request.text)
+    vulns = []
+    for datum in data:
+      vulnerability = vulnerability_pb2.Vulnerability()
+      json_format.ParseDict(datum, vulnerability, ignore_unknown_fields=True)
+      if not vulnerability.id:
+        raise ValueError('Missing id field. Invalid vulnerability.')
```
jess-lowe, line 915: should we be checking for modified here too? it's a required field too.

another-rex: Done.

PR #4036, opened 2025-09-25
fix: oss-fuzz source links using double Git links

jess-lowe:


cuixq (approved):

another-rex (dismissed):

gcp/api/integration_tests.py
```diff
@@ -533,8 +533,9 @@ def test_query_purl(self):
 
     expected_deb = [
         self._get('CVE-2018-25047'),
```
cuixq: this can be removed now I think

jess-lowe:
```suggestion
```

another-rex (approved):

PR #4038, opened 2025-09-25
ci: update link checker based on `osv-scanner` setup

### LINKED ISSUE #4037: markdown-link-check is now deprecated
We depend on [markdown-link-check](https://github.com/marketplace/actions/markdown-link-check) in our [link checker](https://github.com/google/osv.dev/blob/master/.github/workflows/link-checker-on-push.yml#L19C13-L19C60) but it is now deprecated.

G-Rath:
This is the same setup we're using in `osv-scanner`

Resolves #4037


another-rex (approved):

another-rex:
Nice, that was really fast haha.

G-Rath:
oh fwiw though the ci failure that I assume brought your attention to this is because of the ongoing docker incident: https://www.dockerstatus.com/pages/incident/533c6539221ae15e3f000031/68d47a2f93c09e05486d93a9

any workflow that uses docker right now won't work

cuixq:
ah but the alpine version is also end of support :(

PR #4045, opened 2025-09-25
refactor(frontend): streamline ecosystem auto-expand logic

ashmod:
Refactors the Ecosystem Header expand/collapse logic for cases where multi-ecosystem vulnerabilities with large package counts stay collapsed by default. Single-ecosystem vulns behave as before.

This helps avoid endless vertical scrolling for cases like: [RHSA-2025:16669](https://osv.dev/vulnerability/RHSA-2025:16669)

### DIFF
--- a/gcp/website/frontend3/src/templates/vulnerability.html
+++ b/gcp/website/frontend3/src/templates/vulnerability.html
@@ -497,23 +497,36 @@
   document.addEventListener('turbo:load', function() {
     const ECOSYSTEM_EXPAND_THRESHOLD = 10;
     const PACKAGE_EXPAND_THRESHOLD = 7;
+    const ECOSYSTEM_PACKAGE_COLLAPSE_THRESHOLD = 9;
 
   /**
    * Sets up the vertical, grouped layout for vulnerability packages.
-   * - All ecosystem headers are expanded if their total count is below `ECOSYSTEM_EXPAND_THRESHOLD`.
+   * - All ecosystem headers are expanded if their total count is below `ECOSYSTEM_EXPAND_THRESHOLD`,
+   *   and if their total packages count is below `ECOSYSTEM_PACKAGE_COLLAPSE_THRESHOLD`.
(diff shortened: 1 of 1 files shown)


gcp/website/frontend3/src/templates/vulnerability.html
```diff
@@ -497,23 +497,32 @@ <h3 class="mdc-layout-grid__cell--span-3">
...
+      const shouldExpandEcosystems = ecosystemHeaders.length < ECOSYSTEM_EXPAND_THRESHOLD;
+
+      ecosystemHeaders.forEach(header => {
+        const panel = header.nextElementSibling;
+        if (
+          shouldExpandEcosystems &&
+          header.getAttribute('aria-expanded') === 'false' &&
+          panel &&
+          !(ecosystemHeaders.length > 1 &&
+            panel.querySelectorAll('.package-name-title').length > ECOSYSTEM_PACKAGE_COLLAPSE_THRESHOLD)
```
another-rex: This if statement is getting a bit hard to read, can we pull out a few of these conditions into variables and name them so it's clear what each section is checking?

another-rex: Soemthing like 
- shouldExpandEcosystems
- isCurrentlyExpanded
- hasChildrenBeyondThreshold (or maybe break this down even more)

another-rex (commented):
Thanks!

gcp/website/frontend3/src/templates/vulnerability.html
ashmod: I had something like that in an earlier version, reverted back to it in 8cf1724. Let me know if this looks better!

another-rex (approved):

gcp/website/frontend3/src/templates/vulnerability.html
another-rex: Thanks, looks good!

PR #4046, opened 2025-09-25
fix(cron): increase memory for combine-to-osv

jess-lowe:


deployment/clouddeploy/gke-workers/base/combine-to-osv.yaml
```diff
@@ -26,10 +26,10 @@ spec:
             resources:
               requests:
                 cpu: "30"
```
another-rex, line 28: Let's have a look at the current run and see what the max cpu usage gets to, I think we should be bumping this down by quite a bit, as it currently forces a new highend node to spin up specifically for this job, which feels quite wasteful and makes it really slow to start.

PR #4047, opened 2025-09-25
fix: Add vulns bucket to prod crons

michaelkedar:


cuixq (approved):
thanks for fixing this!

PR #4050, opened 2025-09-27
fix(make): run lint via poetry

### LINKED ISSUE #3992: Change `make lint` to just use `poetry run`
The Makefile in the root just needs `poetry run` put in front of `tools/run_linter.sh` (or whatever it's called).

Would make it easier than having to activate up a poetry environment every time.

ashmod:
Closes #3992


jess-lowe (approved):
Thanks!

PR #4052, opened 2025-09-29
refactor(vulnfeeds): git commit resolution code for better reusability

jess-lowe:
Part of splitting logic from #3951

### DIFF
--- a/vulnfeeds/cves/versions.go
+++ b/vulnfeeds/cves/versions.go
@@ -515,8 +515,9 @@ func ValidateAndCanonicalizeLink(link string, httpClient *http.Client) (canonica
 }
 
 // For URLs referencing commits in supported Git repository hosts, return a cloneable AffectedCommit.
-func ExtractGitAffectedCommit(link string, commitType models.CommitType, httpClient *http.Client) (ac models.AffectedCommit, err error) {
-	c, r, err := ExtractGitCommit(link, httpClient)
+func ExtractGitAffectedCommit(link string, commitType models.CommitType, httpClient *http.Client) (models.AffectedCommit, error) {
+	var ac models.AffectedCommit
+	c, r, err := ExtractGitCommit(link, httpClient, 0)
 	if err != nil {
 		return ac, err
 	}
@@ -528,7 +529,10 @@ func ExtractGitAffectedCommit(link string, commitType models.CommitType, httpCli
 	return ac, nil
 }
 
-func ExtractGitCommit(link string, httpClient *http.Client) (string, string, error) {
+func ExtractGitCommit(link string, httpClient *http.Client, depth int) (string, string, error) {
+	if depth > 10 {
+		return "", "", fmt.Errorf("max recursion depth exceeded for %s", link)
+	}
 	var commit string
 	r, err := Repo(link)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


vulnfeeds/cves/versions.go
```diff
@@ -515,44 +515,47 @@ func ValidateAndCanonicalizeLink(link string, httpClient *http.Client) (canonica
 }
 
 // For URLs referencing commits in supported Git repository hosts, return a cloneable AffectedCommit.
-func ExtractGitCommit(link string, commitType models.CommitType, httpClient *http.Client) (ac models.AffectedCommit, err error) {
-	r, err := Repo(link)
+func ExtractGitAffectedCommit(link string, commitType models.CommitType, httpClient *http.Client) (ac models.AffectedCommit, err error) {
```
another-rex: We've mostly avoided having named return values, let's just explicitly name them here.

vulnfeeds/cves/versions.go
```diff
@@ -515,44 +515,47 @@ func ValidateAndCanonicalizeLink(link string, httpClient *http.Client) (canonica
...
-		return ac, err
+		return "", "", err
 	}
 
 	// restart the entire extraction process when the URL changes (i.e. handle a
 	// redirect to a completely different host, instead of a redirect within
 	// GitHub)
 	if possiblyDifferentLink != link {
-		return ExtractGitCommit(possiblyDifferentLink, commitType, httpClient)
+		return ExtractGitCommit(possiblyDifferentLink, httpClient)
```
another-rex: Just noticed a potential infinite loop here, can you add a depth value to limit the recursion here?

vulnfeeds/cves/versions.go
```diff
@@ -515,44 +515,47 @@ func ValidateAndCanonicalizeLink(link string, httpClient *http.Client) (canonica
...
 	}
 
+	ac.SetRepo(r)
+
+	models.SetCommitByType(&ac, commitType, c)
+
+	return ac, nil
+}
+
+func ExtractGitCommit(link string, httpClient *http.Client) (string, string, error) {
```
another-rex: Is this function used outside of here? If not let's just make it private.

vulnfeeds/cves/versions_test.go
```diff
@@ -691,7 +691,7 @@ func TestExtractGitCommit(t *testing.T) {
 			if !tc.disableExpiryDate.IsZero() && time.Now().After(tc.disableExpiryDate) {
 				t.Logf("test %q: extractGitCommit(%q, %q) has been enabled on %s.", tc.description, tc.inputLink, tc.inputCommitType, tc.disableExpiryDate)
```
another-rex: function name need to be updated here

another-rex: and below

another-rex (commented):
Nice!

vulnfeeds/cves/versions.go
jess-lowe: This will be used later by the CVE5 conversion

jess-lowe: sorry nevermind, this one isnt used outside

another-rex (approved):

PR #4061, opened 2025-09-30
fix(linter): handle page routing with trailing slash

ashmod:
Fixes the trailing slash problem, so now [`/linter`](https://test.osv.dev/linter) and [`/linter/`](https://test.osv.dev/linter/) both resolve correctly.


another-rex:
That's meant to point to test instance, the linter is only available on test as it's meant for vulnerability publishers.

another-rex:
We can keep the strict_slashes change though!

ashmod:
> We can keep the strict_slashes change though!

updated.

PR #4078, opened 2025-09-30
fix: assuming one value means fixed should be last affected, and remove CNA qualifier for parsing attempt

jess-lowe:


vulnfeeds/cmd/cvelist2osv/version_extraction.go
```diff
@@ -344,20 +344,19 @@ func findNormalAffectedRanges(affected cves.Affected, cnaAssigner string, metric
 		// in one line instead - like "< 1.5.3" or "< 2.45.4, >= 2.0 " or just "before 1.4.7", so check for that.
 		metrics.Notes = append(metrics.Notes, "Only version exists")
 		// GitHub often encodes the range directly in the version string.
```
another-rex, line 346: Update comment here?

jess-lowe: thisll be fixed in a future pr

PR #4080, opened 2025-10-01
refactor: use AddNotes and AddSource fields in CVEList conversion

jess-lowe:
Tidy up the code a bit to use an AddNotes method to append notes and sources.

### DIFF
--- a/vulnfeeds/cmd/cvelist2osv/converter.go
+++ b/vulnfeeds/cmd/cvelist2osv/converter.go
@@ -31,6 +31,16 @@ type ConversionMetrics struct {
 	Notes          []string                        `json:"notes"`           // A collection of notes and warnings generated during conversion.
 }
 
+// AddNote adds a formatted note to the ConversionMetrics.
+func (m *ConversionMetrics) AddNote(format string, a ...any) {
+	m.Notes = append(m.Notes, fmt.Sprintf(format, a...))
+}
+
+// AddSource appends a source to the ConversionMetrics
+func (m *ConversionMetrics) AddSource(source VersionSource) {
+	m.VersionSources = append(m.VersionSources, source)
+}
+
 // RefTagDenyList contains reference tags that are often associated with unreliable or
 // irrelevant repository URLs. References with these tags are currently ignored
 // to avoid incorrect repository associations.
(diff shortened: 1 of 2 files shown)


vulnfeeds/cmd/cvelist2osv/version_extraction.go
```diff
@@ -147,19 +147,21 @@ func AddVersionInfo(cve cves.CVE5, v *vulns.Vulnerability, metrics *ConversionMe
...
-			metrics.Notes = append(metrics.Notes, err.Error())
+			metrics.AddNote("%s", err.Error())
 		}
 	}
 
 	// As a last resort, try extracting versions from the description text.
 	if !gotVersions {
-		metrics.Notes = append(metrics.Notes, "No versions in CPEs so attempting extraction from description")
+		metrics.AddNote("No versions in CPEs so attempting extraction from description")
 		versions, extractNotes := cves.ExtractVersionsFromText(nil, cves.EnglishDescription(cve.Containers.CNA.Descriptions))
```
another-rex, line 157: I'm guessing refactoring this to not return notes but just taking in a metrics obj will be a pretty big refactor involving nvd-cve-osv?

vulnfeeds/cmd/cvelist2osv/converter.go
```diff
@@ -31,6 +31,16 @@ type ConversionMetrics struct {
 	Notes          []string                        `json:"notes"`           // A collection of notes and warnings generated during conversion.
 }
 
+// AddNote adds a formatted note to the ConversionMetrics.
+func (m *ConversionMetrics) AddNote(format string, a ...any) {
+	m.Notes = append(m.Notes, fmt.Sprintf(format, a...))
```
another-rex, line 37: you can probably do logger.Debug or something here if you want to also log the notes output. You can even put the CNA in a variable.

another-rex:
Looks so clean!

vulnfeeds/cmd/cvelist2osv/version_extraction.go
jess-lowe: It's something I want to do eventually but yeah, it would involve a pretty big refactor

vulnfeeds/cmd/cvelist2osv/converter.go
jess-lowe: good point!

vulnfeeds/cmd/cvelist2osv/converter.go
```diff
@@ -166,7 +178,7 @@ func ConvertAndExportCVEToOSV(cve cves.CVE5, directory string) error {
 	cveID := cve.Metadata.CVEID
 	cnaAssigner := cve.Metadata.AssignerShortName
 	references := identifyPossibleURLs(cve)
-	metrics := &ConversionMetrics{}
+	metrics := &ConversionMetrics{CVEID: cveID}
```
another-rex: You should probably set the CNA here as well right?

PR #4086, opened 2025-10-01
feat: rework combine-to-osv to begin to combine NVD and CVE5 OSV records

jess-lowe:
### NVD Output Changes:
- Updates the NVD conversion to start outputting OSV records instead of `PackageInfo`
- Changes the output location of NVD to no longer be in the 'parts' directory

### `combine-to-osv` changes:
- reads in two file paths for NVD OSV data and CVE5 OSV data
- lists CVEIDs of Alpine and debian vulns and generates the 'upstream' records for them too so they still have an upstream,
- somewhat intelligently tries to combine data between the records, with a bias toward the CVE5 conversion.
- outputs the files.


vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -2,197 +2,338 @@
...
+	defaultNVDOSVPath    = "nvd"
 )
 
 func main() {
 	logger.InitGlobalLogger()
 
-	cvePath := flag.String("cvePath", defaultCvePath, "Path to CVE file")
-	partsInputPath := flag.String("partsPath", defaultPartsInputPath, "Path to CVE file")
+	cve5Path := flag.String("cve5Path", defaultCVE5Path, "Path to CVE file")
+	nvdPath := flag.String("nvdPath", defaultNVDOSVPath, "Path to CVE file")
```
another-rex: Probably should have different descriptions.

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -2,197 +2,338 @@
...
 
-	allCves := vulns.LoadAllCVEs(*cvePath)
-	allParts, cveModifiedMap := loadParts(*partsInputPath)
-	combinedData := combineIntoOSV(allCves, allParts, *cveListPath, cveModifiedMap)
+	// Load CVE5 OSVs
+	allCVE5 := loadOSV(*cve5Path)
+	// Load NVD OSVs
+	allNVD := loadOSV(*nvdPath)
+
+	// nvdCVEs := vulns.LoadAllCVEs(defaultNVDOSVPath)
```
another-rex: should this be removed?

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -2,197 +2,338 @@
...
+	bucket := client.Bucket(bucketName)
+
+	it := bucket.Objects(ctx, &storage.Query{Prefix: prefix})
+
+	var filenames []string
+
+	for {
+		attrs, err := it.Next()
+
+		if err == iterator.Done {
```
another-rex: might want to use errors.Is here,though it might not  matter.

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -2,197 +2,338 @@
...
-		err = json.NewDecoder(file).Decode(&pkgInfos)
 		if err != nil {
-			logger.Fatal("Failed to decode", slog.String("file", file.Name()), slog.Any("err", err))
+			return nil, fmt.Errorf("bucket.Objects: %w", err)
 		}
 
-		// Turns CVE-2022-12345.alpine.json into CVE-2022-12345
-		cveID := cves.CVEID(strings.Split(entryInner.Name(), ".")[0])
-		output[cveID] = append(output[cveID], pkgInfos...)
+		filenames = append(filenames, cleanFilename(attrs.Name, prefix))
```
another-rex: This function should not be modifying the names, since it's called listBucketObjects. Let's pull this out to be called after bucket objects have been listed.

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -2,197 +2,338 @@
...
+	cleaned = strings.TrimSuffix(cleaned, ".json")
+	pre := strings.SplitAfter(cleaned, "-")
+	if pre[0] != "CVE" {
+		cleaned = pre[1]
 	}
-	parsedTime := fileInfo.ModTime()
-
-	return parsedTime, err
+	return cleaned
 }
```
another-rex: Probably name cleanFilename to be something a bit more descriptive, extractCVEName maybe?

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -2,197 +2,338 @@
...
-func getModifiedTime(filePath string) (time.Time, error) {
-	var emptyTime time.Time
-	fileInfo, err := os.Stat(filePath)
-	if err != nil {
-		return emptyTime, err
+func cleanFilename(filename string, prefix string) string {
+	cleaned := strings.TrimPrefix(filename, prefix)
+	cleaned = strings.TrimSuffix(cleaned, ".json")
+	pre := strings.SplitAfter(cleaned, "-")
+	if pre[0] != "CVE" {
```
another-rex, line 89: This seems dangerious assumption, if it isn't expected we should just skip this file. (return "" and then filter in the caller.)

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -2,197 +2,338 @@
...
-			continue
-		}
-		filePath := path.Join(innerPartInputPath, entryInner.Name())
-		file, err := os.Open(filePath)
-		if err != nil {
-			logger.Fatal("Failed to open PackageInfo JSON", slog.String("path", path.Join(innerPartInputPath, entryInner.Name())), slog.Any("err", err))
+	defer client.Close()
+
+	bucket := client.Bucket(bucketName)
+
```
another-rex: We can compact these new lines a bit haha.

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -2,197 +2,338 @@
...
+	// Load CVE5 OSVs
+	allCVE5 := loadOSV(*cve5Path)
+	// Load NVD OSVs
+	allNVD := loadOSV(*nvdPath)
+
+	// nvdCVEs := vulns.LoadAllCVEs(defaultNVDOSVPath)
+	debianCVEs, err := listBucketObjects("osv-test-debian-osv", "/debian-cve-osv")
+	alpineCVEs, err := listBucketObjects("osv-test-cve-osv-conversion", "/alpine")
+
+	noPkg := append(debianCVEs, alpineCVEs...)
```
another-rex: what does noPkg refer to here?

another-rex: Ohh, forced CVE creation even if it's empty. Hmm probably clarify what this part is doing with a comment.

another-rex: And probably something like mandatory CVE Ids, since it's not necessary no affected ranges if there are other (git) ranges present.

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -2,197 +2,338 @@
...
+	"path/filepath"
 	"strings"
-	"time"
 
+	"cloud.google.com/go/storage"
 	"github.com/google/osv/vulnfeeds/cves"
 	"github.com/google/osv/vulnfeeds/utility/logger"
-	"github.com/google/osv/vulnfeeds/vulns"
+	"github.com/ossf/osv-schema/bindings/go/osvschema"
+	"golang.org/x/exp/slices"
```
another-rex: Pretty sure we can just use the stdlib here, if slices.Contains is the only thing you are using.

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -2,197 +2,338 @@
...
-		// Turns CVE-2022-12345.alpine.json into CVE-2022-12345
-		cveID := cves.CVEID(strings.Split(entryInner.Name(), ".")[0])
-		output[cveID] = append(output[cveID], pkgInfos...)
+		filenames = append(filenames, cleanFilename(attrs.Name, prefix))
+	}
+	return filenames, nil
+}
 
-		logger.Info("Loaded "+entryInner.Name(), slog.String("file", entryInner.Name()))
+func loadOSV(osvPath string) map[cves.CVEID]osvschema.Vulnerability {
```
another-rex, line 127: nit: Add some doc comments to these functions

vulnfeeds/cmd/combine-to-osv/run_combine_to_osv_convert.sh
```diff
@@ -18,32 +18,26 @@ set -eu
 
 INPUT_BUCKET="${INPUT_GCS_BUCKET:=cve-osv-conversion}"
 OUTPUT_BUCKET="${OUTPUT_GCS_BUCKET:=cve-osv-conversion}"
-OSV_PARTS_ROOT="parts/"
+
 OSV_OUTPUT="osv_output/"
-CVE_OUTPUT="cve_jsons/"
-CVELIST="${CVELIST_PATH:=cvelistV5/}"
+NVD_OSV_OUTPUT="nvd"
+CVE5_OSV_OUTPUT="cve5"
```
another-rex, line 24: Should these be pulled from env variables as well? Or is it fine to just hard code them as we're unlikely to change it on the fly.

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -2,197 +2,348 @@
...
-	cvePartsModifiedTime := make(map[cves.CVEID]time.Time)
-	for _, entry := range dir {
-		if !entry.IsDir() {
-			logger.Warn("Unexpected file entry", slog.String("file", entry.Name()), slog.String("path", partsInputPath))
-			continue
+
+	// Add any remaining CVEs from NVD that were not in the advisory data.
+	for cveID, nvd := range nvdosv {
+		vulns[cveID] = nvd
+		logger.Info("" + string(cveID))
```
another-rex: remove debugging log?

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -2,197 +2,348 @@
...
-	output := map[cves.CVEID][]vulns.PackageInfo{}
-	cvePartsModifiedTime := make(map[cves.CVEID]time.Time)
-	for _, entry := range dir {
-		if !entry.IsDir() {
-			logger.Warn("Unexpected file entry", slog.String("file", entry.Name()), slog.String("path", partsInputPath))
-			continue
+
+	// Add any remaining CVEs from NVD that were not in the advisory data.
+	for cveID, nvd := range nvdosv {
+		vulns[cveID] = nvd
```
another-rex, line 194: Does this also need the check for combined.Affected == 0?

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -2,197 +2,348 @@
...
+				aliasMap[alias] = true
+			}
+			for _, alias := range nvd.Aliases {
+				if !aliasMap[alias] {
+					combined.Aliases = append(combined.Aliases, alias)
+					aliasMap[alias] = true
+				}
+			}
+
+			// TODO: Elegantly handle combining severity scores
```
another-rex: I think this section could be its own function, that takes in two osvschema.Vulnerabilities and returns a single one.

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -2,197 +2,348 @@
...
+	// Add any remaining CVEs from NVD that were not in the advisory data.
+	for cveID, nvd := range nvdosv {
+		vulns[cveID] = nvd
+		logger.Info("" + string(cveID))
+	}
+
+	return vulns
+}
+
+func getRangeBoundaryVersions(events []osvschema.Event) (introduced, fixed string) {
```
another-rex: This should be placed after the following function in the file right?

another-rex: Also let's add a quick doc comment for this, (e.g. picks non 0 boundary versions...)

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -2,197 +2,348 @@
...
-			// Best-effort attempt to mark a disputed CVE as withdrawn.
-			modified, err := vulns.CVEIsDisputed(convertedCve, cveList)
-			if err != nil {
-				logger.Warn("Unable to determine CVE dispute status", slog.String("id", convertedCve.ID), slog.Any("err", err))
+// pickAffectedInformation merges information from nvdAffected into cve5Affected.
+// It matches affected packages by the repo URL in their version ranges.
+// If a match is found, it merges the version range information, preferring the entry
+// with more ranges. Unmatched nvdAffected packages are appended.
+// cve5Affected is modified in place.
+func pickAffectedInformation(cve5Affected *[]osvschema.Affected, nvdAffected []osvschema.Affected) {
```
another-rex: I think it'll be better to return a new slice here rather than modifying the existing one. It'll make it easier to test as well.

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -2,197 +2,348 @@
...
+	if len(nvdAffected) == 0 {
+		return
+	}
+	// If NVD has more affected packages, prefer it entirely.
```
another-rex, line 253: What if cve5 has additional different packages?

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -2,197 +2,348 @@
...
+	// cve5 := *cve5Affected
+	if len(*cve5Affected) == 0 || len(nvdAffected) > len(*cve5Affected) {
+		*cve5Affected = nvdAffected
+		return
+	}
+
+	nvdRepoMap := make(map[string][]osvschema.Range)
+	for _, affected := range nvdAffected {
+		// Assuming one range per affected for matching purposes.
```
another-rex: This comment doesn't seem to match what's happening below? This is looping through all the ranges

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -2,197 +2,348 @@
...
+		if len(affected.Ranges) > 0 && affected.Ranges[0].Repo != "" {
```
another-rex: Hmm this only checks the first range is not "", but then all the ranges are looped through. Wouldn't it be better to put the check within the loop?

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -2,197 +2,348 @@
...
+			for _, r := range affected.Ranges {
+				x, ok := nvdRepoMap[r.Repo]
+				if ok {
+					nvdRepoMap[r.Repo] = append(x, r)
+				} else {
+					nvdRepoMap[r.Repo] = append([]osvschema.Range{}, r)
```
another-rex: No need for append here.

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -2,197 +2,348 @@
...
 			}
-			if err == nil && !modified.IsZero() {
-				convertedCve.DatabaseSpecific = make(map[string]any)
-				convertedCve.DatabaseSpecific["isDisputed"] = true
+		}
+	}
+
+	cve5RepoMap := make(map[string][]osvschema.Range)
+	for _, affected := range *cve5Affected {
+		// Assuming one range per affected for matching purposes.
```
another-rex: Same as above.

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -2,197 +2,348 @@
...
+				}
 			}
 		}
+	}
 
-		for _, pkgInfo := range allParts[cveID] {
-			// skip debian and alpine parts, but still write out the CVEs.
-			if strings.HasPrefix(pkgInfo.Ecosystem, debianEcosystem) || strings.HasPrefix(pkgInfo.Ecosystem, alpineEcosystem) {
-				continue
+	newAffectedMap := make(map[string]osvschema.Affected)
```
another-rex: newRepoAffectedMap if the string key is the repo name?

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -2,197 +2,348 @@
...
+
+	for repo, cveRanges := range cve5RepoMap {
+		if nvdRanges, ok := nvdRepoMap[repo]; ok {
+			var newAffectedRanges []osvschema.Range
+
+			// Found a match. If NVD has more ranges, use its ranges.
```
another-rex, line 282: Same concern here as above, is this the correct assumption? What if cve5 has different ranges?

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -2,197 +2,348 @@
...
+				if finalIntro != "" || finalFixed != "" {
+					newRange := cves.BuildVersionRange(finalIntro, "", finalFixed)
+					newRange.Repo = repo
+					newRange.Type = osvschema.RangeGit // Preserve the repo
+					newAffectedRanges = append(newAffectedRanges, newRange)
+				}
+			}
+			// Remove from map so we know which NVD packages are left.
+			delete(nvdRepoMap, repo)
+			delete(cve5RepoMap, repo)
```
another-rex: I don't think you can delete from the map while ranging over it?

Also does deleting cve5RepoMap do anything here, I don't think it's used beyond this.

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -2,197 +2,348 @@
...
+		file, err := os.OpenFile(filePath, os.O_CREATE|os.O_WRONLY|os.O_TRUNC, 0644)
 		if err != nil {
 			logger.Fatal("Failed to create/open file to write", slog.Any("err", err))
 		}
 		encoder := json.NewEncoder(file)
 		encoder.SetIndent("", "  ")
 		err = encoder.Encode(osv)
 		if err != nil {
-			logger.Fatal("Failed to encode OSVs")
+			file.Close()
```
another-rex: just defer the file.Close(), and use Panicf instead of Fatal

vulnfeeds/cmd/combine-to-osv/main_test.go
another-rex, line 1: This should add some unit tests for pickAffectedInformation as well.

another-rex (changes requested):

vulnfeeds/cmd/combine-to-osv/run_combine_to_osv_convert.sh
jess-lowe: this is just the local copy so hardcoding should be fine.

vulnfeeds/cmd/combine-to-osv/main.go
jess-lowe: I'm going to look at improving and handling this in a future PR. For now, I just want to have some level of merging intelligence, so we can start populating the records, and will work on improving it later.

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -32,8 +32,8 @@ const (
 func main() {
 	logger.InitGlobalLogger()
 
-	cve5Path := flag.String("cve5Path", defaultCVE5Path, "Path to CVE file")
-	nvdPath := flag.String("nvdPath", defaultNVDOSVPath, "Path to CVE file")
+	cve5Path := flag.String("cve5Path", defaultCVE5Path, "Path to CVE5 OSV files")
+	nvdPath := flag.String("nvdPath", defaultNVDOSVPath, "Path to NVD OSV files")
 	osvOutputPath := flag.String("osvOutputPath", defaultOSVOutputPath, "Path to CVE file")
```
another-rex: This description is still wrong.

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -46,29 +46,47 @@ func main() {
 	allCVE5 := loadOSV(*cve5Path)
 	// Load NVD OSVs
 	allNVD := loadOSV(*nvdPath)
-
-	// nvdCVEs := vulns.LoadAllCVEs(defaultNVDOSVPath)
 	debianCVEs, err := listBucketObjects("osv-test-debian-osv", "/debian-cve-osv")
 	if err != nil {
 		logger.Warn("Failed to list debian cves", slog.Any("err", err))
```
another-rex, line 51: Should this be fatal?

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -46,29 +46,47 @@ func main() {
...
+
+	// run extract file name on each element in debianCVEs and alpineCVEs.
 	alpineCVEs, err := listBucketObjects("osv-test-cve-osv-conversion", "/alpine")
 	if err != nil {
 		logger.Warn("Failed to list alpine cves", slog.Any("err", err))
+	} else {
+		for i, filename := range alpineCVEs {
+			cve := extractCVEName(filename, "ALPINE-")
+			if cve != "" {
+				debianCVEs[i] = cve
```
another-rex: debianCVEs?

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -46,29 +46,47 @@ func main() {
...
+			}
+		}
 	}
 
-	noPkg := append(debianCVEs, alpineCVEs...) //nolint:gocritic
-	// Combine
-	combinedData := combineIntoOSV(allCVE5, allNVD, noPkg)
+	// this ensures the creation of CVEs even if they don't have packages
+	// to ensure Alpine and Debian CVEs have an upstream CVE.
+	mandatoryCVEIDs := append(debianCVEs, alpineCVEs...) //nolint:gocritic
```
another-rex, line 78: Add a comment about the nolint here

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -336,16 +351,18 @@ func writeOSVFile(osvData map[cves.CVEID]osvschema.Vulnerability, osvOutputPath
 		filePath := path.Join(osvOutputPath, string(vID)+".json")
 		file, err := os.OpenFile(filePath, os.O_CREATE|os.O_WRONLY|os.O_TRUNC, 0644)
 		if err != nil {
-			logger.Fatal("Failed to create/open file to write", slog.Any("err", err))
+			logger.Error("Failed to create/open file to write", slog.Any("err", err))
```
another-rex: Can't you use logger.Panic here?

another-rex (commented):
Nice!

vulnfeeds/cmd/combine-to-osv/main.go
jess-lowe: No because making sure non-Debian records are combined and served is considered higher priority than creating upstream-but-empty debian/Alpine CVE records IMO.

vulnfeeds/cmd/combine-to-osv/main.go
another-rex: umm has this been resolved?

jess-lowe: wtf I swear I updated this

jess-lowe: ok done now

another-rex (approved):
LGTM!

PR #4094, opened 2025-10-03
refactor: cve5 conversion to be more modular and extensible per CNA

jess-lowe:
This PR:
- Moves the cve5 conversion code out of the cmd folder
- creates a common.go folder for reused/common functions by the conversion
- initializes a factory-based design pattern for CNA based version extraction methods
    - Linux has its own extractor now
- splits apart some of the version extraction patterns into their own functions.


another-rex:
I think there's a conflict that needs to be resolved before we can review

vulnfeeds/cvelist2osv/extraction.go
```diff
@@ -0,0 +1,26 @@
...
+)
+
+// VersionExtractor defines the interface for different version extraction strategies.
+type VersionExtractor interface {
+	ExtractVersions(cve cves.CVE5, v *vulns.Vulnerability, metrics *ConversionMetrics, repos []string)
+	FindNormalAffectedRanges(affected cves.Affected, metrics *ConversionMetrics) ([]osvschema.Range, VersionRangeType)
+}
+
+// VersionExtractorFactory creates a VersionExtractor based on the CNA.
+type CNAVersionExtractor struct{}
```
another-rex: What's the benefit of having the factory here over a function?

vulnfeeds/cvelist2osv/linux_extractor.go
```diff
@@ -0,0 +1,230 @@
...
+	"strconv"
+	"strings"
+
+	"github.com/google/osv/vulnfeeds/cves"
+	"github.com/google/osv/vulnfeeds/vulns"
+	"github.com/ossf/osv-schema/bindings/go/osvschema"
+)
+
+// LinuxVersionExtractor provides the version extraction logic for Linux kernel CVEs.
+type LinuxVersionExtractor struct {
```
another-rex, line 15: Add a compile time check that this implements the interface.

e.g.

```
var _ VersionExtractor = &LinuxVersionExtractor{}
```

vulnfeeds/cvelist2osv/default_extractor.go
another-rex, line 1: I'm guessing most of this logic is covered by the converter_test.go already?

vulnfeeds/cvelist2osv/extraction.go
jess-lowe: Mostly just that it is easier to extend and separates the logic of choosing the right strategy from the strategies themselves.

another-rex: Hmm... I meant: 

```
func (f *CNAVersionExtractor) GetVersionExtractor(cna string) VersionExtractor {
	switch cna {
	case "Linux":
		return &LinuxVersionExtractor{}
	default:
		return &DefaultVersionExtractor{}
	}
}
```

vs

```
func GetVersionExtractor(cna string) VersionExtractor {
	switch cna {
	case "Linux":
		return &LinuxVersionExtractor{}
	default:
		return &DefaultVersionExtractor{}
	}
}
```

jess-lowe: hmm good point, I was thinking in a Java abstract way, but I guess since Go isnt OOP, it's fine without.

vulnfeeds/cvelist2osv/default_extractor.go
jess-lowe: yeah, it should functionally all work the same

PR #4128, opened 2025-10-07
ci: ensure filenames don't have undesired characters

G-Rath:
This should ensure that files are valid across OSs - I've gone with an inclusive pattern which purposely does not have some characters that are technically legal but that I don't think there's a good reason to use them in a filename.

That technically applies to a few of the included characters too since there are existing files that use them, but I am happy to rename those if folks are ok with banning the characters

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -32,9 +32,9 @@ jobs:
         with:
           persist-credentials: false
       - run: |
-          find . -mindepth 1 ! -regex '.*/[()#@,A-Za-z0-9._+-]*' \
-            | xargs -I '{}' bash -c \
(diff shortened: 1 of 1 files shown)


another-rex (approved):
Neat!

another-rex:
Hmm yeah we probably should ban `()+,` from the list of allowed characters

cuixq:
can we have this for osv-schema as well? I remember we countered some issue for invalid file names.

G-Rath:
> can we have this for osv-schema as well?

Yup I'm planning to apply it to all the osv repos, I just wanted to check opinions on the character list and implementation first

another-rex (approved):
Hmm yeah

PR #4130, opened 2025-10-08
fix: download ALL of the applicable vulns to operate on

jess-lowe:
Bad logic was only downloading the OSV records from CVE5 ending in 4 characters, despite CVEs usually having 4-5 characters. this downloads anything with the CVE structure, unfortunately including the .metrics.json files.

I would like to change this to get downloaded with go but for now, this will fix the cron job.

### DIFF
--- a/vulnfeeds/cmd/combine-to-osv/run_combine_to_osv_convert.sh
+++ b/vulnfeeds/cmd/combine-to-osv/run_combine_to_osv_convert.sh
@@ -19,7 +19,7 @@ set -eu
 INPUT_BUCKET="${INPUT_GCS_BUCKET:=cve-osv-conversion}"
 OUTPUT_BUCKET="${OUTPUT_GCS_BUCKET:=cve-osv-conversion}"
 
-OSV_OUTPUT="osv_output/"
+OSV_OUTPUT="osv-output"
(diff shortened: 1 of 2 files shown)


vulnfeeds/cmd/combine-to-osv/run_combine_to_osv_convert.sh
```diff
@@ -29,11 +29,11 @@ rm -rf $OSV_OUTPUT && mkdir -p $OSV_OUTPUT
 rm -rf $CVE5_OSV_OUTPUT && mkdir -p $CVE5_OSV_OUTPUT
 
 echo "Begin syncing NVD data from GCS bucket ${INPUT_BUCKET}"
-gcloud --no-user-output-enabled storage -q cp "gs://${INPUT_BUCKET}/nvd-osv/*-????.json" "${NVD_OSV_OUTPUT}"
+gcloud --no-user-output-enabled storage -q cp "gs://${INPUT_BUCKET}/nvd-osv/CVE-????-*.json" "${NVD_OSV_OUTPUT}"
```
another-rex, line 32: Bold of you to assume that year will only have 4 digits 😆

PR #4131, opened 2025-10-08
feat: Use a worker uploader instead of gsutils for combine-to-osv

jess-lowe:
This should make sure modified date is updated when we update the record, so a reimport is triggered.

### DIFF
--- a/vulnfeeds/cmd/alpine/main.go
+++ b/vulnfeeds/cmd/alpine/main.go
@@ -63,13 +63,13 @@ func main() {
 		bkt = storageClient.Bucket(*outputBucketName)
 	}
 	var wg sync.WaitGroup
-	vulnChan := make(chan *vulns.Vulnerability)
+	vulnChan := make(chan *osvschema.Vulnerability)
 
 	for range *numWorkers {
 		wg.Add(1)
 		go func() {
 			defer wg.Done()
-			vulns.Worker(ctx, vulnChan, bkt, *alpineOutputPath)
+			vulns.Worker(ctx, vulnChan, bkt, nil, *alpineOutputPath)
 		}()
 	}
 
@@ -78,7 +78,7 @@ func main() {
 			logger.Warn(fmt.Sprintf("Skipping %s as no affected versions found.", v.ID), slog.String("id", v.ID))
 			continue
 		}
-		vulnChan <- v
+		vulnChan <- &v.Vulnerability
 	}
 
 	close(vulnChan)
--- a/vulnfeeds/cmd/combine-to-osv/main.go
+++ b/vulnfeeds/cmd/combine-to-osv/main.go
@@ -10,15 +10,16 @@ import (
 	"io/fs"
 	"log/slog"
 	"os"
-	"path"
 	"path/filepath"
 	"strings"
+	"sync"
 
 	"slices"
 
 	"cloud.google.com/go/storage"
 	"github.com/google/osv/vulnfeeds/cves"
 	"github.com/google/osv/vulnfeeds/utility/logger"
+	"github.com/google/osv/vulnfeeds/vulns"
 	"github.com/ossf/osv-schema/bindings/go/osvschema"
 	"google.golang.org/api/iterator"
 )
@@ -34,7 +35,11 @@ func main() {
 
 	cve5Path := flag.String("cve5Path", defaultCVE5Path, "Path to CVE5 OSV files")
 	nvdPath := flag.String("nvdPath", defaultNVDOSVPath, "Path to NVD OSV files")
-	osvOutputPath := flag.String("osvOutputPath", defaultOSVOutputPath, "Local output path of combined OSV files")
+	osvOutputPath := flag.String("osvOutputPath", defaultOSVOutputPath, "Local output path of combined OSV files, or GCS prefix if uploading.")
+	outputBucketName := flag.String("output_bucket", "cve-osv-conversion", "The GCS bucket to write to.")
+	overridesBucketName := flag.String("overrides_bucket", "cve-osv-conversion", "The GCS bucket to read overrides from.")
+	uploadToGCS := flag.Bool("uploadToGCS", false, "If true, upload to GCS bucket instead of writing to local disk.")
+	numWorkers := flag.Int("num_workers", 64, "Number of workers to process records")
 	flag.Parse()
 
 	err := os.MkdirAll(*osvOutputPath, 0755)
@@ -77,7 +82,36 @@ func main() {
 	// just want to combine these two arrays with a more descriptive name.
 	mandatoryCVEIDs := append(debianCVEs, alpineCVEs...) //nolint:gocritic
 	combinedData := combineIntoOSV(allCVE5, allNVD, mandatoryCVEIDs)
-	writeOSVFile(combinedData, *osvOutputPath)
+
+	ctx := context.Background()
+	var outBkt, overridesBkt *storage.BucketHandle
+	if *uploadToGCS {
+		storageClient, err := storage.NewClient(ctx)
+		if err != nil {
+			logger.Fatal("Failed to create storage client", slog.Any("err", err))
+		}
+		outBkt = storageClient.Bucket(*outputBucketName)
+		overridesBkt = storageClient.Bucket(*overridesBucketName)
+	}
+
+	var wg sync.WaitGroup
+	vulnChan := make(chan *osvschema.Vulnerability, *numWorkers)
+
+	for i := 0; i < *numWorkers; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			vulns.Worker(ctx, vulnChan, outBkt, overridesBkt, *osvOutputPath)
+		}()
+	}
+
+	for _, v := range combinedData {
+		vulnChan <- &v
+	}
+
+	close(vulnChan)
+	wg.Wait()
+	logger.Info("Successfully processed OSV files", slog.Int("count", len(combinedData)))
 }
 
 // extractCVEName extracts the CVE name from a given filename and prefix.
@@ -344,25 +378,4 @@ func getRangeBoundaryVersions(events []osvschema.Event) (introduced, fixed strin
 	}
 
 	return introduced, fixed
-}
-
-// writeOSVFile writes out the given osv objects into individual json files
-func writeOSVFile(osvData map[cves.CVEID]osvschema.Vulnerability, osvOutputPath string) {
-	for vID, osv := range osvData {
-		filePath := path.Join(osvOutputPath, string(vID)+".json")
-		file, err := os.OpenFile(filePath, os.O_CREATE|os.O_WRONLY|os.O_TRUNC, 0644)
-		if err != nil {
-			logger.Panic("Failed to create/open file to write", slog.Any("err", err))
-		}
-		defer file.Close()
-
-		encoder := json.NewEncoder(file)
-		encoder.SetIndent("", "  ")
-		err = encoder.Encode(osv)
-		if err != nil {
-			logger.Panic("Failed to encode OSVs", slog.Any("err", err))
-		}
-	}
-
-	logger.Info("Successfully written OSV files", slog.Int("count", len(osvData)))
 }
--- a/vulnfeeds/cmd/debian/main.go
+++ b/vulnfeeds/cmd/debian/main.go
@@ -69,13 +69,13 @@ func main() {
 	}
 
 	var wg sync.WaitGroup
-	vulnChan := make(chan *vulns.Vulnerability)
+	vulnChan := make(chan *osvschema.Vulnerability)
 
 	for range *numWorkers {
 		wg.Add(1)
 		go func() {
 			defer wg.Done()
-			vulns.Worker(ctx, vulnChan, bkt, *debianOutputPath)
+			vulns.Worker(ctx, vulnChan, bkt, nil, *debianOutputPath)
 		}()
 	}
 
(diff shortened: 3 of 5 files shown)


vulnfeeds/vulns/cveworker.go
```diff
@@ -6,83 +6,104 @@ import (
...
 
-		if err := wc.Close(); err != nil {
-			logger.Error("failed to close GCS writer", slog.String("id", vulnID), slog.Any("err", err))
-			continue
+			if err := wc.Close(); err != nil {
+				logger.Error("failed to close GCS writer", slog.String("id", vulnID), slog.Any("err", err))
+				continue
+			}
+		} else {
+			// Write to local disk
```
another-rex, line 166: Put write to local disk at the top, and continue afterwards to reduce nesting.

vulnfeeds/vulns/cveworker.go
```diff
@@ -6,83 +6,104 @@ import (
...
-		hash := sha256.Sum256(buf)
+		hash := sha256.Sum256(preModifiedBuf)
 		hexHash := hex.EncodeToString(hash[:])
+		filename := vulnID + ".json"
 
-		objName := path.Join(outputDir, vulnID+".json")
-		obj := bkt.Object(objName)
+		if outBkt != nil {
+			// Upload to GCS
+			if overridesBkt != nil {
```
another-rex: This is skipping the file rather than uploading from the override bucket.

another-rex: I think this section should also be independent of the outBkt section (i.e. we should override locally as well if a override bucket is provided.)

vulnfeeds/vulns/cveworker.go
another-rex: Also, this does not need the hashing logic, so can be placed about that.

vulnfeeds/vulns/cveworker.go
```diff
@@ -6,83 +6,104 @@ import (
...
 		}
 
-		hash := sha256.Sum256(buf)
+		hash := sha256.Sum256(preModifiedBuf)
 		hexHash := hex.EncodeToString(hash[:])
+		filename := vulnID + ".json"
 
-		objName := path.Join(outputDir, vulnID+".json")
-		obj := bkt.Object(objName)
+		if outBkt != nil {
```
another-rex: I'm also thinking both write to local file, and upload to bucket, can be their own private functions, which will make this big Worker function easier to read.

vulnfeeds/upload/cveworker.go
```diff
@@ -6,83 +6,104 @@ import (
...
 const (
-	// HashMetadataKey is the key for the sha256 hash in the GCS object metadata.
+	// hashMetadataKey is the key for the sha256 hash in the GCS object metadata.
 	hashMetadataKey = "sha256-hash"
 )
 
-// Worker is a generic worker that processes vulnerabilities from a channel and uploads them to a GCS bucket.
-func Worker(ctx context.Context, vulnChan <-chan *Vulnerability, bkt *storage.BucketHandle, outputDir string) {
+// Worker is a generic worker that processes OSV vulnerabilities from a channel.
+// It can upload them to a GCS bucket or write them to disk.
```
another-rex, line 134: Probably also describe the hashing diff check here as well, and how it only changes the modified time if content is different.

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -77,7 +82,36 @@ func main() {
...
+		}()
+	}
+
+	for _, v := range combinedData {
+		vulnChan <- &v
+	}
+
+	close(vulnChan)
+	wg.Wait()
+	logger.Info("Successfully processed OSV files", slog.Int("count", len(combinedData)))
```
another-rex: Hmm is this logic repeated for Alpine and Debian as well? Should we be refactoring this out into a shared function as well...

another-rex (commented):
Can you describe this override logic in the PR description as well? Thanks!

vulnfeeds/cmd/alpine/main.go
```diff
@@ -53,36 +51,17 @@ func main() {
...
-
-	for range *numWorkers {
-		wg.Add(1)
-		go func() {
-			defer wg.Done()
-			vulns.Worker(ctx, vulnChan, bkt, nil, *alpineOutputPath)
-		}()
-	}
-
+	var vulnerabilities []*osvschema.Vulnerability //nolint:prealloc
```
another-rex: Just prealloc this with the len(osvVulnerabilities)

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -84,34 +83,12 @@ func main() {
...
-
-	for i := 0; i < *numWorkers; i++ {
-		wg.Add(1)
-		go func() {
-			defer wg.Done()
-			vulns.Worker(ctx, vulnChan, outBkt, overridesBkt, *osvOutputPath)
-		}()
-	}
-
+	var vulnerabilities []*osvschema.Vulnerability //nolint:prealloc
```
another-rex: Same here, prealloc please.

vulnfeeds/cmd/debian/main.go
```diff
@@ -57,40 +55,19 @@ func main() {
...
-		wg.Add(1)
-		go func() {
-			defer wg.Done()
-			vulns.Worker(ctx, vulnChan, bkt, nil, *debianOutputPath)
-		}()
-	}
-
 	osvCVEs := generateOSVFromDebianTracker(debianData, debianReleaseMap, allCVEs)
 
+	var vulnerabilities []*osvschema.Vulnerability //nolint:prealloc
```
another-rex: Prealloc

vulnfeeds/upload/cveworker.go
```diff
@@ -6,83 +6,206 @@ import (
...
 	hashMetadataKey = "sha256-hash"
 )
 
-// Worker is a generic worker that processes vulnerabilities from a channel and uploads them to a GCS bucket.
-func Worker(ctx context.Context, vulnChan <-chan *Vulnerability, bkt *storage.BucketHandle, outputDir string) {
+// writeToDisk writes the vulnerability to a local file.
+func writeToDisk(v *osvschema.Vulnerability, preModifiedBuf []byte, outputPrefix string) {
+	filename := v.ID + ".json"
+	filePath := path.Join(outputPrefix, filename)
+	err := os.WriteFile(filePath, preModifiedBuf, 0600)
```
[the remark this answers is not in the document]
another-rex: Yeah use 644 here (but don't apply suggestion, it looks like it got rid of the indentation for some reason :\

vulnfeeds/upload/cveworker.go
```diff
@@ -21,9 +23,118 @@ const (
...
+		}
+
+		var overrideV osvschema.Vulnerability
+		if err := json.Unmarshal(overrideBuf, &overrideV); err != nil {
+			logger.Warn("failed to unmarshal override object, using original", slog.String("id", v.ID), slog.Any("err", err))
+			return v, nil, nil
+		}
+
+		return &overrideV, overrideBuf, nil
+	} else if !errors.Is(err, storage.ErrObjectNotExist) {
```
another-rex: No need for else here.

vulnfeeds/upload/cveworker.go
```diff
@@ -21,9 +23,118 @@ const (
...
+			return v, nil, nil
+		}
+
+		var overrideV osvschema.Vulnerability
+		if err := json.Unmarshal(overrideBuf, &overrideV); err != nil {
+			logger.Warn("failed to unmarshal override object, using original", slog.String("id", v.ID), slog.Any("err", err))
+			return v, nil, nil
+		}
+
+		return &overrideV, overrideBuf, nil
```
another-rex: this is a bit too much logic imo, it should just use the override as it without checking. If it is invalid it will fail at the import stage which is probably what we want anyway.

another-rex: Also, I think we can flip the logic here, err != nil and return early.

vulnfeeds/upload/cveworker.go
another-rex, line 1: This whole thing should probably go into a separate package, `upload` or something.

vulnfeeds/vulns/cveworker.go
```diff
@@ -32,78 +143,69 @@ func Worker(ctx context.Context, vulnChan <-chan *osvschema.Vulnerability, outBk
...
+			writeToDisk(vulnToProcess, preModifiedBuf, outputPrefix)
+			continue
 		}
+		// Upload to GCS
+		uploadToGCS(ctx, vulnToProcess, preModifiedBuf, outBkt, outputPrefix)
 	}
 }
+
+// Run starts the workers to process vulnerabilities.
+func Run(
```
another-rex: Is this comment correct? Is any processing happening here, or is this just uploading? If it's only uploading Run probably isn't a good name either.

vulnfeeds/upload/cveworker.go
```diff
@@ -32,78 +143,69 @@ func Worker(ctx context.Context, vulnChan <-chan *osvschema.Vulnerability, outBk
 			continue
 		}
 
-		// Marshal before setting modified time to generate hash.
-		preModifiedBuf, err := json.MarshalIndent(v, "", "  ")
+		vulnToProcess, preModifiedBuf, err := handleOverride(ctx, v, overridesBkt)
```
another-rex: The if statement for overrideBkt not being nil should be out here.

another-rex: General good rule of thumb is push if's up and for's down.

vulnfeeds/upload/cveworker.go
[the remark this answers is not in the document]
jess-lowe: Linter disagrees so not changing

vulnfeeds/upload/cveworker.go
```diff
@@ -142,10 +142,15 @@ func Worker(ctx context.Context, vulnChan <-chan *osvschema.Vulnerability, outBk
 			logger.Warn("Skipping OSV record as no affected versions found.", slog.String("id", vulnID))
 			continue
 		}
+		var vulnToProcess *osvschema.Vulnerability
```
another-rex: vulnToProcess should be = to v by default if there's no override right?

jess-lowe: good point

vulnfeeds/upload/cveworker.go
```diff
@@ -6,83 +6,206 @@ import (
...
+		if closeErr := wc.Close(); closeErr != nil {
+			logger.Error("failed to close GCS writer after write error", slog.String("id", vulnID), slog.Any("err", closeErr))
+		}
+
+		return
+	}
+
+	if err := wc.Close(); err != nil {
+		logger.Error("failed to close GCS writer", slog.String("id", vulnID), slog.Any("err", err))
+	}
```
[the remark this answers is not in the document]
another-rex: This is fine. Close is not a critical operation

vulnfeeds/upload/cveworker.go
```diff
@@ -0,0 +1,216 @@
...
+			logger.Warn("Skipping OSV record as no affected versions found.", slog.String("id", vulnID))
+			continue
+		}
+		vulnToProcess := v
+		var preModifiedBuf []byte
+		var err error
+
+		if overridesBkt != nil {
+			vulnToProcess, preModifiedBuf, err = handleOverride(ctx, v, overridesBkt)
+			if err != nil {
```
another-rex, line 152: Do we want to log here for this error?

vulnfeeds/upload/cveworker.go
```diff
@@ -0,0 +1,216 @@
...
+				logger.Error("failed to marshal vulnerability", slog.String("id", vulnID), slog.Any("err", err))
+				continue
+			}
+		}
+
+		if outBkt == nil {
+			// Write to local disk
+			writeToDisk(vulnToProcess, preModifiedBuf, outputPrefix)
+			continue
+		}
```
another-rex, line 171: imo this should be if else rather than continue, though the linter might complain about it.

vulnfeeds/upload/cveworker.go
```diff
@@ -0,0 +1,216 @@
...
+		vulnToProcess := v
+		var preModifiedBuf []byte
+		var err error
+
+		if overridesBkt != nil {
+			vulnToProcess, preModifiedBuf, err = handleOverride(ctx, v, overridesBkt)
+			if err != nil {
+				continue
+			}
+		}
```
another-rex: this should probably be if else, rather than checking preModifiedBuf is nil.

vulnfeeds/upload/cveworker.go
```diff
@@ -0,0 +1,216 @@
...
+		logger.Error("failed to close GCS writer", slog.String("id", vulnID), slog.Any("err", err))
+	}
+}
+
+// handleOverride checks for and applies a vulnerability override if it exists.
+// It returns the vulnerability to process, a pre-marshalled buffer if an override was used,
+// and an error if a critical failure occurred.
+func handleOverride(ctx context.Context, v *osvschema.Vulnerability, overridesBkt *storage.BucketHandle) (*osvschema.Vulnerability, []byte, error) {
+	filename := v.ID + ".json"
+	overrideObj := overridesBkt.Object(path.Join("osv-output-overrides", filename))
```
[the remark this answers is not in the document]
another-rex: Just make this a const at the top.

PR #4135, opened 2025-10-08
feat: add Julia as a Semver ecosystem

mbauman:


osv/ecosystems/_ecosystems.py
```diff
@@ -47,6 +47,7 @@
     'Go': SemverEcosystem,
     'Hackage': Hackage,
     'Hex': SemverEcosystem,
+    'Julia': SemverEcosystem,
```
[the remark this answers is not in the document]
mbauman: In what contexts are these used? We could either link to the package information page on JuliaHub or to a page that redirects directly to the package source.

another-rex:
Thanks for the PR! Please ignore gemini-bot's review there, we'll probably make it manually invoked from now on.

> I do not believe we need to extend purl_helpers.py; we are generating our OSV records with purls attached, and the proposed purl spec requires the package UUID. It would be possible to look that up in purl_helpers.py through an HTTP GET to a JSON file, but that would represent a new thing that no other ecosystem does.

I think we still need it for _parsing_ the purls though, as to do queries we convert the purls to `ecosystem/name/version`.
It should be relatively simple, we do not actually need the UUID logic if you provide the purls in the records themselves. So just adding

```
    'Julia':
        EcosystemPURL('julia', None),
```

to the list should be enough.

osv/purl_helpers_test.py
```diff
@@ -98,6 +98,18 @@ def tests_package_to_purl(self):
     self.assertEqual('pkg:hex/acme/foo',
                      purl_helpers.package_to_purl('Hex', 'acme/foo'))
 
+    self.assertEqual('pkg:julia/Example',
+                     purl_helpers.package_to_purl('Julia', 'Example'))
+
+    self.assertEqual('pkg:julia/Example?uuid=7876af07-990d-54b4-ab0e-23690620f79a',
+                     purl_helpers.package_to_purl('Julia', 'Example'))
+
+    self.assertEqual('pkg:julia/Dates?uuid=ade2ca70-3891-5945-98fb-dc099432e06a',
```
another-rex: I don't think this will pass, as we won't generate the uuid when doing package_to_purl. That's fine since we don't use package_to_purl function if there is an existing purl on the record.

osv/purl_helpers_test.py
```diff
@@ -98,6 +98,12 @@ def tests_package_to_purl(self):
     self.assertEqual('pkg:hex/acme/foo',
                      purl_helpers.package_to_purl('Hex', 'acme/foo'))
 
+    self.assertEqual('pkg:julia/Example',
+                     purl_helpers.package_to_purl('Julia', 'Example'))
+
+    self.assertEqual('pkg:julia/Dates@1.9.0',
```
another-rex:
```suggestion
    self.assertEqual('pkg:julia/Dates',
```

mbauman:
Great, added!

another-rex:
Just need to run `yapf -i purl_helpers_test.py`, otherwise LGTM :)

another-rex:
Test error should be fixed when we update our osv-schema submodule, no action required from your side :).

another-rex:
Can you press the merge from main button? The tests should be passing after that. (Also, you can tick a box to give maintainers permission to push to the PR branch, and we can do the updating directly.)

mbauman:
Yeah, I usually do check that box, but I think because this one came from a fork on our company org, we have that option turned off. I can't check it here.  You are absolutely welcome to squash/cherry-pick/merge my work onto a branch of yours if that makes it any easier.

PR #4139, opened 2025-10-09
fix: remove redundance CVEIsDisputed code and add check in cve5 conversion

jess-lowe:
Previous refactors removed 'isDisputed' from databaseSpecific. Adding it back and removing the old redundant code.

This unfortunately doesn't cover NVD records that don't also have a CVE5 record for the combination logic, until we enable all CNAs


another-rex:
Can we still get the isDisputed information from the NVDs?

jess-lowe:
> Can we still get the isDisputed information from the NVDs?

we used to get the data by cloning the cvelist - i hesitate to do that if we are eventually going to enable conversion for all cnas, and will get the data that way anyway

PR #4149, opened 2025-10-12
fix: preserve crates.io build metadata

### LINKED ISSUE #4143: Rust Crate versions missing metadata
**Describe the bug**
Vulnerability reports for Rust crates are missing the metadata value (https://doc.rust-lang.org/cargo/reference/manifest.html#the-version-field) in the version field under 'affected/ranges' (and possibly other places).

**To Reproduce**
Look at the report for RUSTSEC-2024-0013 (https://api.osv.dev/v1/vulns/RUSTSEC-2024-0013). It shows that it was fixed in version 0.16.2 however the correct way to identity this version is with the additional metadata of '+1.7.2'. Without that additional metadata, there is no way to resolve the correct crate.io repo of https://crates.io/crates/libgit2-sys/0.16.2+1.7.2 because just https://crates.io/crates/libgit2-sys/0.16.2 doesn't exist. 

**Expected behaviour**
The metadata must be included in order to resolve the package from other sources as well as perform proper vulnerability management. Because in this case the follow PURLs don't match

pkg:cargo/libgit2-sys@0.16.2 != pkg:cargo/libgit2-sys@0.16.2%2B1.7.2

**Additional context**
Your sister service (deps.dev) lists the versions properly (https://deps.dev/cargo/libgit2-sys/0.16.2%2B1.7.2) so there should be some consistency.

cnaples79:
## Summary
- add a crates.io ecosystem helper that resolves versions with build metadata
- enrich semver ranges for crates.io packages using the helper during impact analysis
- cover the new helper and analysis flow with unit tests

Fixes #4143

### DIFF
--- a/osv/ecosystems/_ecosystems.py
+++ b/osv/ecosystems/_ecosystems.py
@@ -19,6 +19,7 @@ from .ecosystems_base import EnumerableEcosystem, OrderedEcosystem
 from .alpine import Alpine, APK
 from .bioconductor import Bioconductor
 from .cran import CRAN
+from .crates import CratesIO
 from .debian import Debian, DPKG
 from .haskell import Hackage, GHC
 from .maven import Maven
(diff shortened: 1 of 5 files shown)


michaelkedar:
Thanks for your interest in contributing to OSV.
After discussing with the team, we've decided that we do not wish to implement this behaviour.
We do not believe we should be modifying the versions in ranges we receive from upstream data sources.

cnaples79:
@michaelkedar no worries, thanks for the feedback.

PR #4150, opened 2025-10-13
feat(cve5): improve storing of conversion outcomes

jess-lowe:
Started tracking counts of unresolved and resolved version metrics and refactored where versions are extracted. We will be able to use this later to do some analysis on successes, regressions, progression of conversions later.

Moving the version extraction call allows us to make a base OSV record using CVE5 data without extracting version data. This means we are able to better control whether or not we actually want to attempt to extract versioning information, for example if the cna is out of scope for extraction.

### DIFF
--- a/vulnfeeds/cvelist2osv/common.go
+++ b/vulnfeeds/cvelist2osv/common.go
@@ -87,7 +87,7 @@ func resolveVersionToCommit(cveID cves.CVEID, version, versionType, repo string,
 	logger.Info("Attempting to resolve version to commit", slog.String("cve", string(cveID)), slog.String("version", version), slog.String("type", versionType), slog.String("repo", repo))
 	commit, err := git.VersionToCommit(version, normalizedTags)
 	if err != nil {
-		logger.Warn("Failed to get Git commit for version", slog.String("cve", string(cveID)), slog.String("version", version), slog.String("type", versionType), slog.String("repo", repo))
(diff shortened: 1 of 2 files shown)


vulnfeeds/cvelist2osv/common.go
```diff
@@ -73,7 +87,7 @@ func resolveVersionToCommit(cveID cves.CVEID, version, versionType, repo string,
 	logger.Info("Attempting to resolve version to commit", slog.String("cve", string(cveID)), slog.String("version", version), slog.String("type", versionType), slog.String("repo", repo))
 	commit, err := git.VersionToCommit(version, normalizedTags)
 	if err != nil {
-		logger.Warn("Failed to get Git commit for version", slog.String("cve", string(cveID)), slog.String("version", version), slog.String("type", versionType), slog.String("repo", repo), slog.Any("err", err))
```
another-rex, line 76: why was err removed here?

vulnfeeds/cves/versions.go
```diff
@@ -1069,7 +1069,7 @@ func ReposFromReferencesCVEList(cve string, refs []Reference, tagDenyList []stri
 		repos = append(repos, repo)
 	}
 	if len(repos) == 0 {
-		notes = append(notes, "[%s]: Failed to identify any repos using references")
+		notes = append(notes, "[%s]: Failed to identify any repos using references", cve)
```
another-rex: What is [%s] here for? Is this missing a fmt.Sprintf?

vulnfeeds/cvelist2osv/common.go
jess-lowe: Because it clutters my logs with all of the tag to commit combinations and is annoying to skim read

PR #4158, opened 2025-10-14
fix: Add support for CVSS V4 metrics in NVD conversion

### LINKED ISSUE #4157: Support CVSS4 in NVD conversion

jess-lowe:
Closes #4157

### DIFF
--- a/vulnfeeds/cves/nvd2.go
+++ b/vulnfeeds/cves/nvd2.go
@@ -206,26 +206,8 @@ type CVSSV2 struct {
 	UserInteractionRequired bool         `json:"userInteractionRequired,omitempty" mapstructure:"userInteractionRequired,omitempty" yaml:"userInteractionRequired,omitempty"`
 }
 
-// CVSS V3.0 score. (hand-generated)
-type CVSSV30 struct {
-	Source              string       `json:"source"                        mapstructure:"source"                        yaml:"source"`
-	Type                string       `json:"type"                          mapstructure:"type"                          yaml:"type"`
-	CVSSData            CVSS         `json:"cvssData"                      mapstructure:"cvssData"                      yaml:"cvssData"`
-	ExploitabilityScore *DefSubscore `json:"exploitabilityScore,omitempty" mapstructure:"exploitabilityScore,omitempty" yaml:"exploitabilityScore,omitempty"`
-	ImpactScore         *DefSubscore `json:"impactScore,omitempty"         mapstructure:"impactScore,omitempty"         yaml:"impactScore,omitempty"`
-}
-
-// CVSS V3.1 score. (hand-generated)
-type CVSSV31 struct {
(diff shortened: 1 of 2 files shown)


vulnfeeds/cves/nvd2.go
```diff
@@ -224,6 +224,15 @@ type CVSSV31 struct {
 	ImpactScore         *DefSubscore `json:"impactScore,omitempty"         mapstructure:"impactScore,omitempty"         yaml:"impactScore,omitempty"`
 }
 
+// CVSS V4.0 score. (hand-generated)
+type CVSSV40 struct {
```
michaelkedar: Aren't these CVSSVX structs all the same?
Should we just consolidate them?

vulnfeeds/cves/nvd2.go
```diff
@@ -234,6 +243,9 @@ type CVEItemMetrics struct {
 
 	// CVSS V3.1 score.
 	CVSSMetricV31 []CVSSV31 `json:"cvssMetricV31,omitempty" mapstructure:"cvssMetricV31,omitempty" yaml:"cvssMetricV31,omitempty"`
+
+	// CVSS V4 score
+	CVSSMetricV4 []CVSSV40 `json:"cvssMetricV40,omitempty" mapstructure:"cvssMetricV31,omitempty" yaml:"cvssMetricV31,omitempty"`
```
another-rex: map structure and yaml?

vulnfeeds/cves/nvd2.go
```diff
@@ -206,17 +206,8 @@ type CVSSV2 struct {
...
-	Type                string       `json:"type"                          mapstructure:"type"                          yaml:"type"`
-	CVSSData            CVSS         `json:"cvssData"                      mapstructure:"cvssData"                      yaml:"cvssData"`
-	ExploitabilityScore *DefSubscore `json:"exploitabilityScore,omitempty" mapstructure:"exploitabilityScore,omitempty" yaml:"exploitabilityScore,omitempty"`
-	ImpactScore         *DefSubscore `json:"impactScore,omitempty"         mapstructure:"impactScore,omitempty"         yaml:"impactScore,omitempty"`
-}
-
-// CVSS V3.1 score. (hand-generated)
-type CVSSV31 struct {
+// CVSS Score to encapsulate V3.0, V3.1, V4.0 structs
+type BaseCVSSNVD struct {
```
another-rex, line 210: If 3.0, 3.1, and 4.0 are the same fields, what's the difference between the versions? Or are we missing the fields that this actually affects?

jess-lowe: Basically, this is only here afaik, so we can prioritize which severity score we are getting. The actual field is just the vector string which is just a string so doesn't really type that well.

another-rex: Ah ok, we only care about Type and CVSSData.VectorString? If so, let's just comment out the other fields to avoid confusion.

PR #4160, opened 2025-10-14
fix(proto): add database_specific to affected ranges

cuixq:
This was introduced to the schema in https://github.com/ossf/osv-schema/pull/37.
This PR also patches a few typo fixes in https://github.com/ossf/osv-schema/blob/main/proto/vulnerability.proto.

Also updates grpc libraries to version `1.75.0` that aligns with GRPC_GENERATED_VERSION.


another-rex:
Hmm... time to update our grpc library? I thought dependabot should already be updating to the latest version.

another-rex (approved):

PR #4162, opened 2025-10-15
fix: reduce duplication in CVE conversion and combine-to-osv

jess-lowe:
Duplication exists when combining due to unnormalized repo strings: 
image

This PR will also make absolutely sure that only unique ranges get added to the record.


another-rex (approved):
Can you add some description to the PR.

PR #4167, opened 2025-10-15
Extract and add CWE IDs from CVE problem-types

### LINKED ISSUE #4159: Support CWE in CVE5 conversion.
It would be good if we could get at least the CWEID in the database_specific field of the CVE5 converted record. 

https://github.com/google/osv.dev/blob/9025fd1f9db05bb0d2309f616fd668b5b00938f0/vulnfeeds/cvelist2osv/converter.go#L124

Example of where it exists in CVE5 record:
https://github.com/CVEProject/cvelistV5/blob/f24cafc72f1735224e1b18992ef89e2e266fe6e6/cves/2025/9xxx/CVE-2025-9968.json#L79

Mritunjay09:
Closes #4159
Introduces the attachCWEs
 function to extract CWE IDs from both CNA and ADP problem-types in CVE5 records and store them in the Vulnerability's DatabaseSpecific field. Also updates the ProblemTypes struct to include a CWEID field for more accurate parsing.

- Updated ProblemTypes struct to include `CweID` field to match CVE5 JSON schema.
- Updated attachCWEs() function to store both CWE ID and human-readable description.


vulnfeeds/cvelist2osv/converter.go
```diff
@@ -76,6 +77,67 @@ func extractConversionMetrics(cve cves.CVE5, refs []osvschema.Reference, metrics
 	// TODO(jesslowe): Add more analysis based on ADP containers, CVSS, KEV, CWE, etc.
 }
 
+//attachCWEs extracts and adds CWE IDs from the CVE5 problem-types
+func attachCWEs(v *vulns.Vulnerability, cve cves.CVE5, metrics *ConversionMetrics){
+	foundCWEs := make(map[string]struct{})
+	addCWE := func(id string){
+		id = strings.TrimSpace(id)
```
jess-lowe: I'm not sure this is necessary - I just want the value as is. I'm not too concerned about sanity checking beyond it not being a duplicate and that the Type == CWE.

vulnfeeds/cvelist2osv/converter.go
```diff
@@ -76,6 +77,67 @@ func extractConversionMetrics(cve cves.CVE5, refs []osvschema.Reference, metrics
...
+		if id == ""{
+			return
+		}
+		if !strings.HasPrefix(id,"CWE-"){
+			id = "CWE-" + id
+		}
+		foundCWEs[id] = struct{}{}
+	}
+	re := regexp.MustCompile(`CWE-\d+`)
```
jess-lowe: Regex isn't necessary, as above

vulnfeeds/cvelist2osv/converter.go
```diff
@@ -76,6 +77,67 @@ func extractConversionMetrics(cve cves.CVE5, refs []osvschema.Reference, metrics
...
+	// iterate over CNA problem-types
+	for _, pt := range cve.Containers.CNA.ProblemTypes {
+		for _, desc := range pt.Descriptions {
+			// desc is the anonymous struct{Type,Lang,Description string}
+			if desc.Type == "CWE" {
+				if desc.CWEID != ""{
+					addCWE(desc.CWEID)
+				}else{
+					addCWE(desc.Description)
+				}
```
jess-lowe: If it doesn't exist in the CWEID, we should just continue, and not worry about checking description.

vulnfeeds/cvelist2osv/converter.go
```diff
@@ -76,6 +77,67 @@ func extractConversionMetrics(cve cves.CVE5, refs []osvschema.Reference, metrics
...
+			}
+			for _, m := range re.FindAllString(desc.Description, -1) {
+				addCWE(m)
+			}
```
jess-lowe: As above - not necessary.

vulnfeeds/cvelist2osv/converter.go
```diff
@@ -76,6 +77,67 @@ func extractConversionMetrics(cve cves.CVE5, refs []osvschema.Reference, metrics
...
+		}
+	}
+	// iterate over ADP problem-types
+	for _, adp := range cve.Containers.ADP {
+		for _, pt := range adp.ProblemTypes {
+			for _, desc := range pt.Descriptions {
+				if desc.Type == "CWE" {
+					addCWE(desc.Description) // CWE ID is stored in Description field
```
jess-lowe: This should be checking the desc.CWEID field, not the description.

vulnfeeds/cvelist2osv/converter.go
```diff
@@ -76,6 +77,67 @@ func extractConversionMetrics(cve cves.CVE5, refs []osvschema.Reference, metrics
...
+				}
+				for _, m := range re.FindAllString(desc.Description, -1) {
+					addCWE(m)
+				}
```
jess-lowe: As above, not necessary

vulnfeeds/cvelist2osv/converter.go
```diff
@@ -76,6 +77,67 @@ func extractConversionMetrics(cve cves.CVE5, refs []osvschema.Reference, metrics
 	// TODO(jesslowe): Add more analysis based on ADP containers, CVSS, KEV, CWE, etc.
 }
 
+//attachCWEs extracts and adds CWE IDs from the CVE5 problem-types
+func attachCWEs(v *vulns.Vulnerability, cve cves.CVE5, metrics *ConversionMetrics){
+	foundCWEs := make(map[string]struct{})
```
jess-lowe: this can just be a slice of strings, and then to make sure there's no duplicates, just do `slices.Sort` and then `slices.Compact` near the end of the function.

vulnfeeds/cvelist2osv/converter.go
```diff
@@ -76,6 +77,67 @@ func extractConversionMetrics(cve cves.CVE5, refs []osvschema.Reference, metrics
 	// TODO(jesslowe): Add more analysis based on ADP containers, CVSS, KEV, CWE, etc.
 }
 
+//attachCWEs extracts and adds CWE IDs from the CVE5 problem-types
+func attachCWEs(v *vulns.Vulnerability, cve cves.CVE5, metrics *ConversionMetrics){
```
jess-lowe: instead of passing in the whole CVE record, you could just pass in the CVE.Containers.CNA type, and utilize the existing for loop in the FromCVE5 function

vulnfeeds/cvelist2osv/converter.go
```diff
@@ -76,6 +77,67 @@ func extractConversionMetrics(cve cves.CVE5, refs []osvschema.Reference, metrics
...
+	list := make([]string, 0, len(foundCWEs))
+	for id := range foundCWEs{
+		list = append(list, id)
+	}
+	slices.Sort(list)
+	if v.DatabaseSpecific == nil{
+		v.DatabaseSpecific = make(map[string]any)
+	}
+	
+	v.DatabaseSpecific["cwes"] = list
```
jess-lowe:
```suggestion
	v.DatabaseSpecific["cwe_ids"] = list
```
nit just to align with some existing GHSA records

Mritunjay09:
Updated attachCWEs function per review feedback. Now uses only the CWEID field and removes regex/description parsing. The PR is ready for another review.

jess-lowe:
Awesome thanks for that improvement! It looks like the tests are failing as they will need to be updated with the additional data. To run the tests to check, use `make vulnfeed-tests`.

To check for linting errors, run `make lint`

Mritunjay09:
@jess-lowe made all the necessary changes and also checked that lint and vulnfeed-tests return no error

jess-lowe (approved):
LGTM

Mritunjay09:
Let me know if there's anything else I can contribute to.

Issue #4191, opened 2025-10-16
EEF CNA DataSource


maennchen:
- [x] Prepare your data \- refer to the [OSV Schema](https://ossf.github.io/osv-schema/) documentation for information on how to properly format the data so it can be accepted.  
        
- [x] Create a PR to [reserve an ID prefix and define a new ecosystem](https://ossf.github.io/osv-schema/#id-modified-fields) ([example](https://github.com/ossf/osv-schema/pull/219)). We review the records you start publishing for OSV Schema [correctness](https://github.com/ossf/osv-schema/tree/main/validation) and [quality](https://google.github.io/osv.dev/data_quality.html) as part of reviewing and merging this PR.
  
  => https://github.com/ossf/osv-schema/pull/441

- [x] Prepare and publish your records via a Git repository ([example](https://github.com/AlmaLinux/osv-database/tree/master)). If this method isn’t ideal, we also support publishing records from [REST API endpoints](https://google.github.io/osv.dev/data/new/rest-api) or through a GCS bucket([example](https://storage.googleapis.com/android-osv/)).
  
  => https://cna.erlef.org/osv/all.json        

- [x] To support API querying, please create a PR to extend [purl\_helpers.py](https://github.com/google/osv.dev/blob/master/osv/purl_helpers.py) and create a new ecosystem in [\_ecosystems.py](https://github.com/google/osv.dev/blob/master/osv/ecosystems/_ecosystems.py). You can refer to existing examples showing how to implement support for [Semver](https://github.com/google/osv.dev/blob/139de7b69a2ea39e2113309b3a0a47aab920ddcf/osv/ecosystems/_ecosystems.py#L45) and [non-Semver](https://github.com/google/osv.dev/pull/3430) ecosystems.  
  
  => The entry for Hex already exists. Also we're reporting full version lists, so no change should be necessary.
        
- [x] Create a PR to start [importing the records you are publishing into our test instance of OSV.dev](https://github.com/google/osv.dev/blob/master/source_test.yaml) and validate everything is working as intended there.
  
  => #4192

- [x] Create a PR to start [importing the records you are publishing into our production environment](https://github.com/google/osv.dev/blob/master/source.yaml)
  
  => #4272

maennchen:
There's an issue with the test source since #4276 is not yet merged:

```json
[
  {
    "Code": "SCH:001",
[...]

PR #4192, opened 2025-10-16
Add EEF CNA to Test Sources

maennchen:
See #4191

### DIFF
--- a/source_test.yaml
+++ b/source_test.yaml
@@ -220,6 +220,20 @@
   editable: False
   strict_validation: True
 
+- name: 'eef'
+  type: 2
+  rest_api_url: 'https://cna.erlef.org/osv/all.json'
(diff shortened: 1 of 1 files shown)


maennchen:
@another-rex I'm trying to see if this all works properly. I expected to get some details on the `https://api.test.osv.dev/v1experimental/importfindings/eef` endpoint, but there's just an empty response.

Am I looking in the right place?

michaelkedar:
> @another-rex I'm trying to see if this all works properly. I expected to get some details on the `https://api.test.osv.dev/v1experimental/importfindings/eef` endpoint, but there's just an empty response.
> 
> Am I looking in the right place?

There was a small mistake in the config that we didn't catch - I've made #4267 to fix.

PR #4194, opened 2025-10-16
feat(frontend): collapse `database_specific` fields

### LINKED ISSUE #4163: Make the highest level of Affected database_specific field collapsible on frontend
Image

Test example: https://test.osv.dev/vulnerability/CVE-2025-39965

With the addition of vanir signature generation, there's now a lot of metadata type info in the database_specific which is hard/unnecessary to read. Maybe we make the highest level keys within database_specific in the Affected array collapsible on the frontend.

ashmod:
Updates the `database_specific` section to collapse its top-level keys by default. Users can expand individual entries as needed, similar to how the "Affected versions" section works.

Fixes #4163

### DIFF
--- a/gcp/website/frontend3/src/styles.scss
+++ b/gcp/website/frontend3/src/styles.scss
@@ -948,6 +948,46 @@ dl.vulnerability-details,
     overflow: auto;
   }
 
+  .database-specific-list {
+    display: flex;
+    flex-direction: column;
+    gap: 8px;
+  }
+
+  .database-specific-section {
+    --const-mq-affordances: [screen] collapse;
+
+    h2.database-specific-header {
+      background: none;
+      font-family: $osv-heading-font-family;
+      color: #fff;
+      padding-left: 0;
+      font-size: 16px;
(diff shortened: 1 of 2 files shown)


jess-lowe:
image
Hmm... it looks a bit off. I was even thinking just have it like the affected versions above, and then collapsed by default. I think having too many sets of tabs could get confusing.

ashmod:
> image Hmm... it looks a bit off. I was even thinking just have it like the affected versions above, and then collapsed by default. I think having too many sets of tabs could get confusing.

Hmm yes that looks completely broken. Last I tested it was supposed to look exactly like the affected versions, let me check that.

ashmod:
image

@jess-lowe This is how it currently looks like on my end, also did a hard reload to make sure it's not cached. Could you confirm if it changes after a hard refresh maybe?

jess-lowe:
Ok yes that fixed it. How weird!

jess-lowe (approved):
LGTM thanks!

PR #4197, opened 2025-10-17
feat: reimplement the exporter in Go

michaelkedar:
More database migration!
Reimplement the exporter in Go, reading from the GCS proto files instead of from the datastore Bugs.
I've made the whole thing more parallel and all in-memory, which should be a pretty decent performance improvement.

Testing is currently missing - we need a way to mock the GCS buckets.

### DIFF
--- a/go/cmd/exporter/worker.go
+++ b/go/cmd/exporter/worker.go
@@ -67,29 +67,21 @@ func (w *ecosystemWorker) run(ctx context.Context, outCh chan<- writeMsg, wg *sy
 	var allVulns []vulnData
 	var csvData [][]string
 	var vanirVulns []vulnData
+WorkLoop:
 	for {
 		var v *osvschema.Vulnerability
 		var ok bool
 
 		// Wait to receive a vulnerability, or be cancelled.
 		select {
-		case v, ok = <-w.inCh:
-			if !ok {
-				logger.Info("All vulnerabilities processed", slog.String("ecosystem", w.ecosystem))
-				writeModifiedIDCSV(ctx, filepath.Join(w.ecosystem, modifiedCSVFilename), csvData, outCh)
-				writeZIP(ctx, filepath.Join(w.ecosystem, allZipFilename), allVulns, outCh)
-				if w.ecosystem == gitEcosystem {
-					writeVanir(ctx, vanirVulns, outCh)
-				}
-				logger.Info("ecosystem worker finished processing", slog.String("ecosystem", w.ecosystem))
-
-				return
-			}
 		case <-ctx.Done():
 			logger.Warn("ecosystem worker cancelled", slog.String("ecosystem", w.ecosystem), slog.Any("err", ctx.Err()))
 			return
+		case v, ok = <-w.inCh:
+			if !ok {
+				break WorkLoop
+			}
 		}
-
 		// Process vulnerability.
 		b, err := protoMarshaller.Marshal(v)
 		if err != nil {
@@ -119,6 +111,14 @@ func (w *ecosystemWorker) run(ctx context.Context, outCh chan<- writeMsg, wg *sy
 			}
 		}
 	}
+
+	logger.Info("All vulnerabilities processed", slog.String("ecosystem", w.ecosystem))
+	writeModifiedIDCSV(ctx, filepath.Join(w.ecosystem, modifiedCSVFilename), csvData, outCh)
+	writeZIP(ctx, filepath.Join(w.ecosystem, allZipFilename), allVulns, outCh)
+	if w.ecosystem == gitEcosystem {
+		writeVanir(ctx, vanirVulns, outCh)
+	}
+	logger.Info("ecosystem worker finished processing", slog.String("ecosystem", w.ecosystem))
 }
 
 // Finish signals the worker to stop processing by closing its input channel.
@@ -159,19 +159,15 @@ func (w *allEcosystemWorker) run(ctx context.Context, outCh chan<- writeMsg, wg 
 	var allVulns []vulnData
 	var csvData [][]string
 	ecosystems := make(map[string]struct{})
+WorkLoop:
 	for {
 		select {
+		case <-ctx.Done():
+			logger.Warn("all-ecosystem worker cancelled", slog.Any("err", ctx.Err()))
+			return
 		case v, ok := <-w.inCh:
 			if !ok {
-				writeModifiedIDCSV(ctx, modifiedCSVFilename, csvData, outCh)
-				writeZIP(ctx, allZipFilename, allVulns, outCh)
-				ecos := slices.Collect(maps.Keys(ecosystems))
-				slices.Sort(ecos)
-				ecoString := strings.Join(ecos, "\n")
-				write(ctx, ecosystemsFilename, []byte(ecoString), "text/plain", outCh)
-				logger.Info("all-ecosystem worker finished processing")
-
-				return
+				break WorkLoop
 			}
 			b, err := protojson.MarshalOptions{}.Marshal(v.Vulnerability)
 			if err != nil {
@@ -183,11 +179,15 @@ func (w *allEcosystemWorker) run(ctx context.Context, outCh chan<- writeMsg, wg 
 				ecosystems[e] = struct{}{}
 				csvData = append(csvData, []string{v.GetModified().AsTime().Format(time.RFC3339Nano), e + "/" + v.GetId()})
 			}
-		case <-ctx.Done():
-			logger.Warn("all-ecosystem worker cancelled", slog.Any("err", ctx.Err()))
-			return
 		}
 	}
+	writeModifiedIDCSV(ctx, modifiedCSVFilename, csvData, outCh)
+	writeZIP(ctx, allZipFilename, allVulns, outCh)
+	ecos := slices.Collect(maps.Keys(ecosystems))
+	slices.Sort(ecos)
+	ecoString := strings.Join(ecos, "\n")
+	write(ctx, ecosystemsFilename, []byte(ecoString), "text/plain", outCh)
+	logger.Info("all-ecosystem worker finished processing")
 }
 
 // Finish signals the worker to stop processing by closing its input channel.


michaelkedar:
Hm, I'm not sure if 'shortly' was the right word...

deployment/build-and-stage.yaml
```diff
@@ -107,6 +98,20 @@ steps:
   args: ['push', '--all-tags', 'gcr.io/oss-vdb/recoverer']
   waitFor: ['build-recoverer', 'cloud-build-queue']
 
+- name: 'gcr.io/cloud-builders/docker'
```
another-rex, line 102: Add the build exporter image comment down here?

deployment/build-and-stage.yaml
```diff
@@ -107,6 +98,20 @@ steps:
...
+  entrypoint: 'bash'
+  args: ['-c', 'docker pull gcr.io/oss-vdb/exporter:latest || exit 0']
+  id: 'pull-exporter'
+  waitFor: ['setup']
+- name: gcr.io/cloud-builders/docker
+  args: ['build', '-t', 'gcr.io/oss-vdb/exporter:latest', '-t', 'gcr.io/oss-vdb/exporter:$COMMIT_SHA', '-f', 'cmd/exporter/Dockerfile', '--cache-from', 'gcr.io/oss-vdb/record-checker:latest', '--pull', '.']
```
another-rex: Is the cache-from record-check correct?

go/cmd/exporter/exporter.go
```diff
@@ -0,0 +1,170 @@
...
+	var writerWg sync.WaitGroup
+	for range *numWorkers / 2 {
+		writerWg.Add(1)
+		go writer(ctx, cancel, writeCh, outBucket, outPrefix, &writerWg)
+	}
+	var routerWg sync.WaitGroup
+	routerWg.Add(1)
+	go ecosystemRouter(ctx, resultsCh, writeCh, &routerWg)
+
+	it := vulnBucket.Objects(ctx, &storage.Query{Prefix: "all/pb/"})
```
another-rex: nit: make this prefix a const.

go/cmd/exporter/exporter.go
```diff
@@ -0,0 +1,170 @@
...
+	"strings"
+	"sync"
+
+	"cloud.google.com/go/storage"
+	"github.com/google/osv.dev/go/logger"
+	"github.com/ossf/osv-schema/bindings/go/osvschema"
+	"google.golang.org/api/iterator"
+)
+
+func main() {
```
another-rex, line 25: Please add a comment here explaining the graph of all the channels and workers and how they communicate with each other via the channels.

go/cmd/exporter/exporter.go
```diff
@@ -0,0 +1,170 @@
...
+	var outBucket *storage.BucketHandle
+	var outPrefix string
+	if *local {
+		outPrefix = *outBucketName
+	} else {
+		outBucket = cl.Bucket(*outBucketName)
+	}
+
+	objCh := make(chan *storage.ObjectHandle)
+	resultsCh := make(chan *osvschema.Vulnerability)
```
another-rex: Could be clearer to rename these channels like `<source>To<dist>Ch`. e.g. downloaderToRouterCh

go/cmd/exporter/worker.go
```diff
@@ -0,0 +1,246 @@
...
+	close(w.ch)
+}
+
+type vulnAndEcos struct {
+	*osvschema.Vulnerability
+
+	ecosystems []string
+}
+
+type allWorker struct {
```
another-rex: what does allWorker mean?

go/cmd/exporter/exporter.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+	logger.Info("export completed successfully")
+}
+
+func ecosystemRouter(ctx context.Context, inCh <-chan *osvschema.Vulnerability, writeCh chan<- writeMsg, wg *sync.WaitGroup) {
+	defer wg.Done()
+	logger.Info("ecosystem router starting")
+	workers := make(map[string]*ecosystemWorker)
+	var workersWg sync.WaitGroup
+	var vulnCounter int
```
another-rex: nit, explicitly set this to 0

go/cmd/exporter/downloader.go
```diff
@@ -0,0 +1,56 @@
...
+
+func downloader(ctx context.Context, objectCh <-chan *storage.ObjectHandle, resultsCh chan<- *osvschema.Vulnerability, wg *sync.WaitGroup) {
+	defer wg.Done()
+	for {
+		var obj *storage.ObjectHandle
+		var ok bool
+
+		// Wait to receive an object, or be cancelled.
+		select {
+		case obj, ok = <-objectCh:
```
another-rex: What's the reason for using a select here over directly looping objectCh like in ecosystemRouter for example?

go/cmd/exporter/worker.go
another-rex: Oh for all ecosystems. Hmm maybe rename to allEcosystemWorker?

go/cmd/exporter/exporter.go
```diff
@@ -0,0 +1,170 @@
...
+
+	workersWg.Add(1)
```
another-rex: nit: Maybe here and newEcosystemWorker as well the Add(1) should be inside the new function to avoid it from being forgotten (and consistently next to the `go` keyword)

go/cmd/exporter/worker.go
```diff
@@ -0,0 +1,246 @@
...
+	var allVulns []vulnData
+	var csvData [][]string
+	var vanirVulns []vulnData
+	for {
+		var v *osvschema.Vulnerability
+		var ok bool
+
+		// Wait to receive a vulnerability, or be cancelled.
+		select {
+		case v, ok = <-w.ch:
```
another-rex: hmm feels like this should also just be a for loop over w.ch
Then all the final writeCSV and writeZIP functions can be put at the bottom of this function after the loop

go/cmd/exporter/worker.go
```diff
@@ -0,0 +1,246 @@
...
+}
+
+func write(ctx context.Context, path string, data []byte, mimeType string, writeCh chan<- writeMsg) {
+	select {
+	case writeCh <- writeMsg{path: path, mimeType: mimeType, data: data}:
+	case <-ctx.Done():
+	}
+}
+
+func writeCSV(ctx context.Context, path string, csvData [][]string, writeCh chan<- writeMsg) {
```
another-rex: clarify that this is the modified_id csv file

go/cmd/exporter/worker.go
```diff
@@ -0,0 +1,246 @@
...
+	logger.Info("constructing csv file", slog.String("path", path))
+	slices.SortFunc(csvData, func(a, b []string) int {
+		return cmp.Or(
```
another-rex, line 210: Comment here for what is being compared (modified date desc)

another-rex (commented):
Nice!

go/cmd/exporter/downloader.go
michaelkedar:
```go
select {
case obj, ok = <-objectCh:
  // ...
case <- ctx.Done():
  return
}
```
Allows the loop to be cancelled if the context is cancelled (otherwise it'd block on the objectCh).
ecosystemRouter should probably also follow this pattern.

go/cmd/exporter/exporter.go
michaelkedar: Added above where the channels are created.

go/cmd/exporter/worker.go
michaelkedar: This is similar to https://github.com/google/osv.dev/pull/4197#discussion_r2443765315

go/cmd/exporter/worker.go
michaelkedar: renamed

go/cmd/exporter/worker.go
michaelkedar: renamed to `writeModifiedIDCSV`

deployment/build-and-stage.yaml
michaelkedar: no - fixed

go/cmd/exporter/exporter.go
```diff
@@ -0,0 +1,180 @@
...
+)
+
+const gcsProtoPrefix = "all/pb/"
+
+func main() {
+	logger.InitGlobalLogger()
+
+	outBucketName := flag.String("bucket", "osv-vulnerabilities", "Output bucket or directory name. If -local is true, this is a local path; otherwise, it's a GCS bucket name.")
+	vulnBucketName := flag.String("osv_vulns_bucket", os.Getenv("OSV_VULNERABILITIES_BUCKET"), "GCS bucket to read vulnerability protobufs from. Can also be set with the OSV_VULNERABILITIES_BUCKET environment variable.")
+	local := flag.Bool("local", true, "If true, writes the output to a local directory specified by -bucket instead of a GCS bucket.")
```
cuixq: why `local` flag is default to true? we probably don't need to set this flag to false in yaml if default to false.

michaelkedar:
> can you add some high-level documentation to each Go file/function?

done

go/cmd/exporter/exporter.go
michaelkedar: I think it's safer if it defaults to not uploading to the bucket, so someone running it locally is less likely to cause unintended changes to prod (or staging)

cuixq: speaking of this, maybe we should default `outBucketName` to the test bucket.

go/cmd/exporter/exporter.go
```diff
@@ -0,0 +1,185 @@
+// Package main runs the GCS exporter.
+// The exporter is responsible for creating many files in the OSV vulnerabilities bucket:
```
cuixq: these can be in the README.md

go/cmd/exporter/exporter.go
jess-lowe: In the rest of the conversion code we're calling this flag uploadToGCS btw, with the default being false, but same concept applies. Agree not to upload by default.

michaelkedar: changed it to `uploadToGCS`, though the mixing of snake_case and camelCase is upsetting me (CLI flags should probably be kebab-case anyway :upside_down_face:)

go/cmd/exporter/exporter.go
michaelkedar: Moved to new README.md

go/cmd/exporter/worker.go
```diff
@@ -0,0 +1,266 @@
...
+		case v, ok = <-w.inCh:
+			if !ok {
+				logger.Info("All vulnerabilities processed", slog.String("ecosystem", w.ecosystem))
+				writeModifiedIDCSV(ctx, filepath.Join(w.ecosystem, modifiedCSVFilename), csvData, outCh)
+				writeZIP(ctx, filepath.Join(w.ecosystem, allZipFilename), allVulns, outCh)
+				if w.ecosystem == gitEcosystem {
+					writeVanir(ctx, vanirVulns, outCh)
+				}
+				logger.Info("ecosystem worker finished processing", slog.String("ecosystem", w.ecosystem))
+
```
another-rex: Break here and put the finalising code at the bottom of the func (same with the other loops)

michaelkedar: `break`-ed

PR #4203, opened 2025-10-21
fix(frontend): render correct repo names in package headers

### LINKED ISSUE #4196: Repo rendering issue on record frontend.
Frontend package display is only using first repo address as label for top link (see suite-crm vs suite-crm-core)
Image

ashmod:
Updates Git package headers to use the repo associated with each affected range.

Fixes #4196

### DIFF
--- a/gcp/website/frontend3/src/templates/vulnerability.html
+++ b/gcp/website/frontend3/src/templates/vulnerability.html
@@ -193,8 +193,6 @@
                 {% set affected_repo = affected.ranges | default([], true) | selectattr('repo') | map(attribute='repo') | first %}
                 {% if affected_repo %}
                   {{ affected_repo | strip_scheme }}
-                {% elif vulnerability.repo %}
-                  {{ vulnerability.repo | strip_scheme }}
                 {% endif %}
               {% endif %}
             </h3>
@@ -347,8 +345,6 @@
       {% set affected_repo = affected.ranges | default([], true) | selectattr('repo') | map(attribute='repo') | first %}
       {% if affected_repo %}
       {% set package = affected_repo | strip_scheme %}
-      {% elif vulnerability.repo %}
-      {% set package = vulnerability.repo | strip_scheme %}
       {% endif %}
       {% endif %}
       <h2 class="package-header">


gcp/website/frontend3/src/templates/vulnerability.html
```diff
@@ -187,7 +187,16 @@ <h2 class="package-header">
...
           <spicy-sections class="package-accordion{% if is_last_ecosystem and loop.last %} package-accordion--last{% endif %}">
             <h3 class="package-name-title">
-              {% if 'package' in affected %}{{ affected.package.name }}{% else %}{{ vulnerability.repo | strip_scheme }}{% endif %}
+              {% if 'package' in affected %}
+                {{ affected.package.name }}
+              {% else %}
+                {% set affected_repo = affected.ranges | default([], true) | selectattr('repo') | map(attribute='repo') | first %}
+                {% if affected_repo %}
+                  {{ affected_repo | strip_scheme }}
+                {% elif vulnerability.repo %}
```
jess-lowe: I think we can get rid of this elif, because vulnerability.repo is technically a fake field.

gcp/website/frontend3/src/templates/vulnerability.html
```diff
@@ -335,7 +344,10 @@ <h2 class="database-specific-header">{{ key }}</h2>
       {% set package = affected.package.name %}
       {% else %}
       {% set ecosystem = 'Git' %}
-      {% if vulnerability.repo %}
+      {% set affected_repo = affected.ranges | default([], true) | selectattr('repo') | map(attribute='repo') | first %}
+      {% if affected_repo %}
+      {% set package = affected_repo | strip_scheme %}
+      {% elif vulnerability.repo %}
```
jess-lowe: Same here. Otherwise looks good :)

gcp/website/frontend3/src/templates/vulnerability.html
ashmod: Done

gcp/website/frontend3/src/templates/vulnerability.html
ashmod: Should be updated now.

PR #4220, opened 2025-10-21
refactor: Migrate to new bindings

cuixq:
This PR migrates the project from custom Go data structures to Protobuf-generated bindings:
 - Removing manually defined Go models in `bindings/go/osvdev/models.go`.
 - Generating the Go bindings based on `osv_service_v1.proto` and putting them in a new package `bindings/go/api`.
 - Updating the Go client (osvdev.go) and paging logic (paging.go) to use the new generated Go types, including a a new helper function `makeRequest`.
 - Modifying the Makefile to handle the generation of both Go and Python bindings from proto files.
 - Removing the copy of `vulnerability.proto` in osv and using the canonical version from the `osv-schema` submodule.

Currently generated Python codes are checked-in to the `gcp/api` directory - there will be follow-up PR to undo check-in the generated Python code.


Makefile
```diff
@@ -48,9 +48,36 @@ api-server-tests:
...
+
+build-api-protos:
+	cd gcp/api/v1 && $(run-cmd) python -m grpc_tools.protoc \
+      --include_imports \
+      --include_source_info \
+      --proto_path=googleapis \
+      --proto_path=. \
+	  --proto_path=osv \
+	  --proto_path=osv/osv-schema/proto \
+      --descriptor_set_out=api_descriptor.pb \
```
another-rex, line 62: this indentation seems weird.

Makefile
```diff
@@ -48,9 +48,36 @@ api-server-tests:
...
+	  importfinding.proto
+	cd gcp/api/v1 && protoc \
+	  --proto_path=googleapis \
+	  --proto_path=. \
+	  --proto_path=osv \
+      --proto_path=osv/osv-schema/proto \
+	  --go_out=paths=source_relative:../../../bindings/go/api \
+      --go-grpc_out=paths=source_relative:../../../bindings/go/api \
+	  osv_service_v1.proto
+
```
another-rex: indentation

osv/importfinding.proto
```diff
@@ -18,6 +18,8 @@ package osv;
 
 import "google/protobuf/timestamp.proto";
 
+option go_package = "osv.dev/bindings/go/api";
```
another-rex: is this correct for importfinding?

cuixq: is this not correct?

Makefile
cuixq: the indentation looks all the same to me :( but I tried to fix them

another-rex: Looks good now! It seems like it was a mixture of tabs and spaces.

osv/importfinding.proto
another-rex: Ah I guess import findings is an api endpoint

another-rex: let's mark all the bindings/go/api as internal for now.

another-rex (approved):

another-rex (approved):

bindings/go/osvdev/osvdev.go
```diff
@@ -81,15 +85,15 @@ func (c *OSVClient) GetVulnByID(ctx context.Context, id string) (*osvschema.Vuln
...
 	// API has a limit of how many queries are in one batch
 	queryChunks := chunkBy(queries, MaxQueriesPerQueryBatchRequest)
-	totalOsvRespBatched := make([][]MinimalResponse, len(queryChunks))
+	totalOsvRespBatched := make([][]*api.VulnerabilityList, len(queryChunks))
 
 	g, errGrpCtx := errgroup.WithContext(ctx)
 	g.SetLimit(c.Config.MaxConcurrentBatchRequests)
 	for batchIndex, queries := range queryChunks {
-		requestBytes, err := json.Marshal(BatchedQuery{Queries: queries})
+		requestBytes, err := protojson.Marshal(&api.BatchQuery{Queries: queries})
```
michaelkedar, line 96: We probably need to `UseProtoNames` or add the json_name to the protos like https://github.com/ossf/osv-schema/pull/442 (for e.g. `next_page_token`) for all these API translations.

PR #4226, opened 2025-10-23
fix(frontend): affected packages not collapsing when long

jess-lowe:
Git being a fake ecosystem is causing a few edge-case rendering issues. This should help fix this issue where parts are getting cut off:

image

### DIFF
--- a/gcp/website/frontend_handlers.py
+++ b/gcp/website/frontend_handlers.py
@@ -773,9 +773,9 @@ def should_collapse(affected):
   total_text_length_ecosystem = sum(
       len(entry.get('package', {}).get('ecosystem', '')) for entry in affected)
   total_text_length_package = sum(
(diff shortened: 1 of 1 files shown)


gcp/website/frontend_handlers.py
```diff
@@ -773,7 +773,9 @@ def should_collapse(affected):
   total_text_length_ecosystem = sum(
       len(entry.get('package', {}).get('ecosystem', '')) for entry in affected)
   total_text_length_package = sum(
-      len(entry.get('package', {}).get('name', '')) for entry in affected)
+      (len(entry.get('package', {}).get('name', '')) \
+       + len(entry.get('ranges', {})[0].get('repo','')) \
```
another-rex: what if ranges is empty?

jess-lowe: internal server error - woooops. fixed it!

PR #4241, opened 2025-10-28
fix(vulnfeeds): Migrate to latest OSV schema bindings

cuixq:
Unblocks https://github.com/google/osv.dev/pull/4190

This PR updates `vulnfeeds` to use the latest version of `github.com/ossf/osv-schema/bindings/go`.

To better handle ecosystem and database specific fields, a new function `NewStructpbFromMap` is added to `utility` to convert `map[string]any` to `structpb.Struct` so we don't need to worry about conversion every time.


vulnfeeds/cvelist2osv/converter.go
```diff
@@ -97,11 +100,15 @@ func attachCWEs(v *vulns.Vulnerability, cna cves.CNA, metrics *ConversionMetrics
...
+		v.DatabaseSpecific = &structpb.Struct{Fields: make(map[string]*structpb.Value)}
 	}
 
 	// Add CWEs to DatabaseSpecific for consistency with GHSA schema.
-	v.DatabaseSpecific["cwe_ids"] = cwes
+	cweList := make([]*structpb.Value, 0, len(cwes))
+	for _, cwe := range cwes {
+		cweList = append(cweList, structpb.NewStringValue(cwe))
+	}
+	v.DatabaseSpecific.Fields["cwe_ids"] = structpb.NewListValue(&structpb.ListValue{Values: cweList})
```
jess-lowe: I hate this 😭 Are we going to have to do this for every field we want to add to database_specific?

cuixq: we can probably factor out to a helper function that takes the key and the list of strings to construct the struct, so we don't need to repeat this every time.

do we expect to see more variation of fields in `database_specific`?

michaelkedar: Would it be neater to call [`ds := v.DatabaseSpecific.AsMap()`](https://pkg.go.dev/google.golang.org/protobuf/types/known/structpb#Struct.AsMap) at the beginning, then [`v.DatabaseSpecific, err = structpb.NewStruct(ds)`](https://pkg.go.dev/google.golang.org/protobuf/types/known/structpb#NewStruct) at the end of these functions?

michaelkedar: Alternatively, use [`NewValue(cwes)`](https://pkg.go.dev/google.golang.org/protobuf/types/known/structpb#NewValue) instead of `NewXValue`

jess-lowe: I know that at the very least I also want to add the `programFiles` info to `database_specific`

jess-lowe:
Found some issues by running the converter:
published/modified dates aren't showing as timestamps - they're showing as this: 
`  "published": {
    "seconds": 1740622699,
    "nanos": 528000000
  },
  "modified": {
    "seconds": 1746343246,
    "nanos": 575000000
  },`
  
  Also, any enums are being output as the enum value, not the enum string representation

vulnfeeds/cvelist2osv/converter.go
cuixq: @jess-lowe will `programFiles` be a slice of string? e.g.
```
"programFiles": []string{"file1", "file"}
```
?

cuixq: @michaelkedar unfortunately `structpb.NewValue()` only works for `[]any` but not `[]string` or any other known structs like `[]*osvschema.Event` or `[]*osvschema.Range`.

To make it work, we probably need to convert `cwes` to a slice of `any` type, which means a bit of extra code for `Sort` and `Compact`, which is not too bad; however, for more complicated struct like  `osvschema.Event` or `osvschema.Range`, we need to do a deep conversion for each fields in the struct which is almost like the current approach.

Also one thing I don't quite like about `NewValue` is that we need to check the returned error, which causes more chances of unhandled errors. Considering we are certain about the type of structs here, I am leaning towards converting them manually.

michaelkedar: Hm yeah this is actually really annoying.
For the slices I guess we could make a generic function like
```go
func NewListValue[T any](list []T) (*structpb.Value, error) {
	anyList := make([]any, len(list))
	for i, v := range list {
		anyList[i] = v
	}
	return structpb.NewValue(anyList)
}
```

or maybe even go through JSON:
```go
func NewValue(v any) (*structpb.Value, error) {
	b, err := json.Marshal(v)
	if err != nil {
		return nil, err
	}
	var val structpb.Value
	err = json.Unmarshal(b, &val)
	return &val, err
}
```

But this doesn't account for things that need to use `protojson`. Maybe we can be clever with type checking and/or `reflect` but it's starting to get complicated.

And this also doesn't solve the issue of the potential error. Though, I think the only error that would return would be if a type is unconvertable, which is probably a fatal error for the program.

cuixq: I tried the JSON approach and it didn't work for protos as you already pointed out, but I do like the idea of the generic function.

cuixq: hmmm seems the generic solution still has the same problem.
```
WARN:   Failed to convert events to structpb.Value err=proto: invalid type: *osvschema.Event
```

cuixq: I add a function called `newDatabaseSpecific` to convert `map[string]any` to a `structpb.Struct` that should handle the use case we have for now.

vulnfeeds/cvelist2osv/__snapshots__/converter_test.snap
```diff
@@ -1,384 +1,380 @@
...
-  "details": "A disputed vulnerability.",
-  "database_specific": {
-    "isDisputed": true
+  "schema_version":  "1.7.3",
+  "id":  "CVE-2025-9999",
+  "published":  "2025-05-04T07:20:46.575Z",
+  "modified":  "2025-05-04T07:20:46.575Z",
+  "details":  "A disputed vulnerability.",
+  "database_specific":  {
+    "isDisputed":  true
```
michaelkedar: The fact that there seems to be 2 spaces between the key and value makes me suspect we're running into [protojson's unstable output](https://protobuf.dev/reference/go/faq/#unstable-json).

 @jess-lowe do you know if it'd cause problems if the formatting changes from time-to-time? Maybe we should be `json.Compact`-ing these (at least for the tests)

vulnfeeds/cvelist2osv/common.go
```diff
@@ -96,13 +101,95 @@ func resolveVersionToCommit(cveID cves.CVEID, version, versionType, repo string,
...
+	default:
+		return nil, fmt.Errorf("unsupported type: %T", v)
+	}
+}
+
+// structpbValueFromList converts a generic slice into a *structpb.Value that
+// represents a list.
+// It iterates through the list, converting any proto.Message elements into
+// any type via JSON marshalling, while other elements are included as-is.
+func structpbValueFromList[T any](list []T) (*structpb.Value, error) {
```
michaelkedar: I think, as it's currently used, this doesn't need to be a template and can just take a `list []any`, since the only place that's calling this is calling it with `[]any`

vulnfeeds/cvelist2osv/common.go
```diff
@@ -96,13 +101,95 @@ func resolveVersionToCommit(cveID cves.CVEID, version, versionType, repo string,
...
+	case reflect.Slice:
+		var anyList []any
+		for i := range val.Len() {
+			anyList = append(anyList, val.Index(i).Interface())
+		}
+
+		return structpbValueFromList(anyList)
+	default:
+		return nil, fmt.Errorf("unsupported type: %T", v)
+	}
```
michaelkedar: I feel this could possibly be
```go
switch val.Kind() {
case reflect.Slice:
   // current slice logic
  return //...
default:
  return structpb.NewValue(v) // handles bool & string (and int)
}
```

vulnfeeds/vulns/vulns.go
```diff
@@ -292,46 +295,52 @@ func (v *Vulnerability) AddPkgInfo(pkgInfo PackageInfo) {
...
 	})
 
-	affected.EcosystemSpecific = pkgInfo.EcosystemSpecific
+	b, err := json.Marshal(pkgInfo.EcosystemSpecific)
+	if err == nil {
+		pbst := &structpb.Struct{}
+		if json.Unmarshal(b, pbst) == nil {
+			affected.EcosystemSpecific = pbst
+		}
+	}
```
michaelkedar, line 320: This could use `newDatabaseSpecific` also

vulnfeeds/cvelist2osv/__snapshots__/converter_test.snap
cuixq: actually I did notice the test being a bit flaky with/without the space, which I didn't expect for the same version of protojson though.

vulnfeeds/vulns/vulns.go
cuixq: hmmm need a more general function name 🤔

cuixq: also need to move this to utility to make it more usable.

vulnfeeds/cvelist2osv/__snapshots__/converter_test.snap
cuixq: actually I did notice the test being a bit flaky with/without the space, which I didn't expect for the same version of protojson though.

jess-lowe: It'll may cause some issues in that the bucket might be triggered to update more frequently if there is not a stable output.

cuixq: as discussed offline, I use the output from `protojson.Marshal` and call `encoding/json` to marshal and unmarshal to make the output is stable.

another-rex (approved):

Issue #4161, opened 2025-10-14
Discrepancies found with Debian CVE(s)


iam-naveen:
Data Used from: https://osv-vulnerabilities.storage.googleapis.com/Debian/all.zip
Source used to compare against: https://security-tracker.debian.org/tracker

Case 1
---
CVE Entries marked as "NOT-FOR-US" included in the ZIP.

Example: DEBIAN-CVE-2025-0649
https://security-tracker.debian.org/tracker/CVE-2025-0649
https://storage.googleapis.com/debian-osv/debian-cve-osv/DEBIAN-CVE-2025-0649.json

Case 2
---
CVE Entries marked as Rejected by both Debian and NVD included in the ZIP.

Example: DEBIAN-CVE-2023-52979
- Sources
    - https://security-tracker.debian.org/tracker/CVE-2023-52979 (Empty Page)
    - https://salsa.debian.org/security-tracker-team/security-tracker/-/raw/master/data/CVE/list (Specifically marked as Rejected)
- OSV
    - https://storage.googleapis.com/debian-osv/debian-cve-osv/DEBIAN-CVE-2023-52979.json (No Indicating values to show Rejected status)

Case 3
---
Few CVE entries have wrong Debian versions marked as affected in the ZIP.

Example: DEBIAN-CVE-2025-52566
We can see that only the unstable release is mentioned at the source (Debian Security Tracker).
https://security-tracker.debian.org/tracker/CVE-2025-52566
But Debian:14 is mentioned here,
https://storage.googleapis.com/debian-osv/debian-cve-osv/DEBIAN-CVE-2025-52566.json

another-rex:
While we are on the topic of Debian CVE discrepancies: https://osv.dev/vulnerability/DSA-5902-1 doesn't have the correct upstream of debian-cve set.

jess-lowe:
Hey, thanks for finding all these!

Doing some triage:
For 1 & 2: if i delete the respective records from the bucket they don't seem to get regenerated on next run. We likely fixed the generation issue of these a few weeks back, but we haven't implemented proper deletion logic for if this sort of thing changes.
(notes to self:)
**Bandaid solution:** delete everything and do a complete re-ingestion of Debian records.
**Proper solution:** add deletion logic in the upload "worker", that checks for differences between filenames and deletes the ones in the bucket if they haven't been generated. 

For 3: we seem to be mapping unstable incorrectly. I'll look into this separately.

PR #4242, opened 2025-10-28
fix: delete stale Debian/Alpine records from bucket during upload

jess-lowe:
Should fix 1 & 2 of https://github.com/google/osv.dev/issues/4161 

Added a flag for whether to do deletions as CVE5 uses the same function, and we may not want to delete in every case.

### DIFF
--- a/vulnfeeds/cmd/alpine/main.go
+++ b/vulnfeeds/cmd/alpine/main.go
@@ -41,7 +41,7 @@ func main() {
 	outputBucketName := flag.String("output_bucket", outputBucketDefault, "The GCS bucket to write to.")
 	numWorkers := flag.Int("num_workers", 64, "Number of workers to process records")
 	uploadToGCS := flag.Bool("uploadToGCS", false, "If true, do not write to GCS bucket and instead write to local disk.")
-	syncDeletions := flag.Bool("sync_deletions", true, "If false, do not delete files in bucket that are not local")
+	syncDeletions := flag.Bool("sync_deletions", false, "If false, do not delete files in bucket that are not local")
 	flag.Parse()
 
 	err := os.MkdirAll(*alpineOutputPath, 0755)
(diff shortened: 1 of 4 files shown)


vulnfeeds/upload/cveworker.go
```diff
@@ -216,3 +222,46 @@ func Upload(
...
+	for _, v := range vulnerabilities {
+		filename := v.ID + ".json"
+		filePath := path.Join(osvOutputPath, filename)
+		vulnFilenames[filePath] = true
+	}
+	for _, objName := range bucketObjects {
+		if !vulnFilenames[objName] {
+			logger.Info("Deleting stale object from bucket", slog.String("name", objName))
+			obj := outBkt.Object(objName)
+			if err := obj.Delete(ctx); err != nil {
```
michaelkedar, line 243: Not sure how much we'll end up deleting, but it might be worth sending these to a worker pool to do the deletions.

michaelkedar (approved):

vulnfeeds/cmd/alpine/main.go
```diff
@@ -62,7 +62,7 @@ func main() {
 	}
 
 	ctx := context.Background()
-	upload.Upload(ctx, "Alpine CVEs", *uploadToGCS, *outputBucketName, "", *numWorkers, *alpineOutputPath, vulnerabilities)
+	upload.Upload(ctx, "Alpine CVEs", *uploadToGCS, *outputBucketName, "", *numWorkers, *alpineOutputPath, vulnerabilities, true)
```
another-rex: The deletion should be controllled by a flag right?

vulnfeeds/upload/cveworker.go
jess-lowe: For context, currently Debian is about 20.

vulnfeeds/cmd/alpine/main.go
```diff
@@ -41,6 +41,7 @@ func main() {
 	outputBucketName := flag.String("output_bucket", outputBucketDefault, "The GCS bucket to write to.")
 	numWorkers := flag.Int("num_workers", 64, "Number of workers to process records")
 	uploadToGCS := flag.Bool("uploadToGCS", false, "If true, do not write to GCS bucket and instead write to local disk.")
+	syncDeletions := flag.Bool("sync_deletions", true, "If false, do not delete files in bucket that are not local")
```
another-rex: Let's keep this consistently false, and tweak it in either the script or the kubernetes file.

another-rex (approved):
I just noticed the flag names, half are camel case and other half is snake case.  (well just uploadToGCS I guess) 

Probably something to fix in another PR.

jess-lowe:
> I just noticed the flag names, half are camel case and other half is snake case. (well just uploadToGCS I guess)
> 
> Probably something to fix in another PR.

already working on it ;)

PR #4270, opened 2025-10-29
fix: Set Hex ecosystem purl type

maennchen:
See https://github.com/package-url/purl-spec/blob/3e98d4aaa7f4ad3c67eff0b753d12b378ea47a0a/types-doc/hex-definition.md#L4

### DIFF
--- a/osv/purl_helpers.py
+++ b/osv/purl_helpers.py
@@ -56,7 +56,7 @@ ECOSYSTEM_PURL_DATA = {
     'Hackage':
         EcosystemPURL('hackage', None),
     'Hex':
-        EcosystemPURL('hex', None),
+        EcosystemPURL('hex', 'hex'),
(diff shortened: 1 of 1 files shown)


osv/purl_helpers.py
```diff
@@ -56,7 +56,7 @@
     'Hackage':
         EcosystemPURL('hackage', None),
     'Hex':
-        EcosystemPURL('hex', None),
+        EcosystemPURL('hex', 'hex'),
```
michaelkedar, line 59: This EcosystemPURL tuple is (type, namespace), so the namespace should be `None` (it would come from OSV's `package.name`, it's not fixed to `hex`).

maennchen: Oh, sorry. This PR is not relevant then.

PR #4280, opened 2025-10-30
Add utility script to delete Bug entities from Datastore

CoderOggy78:
This script provides a utility to delete Bug entities from Google Cloud Datastore with options for dry-run, verbose logging, and batch deletion.


another-rex:
Please create an issue for why this is necessary before submitting a PR, thanks!

PR #4282, opened 2025-10-31
feat: Add Julia ecosystem to source_test

mbauman:
Our structure is quite similar to Haskell's here.  We don't (yet) have a human-rendered HTML page, but we could link directly to the markdowns on GitHub if you'd prefer to have _something_ for the human_link.  E.g., https://github.com/JuliaLang/SecurityAdvisories.jl/blob/main/advisories/published/2025/JLSEC-2025-1.md is fairly human-friendly.

### DIFF
--- a/source_test.yaml
+++ b/source_test.yaml
@@ -279,6 +279,21 @@
   repo_username: 'git'
   strict_validation: True
 
+- name: 'julia'
+  versions_from_repo: False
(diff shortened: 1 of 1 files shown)


another-rex:
The markdown page definitely works as the human readable link!

another-rex (approved):

PR #4284, opened 2025-11-03
chore: Update the go version to fix vulnerability

another-rex:
OSV-Scanner is reporting new vulns for go 1.24.6, so bumping everything up to 1.25.3. Hopefully in the future renovatebot can do this for us.

### DIFF
--- a/bindings/go/go.mod
+++ b/bindings/go/go.mod
@@ -1,6 +1,6 @@
 module osv.dev/bindings/go
 
-go 1.24.8
+go 1.25.3
 
 require (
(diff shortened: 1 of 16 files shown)


gcp/api/googleapis
michaelkedar, line 1: was this intentional?

cuixq: +1

another-rex: nope

another-rex:
Hmm interestingly most golang ones are on go 1.25. I'll update everything to 1.25.3 to keep things consistent

PR #4306, opened 2025-11-04
test(apitester): rewrite json replacement logic to better handle arrays and existence

G-Rath:
Currently, attempting to replace properties within arrays on objects that don't always have the same structure all the way down can result in undesired behaviours such as the property being added to all objects within arrays on the path, or only half of the potential matches being replaced.

While the problem is a bit awkward to describe, the solution is not: now we fully unroll a replacement path, and apply the replacement logic on each resulting path.

i.e.with the object `{ "arr": [1, 2, 3] }`, we unroll the path `arr.#` to `[]string{"arr.0", "arr.1", "arr.2"}`.


G-Rath:
This is required for #4307 and a sister to https://github.com/google/osv-scanner/pull/2320

PR #4308, opened 2025-11-04
ci: create workflow to automatically update snapshots

G-Rath:
This is based off the robot in `osv-scanner`

### DIFF
--- a/.github/workflows/snapshots.yml
+++ b/.github/workflows/snapshots.yml
@@ -1,0 +1,46 @@
+name: Snapshots
+
+on:
+  schedule:
+    - cron: "47 18 * * *"
+  workflow_dispatch:
+concurrency:
+  # Pushing new changes to a branch will cancel any in-progress CI runs
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+# Restrict jobs in this workflow to have no permissions by default; permissions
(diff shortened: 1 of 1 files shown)


G-Rath:
@another-rex but did you give the bot permission to access the repository? 😅 
image

another-rex:
Hmm... it should have full access to the repo. Maybe it needs some special token perms?

G-Rath:
what about the token itself? its permissions can be scoped to specific repos

another-rex:
Ok, made a fine grained token with I think the right types (For some reason fine grained tokens can't actually be scoped to specific repos that is not directly under your name, e.g. this is under google/osv.dev, not osv-robot/osv.dev, so it can't be scoped 🤦 )

Let's see:

https://github.com/google/osv.dev/actions/runs/19482596515

EDIT: Well that didn't work.

G-Rath:
> For some reason fine grained tokens can't actually be scoped to specific repos that is not directly under your name, e.g. this is under google/osv.dev, not osv-robot/osv.dev, so it can't be scoped

That should be possible so long as you're a member of the org? (its not possible for repos you've an outside collaborator on)

another-rex:
Ayy: https://github.com/google/osv.dev/pull/4356

I really don't understand fine grained tokens, switched back to classical tokens.

PR #4311, opened 2025-11-05
feat: add better structured logging for importer/worker latency

michaelkedar:
Add some structured logs for the (approximate) time it took between a source publishing a change to it being reflected in osv.
- For GIT, this is measured from the timestamp of the commit that introduced the change
- For GCS, this is measured from the modified time of the GCS object
- For REST, this is measured from the modified time of the vulnerability

None of these will notice if more than one update happens between importer runs - it'll only measure from the latest update.

Inserting this into the importer/worker was a little bit awkward. I didn't want to spend too much time on this since I ultimately want to rewrite the importer anyway.

### DIFF
--- a/gcp/workers/importer/importer.py
+++ b/gcp/workers/importer/importer.py
@@ -439,11 +439,11 @@ class Importer:
       vuln_ids.append(vuln.id)
     return vuln_ids
 
-  def _convert_blob_to_vuln(self, storage_client: storage.Client,
-                            ndb_client: ndb.Client,
-                            source_repo: osv.SourceRepository,
-                            blob: storage.Blob,
-                            ignore_last_import_time: bool) -> Optional[Tuple]:
+  def _convert_blob_to_vuln(
+      self, storage_client: storage.Client, ndb_client: ndb.Client,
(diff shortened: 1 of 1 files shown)


gcp/workers/importer/importer_test.py
```diff
@@ -203,7 +203,8 @@ def test_basic(self, unused_mock_time: mock.MagicMock,
             path='2021-111.yaml',
             source='oss-fuzz',
             type='update-oss-fuzz',
-            req_timestamp='12345')
+            req_timestamp='12345',
+            src_timestamp='')
```
cuixq, line 207: maybe a dumb question: what is difference between `None` and an empty string for a timestamp?

michaelkedar: The importer is now always setting `src_timestamp` in the pub/sub message to either the empty string or the timestamp - it shouldn't be None/unset in the message (I found it easier to always set the field than to do it conditionally).
The mock calls must assert that the timestamp field was included, even if empty.

cuixq: gotcha

PR #4312, opened 2025-11-05
feat(source_test): test adding a human link for Julia advisories

mbauman:
I think I got the year extraction correct, but I'm not 100% sure.


cuixq (approved):
LGTM - thanks!

PR #4315, opened 2025-11-06
fix: allow records with inaccesible repos to be ingested

michaelkedar:
Malicious packages is starting to publish git records, which have repositories that have been removed.
Turns out, OSV completely failed to ingest records if the repository is not clone-able, which is bad.
Made it not raise an exception for (what I think are) non-transient git clone errors.

### DIFF
--- a/osv/repos.py
+++ b/osv/repos.py
@@ -124,7 +124,7 @@ def clone(git_url, checkout_dir, git_callbacks=None, blobless=False):
     subprocess.run(cmd, env=env, capture_output=True, check=True)
     return pygit2.Repository(checkout_dir)
   except subprocess.CalledProcessError as e:
-    stderr = e.stderr.decode()
+    stderr = e.stderr.decode(errors='ignore')
     if ('could not read Username' in stderr or
         ('fatal: repository' in stderr and 'not found' in stderr) or
         'Authentication failed' in stderr):


osv/repos.py
```diff
@@ -120,7 +124,14 @@ def clone(git_url, checkout_dir, git_callbacks=None, blobless=False):
     subprocess.run(cmd, env=env, capture_output=True, check=True)
     return pygit2.Repository(checkout_dir)
   except subprocess.CalledProcessError as e:
-    raise GitCloneError(f'Failed to clone repo:\n{e.stderr.decode()}') from e
+    stderr = e.stderr.decode()
+    if ('could not read Username' in stderr or
```
cuixq, line 128: the error messages feel a bit specific - e.g. hardcoded texts, upper/lower cases, but I also don't know a better way to check this. :(

cuixq (dismissed):

osv/repos.py
michaelkedar: For reference, here are the outputs:
```
> clone('https://github.com/not-a-user/not-a-repo.git', '/tmp/repo')
Cloning into '/tmp/repo'...
fatal: could not read Username for 'https://github.com': terminal prompts disabled
```
(This happens both if the repo doesn't exist, and if it's private)

```
> clone('https://github.com/invalid-path.git', '/tmp/repo')
Cloning into '/tmp/repo'...
remote: Not Found
fatal: repository 'https://github.com/invalid-path.git/' not found
```
(and I don't know if `Authentication failed` is actually real - gemini suggested it :upside_down_face:)

cuixq (approved):

PR #4340, opened 2025-11-11
feat(vulnfeeds): create a CVE-CNA allowlist

jess-lowe:
Move the default allowlist for the CNAs whose CVE's we're ingesting to a .txt file. 

This PR will also enable the ingestion of CVEs from the following CNAs:
- GitLab
- Centreon
- @ huntrdev

### DIFF
--- a/vulnfeeds/cmd/cve-bulk-converter/main.go
+++ b/vulnfeeds/cmd/cve-bulk-converter/main.go
@@ -2,6 +2,7 @@
 package main
 
 import (
+	_ "embed"
 	"encoding/json"
 	"flag"
 	"log/slog"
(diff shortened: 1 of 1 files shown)


vulnfeeds/cmd/cve-bulk-converter/main.go
```diff
@@ -35,8 +35,21 @@ func main() {
 
 	jobs := make(chan string)
 	var wg sync.WaitGroup
-
-	cnaList := strings.Split(*cnas, ",")
+	var cnaList []string
+	if *cnas != "" {
+		cnaList = strings.Split(*cnas, ",")
+	} else {
+		cnaAllowlistData, err := os.ReadFile("cna_allowlist.txt")
```
michaelkedar: The relative path here might make this a bit awkward depending on how/where you're running this?

Actually, I'm not sure the path here would be correct in the Dockerfile?

Not sure what the better solution is, though. I thought maybe [//go:embed](https://pkg.go.dev/embed) the list, but that would mean we'd have to compile the list into the program.

another-rex: Hmm... I'm not seeing the benefit of moving it out to a separate file, can we just have a slice at the top of the file that specifies the allowlist?

jess-lowe: the list is going to get a lot longer - up to ~100

PR #4341, opened 2025-11-12
fix: standardise cli flags to kebab-case

jess-lowe:
our flags are a little all over the place, so trying to standardise as many as possible to kebab case

### DIFF
--- a/vulnfeeds/cmd/cve-bulk-converter/main.go
+++ b/vulnfeeds/cmd/cve-bulk-converter/main.go
@@ -22,7 +22,7 @@ var (
 	localOutputDir = flag.String("out-dir", "cvelist2osv", "Path to output results.")
 	years          = flag.String("years", "2022,2023,2024,2025", "A comma-separated list of years to process.")
 	workers        = flag.Int("workers", 30, "The number of concurrent workers to use for processing CVEs.")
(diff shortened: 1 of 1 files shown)


deployment/clouddeploy/gke-indexer/environments/oss-vdb-test/indexer-controller.yaml
```diff
@@ -26,7 +26,7 @@ spec:
             args: [
               "--configs=osv-test-indexer-configs",
               "--repos=osv-test-indexer-repos",
-              "--project_id=oss-vdb-test",
+              "--project-id=oss-vdb-test",
               "--topic=indexer-work",
               "--alsologtostderr",
```
michaelkedar, line 31: since I've noticed it, `alsologtostderr` could be `log-stderr` or similar

jess-lowe: Do we set this or does it come from elsewhere? From what I can tell, the flag comes from the GLog module

michaelkedar: hmm mysterious - just leave it as is then

michaelkedar (approved):

PR #4344, opened 2025-11-12
feat: Have importer write Bugs before sending to worker

michaelkedar:
Added a some logic to make the importer update the database when it receives updates for the data sources, instead of having to wait for the worker to finish enumerating versions / git commits.

To prevent churn, the importer checks if the affected packages have changed in the upstream source first. If they haven't it won't overwrite them (so we can keep our previous worker enrichment). The worker will still run for every record, however.

Bugs put by the importer will use the modified date from the source, while bugs modified by the worker now use the time the worker finished (instead of the source time like before).

I've also modified the REST test to reduce an absolutely massive vulnerability record, and made the `osv.utcnow()` return a time in the future instead of in 2021, which was causing weirdness with time going backwards.

This whole logic is a bit temporary - I eventually want to rewrite the importer (in go) to stop relying on the Bug entities, and remove much of the OSS-Fuzz-specific behaviour.

### DIFF
--- a/.github/workflows/snapshots.yml
+++ b/.github/workflows/snapshots.yml
@@ -1,0 +1,46 @@
+name: Snapshots
+
+on:
+  schedule:
+    - cron: "47 18 * * *"
+  workflow_dispatch:
+concurrency:
+  # Pushing new changes to a branch will cancel any in-progress CI runs
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+# Restrict jobs in this workflow to have no permissions by default; permissions
+# should be granted per job as needed using a dedicated `permissions` block
+permissions: {}
+
+jobs:
+  update:
+    permissions:
+      contents: write # to fetch and commit code
+      actions: write # to manually dispatch checks on the pull request
+      pull-requests: write # Create pull requests
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        with:
+          persist-credentials: false
+      - uses: actions/setup-go@44694675825211faa026b3c33043df3e48a5fa00 # v6.0.0
+        with:
+          go-version: stable
+          check-latest: true
+      - run: go test ./...
+        working-directory: tools/apitester
+        env:
+          OSV_API_BASE_URL: api.test.osv.dev
+          UPDATE_SNAPS: always
+      - uses: peter-evans/create-pull-request@271a8d0340265f705b14b6d32b9829c1cb33d45e # v7.0.8
+        with:
+          token: ${{ secrets.PR_TOKEN_BOT }}
+          title: "test: update snapshots"
+          body: >
+            The snapshots have changed, probably due to OSV advisories being changed.
+
+            Please review the differences to make sure that they're expected!
+          branch: "bot/update-snapshots"
+          author: "osv-robot <osv-robot@google.com>"
+          commit-message: "test: update snapshots"
--- a/gcp/api/integration_tests.py
+++ b/gcp/api/integration_tests.py
@@ -788,6 +788,8 @@ class IntegrationTests(unittest.TestCase,
                     }, {
                         'id': 'CVE-2025-12875',
                     }, {
+                        'id': 'CVE-2025-13120',
+                    }, {
                         'id': 'CVE-2025-7207',
                     }, {
                         'id': 'OSV-2020-744',
--- a/gcp/website/frontend3/src/linter.js
+++ b/gcp/website/frontend3/src/linter.js
@@ -322,7 +322,7 @@ document.addEventListener("DOMContentLoaded", function () {
       paginatedIssues.forEach((issue) => {
         let row = tableBody.insertRow();
         row.classList.add("clickable");
-        row.addEventListener("click", (e) => {
+        row.addEventListener("click", () => {
           openCombinedView(issue.bug_id);
         });
 
@@ -437,7 +437,7 @@ document.addEventListener("DOMContentLoaded", function () {
           </div>
           <div class="details-column">
               <h2>Linter Findings</h2>
-              <pre id="${findingsJsonId}" class="json-pre">Loading...</pre>
+              <div id="${findingsJsonId}" class="json-pre">Loading...</div>
           </div>
       </div>
                 `;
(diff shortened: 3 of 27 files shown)


gcp/workers/importer/importer.py
```diff
@@ -638,6 +641,26 @@ def _process_updates_git(self, source_repo: osv.SourceRepository):
...
+        if (bug.import_last_modified and
+            orig_modified <= bug.import_last_modified):
+          logging.info(
+              'Skipping update for %s because incoming modification time (%s)'
+              ' is not newer than existing record (%s)', vuln.id, orig_modified,
+              bug.import_last_modified)
+        else:
+          update_bug_from_vuln(bug, vuln, source_repo.name, changed_entry)
+          bug.put()
+          log_update_latency(bug)
```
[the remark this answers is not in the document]
michaelkedar: Extracted this to a function

gcp/workers/importer/importer.py
```diff
@@ -1083,6 +1155,131 @@ def export_oss_fuzz(vulnerability, testcase_id, issue_id):
...
+  for affected in vuln.affected:
+    if not affected.HasField('package'):
+      filtered.append(affected)
+    elif osv.ecosystems.is_known(affected.package.ecosystem):
+      filtered.append(affected)
+    else:
+      logging.error('%s contains unknown ecosystem "%s"', vuln.id,
+                    affected.package.ecosystem)
+  del vuln.affected[:]
+  vuln.affected.extend(filtered)
```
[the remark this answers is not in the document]
michaelkedar: Sharing code between the worker and importer is a little bit cumbersome.
I don't really want to spend too much effort refactoring when I want to rewrite it all anyway.

osv/models.py
```diff
@@ -323,6 +323,9 @@ class Bug(ndb.Model):
   # Affected packages and versions.
   affected_packages: list[AffectedPackage] = ndb.LocalStructuredProperty(
       AffectedPackage, repeated=True)
+  # A checksum of the Affected packages and versions for the importer.
+  # TODO(michaelkedar): move this to Vulnerability entity
+  affected_checksum: str = ndb.TextProperty()
```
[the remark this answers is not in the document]
michaelkedar: This will be added when the importer is refactored to use the `Vulnerability` entities

gcp/workers/importer/importer.py
```diff
@@ -1083,6 +1099,158 @@ def export_oss_fuzz(vulnerability, testcase_id, issue_id):
...
+      'Importer put Bug',
+      extra={
+          'json_fields': {
+              'vuln_id': bug.db_id,
+              'latency': str(now - source_time),
+          }
+      })
+
+
+def maybe_add_to_database(vuln: vulnerability_pb2.Vulnerability, source: str,
```
another-rex: nit: put_if_newer?

gcp/workers/importer/importer.py
```diff
@@ -1083,6 +1099,158 @@ def export_oss_fuzz(vulnerability, testcase_id, issue_id):
...
+def new_bug_from_vuln(vuln: vulnerability_pb2.Vulnerability, source: str,
+                      path: str) -> osv.Bug:
+  """Create a new Bug entity from a vulnerability as the worker does."""
+  orig_modified = vuln.modified.ToDatetime(datetime.UTC)
+  bug = osv.Bug(
+      db_id=vuln.id,
+      timestamp=osv.utcnow(),
+      status=osv.BugStatus.PROCESSED,
+      source_of_truth=osv.SourceOfTruth.SOURCE_REPO)
+  bug.update_from_vulnerability(vuln)
```
another-rex, line 1137: Why are the following fields not updated inside update_from_vulnerability? (Also add a comment describing the reason)

gcp/workers/importer/importer.py
```diff
@@ -1083,6 +1099,158 @@ def export_oss_fuzz(vulnerability, testcase_id, issue_id):
...
+    bug.status = osv.BugStatus.INVALID
+
+  if bug.last_modified < prev_modified:
+    # Prevent the modified date going backwards,
+    # e.g. if source updated the bug again before the worker finished
+    bug.last_modified = osv.utcnow()
+
+  new_checksum = compute_raw_affected_checksum(vuln)
+  if bug.affected_checksum == new_checksum:
+    bug.affected_packages = prev_affected
```
another-rex, line 1190: Can you add a comment here describing what this section does?

another-rex: Also do we want to avoid modifying the last_modified if the checksum stays the same?

gcp/workers/importer/importer.py
```diff
@@ -721,12 +726,20 @@ def _process_updates_bucket(self, source_repo: osv.SourceRepository):
               'Failed to parse vulnerability (when considering for import) "' +
               blob.name + '"')
 
-      for cv in converted_vulns:
-        if cv:
+      for cv_result in converted_vulns:
+        if cv_result:
```
another-rex: nit: negate this if statement

another-rex (commented):
Nice!

gcp/workers/importer/importer.py
michaelkedar: I added a small comment.
If the affected_packages did not change, presumably something else had changed e.g. description, references.
We'd still want to update the modified date in that case.

gcp/workers/importer/importer.py
michaelkedar: I mostly copied what [`_do_update`](https://github.com/google/osv.dev/blob/0e41391f9fd7a71c5d061e4313f2bcfb052c3b72/gcp/workers/worker/worker.py#L556) does.
Made this more explicit in the docstring

gcp/workers/importer/importer.py
another-rex: What was confusing me was if checksum was unchanged, isn't setting it back to the previous affected a no op. Then I realized it's the raw vs enriched data. Can you add a comment clarifying that we reset it to keep the enriched affected versions

michaelkedar: the current comment says
```python
    # The checksum of the raw affected packages is unchanged (so no change has
    # been made to the affected packages). Restore the previously enriched,
    # affected_packages from the workers.
```
is that not clear enough?

another-rex: I somehow missed the word enriched here 🤦 , yep this is very clear!

PR #4347, opened 2025-11-13
fix: missing protojson.Marshal

michaelkedar:
I might look into seeing if we can make a linter to automatically detect these problems


cuixq:
thanks!

PR #4352, opened 2025-11-17
fix: Make linter output prettier

jess-lowe:
the right hand side output is hard for a human to read: 
image

So I improved it:
image

### DIFF
--- a/gcp/website/frontend3/src/linter.scss
+++ b/gcp/website/frontend3/src/linter.scss
@@ -471,6 +471,7 @@ tr.clickable {
   vertical-align: top;
   white-space: pre-wrap;
   word-break: break-word;
+  font-family: var(--osv-heading-font-family);
 }
 
 .findings-table th {


gcp/website/frontend3/src/linter.js
```diff
@@ -462,15 +462,31 @@ document.addEventListener("DOMContentLoaded", function () {
 
     // Display finding data
     const details = findingDetails[bugId];
-    const findingsPre = document.getElementById(findingsJsonId);
-    if (details) {
-      findingsPre.textContent = JSON.stringify(details, null, 2);
+    const findingsEl = document.getElementById(findingsJsonId);
+    if (details && details.length > 0) {
```
G-Rath:
```suggestion
    if (details?.length) {
```

Ly-Joey (approved):

another-rex (approved):
Nice! One nit would be to change the font inside the table to a monospace font.

another-rex (approved):

PR #4353, opened 2025-11-17
fix(frontend): render backticks on summary fields

### LINKED ISSUE #4345: Website rendering backticks in summary as diacritics
**Describe the bug**
Vulnerabilities that contain backticks (``` ` ```) in the summary field (intending to do `markdown formatting` even though summary is not meant to be markdown) end up placing the backtick above the following letters as accent marks (like `ò`)

This happens both on vulnerability page and list page.

**To Reproduce**
1. Go to a vulnerability that has a summary containing backticks (e.g. [GHSA-7cx5-254x-cgrq](https://osv.dev/vulnerability/GHSA-7cx5-254x-cgrq))
2. Look at Summary field
3. See error

**Expected behaviour**
Backticks should be rendered as distinct characters. (Or, the summary field should support markdown? but that's not osv-schema-compliant)

**Screenshots**
For [GHSA-7cx5-254x-cgrq](https://osv.dev/vulnerability/GHSA-7cx5-254x-cgrq), the summary is ```Parse Server allows public `explain` queries which may expose sensitive database performance information and schema details```, but ``` `explain` ``` is rendered strangely:

Image

ashmod:
There's an issue with backticks in the Overpass variable-weight font: https://github.com/RedHatOfficial/Overpass/blob/master/fonts/variable/Overpass%5Bwght%5D.ttf , which is the version hosted on Google Fonts and [used by the frontend](https://github.com/google/osv.dev/issues/4345#issuecomment-3541453203)

This change adds a filter to escape summary text backticks and replaces them with the monospace variant that does not have this issue. If the Overpass font is fixed in upstream, this workaround should no longer be needed.

Fixes #4345

### DIFF
--- a/gcp/website/frontend_handlers.py
+++ b/gcp/website/frontend_handlers.py
@@ -926,12 +926,14 @@ def literal_backticks(value: str | None) -> Markup:
   # TODO: This should be no longer needed if the Overpass font bug is fixed
   # (https://github.com/google/osv.dev/issues/4345#issuecomment-3541453203)
   # Summary fields render '`' characters as diacritics because of a bug
-  # with the Overpass font. This escapes them to the mono variant so they render correctly.
+  # with the Overpass font. This escapes them to the mono variant
+  # so they render correctly.
   if not value:
     return Markup('')
 
   escaped = escape(value)
-  if '`' not in (escaped_str := str(escaped)):
+  escaped_str = str(escaped)
+  if '`' not in escaped_str:
     return escaped
 
   replacement = '<span class="literal-backtick">`</span>'


gcp/website/frontend_handlers.py
```diff
@@ -919,6 +920,24 @@ def list_packages(vuln_affected: list[dict]):
...
+  """Escape text and render backticks correctly."""
+  # TODO: This should be no longer needed if the Overpass font bug is fixed
+  # (https://github.com/google/osv.dev/issues/4345#issuecomment-3541453203)
+  # Summary fields render '`' characters as diacritics because of a bug
+  # with the Overpass font. This escapes them to the mono variant so they render correctly.
+  if not value:
+    return Markup('')
+
+  escaped = escape(value)
+  if '`' not in (escaped_str := str(escaped)):
```
another-rex:
```suggestion
  escaped_str = str(escaped)
  if '`' not in escaped_str):
```

nit: Just place it on a normal line to help readability.

another-rex (dismissed):
Nice, thanks!

gcp/website/frontend_handlers.py
ashmod: Done, also lint should be fixed.

PR #4389, opened 2025-11-24
feat(vulnfeeds): Add CNA and source link of CVE5 file to database_specific.

jess-lowe:
For better triaging, CVE5 based records will now include CNA and the GitHub link to the CVE5 record. 

Also refactored so all of the populating of the database_specific happens in the same function.  

Closes #4389

### DIFF
--- a/vulnfeeds/cmd/cve-bulk-converter/main.go
+++ b/vulnfeeds/cmd/cve-bulk-converter/main.go
@@ -127,7 +127,8 @@ func worker(wg *sync.WaitGroup, jobs <-chan string, outDir string, cnas []string
 
 		sourceLink := ""
 		if commitHash != "" {
-			idx := strings.Index(path, "cves/")
+			baseDirCVEList := "cves/" // The base folder for the CVEListV5 repository.
+			idx := strings.Index(path, baseDirCVEList)
 			if idx != -1 {
 				relPath := path[idx:]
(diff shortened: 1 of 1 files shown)


vulnfeeds/cmd/cve-bulk-converter/main.go
```diff
@@ -116,8 +124,18 @@ func worker(wg *sync.WaitGroup, jobs <-chan string, outDir string, cnas []string
 		if errCVE != nil || errMetrics != nil {
 			logger.Fatal("File failed to be created for CVE", slog.String("cve", string(cveID)))
 		}
+
+		sourceLink := ""
+		if commitHash != "" {
+			idx := strings.Index(path, "cves/")
```
another-rex: put cves/ as a constant, and add a comment about where it comes from. Right now it feels like a magic string.

vulnfeeds/cvelist2osv/converter.go
```diff
@@ -304,3 +298,28 @@ func identifyPossibleURLs(cve cves.CVE5) []cves.Reference {
 
 	return refs
 }
+
+func populateDBSpecific(cve cves.CVE5, metrics *ConversionMetrics, sourceLink string) map[string]any {
```
another-rex: populate sounds like this is writing inplace, but it looksl iek this is actually returning dbspecific. Maybe buildDBSpecific or getDBSpecific?

another-rex (commented):
Nice, some nits

PR #4390, opened 2025-11-24
fix(cve5): group by introduced

jess-lowe:
The way we were interpreting version ranges where they have the same introduced is wrong as the schema treats ranges independently like an AND operator rather than an OR operator, meaning that if there is areas of no overlap between the ranges (like branches), all ranges are affected rather than just the ranges specified. 

This PR will group ranges if they have the same Introduced value into one set of events like:
```
{ 
  introduced: x,
  fixed: y,
  fixed: z 
} 
```

shoutout @michaelkedar for spotting this

### DIFF
--- a/vulnfeeds/cvelist2osv/grouping.go
+++ b/vulnfeeds/cvelist2osv/grouping.go
@@ -14,6 +14,7 @@ import (
 // groupAffectedRanges groups ranges that share the same introduced value, type, and repo.
 // This is because having multiple ranges with the same introduced value would act like an
 // OR condition, rather than AND.
+// This function modifies in-place
 func groupAffectedRanges(affected []*osvschema.Affected) {
 	for _, aff := range affected {
 		if len(aff.GetRanges()) <= 1 {
@@ -33,11 +34,16 @@ func groupAffectedRanges(affected []*osvschema.Affected) {
 		for _, r := range aff.GetRanges() {
 			// Find the introduced event
 			var introduced string
+			var introducedCount int
 			for _, e := range r.GetEvents() {
 				if e.GetIntroduced() != "" {
 					introduced = e.GetIntroduced()
-					break
+					introducedCount++
 				}
+			}
+
+			if introducedCount > 1 {
+				logger.Error("Multiple 'introduced' events found in a single range", slog.Any("range", r))
 			}
 
 			// If no introduced event is found, we use an empty string as the introduced value.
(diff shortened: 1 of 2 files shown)


vulnfeeds/cvelist2osv/grouping.go
```diff
@@ -0,0 +1,174 @@
...
+	if source == nil {
+		return
+	}
+	sourceVersions := source.GetFields()["versions"]
+	if sourceVersions == nil {
+		return
+	}
+
+	if target.GetDatabaseSpecific() == nil {
+		target.DatabaseSpecific = &structpb.Struct{Fields: make(map[string]*structpb.Value)}
```
michaelkedar: nit: I feel [`structpb.NewStruct(nil)`](https://pkg.go.dev/google.golang.org/protobuf/types/known/structpb#NewStruct) may be more idiomatic (though I'm unsure if `nil` is okay to use in that function)

vulnfeeds/cvelist2osv/grouping.go
```diff
@@ -0,0 +1,174 @@
...
+		}
+		if uniqueEvents[i].GetIntroduced() == "" && uniqueEvents[j].GetIntroduced() != "" {
+			return false
+		}
+		// If both are introduced or both are not, keep original order (stable)
+		return i < j
+	})
+
+	// Ensure only one Introduced event remains.
+	// Since we grouped by Introduced value, all Introduced events in this group are identical.
```
michaelkedar, line 184: I suspect it's not possible for something to get here with multiple different introduced values, but could you add a sanity check & log if the introduced values don't match just so we don't silently miss some problem.

michaelkedar (commented):
Nice work!

vulnfeeds/cvelist2osv/grouping_test.go
```diff
@@ -0,0 +1,432 @@
...
+								{Introduced: "2.0"},
+								{Fixed: "3.0"},
+							},
+						},
+					},
+				},
+			},
+		},
+		{
+			name: "Different DatabaseSpecific (non-versions) - merge",
```
another-rex: Please document the expected behavior here for database specific (i.e the second one gets  overwritten)

vulnfeeds/cvelist2osv/grouping.go
```diff
@@ -0,0 +1,181 @@
...
+		for _, key := range order {
+			r := groups[key]
+			r.Events = cleanEvents(r.GetEvents())
+			newRanges = append(newRanges, r)
+		}
+		aff.Ranges = newRanges
+	}
+}
+
+func mergeDatabaseSpecificVersions(target *osvschema.Range, source *structpb.Struct) {
```
another-rex, line 101: add a comment here giving some examples of the behavior in different situations.

vulnfeeds/cvelist2osv/grouping.go
```diff
@@ -0,0 +1,181 @@
...
+	"github.com/google/osv/vulnfeeds/utility/logger"
+	"github.com/ossf/osv-schema/bindings/go/osvschema"
+	"google.golang.org/protobuf/encoding/protojson"
+	"google.golang.org/protobuf/types/known/structpb"
+)
+
+// groupAffectedRanges groups ranges that share the same introduced value, type, and repo.
+// This is because having multiple ranges with the same introduced value would act like an
+// OR condition, rather than AND.
+func groupAffectedRanges(affected []*osvschema.Affected) {
```
another-rex, line 18: Specify that this modifies it in place.

vulnfeeds/cvelist2osv/grouping.go
```diff
@@ -0,0 +1,181 @@
...
+
+		groups := make(map[groupKey]*osvschema.Range)
+		var order []groupKey // To maintain deterministic order of first appearance
+
+		for _, r := range aff.GetRanges() {
+			// Find the introduced event
+			var introduced string
+			for _, e := range r.GetEvents() {
+				if e.GetIntroduced() != "" {
+					introduced = e.GetIntroduced()
```
another-rex, line 40: What if there is more than 1 introduced in events? (if that is possible, if not, please panic or throw an error  after adding a check here)

another-rex (approved):
LGTM!

PR #4395, opened 2025-11-25
feat: rewrite alias/upstream computation in Go

### LINKED ISSUE #4355: Add some retries on GCS operations in Go exporter
Transient GCS errors aren't uncommon, and any failure to write a vulnerability in the exporter aborts the whole job. This is causing some alerts that the exporter had not completed recently.

We should add some retry logic (I don't think it exists by default?) to these to prevent it from failing outright as often

michaelkedar:
This is pretty much a direct translation of the existing python code (which makes some of it unidiomatic go), changed to use the new `Vulnerability` instead of `Bug`.

One difference is the alias and upstream computations now run at the same time, and the writes to the Vulnerability protos now happen together. I had to modify the recoverer to be able to handle multiple updating multiple fields when the GCS generation is mismatched to support this properly.

Also ported over the datastore emulator manager, and made some interfaces for the google cloud clients (with the added benefit of introducing write retries which resolves #4355)
This has allowed my to bring over the existing tests.

I've renamed the `alias-computation` cron to `relations` - Cloud Deploy is going to create a new workload for this, and the old one is going to have to be deleted manually.

In a follow-up, I'm going to add another goroutine to calculate the related field, since these aren't being kept up to date at the moment.

### DIFF
--- a/go/osv/clients/gcs_client.go
+++ b/go/osv/clients/gcs_client.go
@@ -99,6 +99,9 @@ func (c *GCSClient) writeObjectOnce(ctx context.Context, path string, data []byt
 	obj := c.bucket.Object(path)
 
 	if opts != nil && opts.IfGenerationMatches != nil {
+		// Use generation matching to ensure we don't overwrite the object
+		// if it has been modified by another process since we last read it.
+		// If the conditions are not met, the write will fail.
 		conds := storage.Conditions{GenerationMatch: *opts.IfGenerationMatches}
 		if *opts.IfGenerationMatches == 0 {
 			conds = storage.Conditions{DoesNotExist: true}


go/cmd/relations/upstream.go
```diff
@@ -0,0 +1,272 @@
...
+				}
+			}
+		}
+		// Add the immediate upstreams of the vuln to the map
+		upstreamMap[vulnID] = group.UpstreamIDs
+		for _, upstream := range group.UpstreamIDs {
+			if _, ok := visited[upstream]; !ok && !slices.Contains(toVisit, upstream) {
+				toVisit = append(toVisit, upstream)
+			}
+		}
```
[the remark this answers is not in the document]
michaelkedar: FWIW, this logic is basically directly translated from the python implementation:
https://github.com/google/osv.dev/blob/3016163aca476ee220ca96450703a3d30411dbbe/gcp/workers/alias/upstream_computation.py#L171-L181

go/cmd/relations/relations.go
```diff
@@ -0,0 +1,114 @@
...
+		if err := ComputeAliasGroups(ctx, gc.datastoreClient, updater.Ch); err != nil {
+			logger.Error("failed to compute alias groups", slog.Any("err", err))
+		}
+	})
+	wg.Go(func() {
+		if err := ComputeUpstreamGroups(ctx, gc.datastoreClient, updater.Ch); err != nil {
+			logger.Error("failed to compute upstream groups", slog.Any("err", err))
+		}
+	})
+	wg.Wait()
```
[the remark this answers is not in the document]
michaelkedar: https://pkg.go.dev/sync#WaitGroup.Go

gcp/workers/recoverer/recoverer.py
```diff
@@ -127,39 +127,41 @@ def transaction():
...
-          aliases_modified = datetime.datetime.now(datetime.UTC)
-        else:
-          aliases = sorted(set(alias_group.bug_ids) - {vuln_id})
-          aliases_modified = alias_group.last_modified
-        # Only update the modified time if it's actually being modified
-        if vuln_proto.aliases != aliases:
-          vuln_proto.aliases[:] = aliases
-          if aliases_modified > modified:
-            modified = aliases_modified
+      fields = field.split(',')
```
another-rex, line 130: Can you describe in the PR description what this change is for?

go/cmd/exporter/exporter.go
```diff
@@ -69,45 +70,40 @@ func main() {
 	// 4. The `ecosystemWorker`s and the `allEcosystemWorker` process the vulnerabilities and
 	//    generate the final files, sending the data to be written to `routerToWriteCh`.
 	// 5. A pool of `writer` workers receive the file data and write it to the output.
-	gcsObjToDownloaderCh := make(chan *storage.ObjectHandle)
+	gcsObjToDownloaderCh := make(chan string)
```
another-rex: nit: rename this variable since it's taking in a path now
nit2: Alias string to a gcsPath type and put that as the channel type.

go/cmd/exporter/exporter.go
```diff
@@ -69,45 +70,40 @@ func main() {
...
 
-	it := vulnBucket.Objects(ctx, &storage.Query{Prefix: gcsProtoPrefix})
 	prevPrefix := ""
 MainLoop:
-	for {
-		attrs, err := it.Next()
-		if errors.Is(err, iterator.Done) {
-			break
-		}
+	for path, err := range vulnClient.Objects(ctx, gcsProtoPrefix) {
```
another-rex, line 94: This is so clean!

go/osv/clients/gcs_client.go
```diff
@@ -0,0 +1,147 @@
...
+			// Exponential backoff: 1s, 2s, 4s
+			time.Sleep(time.Duration(1<<(i-1)) * time.Second)
+		}
+		err = c.writeObjectOnce(ctx, path, data, opts)
+		if err == nil {
+			return nil
+		}
+		// Check if error is not transient and should not be retried
+		var apiErr *googleapi.Error
+		if !errors.As(err, &apiErr) || (apiErr.Code < 500 && apiErr.Code != 429) {
```
another-rex, line 90: Should this be Is? Since you are not using the ouput of apiErr

go/cmd/relations/alias.go
```diff
@@ -0,0 +1,256 @@
...
+	"github.com/google/osv.dev/go/osv/models"
+	"google.golang.org/api/iterator"
+)
+
+const (
+	aliasGroupVulnLimit = 32
+	vulnAliasesLimit    = 5
+)
+
+func updateAliasGroup(ctx context.Context, cl *datastore.Client, vulnIDs []string,
```
another-rex: Can you format this to be one arg per line?

go/cmd/relations/upstream.go
```diff
@@ -0,0 +1,272 @@
...
+	_, err = cl.Put(ctx, targetUpstreamGroup.Key, targetUpstreamGroup)
+
+	return err
+}
+
+func ComputeUpstreamGroups(ctx context.Context, cl *datastore.Client, ch chan<- Update) error {
+	// Query for all vulns that have upstreams.
+	var updatedGroups []*models.UpstreamGroup
+	logger.Info("Retrieving vulns for upstream computation...")
+	query := datastore.NewQuery("Vulnerability").FilterField("upstream_raw", ">", "")
```
another-rex, line 212: Should datastore queries also be wrapped behind an interface?

go/cmd/relations/upstream.go
```diff
@@ -0,0 +1,272 @@
...
+	}
+	group.Key = key
+	updateVulnWithUpstream(ch, vulnID, group)
+
+	return group, nil
+}
+
+func updateUpstreamGroup(ctx context.Context, cl *datastore.Client, group *models.UpstreamGroup, upstreamIDs []string, ch chan<- Update) (*models.UpstreamGroup, error) {
+	if len(upstreamIDs) == 0 {
+		logger.Info("Deleting upstream group due to too few bugs", slog.String("id", group.VulnID))
```
another-rex: too few bugs -> no vulns

go/cmd/relations/upstream_test.go
```diff
@@ -0,0 +1,245 @@
...
+	"slices"
+	"testing"
+
+	"cloud.google.com/go/datastore"
+	"github.com/google/osv.dev/go/osv/models"
+	"github.com/google/osv.dev/go/testutils"
+)
+
+func TestComputeUpstream(t *testing.T) {
+	bugs := map[string][]string{
```
another-rex: We probably should start using vulns as the term in the go version over bugs

go/cmd/relations/upstream.go
```diff
@@ -0,0 +1,272 @@
...
+		}
+		upstreamGroups[group.VulnID] = &group
+	}
+	logger.Info("Upstream groups successfully retrieved", slog.Int("count", len(upstreamGroups)))
+
+	for vulnID, upstreams := range vulns {
+		// Get the specific upstream group ID
+		group, exists := upstreamGroups[vulnID]
+		// Recompute the transitive upstreams and compare with the existing group
+		upstreamIDs := computeUpstream(upstreams, vulns)
```
another-rex: *newUpstreamIDs

go/cmd/relations/upstream.go
```diff
@@ -0,0 +1,272 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to iterate upstream groups: %w", err)
+		}
+		upstreamGroups[group.VulnID] = &group
+	}
+	logger.Info("Upstream groups successfully retrieved", slog.Int("count", len(upstreamGroups)))
+
+	for vulnID, upstreams := range vulns {
+		// Get the specific upstream group ID
+		group, exists := upstreamGroups[vulnID]
```
another-rex: group -> existingUpstreamGroup

go/cmd/relations/upstream.go
```diff
@@ -0,0 +1,272 @@
...
+		logger.Info("Deleting upstream group due to too few bugs", slog.String("id", group.VulnID))
+		if err := cl.Delete(ctx, group.Key); err != nil {
+			return nil, err
+		}
+		updateVulnWithUpstream(ch, group.VulnID, nil)
+
+		return nil, nil //nolint:nilnil
+	}
+
+	if slices.Equal(upstreamIDs, group.UpstreamIDs) {
```
another-rex, line 94: This is already checked outside?

go/cmd/relations/upstream.go
```diff
@@ -0,0 +1,272 @@
...
+			if slices.Equal(upstreamIDs, group.UpstreamIDs) {
+				continue
+			}
+			// Update the existing UpstreamGroup
+			var err error
+			group, err = updateUpstreamGroup(ctx, cl, group, upstreamIDs, ch)
+			if err != nil {
+				return fmt.Errorf("failed to update upstream group: %w", err)
+			}
+			if group == nil {
```
another-rex: Add a comment about nil being the group was deleted.

go/cmd/relations/upstream.go
```diff
@@ -0,0 +1,272 @@
...
+		group, exists := upstreamGroups[vulnID]
+		// Recompute the transitive upstreams and compare with the existing group
+		upstreamIDs := computeUpstream(upstreams, vulns)
+		if exists {
+			if slices.Equal(upstreamIDs, group.UpstreamIDs) {
+				continue
+			}
+			// Update the existing UpstreamGroup
+			var err error
+			group, err = updateUpstreamGroup(ctx, cl, group, upstreamIDs, ch)
```
another-rex: Not sure what to do about this, but it is a bit weird that this function both returns a new value, and also does the datastore updates.

go/cmd/relations/upstream.go
another-rex: We probably can remove the other check

go/cmd/relations/upstream.go
```diff
@@ -0,0 +1,272 @@
...
+	updateVulnWithUpstream(ch, group.VulnID, group)
+
+	return group, nil
+}
+
+func updateVulnWithUpstream(ch chan<- Update, vulnID string, group *models.UpstreamGroup) {
+	update := Update{ID: vulnID, Field: updateFieldUpstream}
+	if group == nil {
+		update.Timestamp = time.Now().UTC()
+		update.Value = []string(nil)
```
another-rex: Should this just be nil?

go/cmd/relations/upstream.go
```diff
@@ -0,0 +1,272 @@
...
+}
+
+func ComputeUpstreamGroups(ctx context.Context, cl *datastore.Client, ch chan<- Update) error {
+	// Query for all vulns that have upstreams.
+	var updatedGroups []*models.UpstreamGroup
+	logger.Info("Retrieving vulns for upstream computation...")
+	query := datastore.NewQuery("Vulnerability").FilterField("upstream_raw", ">", "")
+	it := cl.Run(ctx, query)
+
+	vulns := make(map[string][]string)
```
another-rex: Add a comment here about what vulns is suppose to contain.

go/cmd/relations/upstream_test.go
```diff
@@ -0,0 +1,245 @@
...
+	}{
+		{
+			name:     "Basic transitive",
+			input:    []string{"CVE-1", "CVE-2"},
+			expected: []string{"CVE-1", "CVE-2"},
+		},
+		{
+			name:     "Basic transitive from CVE-3",
+			input:    []string{"CVE-1", "CVE-2"}, // CVE-3's direct upstreams
+			expected: []string{"CVE-1", "CVE-2"},
```
another-rex: These two tests seem to be identical?

go/cmd/relations/upstream.go
```diff
@@ -0,0 +1,272 @@
...
+	"slices"
+	"time"
+
+	"cloud.google.com/go/datastore"
+	"github.com/google/osv.dev/go/logger"
+	"github.com/google/osv.dev/go/osv/models"
+	"google.golang.org/api/iterator"
+)
+
+func computeUpstream(targetBugUpstream []string, bugs map[string][]string) []string {
```
another-rex: Can you add a comment describing what this function does, and what the two args are. it's unclear why targetBugUpstream is a list.

go/cmd/relations/upstream.go
```diff
@@ -0,0 +1,272 @@
...
+		newGroup := make([]string, 0, len(upstreamMap[targetUpstreamGroup.VulnID]))
+		for _, upstream := range upstreamMap[targetUpstreamGroup.VulnID] {
+			if !slices.Contains(v, upstream) {
+				newGroup = append(newGroup, upstream)
+			}
+		}
+		upstreamMap[targetUpstreamGroup.VulnID] = newGroup
+	}
+
+	// Update the datastore entry if hierarchy has changed
```
another-rex: This comment should be below the if statement, or flipped?

go/osv/clients/gcs_client.go
```diff
@@ -0,0 +1,147 @@
...
+
+func (c *GCSClient) writeObjectOnce(ctx context.Context, path string, data []byte, opts *WriteOptions) error {
+	obj := c.bucket.Object(path)
+
+	if opts != nil && opts.IfGenerationMatches != nil {
+		conds := storage.Conditions{GenerationMatch: *opts.IfGenerationMatches}
+		if *opts.IfGenerationMatches == 0 {
+			conds = storage.Conditions{DoesNotExist: true}
+		}
+		obj = obj.If(conds)
```
another-rex, line 109: Add a comment describing what this is doing, specifically with obj.If (I'm guessing it conditionally edits the object when writing?

gcp/workers/recoverer/recoverer.py
michaelkedar: Added:
> One difference is the alias and upstream computations now run at the same time, and the writes to the Vulnerability protos now happen together. I had to modify the recoverer to be able to handle multiple updating multiple fields when the GCS generation is mismatched to support this properly.

go/cmd/exporter/exporter.go
michaelkedar: renamed to `gcsPathToDownloaderCh`
re: nit2 - I don't really see the benefit of doing that - it's fairly standard to use strings as paths in go (e.g. in the `path` package)

go/cmd/relations/alias.go
michaelkedar: formatted

go/cmd/relations/upstream.go
michaelkedar: I'm a bit unsure on the design of the interface/abstraction w.r.t queries and filters, and the datastore keys.
The abstraction wasn't needed right now (since the tests use the datastore emulator instead of mocking the database) so doing the interface wrapping can be done later when I have a better idea on how the interface should look.

go/cmd/relations/upstream_test.go
michaelkedar: I had been doing that in most places, just missed it in a couple
Fixed

go/cmd/relations/upstream.go
michaelkedar: outer check removed

go/cmd/relations/upstream.go
michaelkedar: I renamed this to `rawUpstreams` which is more descriptive

go/cmd/relations/upstream_test.go
michaelkedar: removed duplicate test

go/cmd/relations/upstream.go
michaelkedar: I added back in all the original docstrings from python to the go code, which should make it slightly clearer.

I also changed the function signature to be `computeUpstream(vulnID []string, rawUpstreams map[string][]string) []string` , which is simpler and clearer.

go/osv/clients/gcs_client.go
michaelkedar: added a comment to this block

go/cmd/exporter/exporter.go
another-rex: It's more to clarify whether you are meant to pass in a path, a name, or a vuln ID...etc. This makes it a bit more self documenting. I think you don't even need to cast when passing in a string for aliases

PR #4422, opened 2025-11-26
feat(CVE5): Add link to NVD and computed CVEList file and hash

jess-lowe:
It only makes sense that we include the references to the places we source our CVE data, both for triage, and just record keeping. 

This PR also fixes a bug where empty urls are able to be saved, which shouldn't happen based on the CVE5's quality requirements, but better safe than sorry. 

During this process, identifyPossibleURLs deduplication logic was moved to its own function to be used better elsewhere if necessary.


vulnfeeds/cvelist2osv/converter_test.go
```diff
@@ -85,7 +85,9 @@ func TestIdentifyPossibleURLs(t *testing.T) {
 				{URL: "http://a.com"},
 				{URL: "http://b.com"},
 				{URL: "http://c.com"},
+				{URL: "http://a.com"},
```
another-rex, line 88: Shouldn't this be deduplicated?

vulnfeeds/cvelist2osv/converter.go
```diff
@@ -267,15 +276,19 @@ func ConvertAndExportCVEToOSV(cve cves.CVE5, vulnSink io.Writer, metricsSink io.
...
 // It searches for URLs in the CNA and ADP reference sections, as well as in
 // the 'collectionUrl' and 'repo' fields of the 'affected' entries.
 func identifyPossibleURLs(cve cves.CVE5) []cves.Reference {
 	refs := cve.Containers.CNA.References
 
 	for _, adp := range cve.Containers.ADP {
 		if adp.References != nil {
-			refs = append(refs, adp.References...)
+			for _, ref := range adp.References {
+				if ref.URL != "" {
```
another-rex: Is this necessary if there's a filtering step further down?

vulnfeeds/vulns/vulns.go
```diff
@@ -425,6 +425,10 @@ func ClassifyReferenceLink(link string, tag string) osvschema.Reference_Type {
 	// Index 0 will always be "", so the length must be at least 2 to be relevant
 	if len(pathParts) >= 2 {
 		if u.Host == "github.com" {
+			if len(pathParts) >= 3 && pathParts[1] == "CVEProject" {
```
another-rex, line 430: Please add a example comment like below, and probably another comment describing what this is for, it's not clear to me from just looking at this.

vulnfeeds/cvelist2osv/converter_test.go
jess-lowe: No because deduplication was moved to a different function and no longer called by identifyPossibleURLs

vulnfeeds/cvelist2osv/converter.go
jess-lowe: Good catch

another-rex (approved):

another-rex (approved):

PR #4429, opened 2025-11-27
chore: Update pull request title for snapshot updates

cuixq:
To specify this is for `apitester` snapshots


G-Rath:
Are there other types of snapshots in this codebase that this could be confused with? if you are set on changing it, I think it'd be better to title it "apitester snapshots"

cuixq:
There are snapshot-like tests in worker. I have updated the title to `apitester snapshots`.

another-rex (approved):

PR #4434, opened 2025-12-02
perf(impact): Optimise repo walking when `consider_all_branches` and `detect_cherrypick` == false

jess-lowe:
# Changes
### 1. Deduplicate Branches
In `_get_affected_range`, I added a step to deduplicate the list of branches to process. This prevents redundant analysis when multiple commits belong to the same branch.

```Go
# Deduplicate branches
branches = sorted(set(branches))
```

### 2. Limit Branches
I added a safeguard to limit the number of branches analyzed. If the number of branches exceeds 500, the analysis is restricted to master and main branches only. This prevents excessive processing time for repositories with thousands of branches (e.g., mariadb/server).
```Go
if len(branches) > 500:
      logging.warning('Too many branches (%d), limiting to master/main',
                      len(branches))
      branches = [
          b for b in branches if b.endswith('/master') or b.endswith('/main')
      ]
```
### 3. Optimize `_get_equivalent_commit`
I modified _get_equivalent_commit to use repo.merge_base instead of walking the entire commit history when detect_cherrypicks is False.

```Go
if not detect_cherrypicks:
      try:
        if str(repo.merge_base(search.id, target.id)) == target_commit:
          return target_commit
      except (ValueError, KeyError):
        # Invalid commit or no merge base
        pass
      return None
```
## Verification Results
**MariaDB Benchmark as benchmark**
I ran a benchmark on mariadb/server (which has >3000 branches). with the branch limit on, 
WARNING: root: Too many branches (3299), limiting to master/main
Analysis took 147.1882 seconds.

Without the branch limit, previously `mariadb/server` took 2 hr 22 min. With the `repo.merge_base` optimisation, it is down to 46 mins, roughly a 70% reduction. 

On a smaller repository, there seems to be no differences between the AffectedCommits extracted. 

Currently testing on mariadb/server - both pre and post produce no affected commits?


osv/impact.py
```diff
@@ -216,6 +216,16 @@ def _get_affected_range(self,
         for regress_commit in regress_commits:
           branches.extend(_branches_with_commit(repo, regress_commit))
 
+    # Deduplicate branches
+    branches = sorted(set(branches))
+
+    if len(branches) > 500:
```
another-rex: Make this a const at least.

another-rex (dismissed):

another-rex (dismissed):

PR #4435, opened 2025-12-02
perf(worker): disable vanir generation on kernel vulns

jess-lowe:
It's not using the mirror so it's being blocked and taking 20 mins to do so

### DIFF
--- a/gcp/workers/worker/worker.py
+++ b/gcp/workers/worker/worker.py
@@ -508,7 +508,7 @@ class TaskRunner:
           'Skipping Vanir signature generation for %s as it has no '
           'GIT affected ranges.', vulnerability.id)
       return vulnerability
-    if any(affected.package.name == "Kernel" and 
+    if any(affected.package.name == "Kernel" and
            affected.package.ecosystem == "Linux"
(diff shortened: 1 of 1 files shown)


gcp/workers/worker/worker.py
```diff
@@ -508,6 +508,12 @@ def _generate_vanir_signatures(self, vulnerability):
           'Skipping Vanir signature generation for %s as it has no '
           'GIT affected ranges.', vulnerability.id)
       return vulnerability
+    if any(affected.package.name == "Kernel"
```
cuixq: out of curiosity: what is an example of the kernel vulnerability?

jess-lowe: https://osv.dev/vulnerability/CVE-2025-21652

cuixq: shall we add a check on ecosystem being `Linux`?

jess-lowe:
```suggestion
    if any(affected.package.name == "Kernel" and affected.package.ecosystem == "Linux"
```

PR #4462, opened 2025-12-03
fix(website): add GOOGLE_CLOUD_PROJECT to run_tests.sh

ashmod:
Without this, frontend_handlers_test fails on import because ndb.Client() requires a project to be set.
This matches how [`/worker/run_tests.sh`](https://github.com/google/osv.dev/blob/ff13718c6bcd1e67cba5c29768c4945d643078ba/gcp/workers/worker/run_tests.sh#L16) handles it.


jess-lowe (approved):
LGTM

PR #4468, opened 2025-12-08
fix: non-determinism of affected range output in NVD conversion

jess-lowe:
We have an issue where the CVE record kept being updated by the order of the affected ranges changing. Found the non-determinism was caused by iterating over a map of repositories in the NVD record conversion.

Tested this change by running this test:

``` Go
func TestAttachExtractedVersionInfo_Determinism(t *testing.T) {
	// Create a VersionInfo with multiple repositories to trigger map iteration randomness
	versionInfo := models.VersionInfo{
		AffectedCommits: []models.AffectedCommit{
			{Repo: "https://github.com/repo/A", Fixed: "fixA"},
			{Repo: "https://github.com/repo/B", Fixed: "fixB"},
			{Repo: "https://github.com/repo/C", Fixed: "fixC"},
			{Repo: "https://github.com/repo/D", Fixed: "fixD"},
			{Repo: "https://github.com/repo/E", Fixed: "fixE"},
		},
	}

	// Run multiple times and compare with the first result
	var firstResult *Vulnerability
	
	for i := 0; i < 100; i++ {
		v := &Vulnerability{
			Vulnerability: &osvschema.Vulnerability{},
		}
		AttachExtractedVersionInfo(v, versionInfo)

		if i == 0 {
			firstResult = v
			continue
		}

		if diff := cmp.Diff(firstResult, v, protocmp.Transform()); diff != "" {
			t.Fatalf("Iteration %d produced different result:\n%s", i, diff)
		}
	}
}
```

### DIFF
--- a/vulnfeeds/vulns/vulns_test.go
+++ b/vulnfeeds/vulns/vulns_test.go
@@ -447,3 +447,35 @@ func TestNVD2(t *testing.T) {
 	cve := loadTestData2("CVE-2023-4863")
 	t.Logf("Loaded: %#v", cve)
 }
+
+func TestAttachExtractedVersionInfo_Determinism(t *testing.T) {
+	// Create a VersionInfo with multiple repositories to trigger map iteration randomness
+	versionInfo := models.VersionInfo{
+		AffectedCommits: []models.AffectedCommit{
+			{Repo: "https://github.com/repo/A", Fixed: "fixA"},
+			{Repo: "https://github.com/repo/B", Fixed: "fixB"},
+			{Repo: "https://github.com/repo/C", Fixed: "fixC"},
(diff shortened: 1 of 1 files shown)


vulnfeeds/vulns/vulns.go
```diff
@@ -110,7 +110,14 @@ func AttachExtractedVersionInfo(v *Vulnerability, version models.VersionInfo) {
 		}
 	}
 
-	for repo, commits := range repoToCommits {
+	repos := make([]string, 0, len(repoToCommits))
+	for r := range repoToCommits {
+		repos = append(repos, r)
+	}
+	sort.Strings(repos)
```
another-rex: Use `... := range slices.Sorted(maps.Keys(repoToCommits))`

PR #4469, opened 2025-12-08
feat: Make website no longer depend on Bug entities

michaelkedar:
Make use of the `ListedVulnerability` datastore entity for the `/list` page, and the GCS bucket for the vulnerability page, instead of the hefty, deprecated `Bug` entity for both.

Unfortunately, the `ListedVulnerability` does not quite have enough information as-is to 100% match the current behaviour, and a couple of things will be regressed:
- #3356 because we don't have alias/related/upstream information on the new entity. I've made exact ID matches sort first, which should be okay temporarily.
- #3812 / #3813 since there is no per-ecosystem `is_fixed` field on the `ListedVulnerability` (and computing it currently would require another query to the full vulnerability, which I'm trying to avoid on the list page). I think we could update the `ListedVulnerability` to have a list of fixed ecosystems to re-enable this, but I'd want to do that in a followup PR.
- Also, records that have no affected packages (but are not withdrawn) will now show up on the search page. IDK if that's preferable or not.

### DIFF
--- a/gcp/website/frontend_handlers.py
+++ b/gcp/website/frontend_handlers.py
@@ -434,6 +434,10 @@ def add_cvss_score(record: dict):
   severity_rating = None
   severity_type = None
 
+  # The OSV record (vulnerability page) has the list of severities under the
+  # 'severity' field, while the ListedVulnerability (/list page) has it under
+  # the 'severities' field.
+  # This function is used for both, so check both.
   sev = ''
   if 'severity' in record:
     sev = 'severity'


gcp/website/frontend_handlers.py
```diff
@@ -432,30 +428,35 @@ def calculate_severity_details(
...
   severity_score = None
   severity_rating = None
   severity_type = None
 
-  for severity in bug.get('severity', []):
+  sev = ''
+  if 'severity' in record:
+    sev = 'severity'
+  elif 'severities' in record:
+    sev = 'severities'
```
another-rex, line 445: Is this correct?

michaelkedar: Yeah - I use this function both for the vulnerability page (osv vulns have the `severity` field) and the list page (ListedVulnerabilities use `severities`)

PR #4470, opened 2025-12-08
test: monitor CVE5 conversion regressions through snapshot testing

jess-lowe:
This PR is intended for regression monitoring, and should probably be set up to only run under a specific flag if the conversion code is changed, as it will be doing a lot of converting.

When the data is converted, a snapshot is saved per CNA so that if it changes we can see _what_ exactly has changed.

We also keep a snapshot of the conversion stats, to monitor that on a higher level. 

A follow up PR will include all of the sampled data we want to be testing.


vulnfeeds/cvelist2osv/snapshot_test.go
```diff
@@ -0,0 +1,76 @@
...
+	"strings"
+	"testing"
+
+	"github.com/gkampitakis/go-snaps/snaps"
+)
+
+// TestSnapshotConversion runs the conversion process on a sample of CVEs and
+// creates snapshots of the output for comparison. This is used for monitoring
+// progressions and regressions when making changes to the converter.
+func TestSnapshotConversion(t *testing.T) {
```
another-rex, line 16: t.Parallel()

vulnfeeds/cvelist2osv/snapshot_test.go
```diff
@@ -0,0 +1,76 @@
...
+			snaps.WithConfig(snaps.Filename("cna-snaps/"+cna)).MatchSnapshot(t, vWriter.String())
+		})
+	}
+
+	// Sort keys for deterministic output
+	keys := make([]string, 0, len(stats))
+	for k := range stats {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
```
another-rex: Use the shorter form with maps.Key

vulnfeeds/cvelist2osv/snapshot_test.go
```diff
@@ -0,0 +1,76 @@
...
+
+	var statsOutput strings.Builder
+	statsOutput.WriteString("Conversion Outcomes:\n")
+	for _, k := range keys {
+		statsOutput.WriteString(fmt.Sprintf("%s: %d\n", k, stats[k]))
+	}
+	snaps.WithConfig(snaps.Filename("conversion_outcomes")).MatchSnapshot(t, statsOutput.String())
```
another-rex: Actually is this useful? The snapshot of each individual record would change (including the metrics) without this aggregate right?  Since we are no longer putting every single record in here, it's probably not needed here.

another-rex: Do we have conversion_outcomes aggregation being written normally and uploaded? If not we should do that.

jess-lowe: Personally I'd find it useful because I have the high level overview of numbers going in which direction. If it were numbers going up to Successful, I wouldn't need to look into it as much as NoRanges suddenly spiking. It's just slightly more useful for speedy triage.

jess-lowe: We don't currently have conversion_outcomes being written and uploaded for the CVE5 logic, but there is an equivalent for NVD. The NVD one uses a CSV file that I'm not convinced isn't being truncated at 5000 entries.

It'd be good to use this data though in the future for our metrics dashboard.

vulnfeeds/cvelist2osv/snapshot_test.go
jess-lowe: Can't parallelize this because we are grouping snapshots by CNA, so there would be race conditions causing snapshots to be non-deterministic.

vulnfeeds/cvelist2osv/snapshot_test.go
jess-lowe: maps.Keys isn't deterministic :/

vulnfeeds/cvelist2osv/snapshot_test.go
jess-lowe: Problem is that we're running conversion in parallel and would need to do some analysis using the .metrics.json files, or do icky stuff with mutexes. 

Have got the earlier solution in another WIP PR that I can put up after this one.

vulnfeeds/cvelist2osv/snapshot_test.go
another-rex: I meant use `slices.Sorted(maps.Key(stats))`, same as in the previous PR.

vulnfeeds/cvelist2osv/snapshot_test.go
another-rex: Should be fine to `t.Parallel` here though? This just makes it run at the same time as snapshot_test.go, which should be independent to this I believe.

vulnfeeds/cvelist2osv/snapshot_test.go
another-rex: hmm main concern is I don't want the direction to be part of tests. That'll work in the short term but I feel might get really annoying to update in the long term.

jess-lowe: direction?

vulnfeeds/cvelist2osv/snapshot_test.go
jess-lowe: if you mean the stats output test, then no it can't be run in parallel as the the rest of the files first need to be processed in order to have a stats outcome.

The files can't be parallelized easily either, because of the non-determinism thing grouping them by CNA and that they're all writing to the same stats output.

PR #4481, opened 2025-12-09
fix: remove `index-url` from poetry requirements.txt

michaelkedar:
We don't want this to talk to artifact registry
(I'm not 100% sure if this will work as-is)


cuixq:
I am more curious how this is added? did you just run `pip-compile --generate-hashes`?

michaelkedar:
> I am more curious how this is added? did you just run `pip-compile --generate-hashes`?

I didn't have the airlock exception on my laptop, so I guess it automatically detected it when I ran it

PR #4487, opened 2025-12-15
feat: migrate sitemap generation

michaelkedar:
Migrated sitemap generation to Go & away from Bug entities.
Had to use the entire vulnerability records from GCS in order to chunk them by ecosystem, which is going to make it quite a bit slower from having to download the entirety of all the files (it only runs once per day anyway so it's probably no big deal).
I *could* make it faster by using the `Vulnerability` entities from datastore, but those don't have the ecosystems attached to them. The `ListedVulnerability` does have the ecosystems, but not the modified dates :upside_down_face: (and don't have the withdrawn vulnerabilities).

Some improvements: the sitemap now splits large ecosystems into multiple pages (instead of truncating the vulnerabilities), and withdrawn vulnerabilities are now included in the sitemap (since they do actually have pages you can visit on osv.dev)


go/cmd/generatesitemap/generatesitemap.go
```diff
@@ -0,0 +1,334 @@
...
+	// Channel for GCS paths
+	gcsPathCh := make(chan string)
+	// Channel for parsed entries
+	entryCh := make(chan *osvschema.Vulnerability)
+
+	var wg sync.WaitGroup
+	// Start workers
+	for range *numWorkers {
+		wg.Go(func() { downloader(ctx, vulnClient, gcsPathCh, entryCh) })
+	}
```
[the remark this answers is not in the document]
michaelkedar: lies

deployment/clouddeploy/gke-workers/environments/oss-vdb-test/generate-sitemap.yaml
```diff
@@ -10,9 +10,11 @@ spec:
...
           - name: generate-sitemap-cron
             env:
-            - name: BASE_URL
-              value: "https://test.osv.dev"
             - name: GOOGLE_CLOUD_PROJECT
               value: oss-vdb-test
-            - name: OUTPUT_GCS_BUCKET
-              value: test-osv-dev-sitemap
+            args:
+              - "--base-url='https://test.osv.dev'"
```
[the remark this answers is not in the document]
michaelkedar: This also seems to not actually be true, but I've removed the quotes anyway

go/cmd/generatesitemap/generatesitemap.go
```diff
@@ -0,0 +1,333 @@
...
+
+	// Start listing objects
+	go func() {
```
another-rex, line 107: nit: Move this above the workers, so it's clearer at a glance this is creating stuff for the workers to consume.

go/cmd/generatesitemap/generatesitemap.go
```diff
@@ -0,0 +1,333 @@
...
+
+func writeSitemapFile(ctx context.Context, client clients.CloudStorage, path string, v any) error {
+	data, err := xml.MarshalIndent(v, "", "  ")
+	if err != nil {
+		return err
+	}
+	data = append([]byte(xml.Header), data...)
+
+	if client != nil {
+		return client.WriteObject(ctx, path, data, nil)
```
another-rex: Might be worth it to do a crc32 check here. Most ecosystems probably change often enough that it won't work, but for the smaller ecosystems they don't change that often, so it'll keep the modified date metadata useful.

another-rex (approved):
Minor nits, otherwise LGTM!

go/cmd/generatesitemap/generatesitemap.go
michaelkedar: I've added a checksum check here (and logic for getting the checksum through the client interface)

michaelkedar:
> LGTM, though I thought CRC is already done by google cloud, and we don't need to apply a custom attribute to store it?

It is, and we aren't.
We're wrapping the GCS client read & attributes functions (so we can swap it out for an in-memory client when testing), so I just had to add it to the returned attributes from the wrapper interface (and manually compute/store it in the in-memory client).

PR #4508, opened 2025-12-17
feat: Gitter attempt

another-rex:
The first pass implementation of Gitter, a git proxy/caching service that makes it easier to do fast copies of git repos.

- This implementation only speeds up the initial clone, and not additional fetches (those are pretty fast already (e.g. 1 week behind for the linux kernel takes 18s or so), and that's probably worst worst case)
  - This does leave open options for additional speedups/caches for fetches as well
- Adds a new 2tb disk and the gitter worker.
- Adds support for multi kube configs in a single file in the terraform
- Creates a service exposing port 8888
- Accounts for timeouts by using curl which doesn't have a timeout (and also probably faster than python reading bytes and writing to disk)
- Error propagation from gitter to repo.py

---

For testing:

- Added GITTER_PORT variable to specify what port to run gitter on during tests
- Added a test context manager for gitter to run gitter at the start of tests.

### DIFF
--- a/go/cmd/gitter/gitter_test.go
+++ b/go/cmd/gitter/gitter_test.go
@@ -58,7 +58,7 @@ func TestGitHandler_InvalidURL(t *testing.T) {
 	}{
 		{"", http.StatusBadRequest},
 		{"ftp://example.com/repo.git", http.StatusBadRequest}, // ftp is not allowed (it hangs for some reason)
-		{"file:///etc/passwd", http.StatusBadRequest}, // file protocol not allowed
+		{"file:///etc/passwd", http.StatusBadRequest},         // file protocol not allowed
 	}
 
 	for _, tt := range tests {


osv/tests.py
```diff
@@ -239,3 +240,58 @@ def mock_clone(test, func=None, return_value=None):
   else:
     mocked.side_effect = func
   test.addCleanup(patcher.stop)
+
+
+def setup_gitter(env_var):
```
michaelkedar: can you turn this into a context manager like the [datastore emulator](https://github.com/google/osv.dev/blob/6252d8b238d245c3008c3870d29f1e9ef4f15751/osv/tests.py#L137-L138), and use `unittest.enterModuleContext()` instead of manually closing it in the teardown

osv/tests.py
```diff
@@ -239,3 +240,58 @@ def mock_clone(test, func=None, return_value=None):
...
+  """Setup gitter."""
+  address = os.environ.get(env_var)
+  if not address:
+    raise ValueError(f'{env_var} environment variable is not set')
+
+  port = 80
+  if ':' in address:
+    port = int(address.split(':')[-1])
```
michaelkedar: we're not allowing IPv6 I see :)

osv/tests.py
```diff
@@ -239,3 +240,58 @@ def mock_clone(test, func=None, return_value=None):
...
+
+  repo_root = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+  go_dir = os.path.join(repo_root, 'go')
```
michaelkedar: nit: might be easier to do
```python
go_dir = os.path.abspath(os.path.join(__file__, '..', '..', 'go'))
```

osv/tests.py
```diff
@@ -239,3 +240,58 @@ def mock_clone(test, func=None, return_value=None):
...
+    # If it returns, it exited
+    raise RuntimeError(f'Gitter exited early: {proc.stderr.read().decode()}')
+  except subprocess.TimeoutExpired:
+    # Process is still running
+    pass
+
+  def cleanup():
+    # Kill the process group
+    try:
+      os.killpg(os.getpgid(proc.pid), signal.SIGKILL)
```
michaelkedar: It might be more polite to `SIGINT` to give it a chance to clean up.

osv/repos.py
```diff
@@ -115,23 +117,14 @@ class RepoInaccessibleError(Exception):
 def clone(git_url, checkout_dir, git_callbacks=None, blobless=False):
   """Perform a clone."""
   try:
-    # Use 'git' CLI here as it's much faster than libgit2's clone.
-    env = _set_git_callback_env(git_callbacks)
-    cmd = ['git', 'clone']
-    if blobless:
-      cmd.append('--filter=blob:none')
-    cmd.extend([_git_mirror(git_url), checkout_dir])
-    subprocess.run(cmd, env=env, capture_output=True, check=True)
```
michaelkedar, line 124: Could we maybe keep the original `git clone` logic if the `GITTER_HOST` environment variable isn't set

go/cmd/gitter/gitter.go
```diff
@@ -0,0 +1,150 @@
...
+	archivePath := repoPath + ".zst"
+
+	lastFetchMu.Lock()
+	accessTime, ok := lastFetch[url]
+	lastFetchMu.Unlock()
+
+	// Check if we need to fetch
+	if !ok || time.Since(accessTime) > fetchTimeout {
+		logger.Info("Fetching git blob", slog.String("url", url), slog.Duration("sinceAccessTime", time.Since(accessTime)))
+		if _, err := os.Stat(path.Join(repoPath, ".git")); os.IsNotExist(err) {
```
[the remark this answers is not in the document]
another-rex: Yes, but not really. path.base will strip out any `/`, so no path traversal. And everything else should be fine. The hashes will mean you can't confuse your repo with another one.

osv/repos.py
```diff
@@ -115,23 +117,14 @@ class RepoInaccessibleError(Exception):
...
+    subprocess.run(cmd, check=True)
     return pygit2.Repository(checkout_dir)
   except subprocess.CalledProcessError as e:
-    stderr = e.stderr.decode(errors='ignore')
-    if ('could not read Username' in stderr or
-        ('fatal: repository' in stderr and 'not found' in stderr) or
-        'Authentication failed' in stderr):
-      # Git is asking for username/password, the repository doesn't exist, or
-      # authentication failed.
-      raise RepoInaccessibleError() from e
```
michaelkedar, line 133: You need to make sure you retain similar logic to this in the gitter and pass it through here - it's important for malicious packages where the repo was deleted

go/cmd/gitter/gitter.go
```diff
@@ -0,0 +1,150 @@
...
+			// Clone
+			cmd := exec.Command("git", "clone", url, repoPath)
```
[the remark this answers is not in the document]
another-rex: I'm not sure this is true, each segment of command will be quoted and taken as one, so there's no way to both pass a flag and a url in at the same time. I'll sanitise to check that it starts with `[a-zA-Z]+://`, so it's never a flag.

go/cmd/gitter/gitter.go
```diff
@@ -0,0 +1,150 @@
...
+		val, err, _ := g.Do(url, func() (interface{}, error) {
+			return fetchBlob(r.Context(), url)
+		})
+		file := val.(*os.File)
+		defer file.Close()
+
+		if err != nil {
+			http.Error(w, fmt.Sprintf("Error fetching blob: %v", err), http.StatusInternalServerError)
+			return
+		}
```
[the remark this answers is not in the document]
another-rex: Good point

osv/repos.py
```diff
@@ -115,23 +117,14 @@ class RepoInaccessibleError(Exception):
...
   try:
-    # Use 'git' CLI here as it's much faster than libgit2's clone.
-    env = _set_git_callback_env(git_callbacks)
-    cmd = ['git', 'clone']
-    if blobless:
-      cmd.append('--filter=blob:none')
-    cmd.extend([_git_mirror(git_url), checkout_dir])
-    subprocess.run(cmd, env=env, capture_output=True, check=True)
+    os.makedirs(checkout_dir, exist_ok=True)
+    cmd = ['curl', f"{GITTER_HOST}/getgit?url="+ _git_mirror(git_url), "-o", checkout_dir+".zst", "-s"]
```
michaelkedar: few things:
1. nit: single quotes for python strings
2. nit: instead of concatenating:`f'{GITTER_HOST}/getgit?url={_git_mirror(git_url)}'`
3. I doubt `_git_mirror` / `git_url` is safe to use as a query param without encoding?

go/cmd/gitter/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+COPY ./go.sum /src/go.sum
+RUN go mod download && go mod verify
+
+
+COPY ./ /src/
+RUN CGO_ENABLED=0 go build -o gitter ./cmd/gitter/
+
+FROM alpine:3.23@sha256:51183f2cfa6320055da30872f211093f9ff1d3cf06f39a0bdb212314c5dc7375
+
+RUN apk add --no-cache git zstd
```
michaelkedar, line 29: Is there a reason not to use [`go-git`](https://pkg.go.dev/github.com/go-git/go-git/v5) and [github.com/klauspost/compress/zstd](https://pkg.go.dev/github.com/klauspost/compress/zstd)?

I like having things in pure go :)

deployment/terraform/modules/osv/workers_gke.tf
```diff
@@ -172,3 +175,11 @@ resource "google_project_iam_member" "compute_service_datastore" {
   role    = "roles/datastore.importExportAdmin"
   member  = "serviceAccount:${data.google_compute_default_service_account.default.email}"
 }
+
+resource "google_compute_disk" "gitter_disk" {
+  project = var.project_id
+  name    = "gitter-disk"
+  type    = "pd-ssd"
+  zone    = "us-central1-f"
```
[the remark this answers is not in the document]
another-rex: I think this is fine.

osv/repos.py
```diff
@@ -115,23 +117,14 @@ class RepoInaccessibleError(Exception):
 def clone(git_url, checkout_dir, git_callbacks=None, blobless=False):
```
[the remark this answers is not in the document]
another-rex: I want to avoid touching the signature for now.

go/cmd/gitter/gitter.go
```diff
@@ -0,0 +1,150 @@
...
+	http.HandleFunc(getGitEndpoint, func(w http.ResponseWriter, r *http.Request) {
+		url := r.URL.Query().Get("url")
+		if url == "" {
+			http.Error(w, "Missing url parameter", http.StatusBadRequest)
+			return
+		}
+
+		logger.Info("Received request for %q", url)
+
+		val, err, _ := g.Do(url, func() (interface{}, error) {
```
michaelkedar: I *think* you might need to call `g.Forget(url)` at some point so that it doesn't always return the result of the first time it was ever called.

go/cmd/gitter/Dockerfile
another-rex: It's 2x slower apparently.

osv/repos.py
another-rex: > I doubt _git_mirror / git_url is safe to use as a query param without encoding?

surprisingly it's fine as long as there's no &, which I guess there won't be because git repos generally don't have queries in them.

Still, good idea to encode.

go/cmd/gitter/gitter.go
another-rex: I think that's only needed if you want multiple calls in flight at the same time.

michaelkedar: Ah, you're right about that

osv/repos.py
another-rex: Done.

osv/repos.py
another-rex: Done.

deployment/terraform/modules/osv/workers_gke.tf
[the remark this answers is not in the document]
michaelkedar: The node pools are using `google_container_cluster.workers.location` for this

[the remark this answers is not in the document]
another-rex: Ah good idea, done.

deployment/clouddeploy/gke-workers/base/gitter.yaml
```diff
@@ -0,0 +1,80 @@
...
+        volumeMounts:
+        - mountPath: /work
+          name: disk-data
+      resources:
+        requests:
+          cpu: "28"
+          memory: "200Gi"
+        limits:
+          cpu: "28"
+          memory: "200Gi"
```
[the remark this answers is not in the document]
another-rex: No, we want to adjust down, not up

go/cmd/gitter/gitter.go
```diff
@@ -0,0 +1,171 @@
...
+	lastFetchMu.Lock()
+	accessTime, ok := lastFetch[url]
+	lastFetchMu.Unlock()
+
+	// Check if we need to fetch
+	if !ok || time.Since(accessTime) > fetchTimeout {
+		logger.Info("Fetching git blob", slog.String("url", url), slog.Duration("sinceAccessTime", time.Since(accessTime)))
+		if _, err := os.Stat(path.Join(repoPath, ".git")); os.IsNotExist(err) {
+			// Clone
+			cmd := exec.Command("git", "clone", url, repoPath)
```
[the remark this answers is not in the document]
another-rex: Oh that's cool

go/cmd/gitter/gitter.go
```diff
@@ -0,0 +1,171 @@
...
+		url := r.URL.Query().Get("url")
+		if url == "" {
+			http.Error(w, "Missing url parameter", http.StatusBadRequest)
+			return
+		}
+
+		logger.Info("Received request", slog.String("url", url))
+
+		// Check if url starts with regex [a-zA-Z]+://
+		if match, _ := regexp.MatchString("^[a-zA-Z]+://", url); !match {
```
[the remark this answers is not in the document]
another-rex: Done

osv/repos.py
```diff
@@ -115,23 +117,14 @@ class RepoInaccessibleError(Exception):
...
   except subprocess.CalledProcessError as e:
-    stderr = e.stderr.decode(errors='ignore')
-    if ('could not read Username' in stderr or
-        ('fatal: repository' in stderr and 'not found' in stderr) or
-        'Authentication failed' in stderr):
-      # Git is asking for username/password, the repository doesn't exist, or
-      # authentication failed.
-      raise RepoInaccessibleError() from e
-    raise GitCloneError(f'Failed to clone repo:\n{stderr}') from e
+    raise GitCloneError(f'Failed to clone repo:\n{e}') from e
```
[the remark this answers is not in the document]
another-rex: This is now fixed

osv/tests.py
another-rex: Well removed the whole thing, just forced to be localhost now.

go/cmd/gitter/gitter.go
```diff
@@ -0,0 +1,205 @@
...
+		}
+
+		logger.Info("Archiving git blob", slog.String("url", url))
+		// Archive
+		// tar --zstd -cf <archivePath> -C <gitStorePath> <repoDirName>
+		// using -C to archive the relative path so it unzips nicely
+		cmd := exec.Command("tar", "--zstd", "-cf", archivePath, "-C", path.Join(gitStorePath, repoDirName), ".")
+		if out, err := cmd.CombinedOutput(); err != nil {
+			return nil, fmt.Errorf("tar zstd failed: %w, output: %s", err, out)
+		}
```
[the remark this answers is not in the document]
another-rex: That's the point, as documented in the comments.

osv/repos.py
another-rex: Done

PR #4510, opened 2025-12-17
feat: redirect human Go vanity URL requests to pkg.go.dev

spencerschrock:
This is just a simple QoL change to redirect to the Go module documentation for humans visiting osv.dev/bindings/go.

Other libraries with vanity URLs do this too, like http://go.uber.org/zap


another-rex:
Tested locally, seems to work

PR #4547, opened 2025-12-23
fix: skip uploading unchanged files in exporter, compare CRC32C hash before upload

### LINKED ISSUE #3513: Exporter should check the crc hash before uploading
**Describe the bug**
Currently, the exporter uploads a file no matter if the file has changed or not. 

**Expected behaviour**
We want to only upload if the file has a different file hash.

**Additional context**
Recently with #3491 the all.zips are now reproducable - hashes remain the same across runs. This means we can check the hash and avoid uploading if nothing has changed. This also prevents a new generation from being added to bucket history, making it easier for users to avoid downloading the same files.

Harshit28j:
Fixes #3513 

## Summary
Skip uploading files when content hasn't changed by comparing CRC32C hashes.

## Changes
- Added [_calculate_crc32c()](cci:1://file:///d:/OSS/osv.dev/gcp/workers/exporter/exporter.py:216:0-222:75) helper function
- Modified [upload_single()](cci:1://file:///d:/OSS/osv.dev/gcp/workers/exporter/exporter.py:225:0-243:48) to check existing blob hash before uploading

## How it works
1. Get existing blob from GCS
2. Calculate local file CRC32C hash
3. Compare with remote hash
4. Skip upload if hashes match
image

### DIFF
--- a/gcp/workers/worker/poetry.lock
+++ b/gcp/workers/worker/poetry.lock
@@ -2013,4 +2013,4 @@ type = ["pytest-mypy"]
 [metadata]
 lock-version = "2.1"
 python-versions = ">=3.13,<3.14"
-content-hash = "d415ffbaaa97d13ba0e438119adbe5de03219d01874ca4ec50ad7824a4f738be"
+content-hash = "f8c9f6a79eaf74d56909f471f5ed008434578049c6eef16257214e87082616e9"
--- a/gcp/workers/worker/pyproject.toml
+++ b/gcp/workers/worker/pyproject.toml
@@ -7,6 +7,7 @@ dependencies = [
     "google-cloud-pubsub==2.34.0",
     "google-cloud-ndb==2.4.0",
     "google-cloud-storage==2.19.0",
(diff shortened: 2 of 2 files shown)


Harshit28j:
@another-rex Would really appreciate a review on this PR when you get a chance! 🙂

gcp/workers/exporter/exporter.py
```diff
@@ -22,6 +23,8 @@
 import zipfile
 from typing import List
 
+import google_crc32c
```
another-rex, line 26: Is this already part of the dependencies? I think this should be added to pyproject.toml as a dependency.

another-rex (approved):
LGTM! Though we have mostly moved to a go version of the exporter, though this looks fine to merge in as well.

gcp/workers/exporter/exporter.py
Harshit28j: `google-crc32c` is currently a transitive dependency. I'll add it to `pyproject.toml` and push an update.

PR #4550, opened 2025-12-24
fix(vulnfeeds): add more phrases for extracting versions

### LINKED ISSUE #249: vulnfeeds: Add some more common phrases when extracting versions from CVE description
Current logic is [here](https://github.com/google/osv/blob/9d9bed5f86b223781ac050800bf28796f3d5c903/vulnfeeds/cves/versions.go#L109)

Here are a few I've seen from examining some of my recent manual triages:

- x.x.x up to and including x.x.x
- x.x.x and earlier
- x.x.x or older
- x.x.x and below
- up to and including x.x.x
- prior to x.x.x
- below x.x.x

Harshit28j:
## Overview

Fixes #249

This PR extends the version extraction regex in [ExtractVersionsFromText](cci:1://file:///d:/OSS/osv.dev/vulnfeeds/cves/versions.go:609:0-735:1) to support additional common phrases found in CVE descriptions.

### Problem
The existing regex only matched `through` and `before` keywords, causing version information to be lost when CVE descriptions used other common phrasing.

### Solution
Added support for the following phrases:
- `up to and including` (e.g., "versions 1.0 up to and including 2.3.4")
- `prior to` (e.g., "versions prior to 2.3.4")
- `below` (e.g., "versions below 2.3.4")
- `and earlier` (e.g., "version 2.3.4 and earlier")
- `or older` (e.g., "version 2.3.4 or older")
- `and below` (e.g., "version 2.3.4 and below")

The implementation uses two regex patterns:
1. **Prefix pattern**: For phrases where version comes after the keyword (before, prior to, below, etc.)
2. **Trailing pattern**: For phrases where version comes before the keyword (and earlier, or older, and below)

## Testing

- [x] Existing tests pass (`go test -run TestExtractVersionInfo`)
- [x] Code formatted (`go fmt ./...`)
- [x] Code vetted (`go vet ./...`)

image

### DIFF
--- a/vulnfeeds/cves/versions.go
+++ b/vulnfeeds/cves/versions.go
@@ -608,34 +608,59 @@ func processExtractedVersion(version string) string {
 }
 
 func ExtractVersionsFromText(validVersions []string, text string) ([]models.AffectedVersion, []string) {
-	// Match:
+	// Match patterns where version comes AFTER keyword:
 	//  - x.x.x before x.x.x
 	//  - x.x.x through x.x.x
 	//  - through x.x.x
 	//  - before x.x.x
-	pattern := regexp.MustCompile(`(?i)([\w.+\-]+)?\s+(through|before)\s+(?:version\s+)?([\w.+\-]+)`)
-	matches := pattern.FindAllStringSubmatch(text, -1)
-	if matches == nil {
+	//  - x.x.x up to and including x.x.x
+	//  - prior to x.x.x
+	//  - below x.x.x
+	prefixPattern := regexp.MustCompile(`(?i)([\w.+\-]+)?\s+(through|before|up to and including|prior to|below)\s+(?:version\s+)?([\w.+\-]+)`)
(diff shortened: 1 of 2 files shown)


another-rex:
Please add real world examples of these different version formats, and add tests for them. (i.e. link to the CVEs, add them as test cases.)

another-rex:
Thanks, mostly LGTM, though I'll let @jess-lowe take a look as well. 

1 concern is most of these versions ranges seem to be from closed source CVEs, which we are not that interested in converting, so there's limited value there, but still good to have the functionality.

jess-lowe:
Hey @Harshit28j, sorry for taking so long to review! I merged the changes locally with the changes I've been making for CVE conversion and have found at least one example in which it's picked up another version from the text description, so that is awesome!

If it's possible, would you be able to update it to current level of changes / fix conflicts, and then I'm happy to merge :)

PR #4553, opened 2025-12-27
fix(docs): incorrect api path in faq

Denperidge:
Noticed the FAQ page (https://google.github.io/osv.dev/faq/) links to an non-existing path (https://google.github.io/osv.dev/faq/api/). This commit should fix that!

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -45,7 +45,7 @@ See our blog posts for more details:
 
 The OSV schema and OSV.dev can be used by:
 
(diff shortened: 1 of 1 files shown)


docs/faq.md
ashmod:
```suggestion
1. Open source consumers: By querying [OSV.dev's API](/api/) and using our tooling to find known vulnerabilities in their dependencies.
```
looks like it's also outdated here

Denperidge: Great catch, thank you!

another-rex (approved):
Thanks!

PR #4557, opened 2025-12-29
ci: Add snapshot tests for changes in the API

another-rex:
Also fixes apitester to test against test instance rather than prod by correctly setting req.Host.

### DIFF
--- a/tools/apitester/__snapshots__/cassette_TestCommand.snap
+++ b/tools/apitester/__snapshots__/cassette_TestCommand.snap
@@ -61,35 +61,35 @@
       "vulns": [
         {
           "id": "GO-2024-2598",
-          "modified": "<RFC3339 date with the year 2024>"
+          "modified": "<RFC3339 date with the year 2026>"
         },
         {
           "id": "GO-2024-2599",
-          "modified": "<RFC3339 date with the year 2024>"
+          "modified": "<RFC3339 date with the year 2026>"
         },
         {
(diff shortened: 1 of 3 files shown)


gcp/api/cloudbuild.yaml
```diff
@@ -43,4 +43,17 @@ steps:
     - CLOUDBUILD=1
   waitFor: ['init', 'sync']
 
+- name: 'gcr.io/oss-vdb/ci'
+  id: 'api-snapshot-tests'
+  dir: gcp/api
+  args: ['bash', '-ex', 'run_tests_e2e.sh', '/workspace/dummy.json']
+  env:
+    - CLOUDBUILD=1
```
michaelkedar: with #4616 this is now in the global options

gcp/api/run_tests_e2e.sh
```diff
@@ -0,0 +1,29 @@
...
+fi
+
+export GOOGLE_CLOUD_PROJECT=oss-vdb-test OSV_VULNERABILITIES_BUCKET=osv-test-vulnerabilities
+
+# Try to start docker if not running (mostly for CI)
+service docker start || true
+
+set -e
+
+poetry install
```
michaelkedar: Can you copy the checks added in #4616 here

michaelkedar (approved):

jess-lowe (approved):
LGTM

PR #4558, opened 2025-12-29
feat(source): Add Root source test

chait-slim:
Should be merged after: https://github.com/google/osv.dev/pull/4397

PR #4559, opened 2025-12-29
feat(source): Add Root source prod

chait-slim:
Should be merged after: https://github.com/google/osv.dev/pull/4558


chait-slim:
@michaelkedar @another-rex This is the last PR for adding Root :D
This would really help us if its get merged!

PR #4561, opened 2026-01-01
fix(frontend): navbar search breaking after Turbo navigation

ashmod:
There's a search-bar bug on the website where Turbo’s back/forward cache doesn’t keep the search button’s event wiring, so after returning to a cached page the search icon becomes unresponsive. This can be reproduced by:

  1. Visit the blog page
  2. Go to any blog post, then press the browser's back button
  3. Try clicking the search icon - it won't work (since the blog page is cached by Turbo)

This change ensures the search wiring is torn down before caching and cleanly re‑applied on restore, so the search stays functional after Turbo navigation.


another-rex (approved):
LGTM!

Issue #4581, opened 2026-01-09
Add the OCaml/OPAM advisory database


hannesm:
- [x] Prepare your data \- refer to the [OSV Schema](https://ossf.github.io/osv-schema/) documentation for information on how to properly format the data so it can be accepted.  
        
- [x] Create a PR to [reserve an ID prefix and define a new ecosystem](https://ossf.github.io/osv-schema/#id-modified-fields) ([example](https://github.com/ossf/osv-schema/pull/219)). We review the records you start publishing for OSV Schema [correctness](https://github.com/ossf/osv-schema/tree/main/validation) and [quality](https://google.github.io/osv.dev/data_quality.html) as part of reviewing and merging this PR.

   - done and merged at https://github.com/ossf/osv-schema/pull/473

- [x] Prepare and publish your records via a Git repository ([example](https://github.com/AlmaLinux/osv-database/tree/master)). If this method isn’t ideal, we also support publishing records from [REST API endpoints](https://google.github.io/osv.dev/data/new/rest-api) or through a GCS bucket([example](https://storage.googleapis.com/android-osv/)).
        
   - The datasource is located at https://github.com/ocaml/security-advisories. The branch "generated-osv" is automatically enhanced whenever a new advisory is merged to the main branch. That branch consists of one directory per year which contains the advisories, each in a separate file (OSEC-YEAR-NN).

- [x] To support API querying, please create a PR to extend [purl\_helpers.py](https://github.com/google/osv.dev/blob/master/osv/purl_helpers.py) and create a new ecosystem in [\_ecosystems.py](https://github.com/google/osv.dev/blob/master/osv/ecosystems/_ecosystems.py). You can refer to existing examples showing how to implement support for [Semver](https://github.com/google/osv.dev/blob/139de7b69a2ea39e2113309b3a0a47aab920ddcf/osv/ecosystems/_ecosystems.py#L45) and [non-Semver](https://github.com/google/osv.dev/pull/3430) ecosystems.  
   - See #4582 
        
- [x] Create a PR to start [importing the records you are publishing into our test instance of OSV.dev](https://github.com/google/osv.dev/blob/master/source_test.yaml) and validate everything is working as intended there.
   - See #4632

- [x] Create a PR to start [importing the records you are publishing into our production environment](https://github.com/google/osv.dev/blob/master/source.yaml)
   - See #4651

hannesm:
All done now, it is live \o/

PR #4582, opened 2026-01-09
Add OPAM to osv.dev

hannesm:
For https://github.com/google/osv.dev/issues/4581

Please note that I'm not an experienced Python programmer, happy to get feedback on this.

### DIFF
--- a/tools/apitester/__snapshots__/cassette_single_query.snap
+++ b/tools/apitester/__snapshots__/cassette_single_query.snap
@@ -340,7 +340,7 @@
       "id": "OSV-2020-484",
       "summary": "Heap-buffer-overflow in AAT::KerxSubTableFormat4\u003cAAT::KerxSubTableHeader\u003e::driver_context_t::transition",
       "details": "OSS-Fuzz report: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=12532\n\n```\nCrash type: Heap-buffer-overflow READ 4\nCrash state:\nAAT::KerxSubTableFormat4\u003cAAT::KerxSubTableHeader\u003e::driver_context_t::transition\nvoid AAT::StateTableDriver\u003cAAT::ExtendedTypes, AAT::KerxSubTableFormat4\u003cAAT::Ker\nAAT::KerxSubTableFormat4\u003cAAT::KerxSubTableHeader\u003e::apply\n```\n",
-      "modified": "<RFC3339 date with the year 2022>",
+      "modified": "<RFC3339 date with the year 2026>",
       "published": "2020-07-01T00:00:12.297418Z",
       "references": [
         {


AlexandreEXFO:
@hannesm : The list of data sources in the documentation could also be updated to add this new source. https://github.com/google/osv.dev/blob/master/docs/data.md

This updates this page: https://google.github.io/osv.dev/data/

hannesm:
> @hannesm : The list of data sources in the documentation could also be updated to add this new source. https://github.com/google/osv.dev/blob/master/docs/data.md
> 
> This updates this page: https://google.github.io/osv.dev/data/

Thanks for your comment. FYI, this is not mentioned in the issue template for a new source (see https://github.com/google/osv.dev/issues/4581).

hannesm:
Anything I can do to move this forward? It looks like CI systems await approval (though I can't see the oss-vdb worker output).

AlexandreEXFO:
@cuixq @another-rex

osv/ecosystems/opam.py
```diff
@@ -0,0 +1,49 @@
...
+                           introduced,
+                           fixed=None,
+                           last_affected=None,
+                           limits=None):
+        """Enumerate versions."""
+        response = requests.get(
+            self._REPO + package, timeout=config.timeout)
+        archive_response = requests.get(
+            self._REPO_ARCHIVE + package, timeout=config.timeout)
+        if response.status_code == 404 && archive_response.status_code == 404:
```
another-rex: This is not valid python syntax, need to replace && with and.

another-rex: I believe this is what's causing all the current test failures, everything else looks OK to me.

osv/ecosystems/opam.py
```diff
@@ -0,0 +1,49 @@
...
+    _REPO_ARCHIVE = 'https://api.github.com/repos/ocaml/opam-repository-archive/contents/packages/'
+
+    def enumerate_versions(self,
+                           package,
+                           introduced,
+                           fixed=None,
+                           last_affected=None,
+                           limits=None):
+        """Enumerate versions."""
+        response = requests.get(
```
another-rex: If the initial request 404s, let's avoid making a second request to the archive, and just exit early.

osv/ecosystems/opam.py
hannesm: thanks, fixed in d187e37c

osv/ecosystems/opam.py
hannesm: well, there may be the case that a package gets a security advisory that will then be archived, and no fixed versions will stay live. not sure how useful this is, but I would keep the code as is.

another-rex: Ah I misread line 34, that makes sense.

osv/ecosystems/opam.py
```diff
@@ -0,0 +1,49 @@
+"""OPAM ecosystem helper."""
+import requests
+import json
+
+from ..third_party.univers.debian import Version as DebianVersion
+from . import config
```
another-rex, line 4:
```suggestion
from . import config
from .ecosystems_base import EnumerableEcosystem, EnumerateError
```

another-rex:
LGTM, though please fix the linting issues.

Use `make lint` to run the linter locally.

`yapf -i <file>` to format the file.

hannesm:
@another-rex thanks for your review and insight. I pushed a commit with the formatted file.

A `make lint` didn't work for me - I'm using a FreeBSD laptop, and was afraid of the amount of dependencies. Hope that autoformat is sufficient.

EDIT: Sorry I messed up the git state and did a rebase onto master and a force-push.

osv/ecosystems/opam.py
```diff
@@ -0,0 +1,51 @@
...
+  """OPAM packages ecosystem"""
+
+  def _sort_key(self, version):
+    # OPAM uses debian versioning
+    if not DebianVersion.is_valid(version):
+      # If debian version is not valid, it is most likely an invalid fixed
+      # version then sort it to the last/largest element
+      return DebianVersion(9999999999, '9999999999')
+    return DebianVersion.from_string(version)
+
```
michaelkedar, line 11: Since OPAM uses debian versioning, you can inherit the `_sort_key` implementation from the generic [`DPKG`](https://github.com/google/osv.dev/blob/e780939fee072f082587164609830ec1eee0bc0b/osv/ecosystems/debian.py#L29) implementation:
```suggestion
from .ecosystems_base import EnumerableEcosystem, EnumerateError
from .debian import DPKG


class Opam(EnumerableEcosystem, DPKG):
  """OPAM packages ecosystem"""

```

osv/ecosystems/opam.py
```diff
@@ -0,0 +1,51 @@
+"""OPAM ecosystem helper."""
+import requests
+
+from ..third_party.univers.debian import Version as DebianVersion
+from . import config
+from .ecosystems_base import EnumerableEcosystem, EnumerateError
+
+
+class Opam(EnumerableEcosystem):
```
michaelkedar: nit: Maybe calling this class all-caps `OPAM` would be more consistent stylistically with our other ecosystems (e.g. `APK`, `CRAN`, `RPM`).

On the other hand, the ecosystem in OSV is `opam`, not `OPAM`.

osv/ecosystems/opam.py
```diff
@@ -0,0 +1,51 @@
...
+          f'Failed to get OPAM versions for {package} with: {response.text}')
+
+    responses = {}
+
+    if response.status_code == 200:
+      responses.extend(response.json())
+    if archive_response.status_code == 200:
+      responses.extend(archive_response.json())
+
+    versions = [x["name"] for x in responses]
```
michaelkedar: looking at [the response for `opam-repository`](https://api.github.com/repos/ocaml/opam-repository/contents/packages/opam-repository), the `name` field is in the form `opam-repository.2.0.0`, where I presume the version is supposed to be `2.0.0` (judging by the [OSV record](https://raw.githubusercontent.com/ocaml/security-advisories/refs/heads/generated-osv/2023/OSEC-2023-01.json))

Maybe this should be `x['name'].removeprefix(package + '.')`?
(Are package names case-sensitive, or are they always lower-case?)

another-rex:
> @another-rex thanks for your review and insight. I pushed a commit with the formatted file.
> 
> A `make lint` didn't work for me - I'm using a FreeBSD laptop, and was afraid of the amount of dependencies. Hope that autoformat is sufficient.
> 
> EDIT: Sorry I messed up the git state and did a rebase onto master and a force-push.

No problem, we can do a final format pass on our end to pass the check.

osv/ecosystems/opam.py
hannesm: Indeed, the ecosystem is named `opam`, and as far as I can see, it is common to have the first letter in Python classes uppercase!?

osv/ecosystems/opam.py
hannesm: the removeprefix is done in bacbfa75

>  (Are package names case-sensitive, or are they always lower-case?)

Neither. They are case-sensitive. There are mixed case package names as well in the package repository (such as unionFind wxOCaml CamelCase). The opam-repository git repository is manually maintained (i.e. pull requests are not automatically merged), and care is taken to not have package names that only differ in casing.

osv/ecosystems/opam.py
hannesm: thanks, applied.

hannesm:
Thanks for your review @michaelkedar and @another-rex. Please let me know if there's anything else that needs to be done to get this merged.

osv/ecosystems/opam.py
michaelkedar: Yeah, I was thinking between `Opam` and `OPAM`.
Leaving it as `Opam` is fine.

hannesm:
🎉 CI has a green checkmark now. Is there anything I can do to move this forward?

PR #4590, opened 2026-01-13
feat: implement coarse versioning for future matching against

michaelkedar:
Implemented a string-comparable encoding of ecosystem versions to allow us to do filtering on the database queries.
This should help reduce the number of entities needed to fetch and compare in many ecosystems.

I've also slightly changed the behaviour of `_sort_key`, which previously was returning a 'maximal' value for invalid versions. Now it should raise a `ValueError`, and the `sort_key` wrapper (which was handling `0` values) converts those into maximal version objects.

I've added `hypothesis` to do some fuzzing tests to make sure the logic retains the ordering rules of the ecosystems. (Please let me know if after this is merged you run into more failure cases so I can fix them). Fuzzing also helped identify some potential uncaught errors in our version parsing (particularly, using `isdigit` instead of `isdecimal`), which I've fixed where applicable.

I need to write a script to populate the existing AffectedVersions entities with the newly generated values, then work on making the API use the coarse versions for querying.

We have some PRs open on adding new ecosystems - they don't necessarily need the coarse version method, but I'm happy to merge those first then add the code for this later.

### DIFF
--- a/osv/models.py
+++ b/osv/models.py
@@ -1249,8 +1249,7 @@ def _affected_versions_from_package(affected: AffectedPackage,
       # If we have an ecosystem helper sort the events to help with querying.
       events.sort(key=lambda e, sort_key=e_helper.sort_key:
                   (sort_key(e.value), _EVENT_ORDER.get(e.type, -1)))
-      coarse_min, coarse_max = _get_coarse_min_max(events, e_helper,
-                                                   db_id)
+      coarse_min, coarse_max = _get_coarse_min_max(events, e_helper, db_id)
 
     # If we don't have an ecosystem helper, assume the events are in order.
     for e in all_pkg_ecosystems:
@@ -1324,7 +1323,8 @@ def affected_from_bug(entity: Bug) -> list[AffectedVersions]:
   """Compute the AffectedVersions from a Bug entity."""
   affected_versions = []
   for affected in entity.affected_packages:
-    affected_versions.extend(_affected_versions_from_package(affected, entity.db_id))
+    affected_versions.extend(
+        _affected_versions_from_package(affected, entity.db_id))
 
   # Deduplicate and sort the affected_versions
   unique_affected_dict = {av.sort_key(): av for av in affected_versions}


osv/ecosystems/coarse_version_monotonicity_test.py
```diff
@@ -0,0 +1,169 @@
...
+# Uses standard packaging.version pattern.
+pypi_strategy = st.one_of(
+    st.text(),  # legacy version can be any string
+    st.from_regex(
+        re.compile(r'^' + packaging.version.VERSION_PATTERN + r'$',
+                   re.IGNORECASE | re.VERBOSE | re.ASCII)))
+
+# Matches RPM versions: optional epoch, alternating alphanumeric segments.
+rpm_version_strategy = st.from_regex(
+    re.compile(r'^([0-9]+:)?(([0-9]+|[A-Za-z]+)((?![0-9A-Za-z])[ -~])*)+$',
```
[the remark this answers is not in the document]
michaelkedar: `hypothesis` uses the regex to generate patterns. It doesn't match with them afaik.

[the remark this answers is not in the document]
michaelkedar: `hypothesis` uses the regex to generate patterns. It doesn't match with them afaik.

osv/ecosystems/debian.py
```diff
@@ -31,11 +37,58 @@ class DPKG(OrderedEcosystem):
...
+      int_parts.append(int(couple[0]))
+      if len(couple) == 1:
+        break
+      sep = couple[1]
+      if sep == '.':
+        continue
+      if sep[0] == '.':
+        # Bigger than the max int, so we overflow
+        int_parts.append(9999999999)
+      break
```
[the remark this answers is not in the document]
michaelkedar: > For an input like `'10.20.a30'`, `re.findall(r'^$|\d+|\D+', v)` will produce `['10', '.', '20', '.', 'a', '30']`.

Um, no:
```python
>>> re.findall(r'^$|\d+|\D+', '10.20.a30')
['10', '.', '20', '.a', '30']
```
which would find its way to the `int_parts.append(9999999999)` line

osv/ecosystems/alpine.py
```diff
@@ -33,11 +37,30 @@ class APK(OrderedEcosystem):
...
+    """Coarse version.
+
+    Treats version as dot-separated integers.
+    Trims suffixes (_rc, _p, -r) to ensure monotonicity (e.g. 1.2_rc1 < 1.2).
+    """
+    if not AlpineLinuxVersion.is_valid(version):
+      raise ValueError(f'Invalid version: {version}')
+    # is_valid uses a $ regex anchor (which can match a newline),
+    # so we need to remove the newline if one exists.
+    if version[-1] == '\n':
```
another-rex, line 53: does $ also match `\r\n`? Might be worth it to call strip(), since I don't believe any versioning scheme have white spaces as part of the spec.

osv/ecosystems/ecosystems_base.py
```diff
@@ -210,3 +267,108 @@ def _deps_dev_enumerate(self,
...
+    if overflow or len(components) == 3:
+      break
+
+  # Pad with zeros to ensure 3 components
+  # If we overflowed, we should pad with MAX instead of 0
+  pad_value = 99999999 if overflow else 0
+  while len(components) < 3:
+    components.append(pad_value)
+
+  return f'00:{components[0]:08d}.{components[1]:08d}.{components[2]:08d}'
```
another-rex: nit, let's make this construction a function that just takes in 3 int args. then we can replace 374, 337, and 300 with this func.

osv/ecosystems/ecosystems_base.py
```diff
@@ -210,3 +267,108 @@ def _deps_dev_enumerate(self,
...
+  Format: 00:00000000.00000000.00000000
+  (Epoch:Major.Minor.Patch)
+  
+  The Epoch is always 00.
+  Only the first 3 integer components (Major, Minor, Patch) are used.
+  
+  Args:
+    version: The version string to convert.
+    separators_regex: Regex for separators (default: r'[.]').
+    trim_regex: Regex for characters to trim after (default: r'[-+]').
```
another-rex: rename to trim_suffix_regex?

osv/ecosystems/ecosystems_base.py
```diff
@@ -210,3 +267,108 @@ def _deps_dev_enumerate(self,
...
+    parts = [p for p in parts if p]
+
+  # Extract up to 3 integer components
+  components = []
+  overflow = False
+  for p in parts[:3]:
+    if not p.isdecimal():
+      break
+    val = int(p)
+    if val > 99999999:
```
another-rex: Create a const for MAX

osv/ecosystems/alpine.py
```diff
@@ -33,11 +37,30 @@ class APK(OrderedEcosystem):
...
+      version = version[:-1]
+    return coarse_version_generic(
+        version,
+        separators_regex=r'[.]',
+        # in APK, 1.02.1 < 1.1.1, so we must treat everything after .0x as 0
+        # also split off the _rc, _p, or -r suffixes
+        trim_regex=r'(?:\.0|[_-])',
```
another-rex: does this also trim 1.0.1?

osv/ecosystems/alpine_test.py
```diff
@@ -85,6 +85,24 @@ def test_apk(self):
...
+                     ecosystem.coarse_version('4.5_alpha'))
+    self.assertEqual('00:20200712.00000000.00000000',
+                     ecosystem.coarse_version('20200712-r0'))
+    self.assertEqual('00:00000011.00000003.00000020',
+                     ecosystem.coarse_version('11.3.20.1_p1-r0'))
+    self.assertEqual('00:00000002.00000003.00000000',
+                     ecosystem.coarse_version('02.3'))
+    self.assertEqual('00:00000005.00000000.00000000',
+                     ecosystem.coarse_version('5.06.7'))
+    self.assertEqual('00:00000005.00000000.00000000',
```
another-rex, line 103: Seems like it does trim 5.0.9, should this be happening?

osv/ecosystems/ecosystems_base.py
```diff
@@ -210,3 +267,108 @@ def _deps_dev_enumerate(self,
...
+                           trim_regex=r'[-+]',
+                           implicit_split=False,
+                           empty_as: str | None = None) -> str:
+  """
+  Convert a version string into a coarse, lexicographically comparable string.
+  
+  Format: 00:00000000.00000000.00000000
+  (Epoch:Major.Minor.Patch)
+  
+  The Epoch is always 00.
```
another-rex: Clarify that this is the case only for this function, when I read this section it sounds like it's describing all coarse versions, making it confusing why epoch is there in the first place.

osv/ecosystems/alpine.py
another-rex: Might need `\.0\d`

osv/ecosystems/pypi.py
```diff
@@ -30,6 +34,36 @@ def _sort_key(self, version):
...
+  def coarse_version(self, version: str):
+    """Coarse version.
+
+    Treats version as dot-separated integers.
+    Maps legacy versions to 0 (sort before valid versions).
+    Epochs are preserved.
+    """
+    # legacy versions are less than non-legacy versions, thus mapped to 0
+    ver = packaging_legacy.version.parse(version)
+    if isinstance(ver, packaging_legacy.version.LegacyVersion):
```
another-rex, line 47: Hmm how common are these, might it be worth it to bump all legacy version up by 1 so we can still have legacy versions? (probably not worth it unless it's really common)

osv/third_party/univers/alpine.py
```diff
@@ -20,28 +20,29 @@ class InvalidVersion(ValueError):
...
+  #   return False
+
+  # s = search.group(1)
+  # left, _, _ = s.partition(".")
+  # # handle the suffix case
+  # left, _, _ = left.partition("-")
+  # if not left.isdecimal():
+  #   return True
+  # i = int(left)
+  # return str(i) == left
```
another-rex, line 45: Should this still be here?

osv/models.py
```diff
@@ -40,6 +40,9 @@
 
 _MAX_GIT_VERSIONS_TO_INDEX = 5000
 
+MIN_COARSE_VERSION = '00:00000000.00000000.00000000'
```
another-rex: this can use the helper function suggested in ecosystems.

osv/models.py
```diff
@@ -1165,92 +1175,149 @@ def transaction():
     pubsub.publish_failure(data, type='gcs_retry')
 
 
-def affected_from_bug(entity: Bug) -> list[AffectedVersions]:
-  """Compute the AffectedVersions from a Bug entity."""
+def _get_coarse_min_max(events, e_helper, db_id):
```
another-rex: nit: add types

osv/models.py
```diff
@@ -1165,92 +1175,149 @@ def transaction():
...
+              ecosystem=e,
+              name=pkg_name,
+              coarse_min=coarse_min,
+              coarse_max=coarse_max,
+              events=events,
+          ))
+
+  # Add the enumerated versions
+  # We need at least a package name to perform matching.
+  if pkg_name and affected.versions:
```
another-rex, line 1268: Should we be logging something if there is non pkg_name but got ranges or affected.versions?

osv/ecosystems/alpine.py
michaelkedar: >does $ also match `\r\n`?

No, at least not on Linux.
I would rather not call strip, just in case

osv/ecosystems/alpine.py
michaelkedar: It does trim `1.0.1`, but that's intentional because `1.0.2` < `1.01.1` < `1.1.0`
Will mention that in the comment.

osv/ecosystems/alpine_test.py
michaelkedar: yeah - 5.0.9 < 5.01.0 so everything after the first 0 must be truncated to 0

osv/ecosystems/ecosystems_base.py
michaelkedar: 'suffix' is a bit misleading here, because it implies it'd match the whole end of the string rather than a point at which to truncate from.
I've renamed it to `truncate_regex`, which might be clearer?

osv/ecosystems/ecosystems_base.py
michaelkedar: I've made `coarse_version_from_ints` be the generic construction function for this, and consolidated some of the logic into it.

osv/ecosystems/ecosystems_base.py
michaelkedar: Made epoch an argument to this function, to make setting it less awkward.

osv/ecosystems/pypi.py
michaelkedar: I don't think they're that common.
Plus, legacy versions can be arbitrary strings and I don't want to work out the comparison rules for them.

osv/third_party/univers/alpine.py
michaelkedar: I kinda wanted to signpost the original third-party code that I've changed

osv/models.py
michaelkedar: typified

osv/models.py
michaelkedar: I'm pretty sure this would be expected for GIT ranges, so probably not.

osv/ecosystems/ecosystems_base.py
michaelkedar: const'd

PR #4614, opened 2026-01-14
fix(worker): update modified time if bug was withdrawn.

michaelkedar:
Not updating the modified time on the bug when we're withdrawing it was causing the withdrawn bug to not be updated properly.
The importer-deleter should end up cleaning up the current bugs in invalid state once this is merged.

### DIFF
--- a/gcp/workers/worker/worker.py
+++ b/gcp/workers/worker/worker.py
@@ -450,7 +450,7 @@ class TaskRunner:
 
     logging.info('Marking %s as invalid and withdrawn.', vuln_id)
     bug.status = osv.BugStatus.INVALID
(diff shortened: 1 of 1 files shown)


michaelkedar:
> I would have thought the falsey value for bug.last_modified would be smaller than bug.withdrawn. But I guess this is safer.

Nah, python's annoying w.r.t. this:
```python
>>> max(None, datetime.now())
  File "<python-input-17>", line 1, in <module>
    max(None, datetime.now())
    ~~~^^^^^^^^^^^^^^^^^^^^^^
TypeError: '>' not supported between instances of 'datetime.datetime' and 'NoneType'
```

PR #4615, opened 2026-01-14
chore: update poetry deps

michaelkedar:
dependabot/renovate seem to only update the `requirements-py3.12.txt` file and not this one (because it's alphabetically first?)
Should fix the vulns we're seeing


Ly-Joey (approved):

cuixq:
It is ignored by renovate: https://github.com/google/osv.dev/blob/master/renovate.json#L181

PR #4616, opened 2026-01-15
chore: try fix race conditions in Cloud Build tests

michaelkedar:
Don't try reinstall dependencies in the `run_tests.sh` files if we're running in Cloud Build (which does installs them in a dedicated step).
Also, replaced `poetry install` with `poetry sync`


another-rex (approved):

another-rex:
hmm I don't think this worked

michaelkedar:
> hmm I don't think this worked

:upside_down_face: should be working now

another-rex (approved):

PR #4620, opened 2026-01-15
fix(repos): use bare open method as fallback to checkout repos for analysis

jess-lowe:
Some records like CVE-2025-62408 are no longer importing correctly due to an unsupported extension edgecase in pygit2: `unsupported extension name extensions.refstorage`.

This issue meant that commits weren't being enumerated and neither were tags.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -142,6 +142,17 @@ gcloud auth login --update-adc
 make run-api-server
 ```
 
+#### API E2E Snapshots
+
+If you have made any changes to the API, please update the API query snapshots with
+
+```shell
+gcloud auth login --update-adc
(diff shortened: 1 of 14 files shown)


osv/repos.py
```diff
@@ -115,6 +115,14 @@ class RepoInaccessibleError(Exception):
   """Git repository cannot be cloned due to being deleted or requiring auth."""
 
 
+def checkout_repo(checkout_dir):
```
another-rex: nit: rename to open_repo, checkout has a specific meaning in git.

another-rex:
Also need to update branch to fix the API test I believe.

another-rex (approved):

PR #4631, opened 2026-01-20
refactor(vulnfeeds): move converters/mirrors into subdirectories for cleaner understanding

jess-lowe:
Subdirectories are now as follows:

(crossed out stuff has yet to be moved)
image


another-rex:
Can you explain where combine-to-osv, and pypi went?

jess-lowe:
> Can you explain where combine-to-osv, and pypi went?

I haven't moved these yet because pypi is depended on externally, though I could probably move combine-to-osv, but that can happen in a later PR

PR #4658, opened 2026-01-23
fix: Migrate importer & workers off of Bug entities

michaelkedar:
#3850
This change makes OSV finally stop writing new Bug entities, moving the source of truth to solely be the Vulnerability entity and proto in GCS (for non-oss-fuzz stuff).
- Changed worker to work with / write proto directly, instead of relying on the Bug's _post_put_hook
- Changed importer to use Vulnerability entity in place of Bug during import time
  - Have not made the importer proactively write un-enriched records yet
- Changed recoverer to handle missing GCS files by triggering a forced re-import in the worker, instead of trying to regenerate from the Bug entities. (and added worker logic to support this)
- Updated some datafix tools to use Vulnerability, and moved some Bug-dependent tools to and `old/` directory for reference.

Also added some missing type hints in some places.

### DIFF
--- a/tools/apitester/__snapshots__/cassette_TestCommand_MoreLockfiles.snap
+++ b/tools/apitester/__snapshots__/cassette_TestCommand_MoreLockfiles.snap
@@ -136,6 +136,10 @@
     {
       "vulns": [
         {
+          "id": "GHSA-7gcm-g887-7qv7",
+          "modified": "<RFC3339 date with the year 2026>"
+        },
+        {
           "id": "GHSA-8qvm-5x2c-j2w7",
           "modified": "<RFC3339 date with the year 2026>"
         }
--- a/tools/apitester/__snapshots__/cassette_single_query.snap
+++ b/tools/apitester/__snapshots__/cassette_single_query.snap
@@ -2187,6 +2187,266 @@
       ]
     },
     {
+      "id": "CVE-2025-10966",
+      "details": "curl's code for managing SSH connections when SFTP was done using the wolfSSH\npowered backend was flawed and missed host verification mechanisms.\n\nThis prevents curl from detecting MITM attackers and more.",
+      "aliases": ["CURL-CVE-2025-10966"],
+      "modified": "<RFC3339 date with the year 2026>",
+      "published": "2025-11-07T08:15:39.617Z",
+      "references": [
+        {
+          "type": "FIX",
+          "url": "https://curl.se/docs/CVE-2025-10966.html"
+        },
+        {
+          "type": "ADVISORY",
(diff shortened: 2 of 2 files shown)


gcp/workers/worker/worker.py
```diff
@@ -427,36 +425,72 @@ def _source_update(self, message):
...
+        # in case this was already withdrawn for some reason
+        proto_vuln.withdrawn.FromDatetime(datetime.datetime.now(datetime.UTC))
+      if (not proto_vuln.HasField('modified') or
+          proto_vuln.withdrawn.ToDatetime(
+              datetime.UTC) > proto_vuln.modified.ToDatetime(datetime.UTC)):
+        proto_vuln.modified.CopyFrom(proto_vuln.withdrawn)
+      ds_vuln.is_withdrawn = True
+      ds_vuln.modified = proto_vuln.modified.ToDatetime(datetime.UTC)
+      osv.models.put_entities(ds_vuln, proto_vuln)
+      osv.update_affected_commits(vuln_id, [], False)
```
another-rex: Should this be deleting all affected commits information?
It does not look like the existing implementation is deleting affected commits.

gcp/workers/worker/worker.py
```diff
@@ -491,12 +520,13 @@ def _analyze_vulnerability(self, source_repo, repo, vulnerability, path,
...
-    if not result.has_changes and not added_fix_info:
+    if not result.has_changes:
       return result
 
     if not source_repo.editable:
       return result
-
+    # NB: Only OSS-Fuzz is editable - all other sources are read-only.
+    # This should not be reachable by this worker.
     output_path = os.path.join(osv.repo_path(repo), path)
```
another-rex, line 530: Should this add a warning log if it gets to this point.

osv/models.py
```diff
@@ -1181,6 +1162,30 @@ def transaction():
...
+  to_delete = []
+  old_affected = AffectedVersions.query(
+      AffectedVersions.vuln_id == vuln_pb.id).fetch()
+  if ds_vuln.is_withdrawn:
+    to_delete.append(ndb.Key(ListedVulnerability, vuln_pb.id))
+    to_delete.extend(av.key for av in old_affected)
+  else:
+    to_put.append(ListedVulnerability.from_vulnerability(vuln_pb))
+    new_affected = affected_from_proto(vuln_pb)
+    added, removed = diff_affected_versions(old_affected, new_affected)
```
another-rex, line 1181: Ah I see, all affected versions are deleted when a  record is withdrawn.

gcp/workers/worker/worker.py
michaelkedar: Removed the deletion, to be consistent with how published withdrawals are treated.

PR #4704, opened 2026-01-29
feat: Add `related` computation to cron

michaelkedar:
Make `related` vulns be regularly recomputed, so when `A` updates to related to `B`, `B` will get updated to relate to `A`.

A withdrawn vuln will have non-withdrawn vulns added to its related field, but it will not add it's ID to the related fields of its own related vulns.
(i.e. if `A` is withdrawn and has `related: ["C"]`, `C` will not get `A` added to its related field. If non-withdrawn `B` has `related: ["A"]`, `A` will end up with `related: ["B", "C"]`)

### DIFF
--- a/go/cmd/relations/related.go
+++ b/go/cmd/relations/related.go
@@ -14,12 +14,20 @@ import (
 	"google.golang.org/api/iterator"
 )
 
+// computeRelated computes all related groups for the given vulns.
+// `groups` is a map of vuln IDs to their related IDs.
+// `withdrawnVulns` is a map of withdrawn vulns.
+// Returns a map of vuln IDs to their related IDs, with the inverse relation added.
+// `groups` is modified in place.
 func computeRelated(groups map[string][]string, withdrawnVulns map[string]struct{}) map[string][]string {
 	// Add the inverse relation of the groups to the map
 	for id, group := range groups {
-		// We want to prevent withdrawn vulns from being added to related groups.
-		// But if a non-withdrawn vuln has a withdrawn vuln as a related id, we want to add it.
 		if _, ok := withdrawnVulns[id]; ok {
+			// We want to prevent withdrawn vulns IDs from being added to related groups,
+			// if the withdrawn vuln itself references other non-withdrawn vulns.
+			// For example:
+			// - If A (withdrawn) relates to B (valid), B should NOT list A.
+			// - If A (valid) relates to B (withdrawn), B SHOULD list A.
 			continue
 		}
 		for _, related := range group {
@@ -69,6 +77,8 @@ func updateRelated(ctx context.Context, cl *datastore.Client, id string, related
 
 func ComputeRelatedGroups(ctx context.Context, cl *datastore.Client, ch chan<- Update) error {
 	// Query for all vulns that have related.
+	// It's easier to recompute all groups than to try and figure out which ones
+	// need to be recomputed.
 	logger.Info("Retrieving vulns for related computation...")
 	q := datastore.NewQuery("Vulnerability").FilterField("related_raw", ">", "")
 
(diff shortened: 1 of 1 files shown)


go/cmd/relations/related.go
```diff
@@ -0,0 +1,133 @@
...
+		g, ok := relatedGroups[id]
+		delete(relatedGroups, id)
+		if !ok || !slices.Equal(g.RelatedIDs, relatedIDs) {
+			if err := updateRelated(ctx, cl, id, relatedIDs, ch); err != nil {
+				return fmt.Errorf("failed to update related group: %w", err)
+			}
+		}
+	}
+
+	for id := range relatedGroups {
```
another-rex, line 138: Add some comments above this explaining why the rest of the relatedGroups are being deleted.

go/cmd/relations/related.go
```diff
@@ -0,0 +1,133 @@
...
+		Value:     relatedIDs,
+	}
+
+	return nil
+}
+
+func ComputeRelatedGroups(ctx context.Context, cl *datastore.Client, ch chan<- Update) error {
+	// Query for all vulns that have related.
+	logger.Info("Retrieving vulns for related computation...")
+	q := datastore.NewQuery("Vulnerability").FilterField("related_raw", ">", "")
```
another-rex, line 83: Add comments explaining the reasoning for querying all related every time.

go/cmd/relations/related.go
```diff
@@ -0,0 +1,133 @@
...
+	"slices"
+	"time"
+
+	"cloud.google.com/go/datastore"
+	"github.com/google/osv.dev/go/logger"
+	"github.com/google/osv.dev/go/osv/models"
+	"google.golang.org/api/iterator"
+)
+
+func computeRelated(groups map[string][]string, withdrawnVulns map[string]struct{}) map[string][]string {
```
another-rex, line 22: I think this needs to indicate that it modifies groups inplace.

go/cmd/relations/related.go
```diff
@@ -0,0 +1,133 @@
...
+	// Add the inverse relation of the groups to the map
+	for id, group := range groups {
+		// We want to prevent withdrawn vulns from being added to related groups.
+		// But if a non-withdrawn vuln has a withdrawn vuln as a related id, we want to add it.
```
another-rex: This could be worded a bit differently

We want to prevent withdrawn vulns IDs from being added to related groups, if the withdrawn vuln itself references other non-withdrawn vulns.

another-rex: Maybe give some examples here as well.

go/cmd/relations/related.go
```diff
@@ -0,0 +1,133 @@
...
+	if _, err := cl.Put(ctx, datastore.NameKey("RelatedGroup", id, nil), &group); err != nil {
+		return err
+	}
+	ch <- Update{
+		ID:        id,
+		Timestamp: group.Modified,
+		Field:     updateFieldRelated,
+		Value:     relatedIDs,
+	}
+
```
another-rex, line 74: It does seem a bit confusing to me why the datastore part is done here directly, but there's also an updater channel. Why is everything not done on the receiving end of the channel?

michaelkedar: The channel collates updates for individual vulns across the alias/upstream/related computation (so we only do one write the vulnerability record/gcs).

I think it had made more sense to have them separate for the alias computation in particular, which has a bit of a complicated update flow.

go/cmd/relations/related.go
michaelkedar: comment added

go/cmd/relations/related.go
michaelkedar: the reasoning is basically "It's easier this way", which I added

go/cmd/relations/related.go
michaelkedar: added a comment to this function

go/cmd/relations/related.go
michaelkedar: done

another-rex (approved):

PR #4714, opened 2026-01-29
fix: reduce network requests in nvd conversion

jess-lowe:
in cases like CVE-2016-1255 where it has many CPEs but they're all enumerated versions, conversion took 5mins 39sec. this key optimisation now take 5 sec.

Fixes some minor logs too

### DIFF
--- a/vulnfeeds/conversion/nvd/converter.go
+++ b/vulnfeeds/conversion/nvd/converter.go
@@ -290,7 +290,7 @@ func FindRepos(cve models.NVDCVE, vpRepoCache *cves.VPRepoCache, metrics *models
 		return nil
 	}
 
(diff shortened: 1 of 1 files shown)


vulnfeeds/conversion/nvd/converter.go
```diff
@@ -291,31 +290,34 @@ func FindRepos(cve models.NVDCVE, vpRepoCache *cves.VPRepoCache, metrics *models
 		return nil
 	}
 
+	VendorProductCombinations := make(map[cves.VendorProduct]bool)
```
another-rex: uncap pls

another-rex (approved):

PR #4723, opened 2026-02-01
fix: more nvd optimisations and badrepo cache

jess-lowe:
This change fixes some nits, introduces a bad repository cache and stops the conversion from discarding versions just because a commit was provided in the references. 

- Removed CVEID from error formatting as at time of logging, CVEID is already provided via slog.
- Removed warnings from some logs in favour of adding to metrics.Notes file for better investigation later. There may be value in keeping these at warning levels only if we also are prepared to handle investigating spikes, in which case logs should appear closer to what is causing the errors (network requests generally). 
- made errors lowercase as linter demands

### DIFF
--- a/vulnfeeds/cves/versions.go
+++ b/vulnfeeds/cves/versions.go
@@ -794,7 +794,7 @@ func ExtractVersionsFromCPEs(cve models.NVDCVE, validVersions []string, metrics 
 					Fixed:        fixed,
 					LastAffected: lastaffected,
 				}
-				
+
 				if seen[possibleNewAffectedVersion] {
 					continue
 				}


michaelkedar (approved):

vulnfeeds/cves/versions.go
```diff
@@ -800,21 +787,40 @@ func ExtractVersionInfo(cve models.NVDCVE, validVersions []string, httpClient *h
...
 
-				gotVersions = true
 				possibleNewAffectedVersion := models.AffectedVersion{
 					Introduced:   introduced,
 					Fixed:        fixed,
 					LastAffected: lastaffected,
 				}
-				if slices.Contains(v.AffectedVersions, possibleNewAffectedVersion) {
+
+				if slices.Contains(versions, possibleNewAffectedVersion) {
```
another-rex: Not sure how much time is taken up by this, but this is O(N^2) at least. Probably better to use a map and sort at the end, or do one final sort and compact at the end.

vulnfeeds/conversion/nvd/converter.go
```diff
@@ -48,13 +48,13 @@ func CVEToOSV(cve models.NVDCVE, repos []string, cache *git.RepoTagsCache, direc
 		// There are some AffectedVersions to try and resolve to AffectedCommits.
 		if len(repos) == 0 {
 			metrics.AddNote("No affected ranges for %q, and no repos to try and convert %+v to tags with", maybeProductName, versions.AffectedVersions)
-			return fmt.Errorf("[%s]: No affected ranges for %q, and no repos to try and convert %+v to tags with", cve.ID, maybeProductName, versions.AffectedVersions)
```
another-rex, line 51: Please add an explanation for these error formatting changes to the PR description

vulnfeeds/cves/versions.go
```diff
@@ -1011,12 +1020,13 @@ func GitVersionsToCommits(cveID models.CVEID, versions models.VersionInfo, repos
...
 			// Fixed commits, they're also assumed to be more precise than what may be derived from tag to commit mapping.
 			fixedEquivalentCommit := ""
-			if v.HasFixedCommits(repo) && av.Fixed != "" {
+			if v.HasFixedCommits(repo) && av.Fixed != "" && len(versions.AffectedVersions) == 1 {
+				fixedEquivalentCommit = v.FixedCommits(repo)[0]
 				metrics.AddNote("Using preassumed fixed commits instead of deriving from fixed version %s", av.Fixed)
 			} else if av.Fixed != "" {
 				ac, err := git.VersionToAffectedCommit(av.Fixed, repo, models.Fixed, normalizedTags)
 				if err != nil {
-					logger.Warn("Failed to get a Git commit for fixed version", slog.String("cve", string(cveID)), slog.String("version", av.Fixed), slog.String("repo", repo), slog.Any("err", err))
```
another-rex, line 1019: Add reasoning for this change to the PR description as well.

vulnfeeds/cves/versions.go
jess-lowe: I don't think we want to sort, because then we have versions/commits out of their original order, but changed to use a map.

PR #4728, opened 2026-02-03
feat(gitter): git commit graph and patch ID caching

Ly-Joey:
Building on top of #4508 & co.

Key changes:
- Added foundational structures to save commit details to disk
- We now parse full git history into in-memory graph for commit traversal
- Parallelised patch ID calculation. They are loaded from the cache, meaning subsequent requests to the same repo only need to compute patch IDs for new commits.

API updates:
- `POST /cache`: Triggers a clone / fetch and store the computed commit details
- `/getgit` renamed to `GET /git`
  - [x] change the endpoint in Python

WIP (To be added in follow-up PRs to keep this more manageable to review):
- [ ] `POST /affected-commits`: accepts a list of events and returns the list of affected commits
  - `/affected-commits` should accept a list of ranges as valid request as well
  - Probably refactor the check last fetch time and refetch logic, because every operation relies on it.
- [ ] Check if a repoistory exist (/git-good)
- [ ] Resolve refs to commit hash


go/cmd/gitter/pb/repository/repository.proto
```diff
@@ -0,0 +1,16 @@
...
+
+option go_package = "./repository";
+
+message CommitDetail {
+    bytes hash = 1;
+    bytes patch_id = 2;
+    repeated string tags = 3;
+}
+
+// RepositoryCache is the minimally saved details for a repository
```
another-rex: Probably add a comment here explaining that we don't save the graph, only non mutable information, and the graph is generated per request.

go/cmd/gitter/pb/repository/repository.proto
```diff
@@ -0,0 +1,16 @@
+syntax = "proto3";
+
+package gitter;
+
+option go_package = "./repository";
+
+message CommitDetail {
+    bytes hash = 1;
+    bytes patch_id = 2;
+    repeated string tags = 3;
```
another-rex: I'm not even sure we should save tags here, as tags can be mutable, so our cache might be storing invalid data.

go/cmd/gitter/gitter.go
```diff
@@ -119,22 +153,53 @@ func fetchBlob(ctx context.Context, url string, forceUpdate bool) ([]byte, error
...
+	_, err := os.Stat(path.Join(repoPath, ".git"))
+	if err != nil {
+		if os.IsNotExist(err) {
+			deleteLastFetch(url)
+		}
+
+		return fmt.Errorf("failed to read file: %w", err)
+	}
+
+	logger.Info("Fetch completed", slog.Duration("duration", time.Since(start)))
```
another-rex: log with the context of the url requested

go/cmd/gitter/gitter.go
```diff
@@ -119,22 +153,53 @@ func fetchBlob(ctx context.Context, url string, forceUpdate bool) ([]byte, error
...
+
+	return nil
+}
+
+func archiveRepo(ctx context.Context, url string) ([]byte, error) {
+	repoDirName := getRepoDirName(url)
```
another-rex, line 236: Make sure to do an extra filepath.Base() in getRepoDirName() to suppress this path controlled by user issue.

go/cmd/gitter/gitter.go
```diff
@@ -252,32 +327,45 @@ func gitHandler(w http.ResponseWriter, r *http.Request) {
...
 			return
 		}
 		http.Error(w, fmt.Sprintf("Error fetching blob: %v", err), http.StatusInternalServerError)
 
 		return
 	}
 
+	// Archive repo
+	//nolint:contextcheck // I can't change singleflight's interface
+	fileDataAny, err, _ := gArchive.Do(url, func() (any, error) {
```
another-rex, line 401: Hmm...  what if there's a git fetch happening at the same time as another archive request?

another-rex: I think we are missing a global lock per url.

go/cmd/gitter/gitter.go
```diff
@@ -103,10 +135,12 @@ func isAuthError(err error) bool {
 		(strings.Contains(strings.ToLower(errString), "repository") && strings.Contains(strings.ToLower(errString), "not found"))
 }
 
-func fetchBlob(ctx context.Context, url string, forceUpdate bool) ([]byte, error) {
+func fetchRepo(ctx context.Context, url string, forceUpdate bool) error {
```
another-rex: All these top level handlers should probably be public (like LoadRepo)

go/cmd/gitter/gitter.go
```diff
@@ -286,3 +374,165 @@ func gitHandler(w http.ResponseWriter, r *http.Request) {
...
+
+		return
+	}
+
+	repoDirName := getRepoDirName(url)
+	repoPath := path.Join(gitStorePath, repoDirName)
+
+	//nolint:contextcheck // I can't change singleflight's interface
+	_, err, _ = gLoad.Do(repoPath, func() (any, error) {
+		return LoadRepository(r.Context(), repoPath)
```
another-rex: hmm Cache handler and affected commits handler seems very similar for this bit.

go/cmd/gitter/gitter.go
```diff
@@ -286,3 +374,165 @@ func gitHandler(w http.ResponseWriter, r *http.Request) {
...
+		http.Error(w, fmt.Sprintf("Error fetching blob: %v", err), http.StatusInternalServerError)
+
+		return
+	}
+
+	repoDirName := getRepoDirName(url)
+	repoPath := path.Join(gitStorePath, repoDirName)
+
+	//nolint:contextcheck // I can't change singleflight's interface
+	_, err, _ = gLoad.Do(repoPath, func() (any, error) {
```
another-rex, line 468: We need a short term (1 hour?) in memory cache of the results, otherwise this is not really caching that much (iiuc this is only caching the patch ids to disk)

another-rex: Probably something to discuss in the docs first.

go/cmd/gitter/gitter.go
another-rex: Hmm you do have a Mutex on Repository, but I'm not sure it covers this.

go/cmd/gitter/gitter.go
another-rex, line 352: I have been thinking about the running out of memory issue we have been running into. 

There's a few options we have to deal with this. Ideally we want to slow down rather than hit the memory cap and cause unpredictable behavior. 

We can go with a complicated solution where we track our memory usage, but I think the most practical option is to just limit the number of unique repository requests we are handling at once. 

If it's small repos that don't use a lot of memory, they will be fast to process as well. If it's big repos that use a lot of memory, they will be slow to process, so this will slow them down and give us more memory.

More things for the design doc!

go/cmd/gitter/persistence.go
```diff
@@ -78,8 +80,42 @@
...
+	data, err := proto.Marshal(cache)
+	if err != nil {
+		return err
+	}
+
+	return os.WriteFile(cachePath, data, 0644)
+}
+
+func loadRepositoryCache(cachePath string) (*pb.RepositoryCache, error) {
+	data, err := os.ReadFile(cachePath)
```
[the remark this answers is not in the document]
Ly-Joey: User-provided value `url` went through path.Base() to strip the '/', then we append some hash to form the actual cache path. So I'm not sure if it should still be considered "uncontrolled" and manipulable.

go/cmd/gitter/repository.go
```diff
@@ -0,0 +1,548 @@
...
+	var cache *pb.RepositoryCache
+
+	// Load cache pb file of the repo if exist
+	if c, err := loadRepositoryCache(cachePath); err == nil {
+		cache = c
+		logger.Info("Loaded repository cache", slog.Int("commits", len(cache.GetCommits())))
+	} else {
+		if errors.Is(err, os.ErrNotExist) {
+			// It's fine if cache doesn't exist, log it just in case
+			logger.Info("No repository cache found")
```
another-rex: Have all logs take in context so we know what request they belong to. Might want to setup a helper function for this (it's already a helper function, so probably add the ability to take context to our osv logger)

go/cmd/gitter/repository.go
```diff
@@ -0,0 +1,548 @@
...
+	var newCommits []SHA1
+
+	// Temp outFile for git log output
+	tmpFile, err := os.CreateTemp(r.repoPath, "git-log.out")
+	if err != nil {
+		return nil, fmt.Errorf("failed to create temp file: %w", err)
+	}
+	defer os.Remove(tmpFile.Name())
+
+	// Run git log via bash because redirecting to file is faster than using pipe
```
another-rex: also mention that it's faster than using git's own file output

go/cmd/gitter/repository.go
```diff
@@ -0,0 +1,548 @@
...
+	defer file.Close()
+
+	reader := bufio.NewReaderSize(file, 1024*1024)
+	for {
+		line, err := reader.ReadString('\n')
+		if err != nil {
+			break
+		}
+
+		line = strings.TrimSuffix(line, "\n")
```
another-rex: Put some example lines here to make it clear what is happening with the text manipulation.

go/cmd/gitter/repository.go
```diff
@@ -0,0 +1,548 @@
...
+	// Save cache
+	if err := saveRepositoryCache(cachePath, repo); err != nil {
+		logger.Error("Failed to save repository cache", slog.Any("err", err))
+	}
+
+	return repo, nil
+}
+
+// buildCommitGraph builds the commit graph and associate commit details from scratch
+func (r *Repository) buildCommitGraph(ctx context.Context, cache *pb.RepositoryCache) ([]SHA1, error) {
```
another-rex, line 106: Explain what the return values is (I think it's new commits right?)

another-rex: And say that they are mostly in order if that is what's happening and you are happy to make that contract on this function.

go/cmd/gitter/repository.go
```diff
@@ -0,0 +1,548 @@
...
+		}
+	}
+
+	logger.Info("Commit graph completed", slog.Int("commits", len(r.commitDetails)), slog.Int("nodes", len(r.commitGraph)), slog.Int("new_commits", len(newCommits)), slog.Duration("duration", time.Since(start)))
+
+	return newCommits, nil
+}
+
+// calculatePatchIDs calculates patch IDs only for the specific commits provided.
+func (r *Repository) calculatePatchIDs(ctx context.Context, commits []SHA1) error {
```
another-rex, line 233: Comment that commits should be pass in in order if possible.

go/cmd/gitter/repository.go
```diff
@@ -0,0 +1,548 @@
...
+	logger.Info("Starting patch ID calculation", slog.String("repo", r.repoPath))
+	start := time.Now()
+
+	// Number of workers
+	workers := runtime.NumCPU()
```
another-rex: Explain that we are chunking to keep workers generating patch ids in commit order.

another-rex: Make workers num a variable at the top of the file. I feel like we might want to change this to be different from NumCPU.

go/cmd/gitter/repository.go
```diff
@@ -0,0 +1,548 @@
...
+	if len(commits) < workers {
+		workers = len(commits)
+	}
+
+	chunkSize := len(commits) / workers
```
another-rex, line 242: Is this correct? 

5 commits
2 workers

chunk size will be 2

i = 0 
start := 0
end = 2

i = 1
start =2
end = 4

we never process the last commit...?

go/cmd/gitter/repository.go
```diff
@@ -0,0 +1,548 @@
...
+	if err := errg.Wait(); err != nil {
+		return fmt.Errorf("failed to calculate patch IDs: %w", err)
+	}
+
+	logger.Info("Patch ID calculation completed", slog.Int("commits", len(commits)), slog.Duration("duration", time.Since(start)))
+
+	return nil
+}
+
+func (r *Repository) calculatePatchIDsWorker(ctx context.Context, chunk []SHA1) error {
```
another-rex, line 271: Probably a good idea to document in bash syntax what this is trying to run.

go/cmd/gitter/repository.go
```diff
@@ -0,0 +1,548 @@
...
+type Commit struct {
+	Hash    SHA1     `json:"hash"`
+	PatchID SHA1     `json:"patch_id"`
+	Parents []SHA1   `json:"parents"`
+	Tags    []string `json:"tags"`
+}
+
+// Repository holds the commit graph and other details for a git repository.
+type Repository struct {
+	repoMu sync.Mutex
```
another-rex: Add a comment describing what this is locking.

go/cmd/gitter/repository.go
```diff
@@ -0,0 +1,548 @@
...
+	}
+
+	keys := slices.Collect(maps.Keys(unique))
+
+	return keys
+}
+
+// Between walks and returns the commits that are strictly between introduced (inclusive) and limit (exclusive)
+func (r *Repository) Between(introduced, limit []SHA1) []*Commit {
+	r.repoMu.Lock()
```
another-rex: Why are these locking repoMu? It should only be read operations right, and also it should never be happening at the same time as patch id generation?

go/cmd/gitter/gitter.go
another-rex: fyi #4819

go/cmd/gitter/pb/repository/repository.proto
Ly-Joey: Yeah, the logic is actually reconstructing the tags every time in the same `git log` we use to make the git map. This tag cache isn't being used right now, likely we had it in the initial design and then I forgot to clean up after changes.

go/cmd/gitter/repository.go
Ly-Joey: Done. Also changed to use bufio.NewScanner instead of reader because it handles line reading nicer. The only "problem" is that scanner has a buffer size of 64 KB, but it'll be hard to hit that limit with just a couple commit hashes + some refs anyways.

go/cmd/gitter/repository.go
Ly-Joey: I've added comments on the ordering (based on default behaviour in `git log`)

go/cmd/gitter/repository.go
Ly-Joey: Done! Added more comments to explain the git commands and flags we're using. Let me know if they're too detailed(?)

go/cmd/gitter/gitter.go
Ly-Joey: Done in #4846

go/cmd/gitter/gitter.go
Ly-Joey: Added a repo-level RWMutex to lock fetch from other operations (archive, LoadRepo etc)
The repository mutex has been shrunk to only lock for patch ID calculation.

go/cmd/gitter/gitter.go
Ly-Joey: Added this refactoring to the next PR focusing on `affected-commits`

Ly-Joey:
To keep this PR moving along (and so I don't have to keep rebasing and resolving merge conflicts), I've decided to split the affected commit related logic (mainly graph walking and cherrypick detection) into another branch and a future PR.

go/cmd/gitter/gitter.go
```diff
@@ -410,129 +461,23 @@ func cacheHandler(w http.ResponseWriter, r *http.Request) {
...
-	w.Header().Set("Content-Type", "application/json")
 	w.WriteHeader(http.StatusOK)
-	if err := json.NewEncoder(w).Encode(affectedCommits); err != nil {
-		logger.Error("Error encoding affected commits", slog.String("url", url), slog.Any("error", err))
-		http.Error(w, fmt.Sprintf("Error encoding affected commits: %v", err), http.StatusInternalServerError)
-
-		return
-	}
-	logger.Info("Request completed successfully: /affected-commits", slog.String("url", url), slog.Duration("duration", time.Since(start)))
+	logger.InfoContext(ctx, "Request completed successfully: /cache", slog.String("url", url), slog.Duration("duration", time.Since(start)))
```
another-rex: We need to update the context logging library to use the url stored in the context, rather than manually specifying it.

go/cmd/gitter/gitter.go
```diff
@@ -259,7 +298,7 @@ func main() {
 	defer stop()
 
 	for endpoint, handler := range endpointHandlers {
-		http.HandleFunc(endpoint, handler)
+		http.Handle(endpoint, otelhttp.NewHandler(handler, endpoint))
```
another-rex, line 302: err... typo?

Ly-Joey: That's the change from adding log and error trace in #4819 ([ref](https://github.com/google/osv.dev/pull/4819/changes#diff-7ff66276808807fed4832f15dcf45fb300a1fd6c57900d634ba8ff2326b3016aL215))
`otelhttp.NewHandler()` returns a http.Handler now.

go/cmd/gitter/gitter.go
Ly-Joey: PR #4863 addresses this. After that is merged and rebased into this branch, I'll go ahead and remove all the manual "url" args

Ly-Joey: Done

michaelkedar:
Just note if the go importer gets merged first to change the [`/getgit` endpoint here too](https://github.com/google/osv.dev/pull/4808/changes#diff-3ab96f7470e84a0207131f19c15646a55e3c73ebcd205b9fd5a74907adc0a0fcR130)

PR #4729, opened 2026-02-03
Add all CVSS scores and sources from NVD

conorfitch:
Hello - just opening this as a suggestion as something I would find highly useful! Am very interested in discussing anything further.

* Note: This PR is dependent on the following change to the OSV schema (addition of a `severity[].source` field): https://github.com/ossf/osv-schema/pull/510

* Certain data sources, such as NVD and the CVE Program, provide multiple CVSS scores in their records that come from different providers or are different CVSS versions.

* This PR adds all of the available CVSS scores from the NVD record, into the severity object, rather than just a single CVSS score. The sources of the CVSS scores are also added into the OSV record, taken from the `Source` field that is provided in the NVD record.

* Including all the available CVSS scores in the OSV record, along with their source, allows consumers of the data to know who issued the score (e.g. NVD or the CNA), as well as being able to choose between them. For example, a consumer may have a preference for scores issued by NVD or the CNA, or they may wish to use a specific CVSS version for better comparability with other vulnerabilities.

* One CVSS score continues to be selected for the Debian and Alpine records.

* In the NVD and CVE5 combining process, the severities are taken from the NVD record (if it has at least 1 severity). Perhaps instead the severities and their sources should be taken from CVE5, and then add any NVD-issued severity on top of that?

### DIFF
--- a/vulnfeeds/cmd/combine-to-osv/main.go
+++ b/vulnfeeds/cmd/combine-to-osv/main.go
@@ -251,6 +251,10 @@ func combineTwoOSVRecords(cve5 *osvschema.Vulnerability, nvd *osvschema.Vulnerab
 		}
 	}
 
+	if len(nvd.GetSeverity()) > 0 {
+		baseOSV.Severity = nvd.GetSeverity()
+	}
+
 	return baseOSV
 }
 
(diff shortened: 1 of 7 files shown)


jess-lowe:
Hey, thanks for the contribution! Looks great, we're just waiting on discussion time to actually bring up adding the field. 

This may also be out of date because of all my recent NVD refactors (sorry!) but I'll ping when it's a good time to rebase the changes.

jess-lowe:
We released 1.8.0 recently with the severity.source change - you're welcome to rebase/update these changes if you'd like now :)

PR #4730, opened 2026-02-03
feat: add force_update support for git checkouts

another-rex:
This PR introduces a `force_update` mechanism for git repository checkouts.
Specifically:
1. `osv/repos.py` functions `clone`, `clone_with_retries`, `_use_existing_checkout`, and `ensure_updated_checkout` now accept a `force_update` argument.
2. The `importer` worker now uses `force_update=True` when checking out source repositories.
3. The `gitter` service (Go) now parses `force-update=true` and bypasses its 1-hour cache when this flag is present.
4. Local `FETCH_CACHE` in `osv/repos.py` is also bypassed when `force_update=True`.

---
*PR created automatically by Jules for task [6408382110995720439](https://jules.google.com/task/6408382110995720439) started by @another-rex*


go/cmd/gitter/gitter.go
```diff
@@ -251,9 +252,14 @@ func gitHandler(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
+	key := url
```
another-rex: @jules Let's keep the key the same. And add a comment here that occasionally this could be problematic if an existing unforce updated query is already inplace, no force update will happen. That is highly unlikely in our use case, as importer only queries the repo once, and always with force update. 

This is a tradeoff for simplicity to avoid having to setup locks per repo.

michaelkedar (approved):
LGTM

PR #4731, opened 2026-02-03
feat: Create `SourceRepository` interface for Go

michaelkedar:
This is part of the Go rewrite of the Importer, but I thought I'd split up this PR to get some opinions on the database structure.

Adding a `SourceRepository` abstraction, and the bindings for the Datastore object to allow the source repos to be read from Datastore.
There's an abstraction layer to separate out the business logic (`internal/models`) from the Datastore API (`internal/database/datastore`) which should make migrating off of Datastore easier in the future.
I've moved & aliased the existing datastore models in `osv/models` into `internal/database/datastore`, and will eventually try abstract away the datastore usage from the exporter & relations computation.

### DIFF
--- a/go/internal/database/datastore/source_repository.go
+++ b/go/internal/database/datastore/source_repository.go
@@ -54,8 +54,6 @@ func (s *SourceRepositoryStore) Update(ctx context.Context, name string, repo *m
 	return nil
 }
 
-
-
 func (s *SourceRepositoryStore) All(ctx context.Context) iter.Seq2[*models.SourceRepository, error] {
 	return func(yield func(*models.SourceRepository, error) bool) {
 		q := datastore.NewQuery("SourceRepository")


go/internal/database/datastore/internal/validate/validate.go
```diff
@@ -166,12 +175,46 @@ func writeRecords(ctx context.Context, client *datastore.Client) {
...
+		DetectCherrypicks:       true,
+		ConsiderAllBranches:     true,
+		VersionsFromRepo:        true,
+		IgnoreLastImportTime:    true,
+		IgnoreDeletionThreshold: true,
+		Link:                    "https://example.com/go-source",
+		HumanLink:               "https://example.com/go-source/human",
+		DBPrefix:                []string{"GO-TEST", "GO-2-TEST"},
+		StrictValidation:        true,
+	}
```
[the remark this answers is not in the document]
michaelkedar: This test is to test that the Datastore entities (only) can be read/written in both Go and Python, so we populate all fields.
`toModel` isn't used here

go/internal/database/datastore/models.go
```diff
@@ -0,0 +1,105 @@
...
+	SearchIndices    []string       `datastore:"search_indices"`
+}
+
+type SourceRepository struct {
+	Type                    models.SourceRepositoryType `datastore:"type"`
+	Name                    string                      `datastore:"name"`
+	RepoURL                 string                      `datastore:"repo_url"`
+	RepoUsername            string                      `datastore:"repo_username"`
+	RepoBranch              string                      `datastore:"repo_branch"`
+	RestApiUrl              string                      `datastore:"rest_api_url"`
```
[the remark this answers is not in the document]
michaelkedar: since REST is also an initialism, this is `RESTAPIURL` which looks kinda silly but the linter complained so I changed it to that.

go/internal/database/datastore/source_repository.go
```diff
@@ -0,0 +1,151 @@
...
+	case models.SourceRepositoryTypeREST:
+		msr.REST = &models.SourceRepoREST{
+			URL:                     sr.RestApiUrl,
+			LastUpdated:             sr.LastUpdateDate,
+			IgnoreLastImportTime:    sr.IgnoreLastImportTime,
+			IgnoreDeletionThreshold: sr.IgnoreDeletionThreshold,
+		}
+	}
+	return msr
+}
```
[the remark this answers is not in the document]
michaelkedar: I'm intentionally removing Username, since its purpose seems to be for SSH writing, and we no longer want the importer (or worker) to be able to modify repos.
Repos should be public anyway, so there's no need to have a username / key.

go/internal/database/datastore/internal/validate/validate.go
```diff
@@ -30,65 +31,73 @@ func main() {
...
 	fmt.Println("(Go) Getting RelatedGroup")
 	key = datastore.NameKey("RelatedGroup", "CVE-123-456", nil)
-	var relatedGroup models.RelatedGroup
+	var relatedGroup db.RelatedGroup
 	if err := client.Get(ctx, key, &relatedGroup); err != nil {
 		fmt.Printf("(Go) Failed getting RelatedGroup: %v\n", err)
 		os.Exit(1)
 	}
+
+	fmt.Println("(Go) Getting SourceRepository")
```
cuixq, line 88: shall we use log instead?

go/internal/database/datastore/source_repository.go
```diff
@@ -0,0 +1,155 @@
...
+			LastUpdated:             sr.LastUpdateDate,
+			IgnoreLastImportTime:    sr.IgnoreLastImportTime,
+			IgnoreDeletionThreshold: sr.IgnoreDeletionThreshold,
+		}
+	}
+
+	return msr
+}
+
+type SourceRepositoryStore struct {
```
cuixq: maybe move the struct definition to the beginning of this file for better readability?

go/internal/database/datastore/internal/validate/validate.go
```diff
@@ -166,12 +175,46 @@ func writeRecords(ctx context.Context, client *datastore.Client) {
...
+	relatedGroup := db.RelatedGroup{
 		RelatedIDs: []string{"R-1", "R-2"},
 		Modified:   time.Date(2025, time.January, 1, 1, 1, 1, 1, time.UTC),
 	}
 	if _, err := client.Put(ctx, key, &relatedGroup); err != nil {
 		fmt.Printf("(Go) Failed writing RelatedGroup %v: %v\n", key, err)
 		os.Exit(1)
 	}
+
+	fmt.Println("(Go) Writing SourceRepository")
```
cuixq, line 187: can we modify the log message to make it clear that we are writing to the test bucket for validation purpose? the message now sounds like we are actually writing to bucket with the fake data.

go/internal/database/datastore/source_repository.go
michaelkedar: moved

go/internal/database/datastore/internal/validate/validate.go
michaelkedar: I added a print to the beginning of the Python file to indicate that the Datastore emulator is running:
```
Datastore emulator running
(Python) Putting Vulnerability
...
(Python) Putting SourceRepository
(Go) Getting Vulnerability
...
(Go) Getting SourceRepository
(Go) Writing Vulnerability
...
(Go) Writing SourceRepository
(Python) Getting Vulnerability
...
(Python) Getting SourceRepository
```

go/internal/database/datastore/internal/validate/validate.go
michaelkedar: I think I used prints just to keep the output from the python and go parts of this validation somewhat consistent with each other.

go/internal/database/datastore/source_repository.go
```diff
@@ -0,0 +1,155 @@
...
```
another-rex: Add a var _ models.SourceRepositoryStore = SourceRepositoryStore

to make it clear this implements that, and make errors obvious

another-rex: Also, this really should be at the top of the file, and toModel() at the very bottom of the file.

another-rex (commented):
I do quite like the abstracted models.

go/internal/database/datastore/source_repository.go
michaelkedar: reordered and assigned

cuixq (approved):

PR #4747, opened 2026-02-04
feat(NVD): have NVD conversion indiscriminately output records regardless of git resolution outcome

jess-lowe:
This PR:
- Moves resolving git commits logic into its own function to reduce duplication of logic across OSV/PackageInfo records
- Indiscriminately outputs NVD-OSV records regardless of whether versions were able to be resolved to commits. Bare versions are saved in the database_specific field. (NOTE: this is currently the high level `database_specific` field - future work will involve placing these at `affected.ranges.database_specific` level, but more reworking will need to be done to complete that. 
- Outputs a csv file of CVEID, Outcome mappings for analysis, and refactored to focus on Conversion Outcome rather than errors.

### DIFF
--- a/vulnfeeds/conversion/nvd/converter.go
+++ b/vulnfeeds/conversion/nvd/converter.go
@@ -201,7 +201,7 @@ func FindRepos(cve models.NVDCVE, vpRepoCache *cves.VPRepoCache, repoTagsCache *
 			metrics.AddNote("Failed to parse CPE: %v", CPEstr)
 			continue
 		}
-		if CPE.Part != "a" || CPE.Part != "o" {
+		if CPE.Part != "a" && CPE.Part != "o" { // only care about application and operating system CPEs
 			continue
 		}
 		vendorProductCombinations[cves.VendorProduct{Vendor: CPE.Vendor, Product: CPE.Product}] = true


vulnfeeds/conversion/nvd/converter.go
```diff
@@ -126,50 +107,22 @@ func CVEToPackageInfo(cve models.NVDCVE, repos []string, cache *git.RepoTagsCach
...
+	metrics.Outcome = ResolveVersionsToCommits(&versions, repos, cache, metrics)
 
 	if len(versions.AffectedCommits) == 0 {
 		metrics.AddNote("No affected commit ranges determined for %q", maybeProductName)
-		return fmt.Errorf("no affected commit ranges determined for %q %w", maybeProductName, ErrNoRanges)
+		metrics.Outcome = models.NoCommitRanges
+	}
+
+	if rejectFailed && metrics.Outcome != models.Successful {
+		return metrics.Outcome
```
another-rex, line 125: Is it intentional that we don't write the metrics file if outcome fails? e.g. there are no affected commits

Can you add a comment here stating that if it is intentional and why.

jess-lowe: Not quite intentional, but will be handled much better in a future PR

jess-lowe:
I'm going to close this PR in favour of #4853

PR #4769, opened 2026-02-08
fix(docs): api redirects on faq page

ashmod:
Fixes redirects for the API page on [FAQ](https://google.github.io/osv.dev/faq/)


another-rex (approved):
Thanks!

PR #4770, opened 2026-02-08
fix(nvd): repo from commit in refs different from CPE cache resulting in failed conversion

jess-lowe:
CVE-2026-23522 has a different repository in its CPE cache than in the commit extracted from references, causing a weird bug where it considers it a failed conversion.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -27,7 +27,7 @@ OSV consists of:
 1. [The OSV Schema](https://ossf.github.io/osv-schema/): An easy-to-use data
    format that maps precisely to open source versioning schemes.
 2. Reference infrastructure ([OSV.dev website](https://osv.dev/),
-   [API](/api/), and tooling) that aggregates,
+   [API](../api/), and tooling) that aggregates,
    [enriches](#what-does-osvdev-do-to-the-records-it-imports) and indexes
    vulnerability data from databases that use the OSV schema.
 3. [OSV-Scanner](https://github.com/google/osv-scanner), the officially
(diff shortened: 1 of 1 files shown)


vulnfeeds/conversion/nvd/converter.go
```diff
@@ -63,6 +63,14 @@ func CVEToOSV(cve models.NVDCVE, repos []string, cache *git.RepoTagsCache, direc
...
 			}
 		}
+		if !hasAnyFixedCommits {
+			for _, ac := range versions.AffectedCommits {
+				if ac.Fixed != "" {
+					hasAnyFixedCommits = true
+					break
+				}
+			}
+		}
```
another-rex: I'm not sure this is the best fix. What is the point of the first section where we check fixed commits against repos if we are going to do another loop without caring about the repo?

vulnfeeds/conversion/nvd/converter.go
another-rex, line 44: nit: can we rename this to affectedVersionsAndCommits. I was quite confused for a bit, had to look at the definitions to understand what this actually contained.

another-rex (commented):
Add a comment around

vulnfeeds/conversion/nvd/converter.go
another-rex: 1. Can we fix the root cause of the repos being mismatched
2. If 1 is inevitable, do we need to change the logic everywhere HasFixedCommits(repo) is used? I see it called in a few different places
3. Either way I don't think we need both loops? We can just remove the first loop right?

jess-lowe: TBH I want to rewrite this whole thing to rely natively on the OSV schema fields instead and make the logic a lot clearer, but it's not a high priority rn.

Removed the first block.

vulnfeeds/conversion/nvd/converter.go
jess-lowe: Yeah it's confusing cause it's referencing the VersionInfo struct (which consists of AffectedCommits and AffectedVersions). I can refactor to rename this but eventually I do want to move completely off the VersionInfo struct

PR #4779, opened 2026-02-10
fix(nvd): introduced = 0 missing in cases where more than one range

jess-lowe:
After changing the conversion so that versions extracted and resolved to git commits were not overwritten entirely by a commit being extracted from the references, there persisted an issue where introduced = 0 was not adequately represented in the resulting record. 

This fixes that, as well as provides somewhat of a fix to a commit being extracted from the references also existing in the extracted versions and causing a duplicate. 

It orders it so that introduced == 0 appears first and the extracted git commit appears last.


another-rex (dismissed):
Please fill out the PR description with what changed. That goes into the commit message, which would be useful for debugging in the future.

another-rex (approved):

PR #4783, opened 2026-02-10
feat(cve5): Convert the CNA Allowlist into a denylist. 

jess-lowe:
To facilitate new CNAs that might successfully convert, the allowlist is changed into a denylist. All CNAs in the denylist have never successfully converted any ranges to commits, so are assumed out of scope.


another-rex (approved):
Let's wait until the new release before merging this in

PR #4798, opened 2026-02-11
fix: Bump up gitter allocation

another-rex:
Also fixed:
- Replace deprecated gcePersistentDisk with csi driver
- Actually add gitter.yaml to kustomizations.yaml  (I can't believe this was working before)
- Overriding the volume claim in test (I'm not sure if this works, we'll find out)

### DIFF
--- a/deployment/clouddeploy/gke-workers/environments/oss-vdb-test/kustomization.yaml
+++ b/deployment/clouddeploy/gke-workers/environments/oss-vdb-test/kustomization.yaml
@@ -24,3 +24,4 @@ patches:
 - path: recoverer.yaml
 - path: record-checker.yaml
 - path: custommetrics.yaml
+- path: gitter.yaml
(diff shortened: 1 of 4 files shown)


Ly-Joey (approved):

deployment/clouddeploy/gke-workers/environments/oss-vdb-test/kustomization.yaml
```diff
@@ -24,3 +24,4 @@ patches:
 - path: recoverer.yaml
 - path: record-checker.yaml
 - path: custommetrics.yaml
+- path: gitter.yaml
```
michaelkedar, line 27: how was this working before if this wasn't in this file?

another-rex: It just wasn't overriding the defaults, which was the google cloud project env variable, and gitter is project agnostic so it was fine.

PR #4807, opened 2026-02-13
fix(gitter): Add more logs and retry for index.lock

Ly-Joey:
Better logging specifically in runCmd()
Also added a retry logic for error: `fatal: Unable to create 'repo_path/.git/index.lock': File exists.` during `git reset`.

### DIFF
--- a/go/cmd/gitter/gitter.go
+++ b/go/cmd/gitter/gitter.go
@@ -269,8 +269,7 @@ func gitHandler(w http.ResponseWriter, r *http.Request) {
 	}
 	forceUpdate := r.URL.Query().Get("force-update") == "true"
 
-	ctx := r.Context()
-	ctx = context.WithValue(ctx, urlKey, url)
+	ctx := context.WithValue(r.Context(), urlKey, url)
 
 	logger.Info("Received request", slog.String("url", url), slog.Bool("forceUpdate", forceUpdate), slog.String("remoteAddr", r.RemoteAddr))
 	// If request came from a local ip, don't do the check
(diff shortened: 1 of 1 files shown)


go/cmd/gitter/gitter.go
```diff
@@ -43,6 +44,7 @@ const shutdownTimeout = 10 * time.Second
 // runCmd executes a command with context cancellation handled by sending SIGINT.
 // It logs cancellation errors separately as requested.
 func runCmd(ctx context.Context, dir string, env []string, name string, args ...string) error {
+	logger.Debug("Running command", slog.String("cmd", name), slog.String("dir", dir), slog.Any("args", args))
```
another-rex: For the logging, ideally I think we should have the same url in the value in all the logs of the same request. I think a good way to do this could be to add the url value into the context. (`context.WithValue('url', url)` and then `ctx.Value('url')`)

go/cmd/gitter/gitter.go
```diff
@@ -130,6 +139,18 @@ func fetchBlob(ctx context.Context, url string, forceUpdate bool) ([]byte, error
 				return nil, fmt.Errorf("git fetch failed: %w", err)
 			}
 			err = runCmd(ctx, repoPath, nil, "git", "reset", "--hard", "origin/HEAD")
+			if isIndexLockError(err) {
+				// index.lock exists, likely a previous git reset got terminated and wasn't cleaned up properly.
+				// We can remove the file and retry the command
+				logger.Warn("index.lock exists, attempting to remove and retry", slog.String("repoPath", repoPath))
```
another-rex: url as a value here as well.

another-rex (approved):

PR #4808, opened 2026-02-13
feat: Re-implement importer in Go

michaelkedar:
Big re-implementation of the importer (and importer-deleter) into Go:
- Implemented go gitter getter
- Moved distinct logic (git/GCS/REST) into relevant files
- Removed some outdated things (e.g. public_logs_bucket)
  - I've remove the writing of ImportFindings, because I think the linter is currently also doing (and overwriting) these 
- Parallelised processing of source repositories
  - also parallelised downloading records and publishing to pubsub within a source repository via a shared worker pool
- Made interfaces so that the code is testable.

I'd appreciate people reviewing this to see how follow-able/self-documenting the code is, and point out where things may be unclear.

### DIFF
--- a/go/internal/repos/repos.go
+++ b/go/internal/repos/repos.go
@@ -44,20 +44,15 @@ func CloneToDir(ctx context.Context, repoURL string, dir string, forceUpdate boo
 		if err == nil {
 			// It's a git repo, pull the latest changes.
 			if forceUpdate {
-				wt, err := repo.Worktree()
-				if err != nil {
-					return nil, fmt.Errorf("failed to get worktree: %w", err)
+				cmd := exec.CommandContext(ctx, "git", "pull", "--", repoURL, dir)
+				cmd.Env = append(cmd.Env, "GIT_TERMINAL_PROMPT=0")
+				if err := cmd.Run(); err != nil {
+					return nil, fmt.Errorf("failed to pull repo: %w", err)
 				}
-				err = wt.Pull(&git.PullOptions{
-					RemoteName: "origin",
-					Force:      true,
-				})
-				if err != nil && !errors.Is(err, git.NoErrAlreadyUpToDate) {
-					return nil, fmt.Errorf("failed to pull from remote: %w", err)
-				}
-				if err == nil {
-					logger.Info("Pulled latest changes", slog.String("repo_url", repoURL), slog.String("dir", dir))
-				}
+				logger.Info("Pulled latest changes", slog.String("repo_url", repoURL), slog.String("dir", dir))
+
+				// re-open the repo to get the new HEAD
+				return git.PlainOpen(dir)
 			}
 
 			return repo, nil
@@ -84,9 +79,13 @@ func CloneToDir(ctx context.Context, repoURL string, dir string, forceUpdate boo
 	}
 
 	// regular clone
-	return git.PlainCloneContext(ctx, dir, &git.CloneOptions{
-		URL: repoURL,
-	})
+	cmd := exec.CommandContext(ctx, "git", "clone", "--", repoURL, dir)
+	cmd.Env = append(cmd.Env, "GIT_TERMINAL_PROMPT=0")
+	if err := cmd.Run(); err != nil {
+		return nil, fmt.Errorf("failed to clone repo: %w", err)
+	}
+
+	return git.PlainOpen(dir)
 }
 
 func cloneToDirGitter(ctx context.Context, gitterHost, repoURL, dir string, forceUpdate bool) (*git.Repository, error) {
@@ -127,7 +126,7 @@ func cloneToDirGitter(ctx context.Context, gitterHost, repoURL, dir string, forc
 }
 
 func gitterGet(ctx context.Context, gitterHost, repoURL string, forceUpdate bool) (io.ReadCloser, error) {
-	getGitURL, err := url.JoinPath(gitterHost, "getgit")
+	getGitURL, err := url.JoinPath(gitterHost, "git")
 	if err != nil {
 		return nil, fmt.Errorf("failed to join path: %w", err)
 	}


go/internal/importer/schema_generated.json
michaelkedar, line 1: This file gets copied here via `go:generate` from the osv-schema submodule.
I'm not really sure what the best practice is here.
We currently don't have any automation to update this, nor check that it is up to date.

go/cmd/importer/main.go
```diff
@@ -0,0 +1,81 @@
...
+	if *runDelete {
+		if err := importer.RunDeletions(context.Background(), config); err != nil {
+			logger.Fatal("Importer-deleter failed", slog.Any("error", err))
+		}
+	} else {
+		if err := importer.Run(context.Background(), config); err != nil {
+			logger.Fatal("Importer failed", slog.Any("error", err))
+		}
+	}
+}
```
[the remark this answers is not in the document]
michaelkedar: did not know this was a thing - done

go/internal/importer/importer.go
```diff
@@ -0,0 +1,333 @@
...
+		select {
+		case <-ctx.Done():
+			return
+		case sourceRecord, ok := <-ch:
+			if !ok {
+				return
+			}
+			sourceRepoName := sourceRecord.SourceRepository()
+			sourcePath := sourceRecord.SourcePath()
+			strict := config.StrictValidation && sourceRecord.Strictness()
```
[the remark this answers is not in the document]
michaelkedar: This is how strict behaved in python. Added some comment to the CLI flag to clarify

go/internal/importer/importer.go
```diff
@@ -0,0 +1,333 @@
...
+			"original_sha256": hash,
+			"deleted":         strconv.FormatBool(deleted),
+			"req_timestamp":   strconv.FormatInt(time.Now().Unix(), 10),
+		},
+	}
+	if srcTimestamp != nil {
+		msg.Attributes["src_timestamp"] = strconv.FormatInt(srcTimestamp.Unix(), 10)
+	} else {
+		msg.Attributes["src_timestamp"] = ""
+	}
```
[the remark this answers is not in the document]
michaelkedar: The Python code did this, so I'm hesitant to change it without changing the worker first.

go/internal/repos/repos.go
```diff
@@ -0,0 +1,157 @@
...
+	defer f.Close()
+	_, err = io.Copy(f, w)
+	if err != nil {
+		return nil, fmt.Errorf("failed to write tar file: %w", err)
+	}
+	// shell out to tar xf
+	if err := os.MkdirAll(dir, 0755); err != nil {
+		return nil, fmt.Errorf("failed to create directory: %w", err)
+	}
+	cmd := exec.CommandContext(ctx, "tar", "-xf", tarPath, "-C", dir)
```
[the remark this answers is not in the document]
michaelkedar: I'd like to do this in principle, but it's a lot of lines of code to actually do this in go (especially iterating through the tar archive), versus just one for shelling out.

go/cmd/exporter/exporter.go
```diff
@@ -91,19 +91,19 @@ func main() {
...
 	prevPrefix := ""
 MainLoop:
-	for path, err := range vulnClient.Objects(ctx, gcsProtoPrefix) {
+	for obj, err := range vulnClient.Objects(ctx, gcsProtoPrefix) {
 		if err != nil {
 			logger.Fatal("failed to list objects", slog.Any("err", err))
 		}
 		// Only log when we see a new ID prefix (i.e. roughly once per data source)
-		prefix := filepath.Base(path)
+		prefix := filepath.Base(obj.Name)
```
another-rex, line 105: huh? How did this work before?

go/internal/importer/importer.go
```diff
@@ -0,0 +1,333 @@
...
+
+	workCh := make(chan SourceRecord)
+	var workWg sync.WaitGroup
+	for range config.NumWorkers {
+		workWg.Go(func() {
+			importerWorker(ctx, workCh, config)
+		})
+	}
+
+	var wg sync.WaitGroup
```
another-rex: since there is already a workerWg, this should be named as well. sourceWg?

go/internal/importer/importer.go
```diff
@@ -0,0 +1,333 @@
...
+
+func (r RetryableHTTPLeveledLogger) Debug(msg string, keysAndValues ...any) {
+	logger.Debug(msg, keysAndValues...)
+}
+
+func (r RetryableHTTPLeveledLogger) Warn(msg string, keysAndValues ...any) {
+	logger.Warn(msg, keysAndValues...)
+}
+
+func Run(ctx context.Context, config Config) error {
```
another-rex: Code's mostly self documenting, but to aid in comprehension add a comment at the start (or docstring on the function) describing what it does  in general. 
1. Create import workers
3. Loop through repos and send them to workers via workCh

go/internal/importer/helper.go
```diff
@@ -0,0 +1,75 @@
...
+	"log/slog"
+	"regexp"
+	"slices"
+	"strings"
+
+	"github.com/google/osv.dev/go/internal/models"
+	"github.com/google/osv.dev/go/logger"
+)
+
+func negatedPrefix(regex string) (string, bool) {
```
another-rex, line 13: Feels like we should add a TODO and a issue to clean those up from source.yaml, then remove this function.

go/internal/importer/git.go
```diff
@@ -0,0 +1,274 @@
...
+func handleImportGit(ctx context.Context, ch chan<- SourceRecord, config Config, sourceRepo *models.SourceRepository) error {
+	if sourceRepo.Type != models.SourceRepositoryTypeGit || sourceRepo.Git == nil {
+		return errors.New("invalid SourceRepository for git import")
+	}
+	logger.Info("Importing git source repository",
+		slog.String("source", sourceRepo.Name), slog.String("url", sourceRepo.Git.URL))
+
+	compiledIgnorePatterns := compileIgnorePatterns(sourceRepo)
+	repoInterface, err, _ := repoGroup.Do(sourceRepo.Git.URL, func() (any, error) {
+		// Temporary migration from Python to Go
```
another-rex, line 73: TODO for easy tracking?

go/internal/database/datastore/vulnerability.go
```diff
@@ -0,0 +1,66 @@
...
+	client *datastore.Client
+}
+
+var _ models.VulnerabilityStore = (*VulnerabilityStore)(nil)
+
+func NewVulnerabilityStore(client *datastore.Client) *VulnerabilityStore {
+	return &VulnerabilityStore{client: client}
+}
+
+func (s *VulnerabilityStore) ListBySource(ctx context.Context, source string, skipWithdrawn bool) iter.Seq2[*models.VulnSourceRef, error] {
```
another-rex, line 25: Are there plans to split this source thing apart? Or is that more part of the database migration step.

go/internal/importer/git.go
```diff
@@ -0,0 +1,274 @@
...
+)
+
+type gitSourceRecord struct {
+	repo             sharedRepo
+	commit           *object.Commit
+	path             string
+	keyPath          string
+	format           RecordFormat
+	sourceRepository string
+	shouldSendUpdate bool
```
another-rex: Please comment what shouldSendUpdate actually means.

another-rex: Or maybe the variable is actually not named correctly, I think it should be describing what is actually happening, rather than describing what it wants the receiver to do.

go/internal/importer/git.go
```diff
@@ -0,0 +1,274 @@
...
+func (g gitSourceRecord) SourceRepository() string {
+	return g.sourceRepository
+}
+
+func (g gitSourceRecord) SourcePath() string {
+	return g.path
+}
+
+func (g gitSourceRecord) ShouldSendModifiedTime() bool {
+	return g.shouldSendUpdate
```
another-rex: I think this is intentional, but it looks like a typo. I think the variable names need to be reworked here.

go/internal/importer/importer.go
```diff
@@ -0,0 +1,333 @@
...
+	// Format is the format (json/yaml) of the source record
+	Format() RecordFormat
+	// LastUpdated is the last updated time of the source record
+	LastUpdated() (time.Time, bool)
+	// SourceRepository is the name of the source repository
+	SourceRepository() string
+	// SourcePath is path to the record within the source repository
+	SourcePath() string
+	// ShouldSendModifiedTime returns true if the modified time should be sent (for latency monitoring)
+	ShouldSendModifiedTime() bool
```
another-rex: Can this function be renamed to something clearer?

go/internal/importer/git.go
```diff
@@ -0,0 +1,274 @@
...
+			if !strings.HasPrefix(p, dirPath) {
+				return ""
+			}
+		}
+		if shouldIgnore(path.Base(p), sourceRepo.IDPrefixes, compiledIgnorePatterns) {
+			return ""
+		}
+
+		return p
+	}
```
another-rex, line 128:
```suggestion
	filterPath := func(p string) string {
	  // Has Extension suffix?
		if !strings.HasSuffix(p, sourceRepo.Extension) {
			return ""
		}
		// Has Git.Path Prefix?
		if dirPath := sourceRepo.Git.Path; dirPath != "" {
			if !strings.HasSuffix(dirPath, "/") {
				dirPath += "/"
			}
			if !strings.HasPrefix(p, dirPath) {
				return ""
			}
		}
		
		// Does it match an ignore pattern?
		if shouldIgnore(path.Base(p), sourceRepo.IDPrefixes, compiledIgnorePatterns) {
			return ""
		}

		return p
	}
```

go/internal/importer/bucket.go
```diff
@@ -0,0 +1,214 @@
...
+	if !sourceRepo.Bucket.IgnoreLastImportTime && sourceRepo.Bucket.LastUpdated != nil {
+		lastUpdated = *sourceRepo.Bucket.LastUpdated
+		hasUpdateTime = true
+	}
+	format := RecordFormatUnknown
+	if strings.ToLower(sourceRepo.Extension) == ".yaml" || strings.ToLower(sourceRepo.Extension) == ".yml" {
+		format = RecordFormatYAML
+	} else if strings.ToLower(sourceRepo.Extension) == ".json" {
+		format = RecordFormatJSON
+	}
```
another-rex: Refactor this out into a separate function?

go/internal/importer/rest.go
```diff
@@ -0,0 +1,300 @@
...
+	timeOfRun := time.Now()
+	lastModTime := time.Time{}
+	if hasUpdateTime {
+		// HEAD request to check if there are updates
+		req, err := http.NewRequest(http.MethodHead, sourceRepo.REST.URL, nil)
+		if err != nil {
+			return err
+		}
+		req = req.WithContext(ctx)
+		resp, err := config.HTTPClient.Do(req)
```
another-rex: Is there an err if response is not 200? 

If so do we want to do the fallback when HEAD reqs are not  supported?

If not, can we check if it returns 200 and if not log a warning?

go/internal/importer/rest.go
```diff
@@ -0,0 +1,300 @@
...
+	// This is to be more robust in case of misbehaving servers.
+	timeToUpdate := timeOfRun
+	if !maxModified.IsZero() && maxModified.Before(timeToUpdate) {
+		timeToUpdate = maxModified
+	}
+	if !lastModTime.IsZero() && lastModTime.Before(timeToUpdate) {
+		timeToUpdate = lastModTime
+	}
+	sourceRepo.REST.LastUpdated = &timeToUpdate
+	sourceRepo.REST.IgnoreLastImportTime = false
```
another-rex, line 165: Should we do the update to switch this back immediately? I guess it shouldn't matter if another importer run doesn't run at the same time.

go/cmd/exporter/exporter.go
michaelkedar: I changed `Objects` to return metadata instead of just filenames (to save on number of requests that the importer needs to make)

go/internal/database/datastore/vulnerability.go
michaelkedar: In Datastore, it's annoying but I don't think it really impacts our ability to do queries
(I don't think we'd ever need to query a path without knowing the source, and the `< source + ";"` works fine)
I think it's fine to leave it here, but it does make sense to have these fields separate if/when we move to postgres

go/internal/importer/git.go
michaelkedar: Are these comments really adding anything?
like, `Has Extension suffix?` and `strings.HasSuffix(p, sourceRepo.Extension)` basically communicate the same amount of info

go/internal/importer/rest.go
michaelkedar: If there is some transient error in the importer and it doesn't complete, I don't know if we want to write back immediately versus retrying.
And yeah, since only one importer may run at a time, I think this is okay.

go/internal/importer/rest.go
michaelkedar: `err` is nil if there's a non-2xx response.
done, and moved to a `checkHEAD` function

go/internal/importer/git.go
michaelkedar: (For some reason, GitHub has completely lost what this comment was attached to)

This field has been moved into a `WorkItem` struct, and the fields were all named `ShouldSendModified`.
The field exists to determine whether we should be sending the original modified time of the record over pub/sub, which is meant to be for our update latency SLO measuring. We don't want to measure updates when we're doing a reimport (because this would measure a latency of years).

I'll rename this to `IsReimport` (and negate it) which is probably more clear

go/internal/importer/git.go
michaelkedar: see https://github.com/google/osv.dev/pull/4808#discussion_r2825747378

go/internal/importer/importer.go
michaelkedar: see https://github.com/google/osv.dev/pull/4808#discussion_r2825747378

go/cmd/importer/main.go
```diff
@@ -49,33 +62,61 @@ func main() {
...
 	} else {
-		if err := importer.Run(context.Background(), config); err != nil {
-			logger.Fatal("Importer failed", slog.Any("error", err))
+		if err := importer.Run(ctx, config); err != nil {
+			logger.FatalContext(ctx, "Importer failed", slog.Any("error", err))
 		}
 	}
 }
+
+// importerSampleRate returns the sample rate for the importer (not the individual vulnerability entries).
```
another-rex: Can you explain the difference between importer and individual vulnerability entries here?

michaelkedar: Added some more detail here and in `vulnerabilitySampleRate()`

michaelkedar:
> Probably should mention in the PR/commit description that this does not do git fetches anymore

[It does still do a pull](https://github.com/google/osv.dev/pull/4808/changes#diff-3ab96f7470e84a0207131f19c15646a55e3c73ebcd205b9fd5a74907adc0a0fcR51-R54)

go/cmd/importer/main.go
```diff
@@ -0,0 +1,127 @@
...
+)
+
+func main() {
+	logger.InitGlobalLogger()
+	defer logger.Close()
+	ctx, span := otel.Tracer("importer").Start(context.Background(), "importer",
+		trace.WithAttributes(attribute.Float64("override_sample_rate", importerSampleRate())))
+	defer span.End()
+
+	strictValidation := flag.Bool("strict-validation", false, "Fail to import entries that do not pass validation. "+
```
cuixq:
```suggestion
	strictValidation := flag.Bool("strict-validation", false, "Do not import entries if fail to pass validation. "+
```

cuixq (dismissed):

another-rex (approved):

PR #4815, opened 2026-02-16
refactor(vulnfeeds): move GitVersionsToCommits to common dir and refactor to output resolved ranges

jess-lowe:
Move more common functions and remove useless ones.

GitVersionsToCommits now outputs resolved ranges, unresolved ranges and successful repositories. This function will be more heavily used in the next PR for the NVD refactor.

It should output resolved/unresolved ranges instead of affected ranges so that we are able to group the ranges by repository into one set of ranges/events later on, so as not to have duplicate repository/package information in the record.

Renamed another function also called GitVersionsToCommits to VersionInfoToCommits to better reflect what it's actually doing and as to reduce the confusion around the same name.

### DIFF
--- a/vulnfeeds/conversion/common.go
+++ b/vulnfeeds/conversion/common.go
@@ -233,6 +233,6 @@ func resolveVersionToCommit(version string, normalizedTags map[string]git.Normal
 	if err != nil {
 		return ""
 	}
-	
+
 	return commit
 }


vulnfeeds/cvelist2osv/default_extractor.go
```diff
@@ -39,40 +38,81 @@ func (d *DefaultVersionExtractor) ExtractVersions(cve models.CVE5, v *vulns.Vuln
...
-			logger.Error("Failed to convert git versions to commits", slog.Any("err", err))
+		resolvedRanges, unresolvedRanges, _ := conversion.GitVersionsToCommits(ranges, repos, metrics, repoTagsCache)
+		if len(resolvedRanges) == 0 {
+			metrics.AddNote("Failed to convert git versions to commits")
 		} else {
 			gotVersions = true
 		}
-		conversion.AddAffected(v, aff, metrics)
+
+		if len(resolvedRanges) > 0 {
```
another-rex: Should this block of code be extracted into a function? it's repeated 3 times.

jess-lowe: Fixed in latest commit

another-rex (approved):

PR #4816, opened 2026-02-16
fix(gitter): index.lock error fallback

Ly-Joey:


go/cmd/gitter/gitter.go
```diff
@@ -145,20 +145,23 @@ func fetchBlob(ctx context.Context, url string, forceUpdate bool) ([]byte, error
...
-				if err := os.Remove(indexLockPath); err != nil {
-					return nil, fmt.Errorf("failed to remove index.lock in %s: %w", repoPath, err)
-				}
-				// One more attempt at git reset
-				err = runCmd(ctx, repoPath, nil, "git", "reset", "--hard", "origin/HEAD")
-			}
-
 			if err != nil {
-				return nil, fmt.Errorf("git reset failed: %w", err)
+				if isIndexLockError(err) {
```
another-rex: isIndexLockError should have a check within itself that err is not nil.

PR #4817, opened 2026-02-16
feat(nvd): Add functions to help monitor performance - ConductAnalysis and CpuProfile

jess-lowe:
Added CPU profiling to nvd conversion so we can easily check what's taking the most time, and ConductAnalysis, which runs through all of the outputted metrics files after conversion runs and produces a CSV file on the outcomes.

### DIFF
--- a/vulnfeeds/cmd/converters/cve/nvd-cve-osv/main.go
+++ b/vulnfeeds/cmd/converters/cve/nvd-cve-osv/main.go
@@ -30,7 +30,7 @@ var (
 	outFormat           = flag.String("out-format", "OSV", "Format to output {OSV,PackageInfo}")
 	workers             = flag.Int("workers", 30, "The number of concurrent workers to use for processing CVEs.")
 	outputMetrics       = flag.Bool("output-metrics", true, "If true, output the metrics information about the conversion")
-	cpuProfile          = flag.String("cpuprofile", "", "write cpu profile to file")
+	cpuProfile          = flag.String("cpuprofile", "", "Path to write cpu profile to file (default = no output)")
 )
 
 func loadCPEDictionary(productToRepo *cves.VPRepoCache, f string) error {


vulnfeeds/cmd/converters/cve/nvd-cve-osv/main.go
```diff
@@ -25,6 +29,8 @@ var (
 	outDir              = flag.String("out-dir", "", "Path to output results.")
 	outFormat           = flag.String("out-format", "OSV", "Format to output {OSV,PackageInfo}")
 	workers             = flag.Int("workers", 30, "The number of concurrent workers to use for processing CVEs.")
+	outputMetrics       = flag.Bool("output-metrics", true, "If true, output the metrics information about the conversion")
+	cpuProfile          = flag.String("cpuprofile", "", "write cpu profile to file")
```
tobyhawker: The help string is phrased as if this flag were a boolean, but the String argument seems to be the path. Could you rephrase to indicate this is the path the cpu profile will be written to?

vulnfeeds/cmd/converters/cve/nvd-cve-osv/main.go
```diff
@@ -49,6 +55,18 @@ func main() {
...
+			logger.Fatal("could not create CPU profile: ", slog.Any("err", err))
+		}
+		defer f.Close()
+		if err := pprof.StartCPUProfile(f); err != nil {
+			logger.Fatal("could not start CPU profile: ", slog.Any("err", err))
+		}
+		defer pprof.StopCPUProfile()
+	}
+
 	logger.InitGlobalLogger()
```
michaelkedar, line 70: nit: setting up the profiler should probably go after `logger.InitGlobalLogger()` (though it doesn't actually make a difference)

michaelkedar (approved):

PR #4818, opened 2026-02-16
docs: add instructions for running the documentation site locally

cuixq:
This PR updates `docs/README.md` to provide clearer instructions for contributors to run and test the documentation site locally in docker or natively.

To run documentation site in docker, this PR introduces `docs/docs.Dockerfile` to allow consistent local documentation serving without requiring a local Ruby environment.


docs/docs.Dockerfile
```diff
@@ -0,0 +1,23 @@
+# Use an official Ruby runtime as a parent image.
+FROM ruby:4
+
+# Set the working directory in the container.
+WORKDIR /usr/src/app
+
+# Copy the Gemfile and Gemfile.lock, and the bundle config.
+# This is done first to leverage Docker's layer caching.
+COPY ./Gemfile* ./
+COPY .bundle ./.bundle
```
ashmod: this copy will fail if .bundle is not available locally because there is no .bundle directory pre-existing in the repo.

another-rex:
We probably can just remove the .bundle copy part. 

Though @cuixq can you test if you can actually run the docs locally? I seem to be getting conflicts on unable to resolve the dependencies

docs/docs.Dockerfile
cuixq: thanks for pointing this out! I have removed this line.

cuixq:
@another-rex it seems due to the ruby version e.g. [commonmarker 0.23.12](https://rubygems.org/gems/commonmarker/versions/0.23.12) has requirement `>= 2.6, < 4.0`.

I have modified the docker image to be using `ruby:3` for now. We probably need to do this for [osv-scanner](https://github.com/google/osv-scanner/blob/main/docs/docs.Dockerfile) too.

Issue #3547, opened 2025-06-04
Investigate using OpenTelemetry for API trace logging


michaelkedar:
Recently, GCP seems to have improved the trace explorer - it looks like it could be more useful for us to monitor API performance. 

Currently, we have traces from ESPv2 propagated into our API handler with a bit of a [shoddy log wrapper](https://github.com/google/osv.dev/blob/f3aa51f0689214ee9ce4e57173e56b933fda3649/gcp/api/server.py#L122) (that I made). While these show up on the trace explorer, it seems a little bit off.

The GCP tracing documentation points to using OpenTelemetry to do the trace fields. We should look into using it.

Relevant links:
https://cloud.google.com/endpoints/docs/openapi/tracing
https://cloud.google.com/trace/docs/setup/python-ot

michaelkedar:
In our migration to go we've been adding tracing to all our go logs, so I'm marking this as completed

PR #4819, opened 2026-02-16
feat: instrument error reporting and tracing for go jobs

michaelkedar:
1. Makes error level (and the hypothetical levels above) logs create a new error in the error reporting dashboard, which should hopefully link to the logs
2. adds traces to all our go jobs, which should maybe let us use the trace explorer to see how long things take in a fancy waterfall graph (related to #3547)
  - The exporter will create one big span for the exporter, and sub-spans for each ecosystem
  - The relations will create one big span for the relations, and a sub-span for alias/upstream/related
  - gitter traces each http request that is made to it. We can add propagate the same trace from the worker/importer to correlate requests, which should be fun.

This doesn't affect vulnfeeds (yet), since it uses a duplicate logging thing.
To instrument tracing, you have to propagate the context correctly, and the new `logger.[LEVEL]Context()` commands.

### DIFF
--- a/go/logger/init.go
+++ b/go/logger/init.go
@@ -97,6 +97,7 @@ func initTracing(ctx context.Context, projectID, serviceName string) {
 		res = resource.NewWithAttributes(semconv.SchemaURL, semconv.ServiceNameKey.String(serviceName))
 	}
 
+	// If TRACE_SAMPLE_RATE is unset, default to 5% to prevent unintentional cost and performance impact on high-traffic services.
 	sampleRate := 0.05
 	if r := os.Getenv("TRACE_SAMPLE_RATE"); r != "" {
 		if parsed, err := strconv.ParseFloat(r, 64); err == nil {


go/logger/init.go
```diff
@@ -0,0 +1,153 @@
...
+var (
+	slogLogger  *slog.Logger
+	errorClient *errorreporting.Client
+	once        sync.Once
+	tp          *sdktrace.TracerProvider
+	projectID   string
+)
+
+// InitGlobalLogger initializes the global slog logger and GCP observability clients.
+func InitGlobalLogger(ctx context.Context) {
```
another-rex: This is being ran with once, I don't think it makes sense to ask for context here, especially when all callers right now are using `context.Background()`. This might confused the caller to call this with their specific request context, which we don't want.

another-rex: Also, we can be fancy and use `init()` here, and it'll be automatically initialized by anyone that uses any logger functions. Removing the need to have all those if statements in each function. Or people forgetting to init the logger.

michaelkedar: I removed the context from `InitGlobalLogger`, and changed it slightly.
I don't want to chuck it in `init`, because I don't want to connect to GCP as a side effect of importing. Also, you're supposed to call/defer `logger.Close()` to flush any errors/traces, and I don't want that silently being forgotten.

I removed the lazy initialisation of the GCP logger - now if you log without calling `InitGlobalLogger` it will always use the local logger.

go/logger/init.go
```diff
@@ -100,10 +97,17 @@ func initTracing(ctx context.Context, projectID, serviceName string) {
 		res = resource.NewWithAttributes(semconv.SchemaURL, semconv.ServiceNameKey.String(serviceName))
 	}
 
+	sampleRate := 0.05
```
another-rex, line 101: Please add a comment descrbing why this is 0.05,

another-rex (approved):

PR #4852, opened 2026-02-19
feat: `LOG_LEVEL` env var in go

michaelkedar:
discussions were had

### DIFF
--- a/go/logger/init.go
+++ b/go/logger/init.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"path/filepath"
 	"strconv"
+	"strings"
 	"sync"
 
 	"cloud.google.com/go/errorreporting"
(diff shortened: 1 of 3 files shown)


go/logger/init.go
```diff
@@ -119,8 +120,17 @@ func initTracing(ctx context.Context, projectID, serviceName string) {
...
+		switch strings.ToLower(lvl) {
+		case "debug":
+			level = slog.LevelDebug
+		case "info":
+			level = slog.LevelInfo
+		case "warn":
+			level = slog.LevelWarn
+		case "error":
+			level = slog.LevelError
+		}
```
tobyhawker, line 133: Should there be a default for e.g. a typo

go/logger/localhandler.go
```diff
@@ -26,17 +27,35 @@ var (
...
+		switch strings.ToLower(lvl) {
+		case "debug":
+			level = slog.LevelDebug
+		case "info":
+			level = slog.LevelInfo
+		case "warn":
+			level = slog.LevelWarn
+		case "error":
+			level = slog.LevelError
+		}
```
tobyhawker, line 48: Likewise

go/logger/init.go
michaelkedar: It's set to info before the switch

PR #4853, opened 2026-02-19
refactor(nvd): reduce reliance on Version info struct

jess-lowe:
**Refactoring & Core Logic**
   - Deprecation of `VersionInfo`: Removed the reliance on the VersionInfo struct, transitioning to osvschema.Affected for representing affected ranges throughout the conversion pipeline. PackageInfo continues to use VersionInfo as PyPi/Alpine/Debian continue to use these.
   - Enhanced Range Resolution:
       - GitVersionsToCommits now handles both successful resolutions and captures unresolved ranges. (conversion/common.go)
       - Original version events are now preserved within the database_specific field of OSV ranges.
       - Unresolved ranges are explicitly included in the output's database_specific for better transparency and manual review. (`converter.go`)
   - Improved NVD Converter: Substantially refactored CVEToOSV to leverage the new common utilities, resulting in cleaner and more maintainable code.

### DIFF
--- a/vulnfeeds/git/versions.go
+++ b/vulnfeeds/git/versions.go
@@ -15,6 +15,7 @@
 package git
 
 import (
+	"errors"
 	"fmt"
 	"regexp"
 	"slices"
@@ -83,7 +84,7 @@ func VersionToAffectedCommit(version string, repo string, commitType models.Comm
 // Take an unnormalized version string, the pre-normalized mapping of tags to commits and return a commit hash.
 func VersionToCommit(version string, normalizedTags map[string]NormalizedTag) (string, error) {
 	if version == "" {
-		return "", fmt.Errorf("version cannot be empty")
+		return "", errors.New("version cannot be empty")
 	}
 	// TODO: try unnormalized version first.
 	normalizedVersion, err := NormalizeVersion(version)


vulnfeeds/cves/versions.go
```diff
@@ -796,47 +814,122 @@ func ExtractVersionsFromCPEs(cve models.NVDCVE, validVersions []string, metrics
...
+				}
+				if slices.Contains(v.AffectedVersions, possibleNewAffectedVersion) {
+					// Avoid appending duplicates
+					continue
+				}
+				v.AffectedVersions = append(v.AffectedVersions, possibleNewAffectedVersion)
+			}
 		}
 	}
+	// if !gotVersions {
```
another-rex: Should this be removed?

vulnfeeds/conversion/nvd/converter.go
```diff
@@ -15,107 +15,107 @@ import (
...
+		metrics.VersionSources = append(metrics.VersionSources, models.VersionSourceCPE)
+	} else if len(repos) == 0 {
+		affected := MergeRangesAndCreateAffected(resolvedRanges, unresolvedRanges, nil, nil, metrics)
+		v.Affected = append(v.Affected, affected)
+		outputFiles(v, directory, maybeVendorName, maybeProductName, metrics, rejectFailed, outputMetrics)
 
-	if len(v.Affected) == 0 {
-		metrics.AddNote("No affected ranges detected for %q", maybeProductName)
-		return fmt.Errorf("no affected ranges detected for %q %w", maybeProductName, ErrNoRanges)
+		return models.NoRepos
```
another-rex: Should this be an else if? If this is an early exit I think we should put this above the cpeRanges section to make it clear. We can even repeat the len(cpuRanges) check.

vulnfeeds/conversion/nvd/converter.go
```diff
@@ -15,107 +15,107 @@ import (
...
 	maybeVendorName := "ENOCPE"
 	maybeProductName := "ENOCPE"
 
 	if len(CPEs) > 0 {
-		CPE, err := cves.ParseCPE(CPEs[0]) // For naming the subdirectory used for output.
+		CPE, _ := cves.ParseCPE(CPEs[0]) // For naming the subdirectory used for output.
 		maybeVendorName = CPE.Vendor
 		maybeProductName = CPE.Product
-		if err != nil {
-			return errors.New("can't generate an OSV record without valid CPE data")
```
another-rex, line 39: This change doesn't appear to be in the PR description.

another-rex: Also add a comment in the new code to say why the check is being discarded

vulnfeeds/conversion/nvd/converter.go
another-rex: Also add a comment above each section could help readability (e.g Similar to how you have // Extract Commits just below this.

vulnfeeds/conversion/nvd/converter.go
```diff
@@ -294,3 +294,138 @@ func FindRepos(cve models.NVDCVE, vpRepoCache *cves.VPRepoCache, repoTagsCache *
 
 	return reposForCVE
 }
+
+func MergeRangesAndCreateAffected(resolvedRanges []*osvschema.Range, unresolvedRanges []*osvschema.Range, commits []models.AffectedCommit, successfulRepos []string, metrics *models.ConversionMetrics) *osvschema.Affected {
```
another-rex, line 320: nit: Add docstring, specifically what the args should be, it looks like one puts nil in some args, and another call doesn't.

vulnfeeds/conversion/nvd/converter.go
```diff
@@ -294,3 +294,138 @@ func FindRepos(cve models.NVDCVE, vpRepoCache *cves.VPRepoCache, repoTagsCache *
...
+		}
+	}
+
+	// if there are no resolved version but there are commits, we should create a range for each commit
+	if len(resolvedRanges) == 0 && len(commits) > 0 {
+		for _, commit := range commits {
+			newResolvedRanges = append(newResolvedRanges, conversion.BuildVersionRange(commit.Introduced, commit.LastAffected, commit.Fixed))
+			metrics.ResolvedRangesCount++
+		}
+		metrics.Outcome = models.Successful
```
another-rex: This is a bit weird that this is being set inside a function. I really think all instances of metrics.Outcome should be set in roughly the same place.

vulnfeeds/conversion/nvd/converter.go
```diff
@@ -294,3 +294,138 @@ func FindRepos(cve models.NVDCVE, vpRepoCache *cves.VPRepoCache, repoTagsCache *
...
+	if commit.LastAffected != "" {
+		return &osvschema.Event{
+			LastAffected: commit.LastAffected,
+		}
+	}
+
+	return nil
+}
+
+func outputFiles(v *vulns.Vulnerability, dir string, vendor string, product string, metrics *models.ConversionMetrics, rejectFailed bool, outputMetrics bool) {
```
another-rex, line 448: nit: docstring + arg description please

vulnfeeds/conversion/common.go
```diff
@@ -255,6 +257,20 @@ func GitVersionsToCommits(versionRanges []*osvschema.Range, repos []string, metr
...
+	if version == "" {
+		return ""
+	}
+	commit, err := git.VersionToCommit(version, normalizedTags)
+	if err != nil {
+		return ""
+	}
+
+	return commit
+}
```
another-rex: This function doesn't seem to do that much, do we need it? (also it doesn't look like it's logging either like the docstring says). 

I know it's an existing function, but we probably should still clean it up.

vulnfeeds/conversion/common.go
```diff
@@ -280,16 +296,56 @@ func BuildVersionRange(intro string, lastAff string, fixed string) *osvschema.Ra
 	return &versionRange
 }
 
-// resolveVersionToCommit is a helper to convert a version string to a commit hash.
-// It logs the outcome of the conversion attempt and returns an empty string on failure.
-func resolveVersionToCommit(version string, normalizedTags map[string]git.NormalizedTag) string {
-	if version == "" {
-		return ""
+func MergeTwoRanges(range1, range2 *osvschema.Range) *osvschema.Range {
```
another-rex: Add some examples to the docstring for what this does.

vulnfeeds/conversion/common.go
```diff
@@ -280,16 +296,56 @@ func BuildVersionRange(intro string, lastAff string, fixed string) *osvschema.Ra
...
+	// check if the ranges are the same
+	if range1.GetRepo() != range2.GetRepo() || range1.GetType() != range2.GetType() {
+		return nil
```
another-rex: This should be returning an error rather than just nil

another-rex: Or panicing if you think this will never happen

vulnfeeds/conversion/common.go
```diff
@@ -280,16 +296,56 @@ func BuildVersionRange(intro string, lastAff string, fixed string) *osvschema.Ra
...
+	if db1 != nil {
+		for k, v := range db1.GetFields() {
+			mergedMap[k] = v.AsInterface()
+		}
+	}
+
+	if db2 != nil {
+		for k, v := range db2.GetFields() {
+			if existing, ok := mergedMap[k]; ok {
+				// If both are lists, append them
```
another-rex: Hmm what about 2 maps

vulnfeeds/conversion/common.go
another-rex: MergeTwoRanges made me think that it also merges events by order, but that's kind of difficult.

vulnfeeds/cves/versions.go
```diff
@@ -796,47 +814,122 @@ func ExtractVersionsFromCPEs(cve models.NVDCVE, validVersions []string, metrics
...
 	}
 
-	v.AffectedVersions = ExtractVersionsFromCPEs(cve, validVersions, metrics)
-	if len(v.AffectedVersions) > 0 {
-		metrics.AddNote("Extracted versions from CPEs: %v", v.AffectedVersions)
-	} else {
-		v.AffectedVersions = ExtractVersionsFromText(validVersions, models.EnglishDescription(cve.Descriptions), metrics)
-		if len(v.AffectedVersions) > 0 {
-			metrics.AddNote("Extracted versions from description: %v", v.AffectedVersions)
+	// gotVersions := false
```
another-rex: Remove commented code?

vulnfeeds/cves/versions.go
```diff
@@ -796,47 +814,122 @@ func ExtractVersionsFromCPEs(cve models.NVDCVE, validVersions []string, metrics
...
+	for _, config := range cve.Configurations {
+		for _, node := range config.Nodes {
+			if node.Operator != "OR" {
```
another-rex, line 844: Comment explaining why we are skipping OR?

vulnfeeds/cves/versions.go
```diff
@@ -796,47 +814,122 @@ func ExtractVersionsFromCPEs(cve models.NVDCVE, validVersions []string, metrics
...
 	}
+	if len(versions) > 0 {
+		metrics.AddNote("Extracted versions from CPEs: %v", versions)
+	}
 
 	return versions
 }
 
+// ExtractVersionInfo extracts version information from a CVE.
+// Deprecated: Use ExtractVersions instead.
```
another-rex: hmm... this is a lot of changes and new code for a deprecated function?

another-rex: also I don't see any replacements of ExtractVersionInfo with ExtractVersions

another-rex: I'm not sure this new code is described in the PR description.

jess-lowe: Sorry, this was just the old CPE extraction code moved into the ExtractVersionInfo function as the new CPE extraction uses a different return signature. Not actually new code. ExtractVersions was from a previous commit, and then I changed my mind - updated the docstring

vulnfeeds/conversion/common.go
jess-lowe: Fixed - kept the function because of the nil checks.

vulnfeeds/conversion/nvd/converter.go
jess-lowe: I have updated it so they should all be roughly set in CVEToOSV now (and have made a function to handle the setting of outcome a bit better now)

vulnfeeds/cves/versions.go
jess-lowe: `¯\_(ツ)_/¯` this is from the old CPE extraction code, and I'm not entirely sure why it gets skipped (though I should look into it)

jess-lowe: Yeah it's been there since the dawn of time aka 5 years ago: https://github.com/google/osv.dev/blob/848e687538774591edf1025c76471fccea9c4ccc/vulnfeeds/cves/versions.go#L75

It is probably worth looking into, but not in this PR

vulnfeeds/conversion/common.go
another-rex: how often is this used? I don't think the nil checks is worth a function, we should just inline this.

vulnfeeds/conversion/common.go
```diff
@@ -325,17 +320,16 @@ func MergeTwoRanges(range1, range2 *osvschema.Range) *osvschema.Range {
...
 			if existing, ok := mergedMap[k]; ok {
-				// If both are lists, append them
-				if list1, ok := existing.([]any); ok {
-					if list2, ok := v.AsInterface().([]any); ok {
-						mergedMap[k] = append(list1, list2...)
-						continue
-					}
+				mergedVal, err := mergeDatabaseSpecificValues(existing, val2)
+				if err != nil {
+					return nil, fmt.Errorf("failed to merge database specific key %q: %w", k, err)
```
another-rex: If it can't merge, it probably shouldn't error and fail the entire record.

jess-lowe: fixed

another-rex (approved):

Issue #4849, opened 2026-02-18
Questions about the update cycles


hannesm:
Dear everyone,

thanks for your database. I have some observations about updates thereof:
- I noticed that https://test.osv.dev/vulnerability/OSEC-2026-01 has the modified as 2026-02-17T15:18:05.316779Z, while https://osv.dev/vulnerability/OSEC-2026-01 has 2026-02-17T14:45:51.530795Z
- at the same time, osv.dev has been updating the OSEC source more recently (e.g. https://osv.dev/vulnerability/OSEC-2022-01 has 2026-02-18T09:58:24.266403Z)
- there are no issues reported at the linter https://osv.dev/linter

So, my question is when are sources being updated (how often)? And how comes that one data source (https://github.com/ocaml/security-advisories) is not completely used (see OSEC-2022-01 vs OSEC-2026-01), but only partially?

Are there any publicly available logs that can bring some light into this? I'm mainly asking to figure out whether we (from the OSEC source) can develop some best practises so that osv is happy with our update stream.

michaelkedar:
Hi,
This seems to be a problem on our end, related to #4726.
We *should* be checking for updates every 15 minutes, but it seems some of these are randomly getting dropped.

I'll try forcing a full update of the OSEC source on the production instance to see if it will pick up the missed updates.

I have been looking into the problem, hopefully I can have this resolved soon.

Issue #4726, opened 2026-02-02
Multiple data quality issues with GHSA source


Lexxxzy:
https://osv.dev/vulnerability/GHSA-2hw6-4rv9-82fp
https://osv.dev/vulnerability/GHSA-gggm-66rh-pp98
https://osv.dev/vulnerability/GHSA-vvj3-c3rp-c85p
https://osv.dev/vulnerability/GHSA-vw5p-8cq8-m7mv
https://osv.dev/vulnerability/GHSA-7gcm-g887-7qv7 (already mentioned in #4725)
etc

**Describe the data quality issue observed**
The modified time differs between the JSON data and the import source (https://github.com/github/advisory-database); therefore, the remaining data was not updated, such as:
- the fixed version in GHSA-7gcm-g887-7qv7
- the web link to the fix commit in GHSA-vvj3-c3rp-c85p

Is this intended?

michaelkedar:
I've triggered a manual reimport of these records (and records from this year), and they seem to be fixed now.

We still need to investigate the underlying reason these updates are being missed, though.

lshariprasad:
I will make the change and update it

G-Rath:
@michaelkedar not sure if its the same issue, or if there are other advisories, but fwiw GHSA-33mh-2634-fwr2 doesn't seem to have been updated since it was ingested so its missing the fix version that was added a couple of days ago

michaelkedar:
Triggered a reimport of GHSA-33mh-2634-fwr2, which updated it.

Not enjoying playing whack-a-mole with these missed updates 🙃
The worker logs were complaining about a hash mismatch between the files, so maybe the GHSA repo is not being updated properly in the worker.

another-rex:
Just updating this issue, we now have a gke-workers/*/importer-reconciler.yaml to catch anything that gets dropped by the normal pipeline for whatever reason, and retry importing it. Long term fix is that it will alert us on all dropped updates, so we can patch the issue at the source.

PR #4854, opened 2026-02-19
fix(worker): use force_update on gitter clones of source repos

michaelkedar:
I suspect this may alleviate problems seen in #4726 and #4849


another-rex:
hmm... wouldn't importer always run before worker, so gitter should be at least up to date with the version importer pulled...

PR #4863, opened 2026-02-20
feat(logger): Wrapper to extract and log context

Ly-Joey:
Add a contextHandler to wrap around our slog.Handler so that key-value stored in context will be automatically added to the log as attributes. 

Services will have to explicitly register what context keys they want to show up in logs.
To register a context key to be logged (after initializing the logger):
```
logger.RegisterContextKey(ctxKey, logKey)
```

### DIFF
--- a/go/logger/init.go
+++ b/go/logger/init.go
@@ -20,7 +20,7 @@ import (
 )
 
 var (
-	slogLogger = slog.New(newLocalHandler(os.Stdout))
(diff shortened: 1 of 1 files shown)


go/cmd/gitter/gitter.go
another-rex: Let's make this repoURL as well

another-rex (approved):
Very nice!

PR #4891, opened 2026-02-24
fix: improve navbar search dropdown behavior

ashmod:
The previous navbar search had the dropdown appended as a detached element, making it visually fragile.
For example, zooming in while the search dropdown is on would often break it because of it not being properly attached to the search field:

<div align="center">
image
</div>

Additionally, the styles applied on dropdown made it feel somewhat disconnected (e.g. radius would change on dropdown and the dropdown wouldn't feel seamless with the search field), this stood out particularly in light mode.

This PR refactors the dropdown to mount inline as a child of the search form, to flow naturally below the input.

Before/After:

https://github.com/user-attachments/assets/239896b2-0c20-4cd3-9a35-61148d3f18e7

This PR also fixes a few related issues: event listeners that were never cleaned up, the search bar open animation very briefly flashing the submit arrow (see Before at 0.25x) and applied some cleanup to the search module.


jess-lowe (approved):
LGTM, thanks!

PR #4907, opened 2026-02-25
Prevent major ruby docker updates in renovate.json

another-rex:
Updated `renovate.json` to prevent major updates for the `ruby` docker image.
This is in response to PR https://github.com/google/osv.dev/pull/4894 trying to upgrade ruby from 3 to 4.
The affected file is likely `docs/docs.Dockerfile` which uses `FROM ruby:3`.

---
*PR created automatically by Jules for task [2175686184353370820](https://jules.google.com/task/2175686184353370820) started by @another-rex*


G-Rath:
I assume there is a reason why we're not just upgrading to v4...?

cuixq:
We may need to update `commonmarker` to a newer version so that it's compatible with ruby v4 - maybe another cleanup issue?

G-Rath:
That has marked itself as [compatible w/ Ruby v4](https://github.com/gjtorikian/commonmarker/blob/main/commonmarker.gemspec#L16) - I can look into getting that upgraded across both repos

cuixq:
yeah but not the version that we depend on - feel free to work on this.

PR #4918, opened 2026-02-26
fix(gitter): Better handling for context cancel

Ly-Joey:
Add proper handling for context cancellation in semaphore and our longer running git operations.


michaelkedar:
ah #4917

PR #4925, opened 2026-03-02
fix: Make update channel not use Any because we don't need to

another-rex:
Also moved the loop out becasue it was breaking my brain. This part is in a separate commit to make it easier to review.

Tests still pass, though I might have missed something obvious.


michaelkedar (commented):
Looks good to me (Though the diffs makes looking at reordered functions more difficult than it ought to be)
I don't remember why I used any here anyway 🤷 

> Tests still pass, though I might have missed something obvious.

I don't think this function is covered by the tests so...

PR #4956, opened 2026-03-03
fix(docs): restore callout styling and improve highlight block visibility

ashmod:
A recent just-the-docs update [broke callout block styling](https://google.github.io/osv.dev/post-v1-query/#:~:text=dev/v1/query%22-,The%20API,-has%20a%20response) on the docs pages. 
This PR adds a custom override to restore callouts to their original styles.

Also improved highlight block readability to match [the one applied in osv-scanner docs](https://github.com/google/osv-scanner/pull/2106) a while back.

(Warning and Highlight blocks are not used anywhere in the osv.dev docs yet, but this should help ensure the styles are ready for future updates.)


jess-lowe (approved):
LGTM, thanks :)

PR #4966, opened 2026-03-04
feat: upgrade lipgloss & copy logger package back over to vulnfeeds

michaelkedar:
upgrading lipgloss to v2, fixing issues with #4895

Also copied over the changes to the logger made in `go/` back over to `vulnfeeds/`, which means error logs would start showing up in error reporting.
(I didn't instrument traces/spans, because that's too involved for me)


jess-lowe (commented):
🥂

PR #4967, opened 2026-03-04
fix: protojson non-determinism in ID generation

michaelkedar:
My tests in #4966 were failing because `protojson` decided it wanted to add two spaces between the keys and values in its output.
Run the `protojson` output through the `json.Indent` to format it consistently.

That said, does the formatting of the output actually matter here? Should we instead rewrite the tests to do a semantic comparison instead of a string comparison?

### DIFF
--- a/vulnfeeds/cmd/ids/main.go
+++ b/vulnfeeds/cmd/ids/main.go
@@ -2,8 +2,10 @@
 package main
 
 import (
+	"bytes"
 	"crypto/rand"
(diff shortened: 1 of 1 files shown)


cuixq:
I actually thought about doing the semantic comparison but decided to do string comparison considering the use case is to write the updated vulnerability file. However, maybe the actual ordering of each fields does not matter too much as well as whether the diffs are minimized since the file name is going to be updated too.

PR #4971, opened 2026-03-05
chore: for safety assurance - pull_request instead of target

jess-lowe:


another-rex (approved):

.github/workflows/title.yml
```diff
@@ -16,7 +16,7 @@ name: Title
 
 on:
   # `pull_request_target` is only required when editing PRs from forks.
```
cuixq, line 18: the comment needs to be updated?

jess-lowe: Ironically it was there during the last time i changed it (i think its informative / educational?)

PR #4974, opened 2026-03-05
feat(gitter): affected commits graph walking

Ly-Joey:
Add graph related logic and find affected commits to gitter

Definition of affected commits:
> A `commit` is affected when: from at least one `introduced` that is an ancestor of the `commit`, there is no path between them that passes through a fix. A fix can either be a `fixed` commit, or the children of a `lastAffected` commit.

### Key changes
New endpoint: 
- `POST /affected-commits`: 
  - Accepts a list of events (each containing an eventType (introduced, fixed, lastAffected, or limit) and commit hash) 
  - Returns a list of affected commits and refs (tags and branches)

Cherry-pick Detection (optional): 
Expand introduced, fixed, limit commits to include their cherry-picked equivalents (determined with PatchID)


go/cmd/gitter/repository_test.go
```diff
@@ -127,7 +134,612 @@ func TestLoadRepository(t *testing.T) {
...
+			t.Errorf("patch ID mismatch for commit %s", printSHA1(hash))
+		}
+	}
+}
+
+// Helper to decode string into SHA1
+func decodeSHA1(s string) SHA1 {
+	var hash SHA1
+	// Pad with zeros because the test strings are shorter than 40 char
+	padded := fmt.Sprintf("%040s", s)
```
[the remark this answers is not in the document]
Ly-Joey: "%040s" does work. But I'm still changing it because string.Repeat() is clearer.

go/cmd/gitter/gitter.go
```diff
@@ -448,6 +468,15 @@ func cacheHandler(w http.ResponseWriter, r *http.Request) {
 	defer r.Body.Close()
 
 	url := body.URL
+	// If request came from a local ip, don't do the check
+	if !isLocalRequest(r) {
```
another-rex: Expand the comment to add why we want to skip the check. I think I originally added this so that we can clone local git directories for testing.

go/cmd/gitter/gitter.go
```diff
@@ -499,3 +528,136 @@ func cacheHandler(w http.ResponseWriter, r *http.Request) {
...
+func affectedCommitsHandler(w http.ResponseWriter, r *http.Request) {
+	start := time.Now()
+	// POST requets body processing
+	var body struct {
+		URL                         string  `json:"url"`
+		Events                      []Event `json:"events"`
+		DetectCherrypicksIntroduced bool    `json:"detect_cherrypicks_introduced"`
+		DetectCherrypicksFixed      bool    `json:"detect_cherrypicks_fixed"`
+		ForceUpdate                 bool    `json:"force_update"`
+	}
```
another-rex: This should be defined in the proto right?

go/cmd/gitter/gitter.go
```diff
@@ -499,3 +528,136 @@ func cacheHandler(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		http.Error(w, fmt.Sprintf("Error decoding JSON: %v", err), http.StatusBadRequest)
+
+		return
+	}
+	defer r.Body.Close()
+
+	url := body.URL
+	// If request came from a local ip, don't do the check
+	if !isLocalRequest(r) {
```
another-rex: Same as above

go/cmd/gitter/gitter.go
```diff
@@ -499,3 +528,136 @@ func cacheHandler(w http.ResponseWriter, r *http.Request) {
...
+
+	// Limit and fixed/last_affected shouldn't exist in the same request as it doesn't make sense
+	if (len(fixed) > 0 || len(lastAffected) > 0) && len(limit) > 0 {
+		http.Error(w, "Limit and fixed/last_affected shouldn't exist in the same request", http.StatusBadRequest)
+
+		return
+	}
+
+	// Fetch repo if it's not fresh
+	// I can't change singleflight's interface
```
another-rex: is nolint needed here? If not then you can remove the comment entirely.

go/cmd/gitter/gitter.go
```diff
@@ -499,3 +528,136 @@ func cacheHandler(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	repoDirName := getRepoDirName(url)
+	repoPath := filepath.Join(gitStorePath, repoDirName)
+
+	repoLock := GetRepoLock(url)
+	repoLock.RLock()
+	defer repoLock.RUnlock()
+
+	// I can't change singleflight's interface
```
another-rex: Remove comment?

go/cmd/gitter/gitter.go
```diff
@@ -499,3 +528,136 @@ func cacheHandler(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+	repo := repoAny.(*Repository)
+
+	var affectedCommits []*Commit
+	if len(limit) > 0 {
+		affectedCommits = repo.Limit(ctx, introduced, limit)
+	} else {
+		affectedCommits = repo.Affected(ctx, introduced, fixed, lastAffected, cherrypickIntro, cherrypickFixed)
+	}
```
another-rex: This should error if both limit and no limit is requested at the same time

go/cmd/gitter/gitter.go
```diff
@@ -57,6 +75,8 @@ var (
 	semaphore       chan struct{} // Request concurrency control
 )
 
+var validURLRegex = regexp.MustCompile(`^(https?|git|ssh)://`)
```
michaelkedar: (I know this isn't new, but) do we actually support ssh? I don't think the gitter on GKE has any credentials to use with it

go/cmd/gitter/gitter.go
```diff
@@ -366,7 +386,7 @@ func gitHandler(w http.ResponseWriter, r *http.Request) {
 	// If request came from a local ip, don't do the check
 	if !isLocalRequest(r) {
 		// Check if url starts with protocols: http(s)://, git://, ssh://, (s)ftp://
-		if match, _ := regexp.MatchString("^(https?|git|ssh)://", url); !match {
+		if !validURLRegex.MatchString(url) {
```
michaelkedar: nit: the comment says it supports ftp, but the regex does not (+ same on the comments in other endpoints)

go/cmd/gitter/gitter.go
```diff
@@ -499,3 +528,136 @@ func cacheHandler(w http.ResponseWriter, r *http.Request) {
 	w.WriteHeader(http.StatusOK)
 	logger.InfoContext(ctx, "Request completed successfully: /cache", slog.Duration("duration", time.Since(start)))
 }
+
+func affectedCommitsHandler(w http.ResponseWriter, r *http.Request) {
```
another-rex, line 634: This functions seems quite long while only containing glue code. I think we can restructure this a bit.

First we should put these 4 things into a struct:
```
type SeperatedEvents struct {
	introduced := []string{}
	fixed := []string{}
	lastAffected := []string{}
	limit := []string{}
```

and have repo.Limit and repo.Affected just take in that struct. 

This way we can have a single function that turns Events into SeperatedEvents.

That function should probably also error if both limit and last affected / fixed are used at the same time.

go/cmd/gitter/gitter.go
```diff
@@ -499,3 +528,136 @@ func cacheHandler(w http.ResponseWriter, r *http.Request) {
...
+		logger.ErrorContext(ctx, "Error fetching blob", slog.Any("error", err))
+		if isAuthError(err) {
+			http.Error(w, fmt.Sprintf("Error fetching blob: %v", err), http.StatusForbidden)
+
+			return
+		}
+		http.Error(w, fmt.Sprintf("Error fetching blob: %v", err), http.StatusInternalServerError)
+
+		return
+	}
```
another-rex: Place this in a helper function, as I think it'll be reused for most endpoints.

go/cmd/gitter/gitter.go
another-rex: Please make this check a helper function, as it's now used 3 times I think.

tobyhawker:
I'm not sure that the definition of "affected" is quite right:

>> A commit is affected when: from at least one introduced there is no path from the commit through a fix to the introduced

As phrased, I don't think it's made explicit that only commits where the introduced is an ancestor are affected. How does something like this sound (just a suggestion, please feel free to adjust)?

> A `commit` is affected when: from at least one `introduced` that is an ancestor of the `commit`, there is no path between them that passes through a `fixed`.

go/cmd/gitter/gitter.go
```diff
@@ -499,3 +528,136 @@ func cacheHandler(w http.ResponseWriter, r *http.Request) {
...
+		default:
+			logger.ErrorContext(ctx, "Invalid event type", slog.String("event_type", string(event.Type)))
+			http.Error(w, fmt.Sprintf("Invalid event type: %s", event.Type), http.StatusBadRequest)
+
+			return
+		}
+	}
+	logger.InfoContext(ctx, "Received request: /affected-commits", slog.Any("introduced", introduced), slog.Any("fixed", fixed), slog.Any("last_affected", lastAffected), slog.Any("limit", limit), slog.Bool("cherrypickIntro", cherrypickIntro), slog.Bool("cherrypickFixed", cherrypickFixed))
+
+	semaphore <- struct{}{}
```
another-rex: hmm originally I think this semaphore was meant to be inside the singleflight.

go/cmd/gitter/gitter.go
```diff
@@ -499,3 +528,136 @@ func cacheHandler(w http.ResponseWriter, r *http.Request) {
...
+
+	repoDirName := getRepoDirName(url)
+	repoPath := filepath.Join(gitStorePath, repoDirName)
+
+	repoLock := GetRepoLock(url)
+	repoLock.RLock()
+	defer repoLock.RUnlock()
+
+	// I can't change singleflight's interface
+	repoAny, err, _ := gLoad.Do(repoPath, func() (any, error) {
```
michaelkedar: not in this PR, but if we're using singleflight heavily, and the `any` typing is clumsy, I reckon we could make a small wrapper with generics
```golang
type Group[T any] struct {
  delegate singleflight.Group
}

func (g *Group[T]) Do(key string, fn func() (T, error)) (T, error, bool) { // or whatever's preferable
  // call delegate.Do with wrapped functions and do the casting here
}
```

go/cmd/gitter/gitter.go
```diff
@@ -499,3 +528,136 @@ func cacheHandler(w http.ResponseWriter, r *http.Request) {
...
+
+		return
+	}
+
+	repoDirName := getRepoDirName(url)
+	repoPath := filepath.Join(gitStorePath, repoDirName)
+
+	repoLock := GetRepoLock(url)
+	repoLock.RLock()
+	defer repoLock.RUnlock()
```
another-rex: Ideally I think the repo lock can be attached to the repo itself, so you can't access the underlying repo without locking. Can we do something like this:

Make fetchRepo return a repo struct, which contains:
- path to repo
- *Repository (nil by default) and private
- * to the repo mutex in the global map

then Load can be a function on the repo struct, which will do the locking before loading the repository. Just an idea at the moment, you probably can play around with the types a bit to improve it.

go/cmd/gitter/repository.go
```diff
@@ -395,3 +413,220 @@ func (r *Repository) updatePatchID(commitHash, patchID SHA1) {
 
 	r.patchIDToCommits[patchID] = append(r.patchIDToCommits[patchID], commitHash)
 }
+
+// parseHashes converts slice of string hashes input into slice of SHA1
+func (r *Repository) parseHashes(ctx context.Context, hashesStr []string, isIntroduced bool) []SHA1 {
+	hashes := make([]SHA1, 0, len(hashesStr))
+	for _, hash := range hashesStr {
+		if isIntroduced && hash == "0" {
```
michaelkedar: nit: IMO this `isIntroduced` flag is unnecessary.

go/cmd/gitter/repository.go
```diff
@@ -158,16 +163,24 @@ func (r *Repository) buildCommitGraph(ctx context.Context, cache *pb.RepositoryC
...
+					continue
+				}
+				// Remove prefixes from tags, other refs such as branches will be left as is
+				if strings.HasPrefix(ref, "tag: ") {
+					tag := strings.TrimPrefix(ref, "tag: ")
+					refs = append(refs, tag)
+				} else {
+					// clean up HEAD -> branch-name to just keep the branch name
+					ref = strings.TrimPrefix(ref, "HEAD -> ")
+					refs = append(refs, ref)
```
another-rex: nit: I think this can just be sequential

strings.TrimPrefix(ref, "tag: ")
strings.TrimPrefix(ref, "HEAD -> ")
append

go/cmd/gitter/repository.go
```diff
@@ -395,3 +413,220 @@ func (r *Repository) updatePatchID(commitHash, patchID SHA1) {
...
+	introduced := r.parseHashes(ctx, introStrs, true)
+	fixed := r.parseHashes(ctx, fixedStrs, false)
+	lastAffected := r.parseHashes(ctx, laStrs, false)
+
+	// Expands the introduced and fixed commits to include cherrypick equivalents
+	// lastAffected should not be expanded because it does not imply a "fix" commit that can be cherrypicked to other branches
+	if cherrypickIntro {
+		introduced = r.expandByCherrypick(introduced)
+	}
+	if cherrypickFixed {
```
another-rex, line 513: Should be be cherrypicking last-affected +1 commits?

go/cmd/gitter/repository.go
```diff
@@ -395,3 +413,220 @@ func (r *Repository) updatePatchID(commitHash, patchID SHA1) {
...
+	for _, commit := range fixed {
+		fixedMap[commit] = struct{}{}
+	}
+
+	for _, commit := range lastAffected {
+		if _, ok := r.commitGraph[commit]; ok {
+			for _, child := range r.commitGraph[commit] {
+				fixedMap[child] = struct{}{}
+			}
+		}
```
another-rex: Is this if statement necessary? i.e. would a for loop on a non existent map entry be fine?

Either way I think we can remove the if indent, 
```
children, _ := r.commitGraph[commit]
range children {
```

range nil I believe is fine.

another-rex (changes requested):
Nice!

go/cmd/gitter/repository.go
```diff
@@ -395,3 +413,220 @@ func (r *Repository) updatePatchID(commitHash, patchID SHA1) {
...
+				// Inline DFS from current (fixed) node to make all descendants as unaffected / unaffectable
+				// 1. If a previous path added the descendant to affected list, remove it
+				// 2. Add to the unaffectable set to block future paths
+				stack := []SHA1{curr}
+				for len(stack) > 0 {
+					unaffected := stack[len(stack)-1]
+					stack = stack[:len(stack)-1]
+
+					// Remove from affected list if it was reached via a previous non-fixed path.
+					delete(affectedFromIntro, unaffected)
```
michaelkedar: Do you not have to check if any of the commits here were `introduced` commits?

e.g. if it's like
```
I1 -> F1 -> I2 -> F2
```
If you do `I2` first, then `I1` would all the commits after `F1` become unaffected?

go/cmd/gitter/repository.go
```diff
@@ -395,3 +413,220 @@ func (r *Repository) updatePatchID(commitHash, patchID SHA1) {
...
+		}
+
+		hashes = append(hashes, SHA1(hashBytes))
+	}
+
+	return hashes
+}
+
+// Affected returns a list of commits that are affected by the given introduced, fixed and last_affected events
+func (r *Repository) Affected(ctx context.Context, introStrs, fixedStrs, laStrs []string, cherrypickIntro, cherrypickFixed bool) []*Commit {
```
michaelkedar: It might be useful to move some of these structs/functions to a `go/internal/gitter`, so that other binaries may reuse the logic if they want to bypass the gitter service (e.g. for a datafix-type thing).

go/cmd/gitter/repository.go
```diff
@@ -395,3 +413,220 @@ func (r *Repository) updatePatchID(commitHash, patchID SHA1) {
...
+		}
+	}
+
+	keys := slices.Collect(maps.Keys(unique))
+
+	return keys
+}
+
+// Between walks and returns the commits that are strictly between introduced (inclusive) and limit (exclusive)
+func (r *Repository) Limit(ctx context.Context, introStrs, limitStrs []string) []*Commit {
```
michaelkedar: do we want cherrypick detection here?

go/cmd/gitter/repository.go
```diff
@@ -395,3 +413,220 @@ func (r *Repository) updatePatchID(commitHash, patchID SHA1) {
...
+			continue
+		}
+
+		hashes = append(hashes, SHA1(hashBytes))
+	}
+
+	return hashes
+}
+
+// Affected returns a list of commits that are affected by the given introduced, fixed and last_affected events
```
tobyhawker, line 497: Is it worth restating our definition of "affected" here in the code, where we're more likely to see it (and so help the reader understand why the code does what it does)?

Another point that might make sense to recap is just before the loop over the "introduced" commits.

go/cmd/gitter/repository_test.go
```diff
@@ -127,7 +133,612 @@ func TestLoadRepository(t *testing.T) {
...
+
+			if diff := cmp.Diff(tt.expected, got, cmpSHA1Opts...); diff != "" {
+				t.Errorf("expandByCherrypick() mismatch (-want +got):\n%s", diff)
+			}
+		})
+	}
+}
+
+// Testing cases with introduced and fixed only.
+func TestAffected_Introduced_Fixed(t *testing.T) {
```
michaelkedar, line 248: Can we add some tests for the edge cases when introduced, fixed and/or last_affected are equal to each other.
(and I suspect the logic may need to be adjusted to account for these)

go/cmd/gitter/repository.go
```diff
@@ -395,3 +413,220 @@ func (r *Repository) updatePatchID(commitHash, patchID SHA1) {
...
+				// 2. Add to the unaffectable set to block future paths
+				stack := []SHA1{curr}
+				for len(stack) > 0 {
+					unaffected := stack[len(stack)-1]
+					stack = stack[:len(stack)-1]
+
+					// Remove from affected list if it was reached via a previous non-fixed path.
+					delete(affectedFromIntro, unaffected)
+
+					if children, ok := r.commitGraph[unaffected]; ok {
```
another-rex: same as above, probably no need for a if statement before the for loop

go/cmd/gitter/repository.go
```diff
@@ -395,3 +413,220 @@ func (r *Repository) updatePatchID(commitHash, patchID SHA1) {
...
+		}
+
+		// Add equivalent commits with the same Patch ID (including the current commit)
+		equivalents := r.patchIDToCommits[details.PatchID]
+		for _, eq := range equivalents {
+			unique[eq] = struct{}{}
+		}
+	}
+
+	keys := slices.Collect(maps.Keys(unique))
```
another-rex: Does this need to do some sorting for tests to pass? Or is it fine.

go/cmd/gitter/pb/repository/repository.proto
```diff
@@ -14,3 +14,12 @@ message CommitDetail {
...
 }
+
+message AffectedCommit {
+    bytes hash = 1;
+    repeated string refs = 2;
+}
+
+message AffectedCommitsResponse {
+    repeated AffectedCommit commits = 1;
+}
```
michaelkedar, line 30: we probably want to move refs to the top level response (as a mapping of ref -> hash)

go/cmd/gitter/repository_test.go
michaelkedar: Also tests for introduced = 0

go/cmd/gitter/repository.go
another-rex: I think that's fine, since affectedFromIntro is per loop right? Otherwise we'll have problems from fixes and introduced being in parallel as well.

go/cmd/gitter/repository.go
Ly-Joey: I used cmpopts.SortSlices() in the test file to make sure tests pass deterministically.

go/cmd/gitter/gitter.go
Ly-Joey: It is now in the function that separates Events into SeparatedEvents

go/cmd/gitter/repository.go
Ly-Joey: I’m hesitant to do cherrypicking for lastAffected because it doesn't guarantee a single commit that contains the code fix like a `fixed` commit does. 
Especially if the lastAffected commit is derived from a version tag, then all we can be sure is that a fix commit exists in (current version, next version]. 
We can always add this logic later if we have more stats that support it.

deployment/clouddeploy/gke-workers/base/gitter.yaml
```diff
@@ -29,6 +29,8 @@ spec:
           - "--port=8888"
           - "--work-dir=/work/gitter"
           - "--fetch-timeout=1h"
+          - "--repo-cache-ttl=1h"
+          - "--repo-cache-max-cost=107374182400" # 100GB
```
another-rex: GiB? or GB

another-rex: Also, can we have this take strings and use something like https://github.com/dustin/go-humanize to get the bytes out of "100GiB" string

go/cmd/gitter/gitter.go
```diff
@@ -73,12 +61,47 @@ var (
 	gitStorePath    = filepath.Join(defaultGitterWorkDir, gitStoreFileName)
 	fetchTimeout    time.Duration
 	semaphore       chan struct{} // Request concurrency control
+	// LRU cache for recently loaded repositories (key: repo URL)
+	repoCache        *ristretto.Cache[string, *Repository]
+	repoTTL          time.Duration
+	repoCacheMaxCost int64
```
another-rex: Give the unit this variable is recording in, is it bytes or number of commits or something else?

go/cmd/gitter/gitter.go
```diff
@@ -92,10 +115,55 @@ func GetRepoLock(url string) *sync.RWMutex {
 	return lock.(*sync.RWMutex)
 }
 
-// runCmd executes a command with context cancellation handled by sending SIGINT.
-// It logs cancellation errors separately as requested.
-func runCmd(ctx context.Context, dir string, env []string, name string, args ...string) error {
-	logger.DebugContext(ctx, "Running command", slog.String("cmd", name), slog.Any("args", args))
+// repoCost is the cost function for a repository in the LRU cache.
+// The memory cost of a repository is approximated from the num of commits and a base overhead.
+func repoCost(repo *Repository) int64 {
```
another-rex: This is so clear! Really nice.

nit: rename to repoCostBytes

go/cmd/gitter/gitter.go
```diff
@@ -92,10 +115,55 @@ func GetRepoLock(url string) *sync.RWMutex {
...
+	//   = 20 + 24 + 8 ~= 64 bytes
+	// TOTAL: 264 bytes -> We round up to 300 for some buffer
+	costPerCommit := 300
+
+	return int64(repoOverhead + len(repo.commits)*costPerCommit)
+}
+
+// InitRepoCache initializes the LRU cache for repositories.
+func InitRepoCache() {
+	numCounters := repoCacheMaxCost / (300 * 10000)
```
another-rex: Can you add a comment to where these numbers are coming from? Ideally make some of them const so the variable name can explain it.

go/cmd/gitter/repository.go
```diff
@@ -407,19 +420,28 @@ func (r *Repository) updatePatchID(commitHash, patchID SHA1) {
 	r.patchIDMu.Lock()
 	defer r.patchIDMu.Unlock()
 
-	commit := r.commitDetails[commitHash]
+	idx, ok := r.hashToIndex[commitHash]
+	if !ok {
```
another-rex, line 425: What are the situations this will happen? Can you add a comment about this.

go/cmd/gitter/repository_test.go
Ly-Joey: Done!
Current behaviour at edge cases:
introduced = fixed => commit not affected
introduced = lastAffected => only introduced is affected
fixed = lastAffected => lastAffected is effectively ignored

go/cmd/gitter/gitter.go
Ly-Joey: Removed~

go/cmd/gitter/gitter.go
Ly-Joey: I'm tempted to do this in a separate refactoring PR, to do this and move some of the logic to `go/internal/gitter` (suggested by Michael below).

go/cmd/gitter/repository.go
Ly-Joey: Will do it in a separate PR~

go/cmd/gitter/gitter.go
```diff
@@ -64,10 +66,10 @@ var (
 	// LRU cache for recently loaded repositories (key: repo URL)
 	repoCache        *ristretto.Cache[string, *Repository]
 	repoTTL          time.Duration
-	repoCacheMaxCost int64
+	repoCacheMaxCost int64 // max cost in bytes
```
another-rex: Just make this repoCacheMaxCostBytes rather than comment

PR #4975, opened 2026-03-05
fix(vulnfeeds): unmarshalling error in override upload + tests

jess-lowe:
Thanks to the new logger errors, we discovered that the override upload is not using the right marshaller. 

This PR fixes that, and adds some tests for mocking uploads to GCS.


vulnfeeds/upload/cveworker_test.go
```diff
@@ -0,0 +1,321 @@
...
+	"crypto/sha256"
+	"encoding/hex"
+	"encoding/json"
+	"io"
+	"os"
+	"path"
+	"testing"
+	"time"
+
+	"github.com/fsouza/fake-gcs-server/fakestorage"
```
michaelkedar, line 15: It's too much work to do in this PR, but eventually we probably want to wrap all the GCS calls in an interface we can mock for the tests, instead of relying on this module.

PR #4991, opened 2026-03-10
feat: A reconciler to actually keep our records in check

another-rex:
Wrote a reconciler so that we catch any missing records within 30 hours of it happening.

This acts like a last resort to help us identify leaks in our ingestion pipeline, and ideally should not be actually doing any reimporting.

TODO: We might need to keep a list in the datastore of known bad entries so we avoid erroring on them constantly.

### DIFF
--- a/go/internal/importer/importer.go
+++ b/go/internal/importer/importer.go
@@ -174,8 +174,6 @@ var reconcileSkipSources = []string{
 	"root",
 	// cve-osv is in a state of flux at the moment, let's wait a bit
 	"cve-osv",
-	// ignore chainguard for now as well because their record is too big
-	"chainguard",
 }
 
 // ReconcileLeniencyDuration specifies how long of a time difference before it would be considered an outdated record
@@ -498,6 +496,10 @@ func processUpdate(ctx context.Context, config Config, item WorkItem) {
 
 		return
 	}
+
+	// Hash BEFORE editing the raw data from what is read from disk
+	hash := computeHash(data)
+
 	// protojson does not like invalid UTF-8,
 	// so replace invalid UTF-8 with U+FFFD (replacement character)
 	data = bytes.ToValidUTF8(data, []byte("\uFFFD"))
@@ -600,7 +602,6 @@ func processUpdate(ctx context.Context, config Config, item WorkItem) {
 			}
 		}
 	}
-	hash := computeHash(data)
 	if err := sendToWorker(ctx, config, item, hash, modified, &vulnProto); err != nil {
 		logger.ErrorContext(ctx, "Failed to send to worker",
 			slog.Any("error", err),


go/internal/importer/git.go
```diff
@@ -238,3 +239,110 @@ func changedFiles(ctx context.Context, repo sharedRepo, sourceRepo *models.Sourc
...
+			return nil, err
+		}
+
+		if sourceRepo.Git.Branch != "" {
+			wt, err := repo.Worktree()
+			if err == nil {
+				_ = wt.Checkout(&git.CheckoutOptions{
+					Branch: plumbing.NewRemoteReferenceName("origin", sourceRepo.Git.Branch),
+					Force:  true,
+				})
```
michaelkedar, line 288: Have you verified that this actually checks it out on-disk?
I wouldn't be surprised if this was in-memory only.

go/internal/importer/importer.go
```diff
@@ -152,6 +154,237 @@ func RunDeletions(ctx context.Context, config Config) error {
...
+// reconcileSkipSources is a list of source names to skip during reconciliation.
+var reconcileSkipSources = []string{
+	// Upstream records are completely deprecated and archived now.
+	"uvi",
+	// Currently an upstream issue that we can't fix.
+	"almalinux-alsa",
+	
+	// "oss-fuzz",
+	// "debian-cve",
+	// "cve-osv",
```
michaelkedar: why are these commented out.

Also FYI, "oss-fuzz" is explicitly filtered out of the `SourceRepositoryStore` implementation, so it shouldn't even be surfaced here.

go/internal/importer/importer.go
```diff
@@ -152,6 +155,247 @@ func RunDeletions(ctx context.Context, config Config) error {
...
+		if err != nil {
+			return err
+		}
+
+		// Sanity check for multiple sources with same git url but different branches.
+		// This is not currently supported.
+		if sourceRepo.Type == models.SourceRepositoryTypeGit && sourceRepo.Git != nil {
+			if existingBranch, ok := gitBranches[sourceRepo.Git.URL]; ok {
+				if existingBranch != sourceRepo.Git.Branch {
+					return fmt.Errorf("multiple sources with same git url %q but different branches (%q and %q) is not supported",
```
michaelkedar, line 209: Should this stop the entire reconciler?
If not, we could just combine this loop with the one afterwards, and not store the `sourceRepos`

deployment/clouddeploy/gke-workers/base/importer-reconciler.yaml
michaelkedar, line 1: You have this in base, but no patch in the production environment, meaning it'll get deployed to prod without the PersistentVolume?
Either add the prod one (if you want it), or remove this from base, and add it as a resource (instead of a patch) to the test instance

go/internal/importer/importer.go
```diff
@@ -152,6 +155,247 @@ func RunDeletions(ctx context.Context, config Config) error {
...
+// Arguments:
+//   - ctx: Processing context.
+//   - ch: Channel to push the WorkItem to for re-importing.
+//   - sourceRepo: The configuration of the repository being imported.
+//   - dbRecords: A map containing the datastore 'modified_raw' timestamps for this source.
+//   - path: The relative path or ID representing this specific vulnerability record upstream.
+//   - parsed: A pointer to a pre-parsed gjson.Result if already available. If nil, sourceRecord is opened and parsed dynamically.
+//   - sourceRecord: The interface used to lazily retrieve file contents (from a bucket, git, or REST).
+//   - format: RecordFormat representing the format we expect to decode (JSON, YAML, etc.).
+func checkReconcile(
```
michaelkedar, line 298: This blocks the `SourceRepository` walking to download the record from GCS/REST (or open the file when in git), then sends it to the importerWorker which downloads the file again to do validation/send to worker.

IMO most of this function should be in the pool of workers for the reconciler.

go/internal/importer/importer.go
another-rex: Yes, this should stop the entire reconciler, the concurrency model only holds if all ecosystems on the same repo only does read operations. Otherwise one ecosystem would change the branch while the other ecosystem is still walking the fs.

go/internal/importer/importer.go
another-rex: Ok, reworked now to actually use workers effectively, also kinda refactored importer a little bit, turns out changes are smaller than I expected.

go/internal/importer/git.go
another-rex: Just double checked ,it does checkout on disk.

deployment/clouddeploy/gke-workers/base/importer-reconciler.yaml
another-rex: Now fixed.

go/internal/importer/importer.go
another-rex: Fixed.

go/internal/importer/importer.go
michaelkedar: I guess I meant we could still try reconcile the other source repositories that don't use this same git URL (and also try reconcile the first one encountered).

go/internal/importer/importer.go
```diff
@@ -154,7 +159,214 @@ func RunDeletions(ctx context.Context, config Config) error {
...
+		ch <- recordTemplate
+
+		return
+	}
+
+	// If we already have a parsed result, use it directly
+	if parsed != nil {
+		modified, err := getModifiedTime(parsed)
+		if err != nil {
+			logger.WarnContext(ctx, "Failed to get modified time in reconcile",
```
michaelkedar, line 339: should this be more than a warning?

go/internal/importer/importer.go
```diff
@@ -165,6 +377,14 @@ const (
 	RecordFormatYAML
 )
 
+type Action int
+
+const (
+	Import Action = iota
+	Reconcile
+	Withdraw
```
michaelkedar: should these be named `ActionImport`, etc?

go/internal/importer/importer.go
```diff
@@ -181,51 +401,43 @@ type WorkItem struct {
...
-							slog.Any("error", err),
-							slog.String("source", item.SourceRepository),
-							slog.String("path", item.SourcePath))
-
-						return
-					}
-					logger.InfoContext(ctx, "Sent deletion to worker",
+func importerWorker(_ context.Context, ch <-chan WorkItem, config Config) {
+	// We keep reading from the channel (even if context is closed)
+	// as the channel will be closed by the producer side.
```
michaelkedar: Hmm the explanation here is a bit spurious.
Closing a channel does not discard unpulled messages.
IIRC, what's actually happening is the senders block on the channel being read, and presumably the senders will exit early when the context is cancelled.

go/internal/importer/importer.go
```diff
@@ -154,7 +159,214 @@ func RunDeletions(ctx context.Context, config Config) error {
...
+		SourcePath:       path,
+		Format:           format,
+		KeyPath:          sourceRepo.KeyPath,
+		Strict:           sourceRepo.Strictness,
+		IsReimport:       true,
+	}
+
+	dbMod, exists := dbRecords[path]
+	if !exists {
+		logger.ErrorContext(ctx, "Found missing vulnerability in database during reconcile",
```
michaelkedar: I'm a bit ambivalent on using Error logs here (and for the `Found outdated vulnerability` logs).
On one hand, this is not an error - this is what we want the reconciler to do.
On the other, it's probably good to alert us when our database is wrong.

go/internal/importer/importer.go
another-rex: Yes

go/internal/importer/importer.go
another-rex: I think the warning is correct, we can't guarantee upstream records to be correct, so if they have issues, they I think we need other mechanisms like linter to fix it.

go/internal/importer/importer.go
another-rex: Yeah, for now I made this warning, we can switch to error once upstream errors are lower.

go/internal/importer/importer.go
another-rex: I think I just want it to completely break if that assumption is broken so we can go and fix it.

PR #4992, opened 2026-03-10
fix: Optimize OSS-Fuzz export by querying public bugs directly in Datastore

jess-lowe:
Modified the Datastore query in gcp/workers/oss_fuzz_importer/importer.py to filter by osv.Bug.public == True directly on the server side, rather than fetching all bugs for the OSS-Fuzz ecosystem and discarding non-public ones in memory.

The previous code fetched potentially thousands of osv.Bug entities from Google Cloud Datastore across the network, instantiated full Python objects for each one, and then immediately dropped the non-public ones. Pushing this filter down to the database avoids unnecessary network transfer, memory allocation, and CPU cycles spent on deserialization.

PR created automatically by Jules for task [4719671399546247123](https://jules.google.com/task/4719671399546247123) started by @jess-lowe

### DIFF
--- a/gcp/workers/oss_fuzz_importer/importer.py
+++ b/gcp/workers/oss_fuzz_importer/importer.py
@@ -1106,7 +1106,7 @@ class Importer:
     with concurrent.futures.ThreadPoolExecutor(
         max_workers=_EXPORT_WORKERS) as executor:
       for bug in osv.Bug.query(osv.Bug.ecosystem == 'OSS-Fuzz',
-                               osv.Bug.public == True):
+                               osv.Bug.public == True):  # pylint: disable=singleton-comparison
         _, source_id = osv.parse_source_id(bug.source_id)
         executor.submit(export_oss_fuzz, bug.to_vulnerability(), source_id,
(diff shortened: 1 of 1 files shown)


gcp/workers/oss_fuzz_importer/importer.py
```diff
@@ -1105,10 +1105,8 @@ def export_oss_fuzz(vulnerability, testcase_id, issue_id):
 
     with concurrent.futures.ThreadPoolExecutor(
         max_workers=_EXPORT_WORKERS) as executor:
-      for bug in osv.Bug.query(osv.Bug.ecosystem == 'OSS-Fuzz'):
-        if not bug.public:
-          continue
-
+      for bug in osv.Bug.query(osv.Bug.ecosystem == 'OSS-Fuzz', osv.Bug.public \
+                               is True):
```
michaelkedar: I actually think this needs to be `osv.Bug.public == True` (with a `# pylint: disable=singleton-comparison`) because I'm pretty sure ndb overloads the `__eq__` function to make this filter work.

PR #4993, opened 2026-03-10
fix: importer issues for bad records

michaelkedar:
- Pub/Sub has a data limit of 10MB, and a couple of vulns are larger than that and thus cannot be sent over Pub/Sub. Use zstd to compress them, and if they're still too big, omit them, and let the worker re-download them when it gets to it.
- A few records have invalid UTF-8 characters in them, which protojson does not handle (and is invalid for proto strings). Replace the invalid characters with �


another-rex (dismissed):

another-rex (dismissed):

go/internal/importer/importer.go
```diff
@@ -349,22 +357,39 @@ func sendDeletionToWorker(ctx context.Context, config Config, item WorkItem) err
...
 	if vuln != nil {
 		var err error
 		data, err = proto.Marshal(vuln)
 		if err != nil {
 			return err
 		}
+		// make sure we have enough capacity for the compressed data
+		// to avoid reallocations
+		buf := make([]byte, 0, len(data))
+		data = zstd.EncodeTo(buf, data)
```
another-rex, line 370: I'm confused at what data contains here... looking at the documentation of EncodeTo also doesn't help. 

Can we just remove the data assignment and use buf.

michaelkedar: I want to use the `data` variable to make the logic a bit easier for the empty data case.

`EncodeTo` behaves like `append` - would it be clearer to have it as:
```suggestion
		buf = zstd.EncodeTo(buf, data)
		data = buf
```

another-rex (approved):

PR #5020, opened 2026-03-11
fix: More packages are kernel it turns out

another-rex:


gcp/workers/worker/worker.py
```diff
@@ -107,6 +107,18 @@ def _setup_logging_extra_info():
   logging.getLogger().addFilter(_ContextFilter())
 
 
+def affected_is_kernel(affected: vulnerability_pb2.Affected) -> bool:
+  if affected.package.name == 'Kernel' and \
+     affected.package.ecosystem == 'Linux':
+    return True
+
+  if any('git.kernel.org/pub/scm/linux/kernel/git' in ar.repo
+         for ar in affected.ranges):
```
michaelkedar, line 116: realistically, this is the only thing we actually care about

michaelkedar (approved):

PR #5022, opened 2026-03-11
fix: too many logs about version being empty

jess-lowe:
The functions calling this one don't care if the version is empty and it's not really actually a problem we care about.


michaelkedar (approved):
okay....

vulnfeeds/git/versions.go
jess-lowe, line 18:
```suggestion
```

PR #5028, opened 2026-03-12
fix(website): correct external nav link targets

ashmod:
The target attributes in the website header were using curly quotes instead of normal ones which made the new tab behavior broken (both links would open in the same new tab).


another-rex:
Huh interesting, wonder how those got there haha.

PR #5029, opened 2026-03-12
docs: mention Drupal advisory database as a data source

G-Rath:
I feel like this list is missing a couple of other databases - maybe it'd be worth adding a generator or lint check for changes to `source.yaml` and friends..?


another-rex (approved):

another-rex:
Could be good to have a linter, though markdown is kind of hard to lint. Might just be a change in process needed here to keep it in sync.

G-Rath:
I think there's a couple of options that'd be straightforward, we just might have to compromise on links or something depending on what is available in https://github.com/ackama/osv.dev/blob/master/source.yaml

jess-lowe:
I think we should merge this, and Ill make an issue for keeping the other databases in sync.

PR #4961, opened 2026-03-04
feat(vulnfeeds): rewrite debian first version finder and DSA/DLA/DTSA converters in Go

jess-lowe:
I have checked whether semantically the new versions are producing the same output as the last, and they are.

Also moved these out of the vulnfeeds/tools directory and into go/cmd/first-version-finder and vulnfeeds/cmd/converters/dsa-dla-dtsa respectively

PR #5030, opened 2026-03-12
docs: update converter links

G-Rath:
I'm guessing these were missed when things got rewritten? The current Debian tool link does still exist but looks like the Python version, which I'm assuming has been rewritten into Go living next to the other `cmd/converters`, but let me know if I've got that one wrong


jess-lowe:
Oh yeah this definitely needs updating. We should probably add something about the CVE5 conversion here, and also the DSA/DLA/DTSA records, but they're in the process of being rewritten here: #4961, which should be merged after the next release.

PR #5046, opened 2026-03-13
chore: Updater tool

another-rex:
Tool to update modified date back to the original value from local snapshot

### DIFF
--- a/go/cmd/vuln-modified-sync/main.go
+++ b/go/cmd/vuln-modified-sync/main.go
@@ -317,7 +317,7 @@ func parseTime(timeStr string) (time.Time, error) {
 	return time.Time{}, fmt.Errorf("unable to parse time string %q", timeStr)
 }
 
-func removeEmptySlicesAndMaps(v interface{}) interface{} {
+func removeEmptySlicesAndMaps(v any) any {
 	val := reflect.ValueOf(v)
 	if !val.IsValid() {
 		return v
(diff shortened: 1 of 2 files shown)


go/cmd/vuln-modified-sync/main.go
```diff
@@ -0,0 +1,203 @@
...
+			slog.String("id", task.VulnID),
+			slog.Time("datastore_modified", vuln.Modified),
+			slog.Time("file_modified", fileModifiedTime))
+
+		if !dryRun {
+			// Update Datastore
+			vuln.Modified = fileModifiedTime
+			vuln.ModifiedRaw = fileModifiedTime
+
+			if _, err := client.Put(ctx, key, &vuln); err != nil {
```
michaelkedar, line 292: you also need to write the vuln to GCS (and set the CustomTime field)

another-rex: Ok fixed!

go/cmd/vuln-modified-sync/main.go
```diff
@@ -0,0 +1,379 @@
...
+		if err := json.Unmarshal(compareData, &m2); err != nil {
+			logger.ErrorContext(ctx, "Failed to unmarshal compare data", slog.Any("error", err))
+			return
+		}
+
+		delete(m1, "modified")
+		delete(m2, "modified")
+
+		m1Strip := removeEmptySlicesAndMaps(m1)
+		m2Strip := removeEmptySlicesAndMaps(m2)
```
michaelkedar, line 235: I don't think this matters in practice, but I think differences in empty slices/maps vs omitted fields should actually be considered different

another-rex:
Done.

PR #5047, opened 2026-03-13
fix: markdown exception causing website 500

another-rex:
If markdown rendering fails due to an exception in the underlying library, we should just return the raw markdown text rather than 500ing

Currently record: GHSA-9rp6-23gf-4c3h


jess-lowe (approved):
LGTM except we should probably look into whether those TODOs are fixed now

PR #5050, opened 2026-03-15
feat: log successful uploads in combine-to-osv

jess-lowe:
It's nice to know how many records were actually changed on the combine to osv run

### DIFF
--- a/vulnfeeds/upload/cveworker.go
+++ b/vulnfeeds/upload/cveworker.go
@@ -41,6 +41,7 @@ func writeToDisk(v *osvschema.Vulnerability, preModifiedBuf []byte, outputPrefix
 		logger.Error("Failed to write OSV file", slog.Any("err", err), slog.String("path", filePath))
 		return false
 	}
+
 	return true
 }
 
--- a/vulnfeeds/upload/cveworker.go
+++ b/vulnfeeds/upload/cveworker.go
@@ -13,6 +13,7 @@ import (
 	"os"
 	"path"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"cloud.google.com/go/storage"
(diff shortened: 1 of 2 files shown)


vulnfeeds/upload/cveworker.go
```diff
@@ -32,17 +33,20 @@ const (
...
 		logger.Error("Failed to write OSV file", slog.Any("err", err), slog.String("path", filePath))
+		return false
 	}
+
+	return true
 }
 
 // uploadToGCS uploads the vulnerability to a GCS bucket.
-func uploadToGCS(ctx context.Context, v *osvschema.Vulnerability, preModifiedBuf []byte, outBkt *storage.BucketHandle, outputPrefix string) {
+func uploadToGCS(ctx context.Context, v *osvschema.Vulnerability, preModifiedBuf []byte, outBkt *storage.BucketHandle, outputPrefix string) bool {
```
another-rex: Update comment on what the bool means

michaelkedar: I think it'd be more idiomatic to return the error (and have a special `errUploadSkipped` or something for the hash match case) and maybe even do the error logging outside of the upload/write function?

jess-lowe: done :)

vulnfeeds/upload/cveworker.go
```diff
@@ -174,12 +181,24 @@ func Worker(ctx context.Context, vulnChan <-chan *osvschema.Vulnerability, outBk
...
 			// Upload to GCS
-			uploadToGCS(ctx, vulnToProcess, preModifiedBuf, outBkt, outputPrefix)
+			writeErr = uploadToGCS(ctx, vulnToProcess, preModifiedBuf, outBkt, outputPrefix)
+		}
+
+		if writeErr == nil {
+			logger.Info("Uploaded successfully", slog.String("id", vulnID))
+			if counter != nil {
+				counter.Add(1)
+			}
```
another-rex, line 197: Are we expecting the counter to ever be nil? I think we can get rid of this check, and just panic if we forgot to pass in a counter.

another-rex (approved):

PR #5058, opened 2026-03-17
fix(importer): Check modified times against database instead of importer run.

michaelkedar:
Bucket updates were ended up being skipped due to records with modified times before the importer last run being uploaded after that importer run.

Now, check each record against what we have in database, instead of against the importer time.

### DIFF
--- a/go/internal/importer/bucket.go
+++ b/go/internal/importer/bucket.go
@@ -40,11 +40,11 @@ func handleImportBucket(ctx context.Context, ch chan<- WorkItem, config Config, 
 
 	compiledIgnorePatterns := compileIgnorePatterns(sourceRepo)
 	bucket := config.GCSProvider.Bucket(sourceRepo.Bucket.Name)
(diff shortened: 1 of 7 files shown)


go/internal/importer/bucket.go
```diff
@@ -70,14 +70,13 @@ func handleImportBucket(ctx context.Context, ch chan<- WorkItem, config Config,
...
-			SourcePath:       obj.Name,
-			LastUpdated:      lastUpdated,
-			HasLastUpdated:   hasUpdateTime,
-			Format:           format,
-			KeyPath:          sourceRepo.KeyPath,
-			Strict:           sourceRepo.Strictness,
-			IsReimport:       !hasUpdateTime,
+			SourceRepository:       sourceRepo.Name,
+			SourcePath:             obj.Name,
+			CompareAgainstDatabase: !reimport,
```
another-rex, line 75: Isn't this just the !IsReimport field?

PR #5083, opened 2026-03-18
feat: script for deleting from datastore

jess-lowe:


another-rex (commented):
A couple of things:

1. Make it all parallel, otherwise this will take forever to run locally.
2. Move this into go/ folder, then you can use the logger there + all the datastore clients and stuff already setup  in libraries there.
3. You are missing deletions from ListedVulnerabilites, and maybe relations? @michaelkedar

another-rex (approved):

PR #5089, opened 2026-03-19
feat: OAuth endpoint to gate management tools

another-rex:
Added a OAuth authentication endpoint to put our management tools behind

### DIFF
--- a/gcp/website/auth.py
+++ b/gcp/website/auth.py
@@ -12,9 +12,8 @@ blueprint = Blueprint('auth', __name__)
 
 GOOGLE_CLIENT_ID = os.environ.get("GOOGLE_OAUTH_CLIENT_ID")
 GOOGLE_CLIENT_SECRET = os.environ.get("GOOGLE_OAUTH_CLIENT_SECRET")
-GOOGLE_DISCOVERY_URL = (
-    "https://accounts.google.com/"
-    ".well-known/openid-configuration")
+GOOGLE_DISCOVERY_URL = ("https://accounts.google.com/"
+                        ".well-known/openid-configuration")
 
 # Easy bypass for local development testing
 BYPASS_OAUTH_FOR_LOCAL_DEV = os.environ.get("BYPASS_OAUTH_FOR_LOCAL_DEV",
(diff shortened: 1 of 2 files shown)


deployment/clouddeploy/osv-website/run-staging.yaml
```diff
@@ -18,6 +18,20 @@ spec:
           value: '10.189.34.180'
         - name: REDISPORT
           value: '6379'
+        
+        # --- Authentication Variables ---
+        - name: GOOGLE_OAUTH_CLIENT_ID
+          value: '175272184261-plkdbn2vea10d769llle6e965u8hupat.apps.googleusercontent.com'
```
jess-lowe, line 25: should this also be a config variable instead?

another-rex: Perhaps? The issue is the google cloud oauth is really manual from what I can tell (e.g. I can't add it as a terraform config). So we kind of have to copy and paste this value somewhere.

This yaml file is essentially the config variable.

cuixq: actually, where does this value come from?

gcp/website/triage_handlers.py
```diff
@@ -2,19 +2,20 @@
 import logging
 import re
 import requests
-import utils
+
```
cuixq, line 5: delete?

deployment/clouddeploy/osv-website/run-staging.yaml
another-rex: Look in our GCP oauth clients page.

gcp/website/triage_handlers.py
another-rex: the extra new line?

cuixq: yes

another-rex: Fixed!

deployment/clouddeploy/osv-website/run-staging.yaml
another-rex: Added a link here to make it easier to refer to in the future

PR #5098, opened 2026-03-20
fix: website deployment allow access to secrets

another-rex:
Allow access to secrets. This is needed because we need the oauth client secrets.

### DIFF
--- a/deployment/terraform/modules/osv/website.tf
+++ b/deployment/terraform/modules/osv/website.tf
@@ -22,6 +22,12 @@ resource "google_cloud_run_v2_service" "website" {
   }
 }
 
+resource "google_project_iam_member" "website_secret_accessor" {
+  project = var.project_id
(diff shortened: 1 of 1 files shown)


jess-lowe (approved):

deployment/terraform/modules/osv/website.tf
```diff
@@ -22,6 +22,12 @@ resource "google_cloud_run_v2_service" "website" {
   }
 }
 
+resource "google_project_iam_member" "website_secret_accessor" {
+  project = var.project_id
+  role    = "roles/secretmanager.secretAccessor"
+  member  = "serviceAccount:${data.google_compute_default_service_account.default.email}"
```
michaelkedar, line 28: though, maybe we should make a specific service account for this

another-rex: Yeah... I'll have a look at how easy that'll be to do.

PR #5099, opened 2026-03-20
feat(nvd): use go to upload NVD conversion to gcs upon conversion

jess-lowe:
This pull request introduces a highly optimized, Go-native asynchronous Google Cloud Storage (GCS) upload pipeline for vulnerability converters (`nvd-cve-osv` and `cve-bulk-converter`). By transitioning the GCS uploading and synchronization logic from external bash scripts (`gcloud storage rsync` / sequential CLI wrappers) into a concurrent, thread-safe Go worker pool, this change drastically improves execution speed, reduces resource consumption, and enables unified in-memory caching across multiple years.

Additionally, this refactoring completely decouples the core conversion logic (`nvd.CVEToOSV`) from the file-system writing side-effects, making the system significantly cleaner, more modular, and easier to test.

---

### 1. Highly Concurrent Go-Native GCS Uploader (`gcs` Package)
- **Worker Pool Architecture**: Introduced the `gcs.Helper` struct which manages a pool of concurrent goroutines (`bucketWorker`) communicating via an internal upload queue channel.
- **Hash-Based Smart Uploading**: Before pushing any JSON payload to GCS, the worker checks the `sha256-hash` metadata attribute of the existing object on GCS. If it matches the computed SHA256 of the new record, the upload is skipped. This saves significant network bandwidth, operation count, and time.
- **Resource Cleanup**: Robust context handling and graceful termination via `CloseAndWait()` ensures all active writers and storage client handles are cleanly shut down.

### 2. Unified Local & Cloud Output Manager (`writer` Package)
- Replaced the deprecated `upload` package with a modern `writer` package that encapsulates all output-oriented operations.
- Employs a centralized `VulnWorker` which handles:
  - Enqueuing asynchronous uploads to GCS via `gcs.Helper`.
  - Falling back to local disk writes cleanly.
  - Fetching and applying overrides from an overrides GCS bucket path.

### 3. Decoupled & Optimized NVD CVE Conversion
- **Side-Effect Elimination**: Modified `nvd.CVEToOSV` to return a `vulns.Vulnerability` object structurally rather than writing directly to disk, facilitating unit testing.
- **Shared In-Memory Caching**: The year-by-year loop has been moved from the shell script into Go (`main.go`). The `nvd-cve-osv` tool now processes the entire `nvd-json-dir` directory in a single execution run (processing files in reverse chronological order). This enables the vendor-product repo cache (`vpRepoCache`) and git repo tags cache (`repoTagsCache`) to be **shared across all years**, avoiding redundant git listings and fetching overhead.
- **Deterministic JSON & Hashes**: To prevent calculated SHA256 hashes from fluctuating due to map iteration randomness, `v.Affected` records are now sorted alphabetically by repository name before serializing, guaranteeing reliable cache hit rates.

### 4. Drastically Simplified Orchestration Scripts
- **Removed Complex Shell logic**: Fully eliminated the sequential year-by-year bash loop, local staging directories (`gcs_stage`), manual `find -exec cp` operations, and the `gcloud storage rsync` command pipelines.
- **Streamlined Execution**: Both NVD and CVE5 bulk converters now invoke their corresponding Go binaries once, passing `--upload-to-gcs=true` along with the target bucket name, allowing the Go runtime to manage the entire lifecycle seamlessly.
---

### Automated Unit Tests
A comprehensive suite of unit and integration tests has been implemented and verified to pass:
- **GCS Worker Pool Tests**: `vulnfeeds/gcs-tools/gcs_test.go` covers worker initialization, concurrent scheduling, and object metadata validation.
- **Writer Tests**: `vulnfeeds/conversion/writer/writer_test.go` validates concurrent local disk writes, remote GCS uploads, override resolution, and metrics generation.
- **Converter Tests**: Verified that decoupled NVD and CVE5 conversion tests pass without side-effects.

To run the tests locally:
```bash
cd vulnfeeds
go test ./gcs-tools/...
go test ./conversion/writer/...
go test ./conversion/nvd/...
```

### Manual Verification
- Ran the `run_cve_to_osv_generation.sh` script locally with `--upload-to-gcs=true` to confirm:
  1. Correct execution flow across all years.
  2. Successful caching and lookup of CPE and git repositories.
  3. Parallel execution of 30 concurrent uploader workers enqueuing and transferring objects.
  4. Skipping of identical records where GCS object hashes matched computed local hashes.


vulnfeeds/gcs-tools/gcs.go
another-rex, line 1: We should make the gcs-tools repo generic to only uploading to GCS, but we shouldn't put CVE specific logic into here.

another-rex: If uploading to GCS is going to take a while, I would even put the multithreading / concurrency logic in here. 
E.g. provide a function that will spin up X number of works, and a "gcs client" that just contains a channel. 

Other code can pass the client to their code to upload. 

Probably for a separate PR though.

vulnfeeds/gcs-tools/gcs.go
```diff
@@ -0,0 +1,154 @@
...
+	"cloud.google.com/go/storage"
+	"github.com/google/osv/vulnfeeds/models"
+	"github.com/ossf/osv-schema/bindings/go/osvschema"
+	"golang.org/x/sync/errgroup"
+	"google.golang.org/api/iterator"
+	"google.golang.org/protobuf/encoding/protojson"
+)
+
+// ToGCS uploads data from an io.Reader to a GCS bucket.
+func ToGCS(ctx context.Context, bkt *storage.BucketHandle, objectName string, data io.Reader, contentType string) error {
```
another-rex: If we have a ToGCS and UploadFile what is the Upload() function for? Uploading an entire folder?

If so, can we have it all unified in this file, and have Upload use UploadFile for individual files. Then probably update the names so they all follow the same convention.

jess-lowe: Upload should be renamed to UploadVulnerabilities (cause that's what it does)

vulnfeeds/gcs-tools/gcs.go
jess-lowe: > We should make the gcs-tools repo generic to only uploading to GCS, but we shouldn't put CVE specific logic into here.

Moved these into their own thing in conversion/writer

jess-lowe: > If uploading to GCS is going to take a while, I would even put the multithreading / concurrency logic in here. E.g. provide a function that will spin up X number of works, and a "gcs client" that just contains a channel.
> 
> Other code can pass the client to their code to upload.
> 
> Probably for a separate PR though.

For uploading vulnerability records, this is too nuanced, hence it has its own thing in writer.VulnWorker, but with the NVD data this will be happening in the same thread that converts the record

vulnfeeds/cmd/converters/cve/nvd-cve-osv/main.go
```diff
@@ -116,24 +135,73 @@ func processCVE(cve models.NVDCVE, vpRepoCache *c.VPRepoCache, repoTagsCache *gi
...
-func worker(wg *sync.WaitGroup, jobs <-chan models.NVDCVE, _ string, vpRepoCache *c.VPRepoCache, repoTagsCache *git.RepoTagsCache) {
+func worker(ctx context.Context, wg *sync.WaitGroup, jobs <-chan models.NVDCVE, bkt *storage.BucketHandle, outDir string, vpRepoCache *c.VPRepoCache, repoTagsCache *git.RepoTagsCache) {
 	defer wg.Done()
 	for cve := range jobs {
-		outcome := processCVE(cve, vpRepoCache, repoTagsCache)
+		vuln, metrics, outcome := processCVE(cve, vpRepoCache, repoTagsCache)
+		cveID := string(cve.ID)
+		if outcome == models.Error {
+			logger.Error("Error generating OSV record", slog.String("cve", cveID), slog.String("outcome", outcome.String()))
+			return // Don't attempt to output files if there was an error
```
another-rex: Should these be continues? (To skip this CVE, but not kill the worker entirely?)

vulnfeeds/conversion/writer/writer.go
another-rex: nit: Update doc comment name

another-rex (commented):
Nice, mostly looks good. Have you tested it locally and see how much faster it is compared to the script? (Probably not a big impact here, since locally we have a lot of threads compared to the cronjob)

vulnfeeds/conversion/writer/writer.go
```diff
@@ -276,21 +279,79 @@ func handleDeletion(ctx context.Context, outBkt *storage.BucketHandle, osvOutput
...
+	if vuln == nil || vuln.GetId() == "" {
+		return errors.New("invalid vulnerability provided")
+	}
+
+	data, err := protojson.MarshalOptions{Indent: "  "}.Marshal(vuln)
+	if err != nil {
+		return fmt.Errorf("failed to marshal vulnerability %s: %w", vuln.GetId(), err)
+	}
+
+	objectName := filepath.Join(prefix, vuln.GetId()+".json")
```
another-rex: For object names related to GCS, we should be using path.Join for everything. filepath is *only* for interacting with disk. If we run this on windows (which to be fair we don't), I think this might break.

vulnfeeds/conversion/writer/writer.go
another-rex, line 247: just noticed this, should this be defer closed?

another-rex (dismissed):
Please update the PR description. Otherwise LGTM

another-rex (approved):

PR #5104, opened 2026-03-22
refactor(linter): rewrite linter cronjob in go

jess-lowe:
The linter cronjob is currently timing out and not completing due to some error. Instead of fixing that, I'll take the opportunity to rewrite the linter in Go instead.


go/cmd/osv-linter-worker/main.go
```diff
@@ -0,0 +1,647 @@
+// osv-linter-worker
+//
+// # Worker for osv-linter
+//
+// This worker is responsible for running osv-linter and uploading the results to GCS.
+// It also writes the results to Datastore.
+//
+// Usage:
+//
+//	go run main.go -work_dir=/tmp -local_data=/path/to/all.zip -linter_binary=osv-linter -dry_run=true
```
jess-lowe:
```suggestion
//	go run main.go -work-dir=/tmp -local-data=/path/to/all.zip -linter-bin=osv-linter -dry-run=true
```

go/cmd/osv-linter-worker/main.go
```diff
@@ -0,0 +1,647 @@
...
+//
+// Options:
+//
+//	work_dir: Working directory
+//	local_data: Path to local all.zip or directory containing OSV data
+//	linter_binary: Path to osv-linter binary
+//	dry_run: Dry run mode (no GCS upload or Datastore writes)
```
jess-lowe:
```suggestion
//	work-dir: Working directory
//	local-data: Path to local all.zip or directory containing OSV data
//	linter-bin: Path to osv-linter binary
//	dry-run: Dry run mode (no GCS upload or Datastore writes)
```

go/cmd/osv-linter-worker/Dockerfile
michaelkedar, line 1: Maybe move this file to `go/cmd/osv-linter-worker` (and make sure all the paths here & in build-and-stage still work)

go/cmd/osv-linter-worker/main.go
```diff
@@ -0,0 +1,671 @@
...
+func main() {
+	flag.Parse()
+	if err := run(); err != nil {
+		logger.Fatal("error running linter worker", slog.Any("err", err))
+	}
+}
+
+func run() error {
+	ctx := context.Background()
+	dsClient, err := datastore.NewClient(ctx, gcpProject)
```
michaelkedar, line 89: Can we abstract away the Datastore stuff behind an interface like the importer does so this is easier to migrate in the future.
We have a [SourceRepositoryStore](https://github.com/google/osv.dev/blob/master/go/internal/models/source_repository.go) already that is probably already sufficient for this, and you could make an ImportFindingsStore + struct that has the methods/fields you need

michaelkedar: Maybe the ImportFindingsStore could also do the GCS bucket things too? Or are the bucket outputs unlikely to be migrated?

michaelkedar (commented):
I think this is good, just a couple of things

michaelkedar (approved):
neat

Issue #4611, opened 2026-01-13
Add a job for Vanir signature generation


cuixq:
We are planning to migrate worker to Go; however [Vanir](https://github.com/google/vanir) is written in Python so we probably need a separate job to generate Vanir signature which we can keep the implementation in Python.

The plan is to add a separate job to generate Vanir signatures:
 - We are going to store the timestamp where the last job was run
 - Vanir signature generation will be run against GIT vulnerabilities modified since the last timestamp

 We may want to optimize the performance by using `gitter` for git cloning, which requires a bit work on Vanir's code extractor.

PR #5126, opened 2026-03-25
feat: move Vanir signature generation to a cron job

cuixq:
https://github.com/google/osv.dev/issues/4611

This PR introduces a new `vanir_signatures` worker to enrich OSV records with Vanir signatures. This cron job identifies modified vulnerabilities, generates the signatures, and updates both the Datastore and GCS with the enriched data.

   * Identifies vulnerabilities modified since the last successful run by querying the Vulnerability Datastore index, ensuring efficient updates.
   * Utilizes the vanir library to perform batch processing of OSV records, generating signatures for any vulnerability that includes a GIT affected range.
   * Filtering: 
       * Excludes records without GIT-based repository information.
       * Excludes Linux Kernel vulnerabilities (identified by package name or repository URL).
       * Ensures signatures are only generated for active (non-withdrawn) records.
   * Uses `ThreadPoolExecutor` to handle parallel I/O for fetching records from GCS and managing batch processing.
   * Enriched records are synchronized back to GCS (as the primary storage for full protos) and the Datastore index (to maintain accurate modification timestamps and status metadata).
   * Includes a `--dry-run` flag for testing the workflow without modifying production data.
   * Introduces `modify_storage_client_adapters` to configure the GCS client with an increased `HTTPAdapter` connection pool and retries.

Currently the crob job is scheduled daily at 9:00 AM Sydney time which we may want to adjust to run more frequent later.

### DIFF
--- a/gcp/workers/vanir_signatures/vanir_signatures.py
+++ b/gcp/workers/vanir_signatures/vanir_signatures.py
@@ -136,7 +136,8 @@ def process_batch(vuln_ids: list[str],
         continue
 
       if any(
-          affected_is_kernel(affected) for affected in vulnerability_pb.affected):
+          affected_is_kernel(affected)
+          for affected in vulnerability_pb.affected):
         logging.debug('Skipping %s: it is a Kernel vulnerability', vuln_id)
         continue
 
@@ -181,7 +182,8 @@ def process_batch(vuln_ids: list[str],
 
   # Batch fetch Vulnerabilities from Datastore
   vuln_keys = [
-      ndb.Key(osv.models.Vulnerability, vuln_pb.id) for vuln_pb in all_enriched_pbs
+      ndb.Key(osv.models.Vulnerability, vuln_pb.id)
+      for vuln_pb in all_enriched_pbs
   ]
   vulns_ds = ndb.get_multi(vuln_keys)
 
@@ -207,7 +209,8 @@ def process_batch(vuln_ids: list[str],
       osv.gcs.upload_vulnerability(vuln_pb, gen)
       updated_count += 1
     except Exception:
-      logging.exception('Failed upload for %s. Adding to retry list.', vuln_pb.id)
+      logging.exception('Failed upload for %s. Adding to retry list.',
+                        vuln_pb.id)
       failed_ids.append(vuln_pb.id)
       continue
 


gcp/workers/vanir_signatures/vanir_signatures.py
```diff
@@ -0,0 +1,295 @@
...
+    query = query.filter(osv.models.Vulnerability.modified > last_run)
+
+  total_generated_count = 0
+  total_processed_count = 0
+
+  with futures.ThreadPoolExecutor(max_workers=args.max_workers) as executor:
+
+    def process_with_context(batch):
+      with ndb.Client().context():
+        return process_batch(batch, args.dry_run, args.max_workers)
```
michaelkedar, line 286: It seems like we spawn `max_workers * max_workers` workers, which is okay, but maybe worth calling out somehow that it'll increase with the square of the `max_workers` args

gcp/workers/vanir_signatures/vanir_signatures.py
```diff
@@ -0,0 +1,295 @@
...
+               len(vulnerabilities))
+
+  try:
+    vuln_dicts = [
+        json_format.MessageToDict(v, preserving_proto_field_name=True)
+        for v in vulnerabilities
+    ]
+    vuln_manager = vulnerability_manager.generate_from_json_string(
+        content=json.dumps(vuln_dicts))
+    vuln_manager.generate_signatures()
```
michaelkedar, line 56: Do you know if vanir is actually okay with multiples of these running at the same time?
e.g. if two vulns for the same git repo are in different batches, it's not going to try clone the same repo to the same directory at the same time?

gcp/workers/vanir_signatures/vanir_signatures.py
```diff
@@ -0,0 +1,295 @@
...
+    # Update Vulnerability entity in Datastore
+    vuln.put()
+
+    # Use gcs_generations[original_id] ONLY if it matches enriched ID.
+    gen = gcs_generations.get(v.id)
+    try:
+      osv.gcs.upload_vulnerability(v, gen)
+      updated_count += 1
+    except Exception:
+      logging.exception('Failed to upload vulnerability %s to GCS', v.id)
```
michaelkedar: We may want to flag these failures for retry somehow if they fail.
I'm a bit worried that another job will update the record while the signatures are being generated (since it takes a while), and datastore will be out of sync with GCS (which will trigger a reimport, which will just change the modified date on the record without adding the signatures). 

It might be nice to store the generated signatures in the database separately so we could try recover quickly without waiting for the job to restart, but I understand this is quite awkward to do with how vanir sticks the signatures into the vuln directly. It probably would warrant having a proper database design, though.

gcp/workers/vanir_signatures/vanir_signatures.py
cuixq: good point! updated the arg message as well as the comment

gcp/workers/vanir_signatures/vanir_signatures.py
cuixq: In [GitCommit class](https://github.com/google/vanir/blob/main/vanir/code_extractors/git_commit.py#L114) it seems a temp directory will be made if no working directory is specified.

[GitCodeExtractor](https://github.com/google/vanir/blob/main/vanir/code_extractors/code_extractor_git.py#L43) is the code extractor for OSV GIT vulns which does not pass through `git_working_dir` so I guess a temp directory will be made for each unique commit? 

Anyway it should be fine to have multiple of this running at the same time.

gcp/workers/vanir_signatures/vanir_signatures.py
cuixq: actually let's start with a default value of 10 max workers - we can adjust it by a flag later anyway

gcp/workers/vanir_signatures/vanir_signatures.py
cuixq: We probably can do something similar to record checker: record the list of vuln IDs that require a retry and for the next run we add these vulns to the "TODO" list so they are retried. 

Agree that it will be nice that we could store the generated signatures somewhere so we could recover them immediately instead of generating again. Even without a database, we probably can still store them in DataStore? Though considering we are migrating to database soon, we may only want to do that if the work required is not that much, which can be a follow up.

michaelkedar: Happy to merge first to see how common the errors are before deciding if we want to do the work to store them.

michaelkedar (approved):

gcp/workers/vanir_signatures/vanir_signatures.py
```diff
@@ -176,6 +191,12 @@ def process_batch(vuln_ids: list[str],
     # We will only apply to Datastore entity and put() on successful GCS upload.
     now = osv.utcnow()
     v.modified.FromDatetime(now)
```
another-rex: Hmm what does this line mean?, how is this different from the `    vuln.modified = now` below.

cuixq: `v` is the vulnerability proto to be uploaded to GCS and `vuln` is the datastore entity. I will rename them to make this clear.

PR #5140, opened 2026-03-27
feat: Implement ecosystem enumeration in Go

michaelkedar:
Translated all the ecosystem queries for version enumeration into go
- Bioconductor seems to be broken (the newer releases 500 on for the requests). We have 0 Bioconductor vulns, so I just disabled enumeration for them (but left the code in for reference)
- Removed the Alpine enumeration, because querying the aports repo is quite heavy
-  opam is still Implemented and disabled, as it is in python
- Most of the ecosystems are uninteresting - they're mostly just parsing versions out of JSON (though please look at them!). Debian does the most work in its thing
- Added go-vcr so the tests don't depend on network requests. We'd want to have something regenerate these once in a while/test against live to make sure nothing starts breaking.

### DIFF
--- a/go/osv/ecosystem/bioconductor.go
+++ b/go/osv/ecosystem/bioconductor.go
@@ -80,5 +80,6 @@ func (e bioconductorEcosystem) getBiocVersions() ([]string, error) {
 	if err != nil {
 		return nil, fmt.Errorf("failed to get Bioconductor versions: %w", err)
 	}
+
 	return versions, nil
 }
--- a/go/osv/ecosystem/debian.go
+++ b/go/osv/ecosystem/debian.go
@@ -40,7 +40,7 @@ func (e debianEcosystem) getDebianFirstPackageVersion(pkg string) (string, error
 		return "0", nil
 	}
 	urlStr := fmt.Sprintf(debianFirstPackageOutputURL, url.PathEscape(e.release))
-	
+
 	escapedPkg := strings.ReplaceAll(pkg, ".", `\.`)
 	res, err := e.p.fetchJSONPaths(urlStr, escapedPkg)
 	if err != nil {
--- a/go/osv/ecosystem/hackage.go
+++ b/go/osv/ecosystem/hackage.go
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//nolint:dupl
 package ecosystem
 
 import (
--- a/go/osv/ecosystem/pypi.go
+++ b/go/osv/ecosystem/pypi.go
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//nolint:dupl
 package ecosystem
 
 import (


go/osv/ecosystem/packagist.go
```diff
@@ -32,6 +36,35 @@ func (e packagistEcosystem) IsSemver() bool {
 	return false
 }
 
-func (e packagistEcosystem) GetVersions(_ string) ([]string, error) {
-	panic("not yet implemented")
+func packagistAPIURL(pkg string) string {
```
G-Rath, line 41: that reminds me, I think ideally we should (as a follow up PR) have this logic mirror the linter to account [for alternative repos](https://github.com/ossf/osv-schema/blob/main/tools/osv-linter/internal/pkgchecker/packagist.go) i.e. Drupal

(assuming I've correctly understood that this is the right place to be doing that)

go/osv/ecosystem/ghc.go
```diff
@@ -14,12 +14,122 @@
...
+	"7.0.3", "7.0.4-rc1", "7.0.4",
+}
+
+const ghcAPIURL = "https://gitlab.haskell.org/api/v4/projects/3561/repository/tags?per_page=100"
+
+var ghcLinkNextRe = regexp.MustCompile(`<([^>]+)>;\s*rel="next"`)
+
+// GetVersions enumerates GHC versions.
+// Different components of GHC are part of the same software release, so we ignore the package name.
 func (e ghcEcosystem) GetVersions(_ string) ([]string, error) {
```
another-rex, line 53: Probably explain here that GHC ecosystem is just essentially one package.

go/osv/ecosystem/ghc_test.go
```diff
@@ -0,0 +1,23 @@
+package ecosystem
+
+import (
+	"testing"
+)
+
+func TestGHC_GetVersions(t *testing.T) {
```
another-rex, line 7: All these tests should have t.Parallel(), since they are all in the ecosystem package, we don't want to test sequentially.

go/osv/ecosystem/util.go
```diff
@@ -0,0 +1,108 @@
...
+		}
+		result = append(result, p.str)
+		last = p.str
+	}
+
+	return result, sortErr
+}
+
+// getVersionsDepsDev enumerates versions for a package using the deps.dev API.
+func getVersionsDepsDev(e Ecosystem, depsDevSystem string, pkg string) ([]string, error) {
```
another-rex: Should we be using the depsdev package for grpc requests?

Also replace depsDevSystem with the depsdev enum maybe

go/osv/ecosystem/testdata/TestNuGet_GetVersions_NotFound.yaml
```diff
@@ -0,0 +1,52 @@
...
+        content_length: 0
+        host: api.nuget.org
+        url: https://api.nuget.org/v3/registration5-semver1/doesnotexist.123456/index.json
+        method: GET
+      response:
+        proto: HTTP/1.1
+        proto_major: 1
+        proto_minor: 1
+        content_length: 215
+        body: "\uFEFF\x3C\x3F\x78\x6D\x6C\x20\x76\x65\x72\x73\x69\x6F\x6E\x3D\"\x31\x2E\x30\"\x20\x65\x6E\x63\x6F\x64\x69\x6E\x67\x3D\"\x75\x74\x66\x2D\x38\"\x3F\x3E\x3C\x45\x72\x72\x6F\x72\x3E\x3C\x43\x6F\x64\x65\x3E\x42\x6C\x6F\x62\x4E\x6F\x74\x46\x6F\x75\x6E\x64\x3C\x2F\x43\x6F\x64\x65\x3E\x3C\x4D\x65\x73\x73\x61\x67\x65\x3E\x54\x68\x65\x20\x73\x70\x65\x63\x69\x66\x69\x65\x64\x20\x62\x6C\x6F\x62\x20\x64\x6F\x65\x73\x20\x6E\x6F\x74\x20\x65\x78\x69\x73\x74\x2E\n\x52\x65\x71\x75\x65\x73\x74\x49\x64\x3A\x36\x31\x38\x62\x33\x30\x36\x35\x2D\x39\x30\x31\x65\x2D\x30\x30\x33\x66\x2D\x31\x32\x39\x38\x2D\x62\x64\x37\x35\x35\x34\x30\x30\x30\x30\x30\x30\n\x54\x69\x6D\x65\x3A\x32\x30\x32\x36\x2D\x30\x33\x2D\x32\x37\x54\x30\x33\x3A\x31\x39\x3A\x32\x30\x2E\x34\x30\x38\x31\x38\x36\x38\x5A\x3C\x2F\x4D\x65\x73\x73\x61\x67\x65\x3E\x3C\x2F\x45\x72\x72\x6F\x72\x3E"
```
another-rex, line 18: Hmm bit odd this is completely binary encoded when most of these are ascii characters

another-rex: hallucinated conversion:
```
<?xml version="1.0" encoding="utf-8"?>
<Error><Code>BlobNotFound</Code><Message>The specified blob does not exist.
RequestId:618b3065-901e-003f-1298-bd7554000000
Time:2026-03-27T03:19:20.4081868Z</Message></Error>
```

go/osv/ecosystem/nuget_test.go
```diff
@@ -0,0 +1,56 @@
+package ecosystem
+
+import (
+	"errors"
+	"testing"
+)
+
+func TestNuGet_GetVersions(t *testing.T) {
```
another-rex, line 8: I'm guessing this tests both paging logic (page.Items and page.[No]Items)? The cassette is so big it seems impossible to see.

go/osv/ecosystem/testdata/TestNuGet_GetVersions.yaml
another-rex, line 1: 3Mb file :O

go/osv/ecosystem/opam.go
```diff
@@ -14,12 +14,65 @@
...
+)
+
 type opamEcosystem struct {
 	dpkgEcosystem
 }
 
-var _ Enumerable = opamEcosystem{}
+var _ Ecosystem = opamEcosystem{}
+
+// Opam disables the Enumerable ecosystem interface as the record is pre-enumerated on import.
```
another-rex, line 30:
```suggestion
var _ Ecosystem = opamEcosystem{}

// Opam disables the Enumerable ecosystem interface as the record is pre-enumerated on import.
// var _ Enumerable = opamEcosystem{}
```

go/osv/ecosystem/packagist.go
```diff
@@ -32,6 +36,35 @@ func (e packagistEcosystem) IsSemver() bool {
...
+	// Packagist API uses literal slashes for scoped packages (e.g. monolog/monolog).
+	// Standard url.PathEscape will break routing because the forward slash is expected intact.
+	return fmt.Sprintf("https://repo.packagist.org/p2/%s.json", pkg)
```
another-rex, line 44: Should we be more consistent in using url.PathEscape for everything?

another-rex (commented):
Looks pretty good!

go/osv/ecosystem/testdata/TestNuGet_GetVersions.yaml
michaelkedar: oh 😅
The [python one](https://github.com/google/osv.dev/blob/master/osv/ecosystems/cassettes/NuGetEcosystemTest.test_next_version.yaml) is a similar size so that's extra nice

go/osv/ecosystem/nuget_test.go
michaelkedar: tbh I'm not actually sure - these test cases were copied from Python

go/osv/ecosystem/ghc_test.go
michaelkedar: The tests all call [`setupHTTPClientForTest`](https://github.com/michaelkedar/osv.dev/blob/d062d40c6cd0684c925e128d3fbd6245a7500817/go/osv/ecosystem/testutils.go#L10-L20) which messes with the global http client to use the go-vcr client.

We can't run them in parallel because they'd all interfere with each other.

I don't really want to *have* to pass in a http client to `GetVersions`

go/osv/ecosystem/util.go
michaelkedar: Using the REST API lets us use go-vcr for tests, which I think I'd prefer.

go/osv/ecosystem/ghc.go
michaelkedar: According to the schema, GHC is 
> The Haskell compiler ecosystem. The `name` field is the name of a component of the GHC compiler ecosystem (e.g., compiler, GHCI, RTS).

So it's not just *one* package, but each of the packages are released together with the new compiler release, which I guess is what the comment was already saying (which was transplanted from Python)

go/osv/ecosystem/packagist.go
michaelkedar: I don't really understand - this is the only place where we don't do `url.PathEscape` and the comment here is already explicitly explaining why?

another-rex: Hmm... nevermind ignore this comment, I thought I saw it not using url.PathEscape in a few places, my bad 😅

go/osv/ecosystem/ghc_test.go
another-rex: do we know how long this test is taking right now? Another option is storing the httpClient (or just a httpClientOverride) in each ecosystem struct.

michaelkedar: On the latest Cloud Build run on this PR it took 21.617s, which *I* think is tolerable, but if you don't like it I can try adding a per-struct client

another-rex (approved):

go/osv/ecosystem/ghc_test.go
michaelkedar: I did some factoring to add a `Provider` to all the ecosystem implementations to allow the http client to be set separately. Running the tests in parallel now took 11.195s.
PTAL

another-rex: hmm that still feels slow no? Especially with everything cassetted it should be almost instant...

go/osv/ecosystem/provider.go
```diff
@@ -0,0 +1,85 @@
...
+	return &Provider{
+		Client: client,
+	}
+}
+
+var DefaultProvider = NewProvider(http.DefaultClient)
+
+// Get returns an ecosystem for the given ecosystem name.
+// If the ecosystem is not found, it returns nil, false.
+func (p *Provider) Get(ecosystem string) (Ecosystem, bool) {
```
another-rex, line 30: Hmm this feels very... circular. 

You use the provider to get the ecosystem which uses the provider to call fetch.

Few ideas, not sure about these:
- The functions in the ecosystems map seems very ugly
  - Maybe we can use an interface for setting the providers on construction? (this might be too much work for not really much gain)
- An option for defaulting to using the default http client would be to change the type of Provider from a pointer to just the struct `Provider` (For both in these signatures and each ecosystem field). And in the fetchJSON function we can do a check to see if client pointer exists, if not use the default http client.

michaelkedar: > Hmm this feels very... circular.
> 
> You use the provider to get the ecosystem which uses the provider to call fetch.

It's more like the Provider is the environment/context that the ecosystems exist in, so the ecosystems retain a handle on the environment that produced them to make requests within that environment.

> The functions in the ecosystems map seems very ugly

Agreed, but setting up the interface adds more boilerplate that's not worth it IMO.
I would love it if I could have a generic function to do it, but go generics won't let you set fields on a generic struct.

> An option for defaulting to using the default http client...

I feel like having the `DefaultProvider` is sufficient for this.

go/osv/ecosystem/ghc_test.go
michaelkedar: idk - it *does* have to load megabytes of test data

michaelkedar: I switched many ecosystems to using gjson, which made it ~0.5s faster, but the biggest bottleneck is probably parsing all the yaml cassettes that I can't really do much about

PR #5147, opened 2026-03-29
fix(gitter): Update affected commit logic when introduced=0

Ly-Joey:
When `introduced=0`, we now only evaluate root commits that are ancestors of a fixed or last_affected commit. This prevents disjoint trees in multi-root repos from being incorrectly marked as fully affected when a vuln is only applicable to the tree with the fix.

If no fix / last affected commit, `introduced=0` still defaults to evaluating all roots.

Also extracted the event parsing and cherry-pick expansion into a `resolveEvents` helper function for clarity.

### DIFF
--- a/go/cmd/gitter/repository.go
+++ b/go/cmd/gitter/repository.go
@@ -547,7 +547,6 @@ func (r *Repository) findAncestorRoots(from []int) []int {
 	return foundRoots
 }
 
-// Hash strings of commits cherrypicked from input (excluding the input commits)
 type cherrypickedHashes struct {
 	Introduced []string
 	Fixed      []string
@@ -555,8 +554,11 @@ type cherrypickedHashes struct {
 }
 
 type resolvedEvents struct {
-	introduced   []int
-	allFixes     []int
+	// Introduced commits (from input) + cherrypicked intro commits (if enabled) + resolved roots (if there is intro=0)
+	introduced []int
+	// Fixed commits (from input) + children of last_affected commits (from input) + cherrypick of fixed commit (if enabled)
+	allFixes []int
+	// Cherrypicked commit hash strings (excluding original input commits)
 	cherrypicked cherrypickedHashes
 }
 


go/cmd/gitter/repository.go
```diff
@@ -509,50 +509,115 @@ func (r *Repository) expandByCherrypick(commits []int) []int {
...
 		}
 	}
 
+	if cherrypickIntro {
+		newIntro := r.expandByCherrypick(introduced)
+		newIntroHashes = r.hexHashes(newIntro)
+		introduced = append(introduced, newIntro...)
+	}
+
+	return introduced, allFixes, newIntroHashes, newFixedHashes
```
another-rex: nit: these variables are a bit confusing.

are the hashes the same commits as the other variable?

e.g. is introduced=newIntroHashes
and allFixes=newFixedHashes?

If not can we make them the same?

And if they are the same, can we update the name to make it clearer

another-rex (approved):
Nice! Looks great

go/cmd/gitter/repository.go
Ly-Joey: They're not really the same

`introduced` = intro from input + cherrypicked intro + roots resolved from intro=0
`allFixes` = fixed from input + children of lastAffected + cherrypicked fixed

The `new___Hashes` contains only the cherrypicked commit hashes (string) and it was added in #5101 because worker needs them to update the record (with minimal change to existing logic).

I've renamed `new___Hashes` -> `cherrypicked___Hashes` in an attempt to make it clearer(?)

go/cmd/gitter/repository.go
```diff
@@ -549,15 +549,20 @@ func (r *Repository) findAncestorRoots(from []int) []int {
 
 // resolveEvents parses and expands SeparatedEvents into lists of introduced and fixed commits
 // In case of intro=0, it will not include root commits that are not ancestors of any fixed commits
-func (r *Repository) resolveEvents(ctx context.Context, se *SeparatedEvents, cherrypickIntro, cherrypickFixed bool) (introduced []int, allFixes []int, newIntroHashes []string, newFixedHashes []string) {
+func (r *Repository) resolveEvents(ctx context.Context, se *SeparatedEvents, cherrypickIntro, cherrypickFixed bool) ([]int, []int, []string, []string) {
```
michaelkedar: nit: not having the names here makes these return values especially hard to interpret.
Maybe return a `resolvedEvents` struct that contains all these values?

go/cmd/gitter/repository.go
another-rex: Hmm yeah I think it might be a good idea to return a struct then, and have each field commented to explain what they contain.

go/cmd/gitter/repository.go
Ly-Joey: Very good point! Now I'm going into the deep end 📦 📦 📦

go/cmd/gitter/repository.go
Ly-Joey: Done!

PR #5150, opened 2026-03-30
feat: inject original query into apitester snapshots

another-rex:
Modifies the APITester to attach the original request query payload directly into the response JSON output before creating snapshots. This is particularly useful for debugging and tracing batch queries, allowing users to map the output to its input in a 1-to-1 manner.

- `/v1/query` responses now include a top-level `"query"` field reflecting the original payload.
- `/v1/querybatch` responses map each item in `queries` into its corresponding object within the `results` array under `"query"`.
- Added required dependency `github.com/tidwall/sjson` to modify the JSON bodies efficiently.
- Regenerated all snapshot test files to capture these updates.

**Note: ** This does result in the snapshots not being exactly 1:1 with what our API actually returns. But since this script will likely break if it isn't in the format we expect, I think that's probably fine tradeoff. 

---
*PR created automatically by Jules for task [730514506246798316](https://jules.google.com/task/730514506246798316) started by @another-rex*


another-rex:
@jules For clarity, the responses with no vulns should still have the field "vulns: []" to clearly show there are no vulns for that query.

another-rex:
@jules Please rebase and update this branch

another-rex:
@jules rebase again on the latest main branch, apply the changes, and rebuild the snapshots.

another-rex:
There seems to still be conflicts?

another-rex:
I have no idea why jules is allowed to overwrite my changes.

jess-lowe (approved):
Seems fine to me

Issue #5152, opened 2026-03-30
Change echo test advisory URL


orizerah:
**Is your feature request related to a problem? Please describe.**
We (echo) want to test a change to our OSV advisory so we want to change the advisory URL of the test instance to a different endpoint.

orizerah:
@another-rex Is it possible to address the PR? It's fairly small.

another-rex:
Yes, this should be fine to do. I'll merge the PR.

orizerah:
Thank you

PR #5153, opened 2026-03-30
feat: Change echo test advisory url

orizerah:
https://github.com/google/osv.dev/issues/5152


another-rex (approved):

another-rex:
New records from that endpoint should get picked up within the hour. We'll also want to do a reimport in the test instance to properly ingest the changes though.

PR #5157, opened 2026-03-30
chore: Add Azure Linux (AZL) source to test instance

JamieMagee:
## Overview

Add Azure Linux as a source in the test instance so we can start importing AZL advisories.

## Details

We publish Azure Linux vulnerability data as OSV-format JSON files in [microsoft/AzureLinuxVulnerabilityData](https://github.com/microsoft/AzureLinuxVulnerabilityData). The repo has ~12,000 advisories in an osv directory, all prefixed `AZL-`.

The ecosystem (`Azure Linux`) and its RPM-based version comparison already exist in the codebase, so this is just the source config to wire up the import.

This only touches source_test.yaml (test instance). A follow-up PR will add it to source.yaml for production once we confirm the import works.

## Testing

- Verified the entry matches the structure of other Git-based sources (type 0) like `almalinux`, `bellsoft`, etc.
- Confirmed the repo URL, directory path (osv), prefix (`AZL-`), and extension (`.json`) match what's actually in the upstream repo.
- Confirmed `Azure Linux` ecosystem support already exists in _ecosystems.py, purl_helpers.py, and the Go importer schema.

### DIFF
--- a/source_test.yaml
+++ b/source_test.yaml
@@ -72,6 +72,20 @@
   editable: False
   strict_validation: True
 
+- name: 'azurelinux'
+  versions_from_repo: False
+  type: 0
+  ignore_patterns: ['^(?!AZL-).*$']
+  directory_path: 'osv'
+  repo_url: 'https://github.com/microsoft/AzureLinuxVulnerabilityData.git'
+  detect_cherrypicks: False
+  extension: '.json'
+  db_prefix: ['AZL-']
+  ignore_git: False
+  link: 'https://github.com/microsoft/AzureLinuxVulnerabilityData/blob/main/'
+  editable: False
+  strict_validation: True
+
 - name: 'bellsoft'
   versions_from_repo: False
   type: 0


another-rex:
This should start showing up on test.osv.dev soon (today) if everything goes correctly.

JamieMagee:
> This should start showing up on test.osv.dev soon (today) if everything goes correctly.

@another-rex Thank you! I can see the vulnerability information on there, but everything shows as "No fix available" despite `last_affected` being set. For example: https://test.osv.dev/vulnerability/AZL-79538

another-rex:
We display last_affected as No fix available, as we don't know if a newer version has been released after the "last_affected" version.

We strongly recommend switching to fixed once an actual fix version is released: https://ossf.github.io/osv-schema/#requirements

Looking at the contents of the advisory git repo, it looks like the majority of your records do have "fixed" (11.1k vs 3.3k last affected), and when they do have fixed, it does show up correctly (you'll need to press Load More a few times, or for example this one: https://test.osv.dev/list?q=AZL-78305&ecosystem=Azure+Linux).

I'm guessing the difference is because there are actually no fixed versions for those packages?

PR #5158, opened 2026-03-31
fix(gitter): Add git -> https protocol change for github urls

Ly-Joey:
Add a function to change `git://` to `https://` when repo is from github.com because it hangs and times out currently.

It works fine for other repo urls though (e.g.: `git://code.qt.io/qt/qt5.git` in OSV-2021-1477) so the normalization is currently only applied to `git://github.com/<repo>`.

### DIFF
--- a/go/cmd/gitter/gitter.go
+++ b/go/cmd/gitter/gitter.go
@@ -244,8 +244,8 @@ func prepareURL(r *http.Request, url string) (string, error) {
 func normalizeURL(url string) string {
 	// Convert git://github.com/ to https://github.com/ because it times out for some reason
 	// git protocol on non-github urls works fine
-	if strings.HasPrefix(url, "git://github.com/") {
-		return "https://github.com/" + strings.TrimPrefix(url, "git://github.com/")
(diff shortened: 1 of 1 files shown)


go/cmd/gitter/gitter.go
```diff
@@ -223,19 +223,32 @@ func isLocalRequest(r *http.Request) bool {
...
 
-	return nil
+	return url, nil
+}
+
+func normalizeURL(url string) string {
+	// Convert git://github.com/ to https://github.com/ because it times out for some reason
+	// git protocol on non-github urls works fine
+	if strings.HasPrefix(url, "git://github.com/") {
+		return "https://github.com/" + strings.TrimPrefix(url, "git://github.com/")
```
michaelkedar: nit: using [`strings.CutPrefix`](https://pkg.go.dev/strings#CutPrefix) can save on one function call

michaelkedar (approved):

another-rex (dismissed):

michaelkedar (approved):

PR #5181, opened 2026-04-02
feat(gitter): tags endpoint

Ly-Joey:
Add /tags endpoint to gitter to serve CVE conversion.
Given repo url, it returns the full list of tags (stripped of the "refs/tag/" prefix) and the associated hash.

Response Codes:
200: If repo has tags
204 (No content): If repo exists but there are no tags
404: Repo not found

### Example response:
```json
curl -s "http://localhost:8888/tags?url=https://github.com/oliverchang/osv-test.git" -H "Content-Type: application/json"
{"tags":[{"label":"branch-v0.1.1-with-fix","hash":"ubP9RzJpW4PDBot7ahS7Ny7DH5g="},{"label":"branch_1_cherrypick_regress","hash":"/r+sGUAIa8H209wz/aCh0bozYgk="},{"label":"v0.1","hash":"orqUkpCRXURdNNDo6d4ufOOBmPw="},{"label":"v0.1.1","hash":"sclaGW8i0G/PgN+MZpHNET2P7/8="},{"label":"v0.2","hash":"jYJC9UXpzsPm0NLj9b3ovhxllzU="},{"label":"branch-v0.1.1","hash":"TBVXlUJnJ+oFV1vVkEMh3vI8A/Q="}]}
```

### Note on annotated vs lightweight tags:
When tags are annotated (`git tag` with -a, -s, or -u), git creates a separate tag object to contain metadata (tagger, date, etc) with its own hash. 
This results in annotated tags effectively resolving to two objects when we do `ls-remote` or `show-ref`
- ref/tags/tag-name: the tag object
- ref/tags/tag-name^{}:  the peeled commit hash the tag is actually pointing to

Take https://github.com/ossf/scorecard as an example, gitter's response will contain:
```json
{ "label": "v5.4.0", "hash": "FULSTmQUBv/amP3sUMXuB9JHJiY=" } <- Don't use this
{ "label": "v5.4.0^{}", "hash": "gO4+z+34sZq4mRcTqf2y59zXJi4=" } <- Use this
```

When it's a lightweight tag, the resolved hash **is** the commit.

---

TODO for this PR:
- [x] Find a way to test the ls-remote code path locally. 
- [x] And more tests in general

Some possible improvements I can think of (to be discussed):
- Populate repository struct with show-ref / ls-remote output
  - This requires some changes to the existing loadRepository function. I also want the separate PR to do some small refactoring (e.g. putting repo struct behind locks)
- Possibly spin up a background process to pre-load the repo if it is valid
---

_This description might look AI generated, but I actually wrote it myself!_

### DIFF
--- a/go/cmd/gitter/gitter.go
+++ b/go/cmd/gitter/gitter.go
@@ -72,7 +72,9 @@ var (
 	repoCacheMaxCostBytes int64
 	// Cache for invalid (does not exist, or does not have tags) repos
 	// Maps repo URL to the HTTP status code (404 or 204) to return
-	invalidRepoCache *ristretto.Cache[string, int]
+	invalidRepoCache           *ristretto.Cache[string, int]
+	invalidRepoTTL             time.Duration
+	invalidRepoCacheMaxEntries int64
 )
 
 const shutdownTimeout = 10 * time.Second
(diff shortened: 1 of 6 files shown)


jess-lowe (approved):
This LGTM - if we want to be more nuanced in the 500 codes, that's something we can determine later.

PR #5194, opened 2026-04-07
fix: use ndb.Expando for JobData to support different types of value

cuixq:
https://github.com/google/osv.dev/issues/4611

This PR aims to fix `TypeError: ('Lists cannot be wrapped. Received', [])` in vanir signature generation job.
 
`JobData` model was previously defined as a standard `ndb.Model` with a fixed `ndb.GenericProperty` field. While this worked for single values (like `last_run` timestamp), it caused a crash when attempting to store a list of IDs in `retry_list`. 

To fix this, this PR refactored `JobData` to `ndb.Expando` - by switching to `Expando` and removing the explicit property definition, we allow the value property to be dynamically typed.

This PR also added try-except blocks around `JobData` updates to ensure that if there is a transient failure updating the job's metadata, the primary work of the job is not lost and the process doesn't crash.

### DIFF
--- a/bindings/go/osvdev/osvdev_test.go
+++ b/bindings/go/osvdev/osvdev_test.go
@@ -107,6 +107,7 @@ func TestOSVClient_QueryBatch(t *testing.T) {
 					"GHSA-5w9c-rv96-fr7g",
 				},
 				{ // Commit
+					"CVE-2024-2002",
 					"OSV-2023-890",
 				},
 				// non-existent package
@@ -280,6 +281,7 @@ func TestOSVClient_Query(t *testing.T) {
 				},
 			},
 			wantIDs: []string{
+				"CVE-2024-2002",
 				"OSV-2023-890",
 			},
 		},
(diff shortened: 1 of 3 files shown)


gcp/workers/vanir_signatures/vanir_signatures.py
```diff
@@ -334,21 +334,23 @@ def process_with_context(batch):
...
-  if not last_run_data:
+  try:
     last_run_data = osv.models.JobData(id=JOB_DATA_LAST_RUN)
-
-  last_run_data.value = current_run
-  last_run_data.put()
+    last_run_data.value = current_run
+    last_run_data.put()
+  except Exception:
+    logging.exception('Failed to update last run timestamp.')
```
another-rex: Will this automatically log the actual exception that this is try excepting? Or just trigger a separate exception.

cuixq: I don't think `logging.exception` raises a new exception; it simply logs the error and allows
the process to continue. To improve log readability, I've updated the code to include the exception directly in the log message.

another-rex: Ah I searched it up, apparently it does automatically capture exception when ran inside an except block.

Issue #5193, opened 2026-04-07
Gitter Mega Issue


Ly-Joey:
Mega issue to track todos for gitter

PR #5196, opened 2026-04-07
feat(gitter): Accept consider_all_branches in /affected-commits

Ly-Joey:
Part of (many) for #5193 

Accepts bool `consider_all_branches` in `/affected-commits` endpoint . 
If set to false, the affected commit walk will be similar to that of limit (following only first-parent).
I've opted to do this semi-limit affected walking in a new function to not make the original Affected() function more confusing.

Note that if cherrypick detection is true, consider all branches should also be true. 

There is also a change to the limit / single-branch affected walking behaviour:
- When introduced and limit are not on the same branch
  - New behaviour: Limit walk that does not encounter an introduced commit will be discarded. 
  - Original behaviour: The full Limit -> Root path will be considered affected.

### DIFF
--- a/gcp/workers/worker/worker_test.py
+++ b/gcp/workers/worker/worker_test.py
@@ -449,6 +449,7 @@ class UpdateTest(unittest.TestCase, tests.ExpectationTest(TEST_DATA_DIR)):
     self.source_repo.ignore_git = False
     self.source_repo.versions_from_repo = False
     self.source_repo.detect_cherrypicks = False
+    self.source_repo.consider_all_branches = True
     self.source_repo.db_prefix.append('PYSEC-')
     self.source_repo.put()
 
@@ -502,6 +503,7 @@ class UpdateTest(unittest.TestCase, tests.ExpectationTest(TEST_DATA_DIR)):
     self.source_repo.ignore_git = False
     self.source_repo.versions_from_repo = False
     self.source_repo.detect_cherrypicks = False
+    self.source_repo.consider_all_branches = True
     self.source_repo.db_prefix.append('PYSEC-')
     self.source_repo.put()
 


go/cmd/gitter/repository.go
```diff
@@ -732,14 +732,71 @@ func (r *Repository) Affected(ctx context.Context, se *SeparatedEvents, cherrypi
...
+// walkRevFirstParent walks backward from the walkFrom commits following the first parent until it hits an introduced commit.
+// If a path does not encounter an introduced, it is discarded.
+// Returns a list of affected commits
+func (r *Repository) walkRevFirstParent(introduced []int, walkFrom []int) []*Commit {
+	introMap := make([]bool, len(r.commits))
+	for _, idx := range introduced {
+		introMap[idx] = true
+	}
+
+	affectedMap := make([]bool, len(r.commits))
```
another-rex, line 749: What's the reason for using a map here?

I think we can just have a list here to append all the affected commits, and do a sort compact to get rid of duplicate commits.

go/cmd/gitter/repository.go
```diff
@@ -757,51 +814,47 @@ func (r *Repository) Limit(ctx context.Context, se *SeparatedEvents, cherrypickI
...
+	return affectedCommits, cherrypickedHashes{
+		Introduced: cherrypickedIntroHashes,
+		Limit:      cherrypickedLimitHashes,
+	}
+}
 
-		// Add current node to affected commits
-		commit := r.commits[curr]
-		affectedCommits = append(affectedCommits, commit)
+// AffectedSingleBranch walks backward from the lastAffected commits following the first parent.
```
another-rex: This docstring comment seems incorrect? It is covering more than just lastAffected.

go/cmd/gitter/repository.go
another-rex: Actually, something that can be useful for the affectedMap, is that it stores paths that reach a introduced right, so we can reuse that for other walkFroms. E.g. on line 764, we can change that to be 
```
if introMap[curr] || affectedMap[curr] {
```

which can save a lot of re-walking.

Ly-Joey: Done!

another-rex (dismissed):

michaelkedar (approved):
neat!

PR #5211, opened 2026-04-08
fix(gitter): better traffic handling and patch ID caching

Ly-Joey:
Makes 2 changes to gitter traffic handling and Patch ID calculation:
1. If context cancels in the middle of loadRepository(), patch ID calculated so far will be saved so we don't need to redo them next time
2. Move semaphore acquisition into singleflight 
This means affected commit graph walking will not have concurrency limit, but I think (hope?) graph walking is fast enough once a repo is loaded.

### DIFF
--- a/go/cmd/gitter/repository.go
+++ b/go/cmd/gitter/repository.go
@@ -95,6 +95,11 @@ func LoadRepository(ctx context.Context, repoPath string) (*Repository, error) {
 		patchIDErr = repo.calculatePatchIDs(ctx, newCommits)
 	}
 
+	// If error is anything other than context cancel, exit early without saving
+	if patchIDErr != nil && !errors.Is(ctx.Err(), context.Canceled) {
+		return nil, fmt.Errorf("failed to calculate patch id for commits: %w", patchIDErr)
+	}
+
 	// Save cache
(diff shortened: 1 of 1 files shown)


go/cmd/gitter/repository.go
```diff
@@ -90,17 +90,21 @@ func LoadRepository(ctx context.Context, repoPath string) (*Repository, error) {
...
 	if len(newCommits) > 0 {
-		if err := repo.calculatePatchIDs(ctx, newCommits); err != nil {
-			return nil, fmt.Errorf("failed to calculate patch id for commits: %w", err)
-		}
+		patchIDErr = repo.calculatePatchIDs(ctx, newCommits)
 	}
 
 	// Save cache
-	if err := saveRepositoryCache(cachePath, repo); err != nil {
+	patchIDCount, err := saveRepositoryCache(cachePath, repo)
```
another-rex, line 104: Do we want to exit early (and not save) if the error is something other than a context cancelled error?

another-rex (dismissed):

go/cmd/gitter/repository.go
Ly-Joey: Done :)

another-rex (approved):

PR #5225, opened 2026-04-10
feat: implement coarse versioning in go

michaelkedar:
Re-implements the `EE:XXXXXXXX.YYYYYYYY.ZZZZZZZZ` comparable version string generation in Go from Python.

The implementations are mostly the same as the Python ones, but the semantic submodule in scalibr is generally much more lenient in accepting technically invalid versions (dunno if we should be being more strict about this).

Differences are mostly due to how we want to handle invalid versions, which I don't think really appears in the OSV database. I'll do a pass over the AffectedVersions in the database once the worker is migrated to Go to make sure this is all consistent anyway.

I created a program & test that compiles every single unique version (from `affected[].versions` and `affected[].ranges[]`) in every single OSV record, and verifies that `Parse` and `Coarse` both error on the same strings, and `Coarse` maintains monotonicity. This test doesn't run by default because a) the versions list is 22MB big and b) it takes a while to generate and run.

I've also added some fuzzing tests to help catch edge cases if we want to run them for a bit. It's already helped me find a few edge cases in dpkg and packagist, which is nice. The regular tests run the fuzzers with only the explicit seed corpus. To do full fuzzing, you need to run it manually.

APK is currently not implemented due to transitivity issues in scalibr that I'm looking to fix (https://github.com/google/osv-scalibr/pull/1932)

### DIFF
--- a/go/osv/ecosystem/apk.go
+++ b/go/osv/ecosystem/apk.go
@@ -14,7 +14,11 @@
 
 package ecosystem
 
-import "github.com/google/osv-scalibr/semantic"
+import (
+	"regexp"
+
+	"github.com/google/osv-scalibr/semantic"
+)
 
 // apkEcosystem is an ecosystem for ecosystems using Alpine Package Keeper versioning.
 type apkEcosystem struct{}
@@ -30,7 +34,18 @@ func (e apkEcosystem) Parse(version string) (Version, error) {
 	return SemanticVersionWrapper[semantic.AlpineVersion]{ver}, nil
 }
 
+//nolint:unused
+var apkCoarseVersioner = CoarseVersioner{
+	Separators:    regexp.MustCompile(`[.]`),
+	Truncate:      regexp.MustCompile(`(?:\.0|[_-])`),
+	ImplicitSplit: true,
+	EmptyAs:       &[]string{""}[0],
+}
+
 func (e apkEcosystem) Coarse(_ string) (string, error) {
+	// TODO(michaelkedar): semantic.AlpineVersion currently breaks transitivity rules
+	// (a < b, b < c, c < a) in some cases with invalid versions.
+	// Which makes coarse versions kinda broken.
 	return "", ErrCoarseNotSupported
 }
 
--- a/go/osv/ecosystem/coarse_large_test.go
+++ b/go/osv/ecosystem/coarse_large_test.go
@@ -1,0 +1,104 @@
+package ecosystem
+
+import (
+	"bufio"
+	"os"
+	"slices"
+	"testing"
+)
+
+func TestCoarseMonotonicityLarge(t *testing.T) {
+	if os.Getenv("RUN_COARSE_LARGE_TEST") != "1" {
+		t.Skip("Skipping large test: RUN_COARSE_LARGE_TEST=1 not set")
+	}
+
+	filePath := "testdata/all_versions.txt"
+	f, err := os.Open(filePath)
+	if os.IsNotExist(err) {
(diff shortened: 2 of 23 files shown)


go/osv/ecosystem/apk.go
```diff
@@ -30,7 +34,18 @@ func (e apkEcosystem) Parse(version string) (Version, error) {
...
+//nolint:unused
+var apkCoarseVersioner = CoarseVersioner{
+	Separators:    regexp.MustCompile(`[.]`),
+	Truncate:      regexp.MustCompile(`(?:\.0|[_-])`),
+	ImplicitSplit: true,
+	EmptyAs:       &[]string{""}[0],
+}
+
 func (e apkEcosystem) Coarse(_ string) (string, error) {
+	// TODO(michaelkedar): semantic.AlpineVersion currently breaks transitivity rules
```
another-rex, line 46: What happens if coarse is not supported when querying at the moment? Is it just everything for a specific package has to be loaded into memory?

go/osv/ecosystem/coarse_large_test.go
```diff
@@ -0,0 +1,104 @@
...
+
+import (
+	"bufio"
+	"os"
+	"slices"
+	"testing"
+)
+
+func TestCoarseMonotonicityLarge(t *testing.T) {
+	if os.Getenv("RUN_COARSE_LARGE_TEST") != "1" {
```
another-rex, line 11: Can you document this env somewhere. 

Also ideally add this to the Makefile as a help option similar to osv-scanner, though I think that can wait till we moved everything over to go.

go/osv/ecosystem/util.go
```diff
@@ -58,3 +61,149 @@ func sortVersions(e Ecosystem, versions []string) ([]string, error) {
...
+
+// CoarseVersioner contains configuration for generating coarse versions.
+type CoarseVersioner struct {
+	Separators    *regexp.Regexp // Regex for separators (e.g. types using dot notation).
+	Truncate      *regexp.Regexp // Regex for characters to truncate after (e.g. prerelease/build suffixes). If nil, no truncation.
+	ImplicitSplit bool           // If True, splits on transitions between digits and non-digits.
+	EmptyAs       *string        // If not nil, treats empty parts as the given string instead of removing them. If nil, removes them.
+}
+
+var implicitRegex = regexp.MustCompile(`\d+|\D+`)
```
another-rex, line 79: what does implicit mean here...?

another-rex (approved):
Looks good! Just some minor questions.

go/osv/ecosystem/apk.go
michaelkedar: [The API just does the query without the filter](https://github.com/google/osv.dev/blob/184085f60f010d9def100d2a3f2b421057ec5d95/gcp/api/server.py#L975-L981)
[The python worker sets min/max to 000..., 999...](https://github.com/google/osv.dev/blob/184085f60f010d9def100d2a3f2b421057ec5d95/osv/models.py#L1246)

go/osv/ecosystem/util.go
michaelkedar: Renamed it `implicitSplitRegex` and added a comment that it's for splitting transitions between letters and numbers

go/osv/ecosystem/coarse_large_test.go
michaelkedar: Added some instructions to CONTRIBUTING.md

I don't know about adding this specifically to the Makefile - it's a bit specific and generating the data takes quite a while.

another-rex (approved):

PR #5228, opened 2026-04-10
fix(website): escape vulnerability IDs in hierarchy display

TristanInSec:
## Summary

`construct_hierarchy_string()` in `gcp/website/frontend_handlers.py` builds the upstream/downstream hierarchy HTML for vulnerability pages by concatenating raw `vuln_id` strings into `<li>` and `<a href=...>` fragments. The resulting string is rendered with Jinja2's `| safe` filter in `vulnerability.html`, so any markup characters that reach this code path are emitted verbatim.

This PR wraps each id with `markupsafe.escape` and assembles the output as `markupsafe.Markup`, so ids drawn from source records are always HTML-escaped in the rendered hierarchy — matching how the rest of the template handles data-derived text.

## Changes

- `gcp/website/frontend_handlers.py` — `construct_hierarchy_string` now returns `Markup`; each `vuln_id` is passed through `escape()` before being spliced into the `<li>` / `<a>` fragments, using `Markup(...).format(...)` so nested `Markup` fragments compose safely.
- `gcp/website/frontend_handlers_test.py` — new `ConstructHierarchyStringTest` with three cases:
  - a known id containing `<script>` — verified not present verbatim, present as `&lt;script&gt;`
  - an unknown id containing `">` — verified escaped
  - a plain `CVE-2024-0001` — verified the anchor still renders identically to before

## Testing

- `python3 -m unittest frontend_handlers_test.ConstructHierarchyStringTest` — passes locally against the patched function.
- Existing hierarchy rendering is byte-identical for ids containing only `[A-Za-z0-9:._-]`, so no change is expected for any well-formed OSV id.

### DIFF
--- a/gcp/website/frontend_handlers.py
+++ b/gcp/website/frontend_handlers.py
@@ -1103,7 +1103,7 @@ class ComputedHierarchy(NamedTuple):
 
 
 def construct_hierarchy_string(target_bug_id: str, hierarchy: ComputedHierarchy,
-                               known_ids: set[str]) -> str:
+                               known_ids: set[str]) -> Markup:
   """Constructs a hierarchy string for display.
 
   Args:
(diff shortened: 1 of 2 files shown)


michaelkedar (approved):
Thanks!

TristanInSec:
PS: This PR fixes the issue reported in https://issuetracker.google.com/issues/502293433

PR #5233, opened 2026-04-14
feat: implement PURL converters in Go

michaelkedar:
(courtesy of Gemini)
Implement helpers to convert to and from PURLs/OSV-Schema ecosystems in Go ([from Python](https://github.com/google/osv.dev/blob/master/osv/purl_helpers.py)):
- Implement a registry-based system using `osvconstants.Ecosystem` types.
- Ported all supported ecosystems
- Implement specialised logic for special ecosystems (Go, Maven, Debian)
- Tentatively supporting Debian release conversions (e.g. `distro=bookworm` -> `Debian:12`
- added a dependency on the official `github.com/package-url/packageurl-go`

The worker only really needs OSV -> PURL, but I've added PURL -> OSV as a surprise tool to help us later (for the API)


go/purl/ecosystem_debian.go
```diff
@@ -0,0 +1,72 @@
...
+
+import (
+	"strings"
+
+	"github.com/ossf/osv-schema/bindings/go/osvconstants"
+	"github.com/package-url/packageurl-go"
+)
+
+var debianCodenames = map[string]string{
+	"7":  "wheezy",
```
tobyhawker: Why does this start at 7? Do we have no records referring to versions earlier than that?

Also, according to https://wiki.debian.org/DebianReleases, release 15 will be "Duke" so, you could consider adding that now.

go/purl/purl.go
```diff
@@ -0,0 +1,133 @@
...
+	reverseLookup    = make(map[string]osvconstants.Ecosystem) // "type/namespace" -> "EcosystemName"
+	registryOnce     sync.Once
+)
+
+func ensureRegistryPopulated() {
+	registryOnce.Do(func() {
+		registerSimpleEcosystems()
+		registerDebian()
+		registerGo()
+		registerMaven()
```
tobyhawker: Out of curiosity: is there an idiomatic way to register each ecosystem from the source file that implements it, rather than needing to do it centrally here?

I'm thinking that there is a dependency in both directions, in that the implementations in the ecosystem-specific source files have to conform to the Adapter and Parser types, and call the register function directly, but they're then imported and used explicitly here.

go/purl/purl.go
```diff
@@ -0,0 +1,133 @@
...
+		panic(fmt.Sprintf("PURL collision: %s and %s both map to %s", name, existingEco, key))
+	}
+
+	reverseLookup[key] = name
+}
+
+// PackageToPURL converts an ecosystem and package name to a PURL string.
+func PackageToPURL(ecosystem, packageName string) (string, error) {
+	ensureRegistryPopulated()
+	baseEcosystem, _, _ := strings.Cut(ecosystem, ":")
```
tobyhawker, line 14: Would CutSuffix also work here, with one fewer discarded output arguments?

go/purl/purl_test.go
```diff
@@ -0,0 +1,66 @@
...
+		{"crates.io", "rand", "pkg:cargo/rand"},
+		{"Debian", "curl", "pkg:deb/debian/curl?arch=source"},
+		{"Debian:11", "curl", "pkg:deb/debian/curl?arch=source&distro=bullseye"},
+		{"Debian:12", "curl", "pkg:deb/debian/curl?arch=source&distro=bookworm"},
+		{"Debian:99", "curl", "pkg:deb/debian/curl?arch=source&distro=99"}, // fallback
+		{"Alpine:v2.23", "curl", "pkg:apk/alpine/curl?arch=source"},
+		{"Go", "github.com/gorilla/mux", "pkg:golang/github.com/gorilla/mux"},
+		{"Go", "stdlib", "pkg:golang/stdlib"},
+		{"Maven", "org.apache.commons:commons-lang3", "pkg:maven/org.apache.commons/commons-lang3"},
+	}
```
tobyhawker, line 26: Is it practical to add tests for error conditions? Specifically I'm thinking of testing what happens when an unregistered ecosystem is used.

go/purl/purl_test.go
```diff
@@ -0,0 +1,66 @@
...
+		{"pkg:cargo/rand@0.8.5", "crates.io", "rand", "0.8.5"},
+		{"pkg:deb/debian/curl@7.74.0-1.3+deb11u1?arch=source", "Debian", "curl", "7.74.0-1.3+deb11u1"},
+		{"pkg:deb/debian/curl@7.74.0-1.3+deb11u1?arch=source&distro=bullseye", "Debian:11", "curl", "7.74.0-1.3+deb11u1"},
+		{"pkg:deb/debian/curl@7.74.0-1.3+deb11u1?arch=source&distro=11", "Debian:11", "curl", "7.74.0-1.3+deb11u1"},       // lenient
+		{"pkg:deb/debian/curl@7.74.0-1.3+deb11u1?arch=source&distro=Bullseye", "Debian:11", "curl", "7.74.0-1.3+deb11u1"}, // case-insensitive
+		{"pkg:golang/github.com/gorilla/mux@v1.8.0", "Go", "github.com/gorilla/mux", "v1.8.0"},
+		{"pkg:golang/stdlib@1.18", "Go", "stdlib", "1.18"},
+		{"pkg:maven/org.apache.commons/commons-lang3@3.12.0", "Maven", "org.apache.commons:commons-lang3", "3.12.0"},
+		{"pkg:gradle/org.apache.commons/commons-lang3@3.12.0", "Maven", "org.apache.commons:commons-lang3", "3.12.0"}, // alias
+	}
```
tobyhawker, line 64: Same here: can you test for some error handling?

go/go.mod
```diff
@@ -17,6 +17,7 @@ require (
 	github.com/hashicorp/go-retryablehttp v0.7.8
 	github.com/klauspost/compress v1.18.5
 	github.com/ossf/osv-schema/bindings/go v0.0.0-20260331231022-2a6a0b9b0ccd
+	github.com/package-url/packageurl-go v0.1.3
```
tobyhawker, line 20: I don't know if it's common practice, but I think adding a new third-party dependency is worth calling out in the top-level PR description.

tobyhawker (dismissed):

go/purl/purl.go
michaelkedar: `CutSuffix` requires the entire suffix to match (e.g. `strings.CutSuffix("Debian:10", ":10")`)
`Cut` here is basically split on the first `":"`

go/purl/ecosystem_debian.go
michaelkedar: 7 was the earliest in our database, but I've added all the way back to 1.1 (buzz), and up to 15 (duke) now.

(To be fair, if you're using a Debian version below 11 you already have security issues)

It might make sense to generate this with some fancy `//go:generate` from https://debian.pages.debian.net/distro-info-data/debian.csv

go/purl/purl.go
michaelkedar: The `init()` function is the idiomatic way, but the linter whinges about using it.
I've made it use `init()` with a nolint directive now, because imo it's much neater and pretty safe in this instance.

go/purl/purl_test.go
michaelkedar: added a couple of tests for invalid stuff

go/go.mod
michaelkedar: mentioned this in PR description

go/purl/purl.go
```diff
@@ -0,0 +1,105 @@
...
+
+	return ecosystem, packageName, purl.Version, nil
+}
+
+var (
+	generators = make(map[osvconstants.Ecosystem]purlGenerator)
+	parsers    = make(map[string]purlParser) // "type/namespace" -> purlParser
+)
+
+// registerGenerator adds a new configuration for OSV -> PURL generation.
```
tobyhawker, line 63: Is it worth adding in the comment here (and for registerParser) that the individual ecosystem implementations should register in init. Also consider referring to the guidance in the litmus tests for global state/init: https://google.github.io/styleguide/go/best-practices.html#litmus-tests

tobyhawker (approved):

PR #5251, opened 2026-04-16
fix(frontend): theme toggle breaking after Turbo navigation

ashmod:
The theme toggle button stopped working on Turbo-cached pages. This follows the same pattern as #4561


another-rex (approved):
Thanks!

PR #5252, opened 2026-04-16
fix: improve styling for list page and accordion view

ashmod:
Applies some tiny improvements for styling after light-mode implementation

---

Fixes 'Published' sort arrow icon visibility in dark mode on the /list page:

**(Before)** 
image

**(After)**
image

---

Fixes accordion open state styling in light mode with proper background/text contrast:

**(Before)**
image

**(After)**
image

---

Also, unrelated to theming, reduced the tree structure ending line distance from last child item in accordion view because it looked too spacious.

### DIFF
--- a/gcp/website/frontend3/src/styles.scss
+++ b/gcp/website/frontend3/src/styles.scss
@@ -636,7 +636,7 @@ md-icon-button.mdc-data-table__sort-icon-button {
   --md-icon-button-disabled-icon-color: $osv-grey-600;
   --md-icon-button-disabled-icon-opacity: 1;
 
-  md-icon {
+  &[disabled] md-icon {
     color: $osv-grey-600;
   }
 }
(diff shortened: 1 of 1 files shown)


michaelkedar (approved):
I like it, though I'm wondering if there's a way to keep the red accent colour somehow (or red text on black is just always bad contrast)

ashmod:
I've tried out a few other combinations before I went with the white on black, one of the options would be light red on black, because osv-red wouldn't have the best contrast here:

image

<br />
I also tried white on red:
image

but colors felt too vibrant I guess.

image

and toning down the red did not help much either.

I eventually went with the current palette because it felt easier on the eyes, but lmk what you think.

another-rex:
Thanks for the alternative screenshots, I think I'm still a fan of the current gray and red, maybe we can make the red text darker for better contrast?
I'm happy with the other changes though, so going to merge this in first and we can follow up with any other color changes.

ashmod:
> Thanks for the alternative screenshots, I think I'm still a fan of the current gray and red, maybe we can make the red text darker for better contrast?
> I'm happy with the other changes though, so going to merge this in first and we can follow up with any other color changes.

Yeah I think a lighter gray + darker red could work better too. Will look into this in the follow up.

PR #5253, opened 2026-04-17
fix(go-relations): ListedVuln ecosystems not being correctly generated

michaelkedar:
[This line](https://github.com/google/osv.dev/blob/3d5779cd03c8c104ce60680b90b623a30f95040c/osv/models.py#L1169) is important and was missing from the go implementation


Ly-Joey (approved):
😶‍🌫

PR #5275, opened 2026-04-22
feat: begin laying out things for go worker

michaelkedar:
Very much not complete, but think of this as a design doc without the doc part.

The basic idea is to have a bunch of Enrichers in charge of populating fields in the records as the worker currently does, but in a more concrete and modular pipeline.
I've Implemented the thing that adds the source link into database_specific as an example, but Enrichers will include version enumeration, PURL generated, etc.

Affected commit computation and enumeration will happen after the enricher pipeline, since it needs to return the list of commits that are not part of the OSV record.

I've made two separate structs for the pub/sub subscriber/parser and the actual vuln processing struct (the 'engine') which hopefully makes it less coupled to pub/sub.


go/internal/worker/subscriber.go
```diff
@@ -0,0 +1,103 @@
...
+	"google.golang.org/protobuf/proto"
+)
+
+type Subscriber struct {
+	Engine    Engine
+	PubSubSub *pubsub.Subscriber
+}
+
+func (s *Subscriber) Run(ctx context.Context) error {
+	return s.PubSubSub.Receive(ctx, func(ctx context.Context, m *pubsub.Message) {
```
michaelkedar, line 24: by they way, you can configure parallelism on this pubsub.Subscriber, so we wouldn't have to manually manage a worker pool

go/internal/worker/subscriber.go
```diff
@@ -0,0 +1,103 @@
...
+			slog.String("path", task.PathInSource),
+		}
+		if len(m.Data) != 0 {
+			if m.Attributes["content_encoding"] != "zstd" {
+				logger.ErrorContext(taskCtx, "Unrecognized content encoding", append(logInfo, slog.String("encoding", m.Attributes["content_encoding"]))...)
+				m.Nack()
+
+				return
+			}
+			buf := make([]byte, 0, len(m.Data)*3) // let's guess 3x compression
```
another-rex: zstd probably has the full size in the encoded data's metadata, it's apparently an optional header, but we can just make sure we encode that on the encoding end.

go/internal/worker/subscriber.go
```diff
@@ -0,0 +1,103 @@
...
+		defer span.End()
+		task := Task{
+			SourceID:     m.Attributes["source"],
+			PathInSource: m.Attributes["path"],
+		}
+		logInfo := []any{
+			slog.String("source", task.SourceID),
+			slog.String("path", task.PathInSource),
+		}
+		if len(m.Data) != 0 {
```
another-rex: Can we add some new lines and comments to break up this function a bit. (Or probably better to put it into separate functions)

go/internal/worker/worker.go
another-rex, line 1: Should we just call this file engine?

go/internal/worker/worker.go
```diff
@@ -0,0 +1,106 @@
...
+	// TODO: affected commits
+
+	// Get the current state of the vuln to check against
+	current, err := e.Stores.Vulnerability.Get(ctx, enriched.GetId())
+	if errors.Is(err, models.ErrNotFound) {
+		enriched.Modified = timestamppb.Now()
+	} else if err != nil {
+		logger.ErrorContext(ctx, "Failed to get current vuln state", slog.String("vuln_id", enriched.GetId()), slog.Any("error", err))
+		return fmt.Errorf("failed to get current vuln state: %w", err)
+	} else if e.isSemanticallyDifferent(current, enriched) {
```
another-rex: nit: remove else since it's already returngin

go/internal/worker/worker.go
```diff
@@ -0,0 +1,106 @@
...
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/testing/protocmp"
+	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+type Task struct {
+	Vuln         *osvschema.Vulnerability
+	SourceID     string
+	PathInSource string
+	IsDeleted    bool
```
another-rex: Should we make this a enum instead called TaskType?

go/internal/worker/worker.go
```diff
@@ -0,0 +1,106 @@
...
+		PathInSource: task.PathInSource,
+	}
+	var err error
+	params.SourceRepo, err = e.Stores.SourceRepo.Get(ctx, task.SourceID)
+	if err != nil {
+		return err
+	}
+	if task.Vuln == nil {
+		// TODO: Download Vuln from source
+		return errors.New("vuln not provided")
```
another-rex: I think returning the error is the correct move. We shouldn't try to  download the vuln in worker, that should be importer's job.

another-rex (commented):
Architecture LGTM!

go/internal/worker/subscriber.go
another-rex: Do you have an example of how this is done? It just calls the Receive function in multiple goroutines?

go/internal/worker/subscriber.go
another-rex: Actually had a quick look at how to do this, seems kind of annoying to do, so probably just add a TODO here and leave it.

go/internal/worker/subscriber.go
michaelkedar: [Yeah](https://pkg.go.dev/cloud.google.com/go/pubsub/v2#hdr-Receiving:~:text=The%20callback%20is%20invoked%20concurrently%20by%20multiple%20goroutines%2C%20maximizing%20throughput.%20To%20terminate%20a%20call%20to%20Subscriber.Receive%2C%20cancel%20its%20context.)

There's `Subscriber.ReceiveSettings.MaxOutstandingMessages` to configure the number of messages (and therefore goroutines) to be processed at once

go/internal/worker/worker.go
michaelkedar: This still needs the else since the first if doesn't return

maybe this would be neater as a switch statement?

go/internal/worker/worker.go
michaelkedar: I guess I'm worried about what if the compressed vuln is too big to send over pub/sub (which, I guess would be *really* big), and also if we ever want to request an update to a record from a datafix tool or something we'd need to download it there as well.

another-rex: Hmm still think the vuln should be sent along by the datafix tool right?

For the really big records, I think that's something we should solve with a side channel to pass the vuln along. I just think it's going to get really messy / buggy if we start adding logic for getting the record from upstream into the worker. Probably something we don't need to worry about for now as I think with zstd compression it has to be something truly monstrous for it to be a problem, and it'll likely cause other breakages as well.

go/internal/worker/enrich/source_link_adder.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/google/osv.dev/go/internal/worker"
+	"github.com/ossf/osv-schema/bindings/go/osvschema"
+	"google.golang.org/protobuf/types/known/structpb"
+)
+
+type SourceLinkAdder struct{}
+
+var _ worker.Enricher = (*SourceLinkAdder)(nil)
+
+func (*SourceLinkAdder) Enrich(_ context.Context, vuln *osvschema.Vulnerability, params *worker.EnrichParams) error {
```
tobyhawker: From the package comment, I think this is likely intended as an example, rather than the full complement of enrichers? Perhaps say so explicitly in a comment.

go/internal/worker/interfaces.go
```diff
@@ -0,0 +1,22 @@
...
+	"github.com/google/osv.dev/go/internal/models"
+	"github.com/ossf/osv-schema/bindings/go/osvschema"
+)
+
+type Stores struct {
+	SourceRepo    models.SourceRepositoryStore
+	Vulnerability models.VulnerabilityStore
+}
+
+type EnrichParams struct {
```
tobyhawker, line 11: Are these the only things that would be needed for any possible enricher? If not, would you expect that we would include all possible types as (optional, I guess) fields here?

go/internal/models/vulnerability.go
```diff
@@ -15,10 +17,38 @@ type VulnSourceRef struct {
 	ModifiedRaw time.Time
 }
 
+// WriteRequest bundles everything needed to perform a permanent update to an OSV record.
+type WriteRequest struct {
+	ID              string
+	Source          string                   // The source name (e.g. "debian")
+	Path            string                   // The relative path in the source (e.g. "CVE-2023.json")
+	Raw             *osvschema.Vulnerability // The original input proto
+	Processed       *osvschema.Vulnerability // The final enriched proto
```
tobyhawker: Elsewhere you seem to have called this "enriched". Is the difference meaningful?

go/internal/worker/subscriber.go
michaelkedar: functionized bits

go/internal/worker/worker.go
michaelkedar: renamed this to engine; moved some other things to what once was `interfaces` and renamed that to `worker` 🙂

go/internal/worker/worker.go
michaelkedar: didn't make it a switch, but changed the logic to be a bit easier to follow

go/internal/worker/enrich/source_link_adder.go
michaelkedar: moved some things around and moved each enricher to its own package so that the package comments are more specific to the package

go/internal/worker/interfaces.go
michaelkedar: This I don't have a great design for.

This struct should contain all things needed across all enrichers, so if a new enricher needs something else we'd have to add it here and populate it.

I guess technically all the fields are optional, (if a specific enricher doesn't need it) but practically, we'd be running every enricher so all the fields become required.

go/internal/models/vulnerability.go
michaelkedar: Processed felt a bit more generic, but I can change it to enriched for consistency

another-rex (approved):

PR #5276, opened 2026-04-22
feat: Agent skill to help dependency reviews

another-rex:


.agents/skills/dependency_review_plan.md
```diff
@@ -0,0 +1,37 @@
+# Dependency Update Review Plan - google/osv.dev
+
+This document outlines the automated workflow for reviewing and managing dependency update Pull Requests in the `google/osv.dev` repository.
+
+## 1. Discovery & Triage
+The goal is to identify all open dependency updates.
```
tobyhawker, line 6: AI overview tells me that skills are best in the third-person imperative. This is already true of most of the detailed steps, but you could also try it for the overall component. So perhaps consider "Identify all open dependency updates".

Similarly, # 2 could be changed to use "Analyze" and "use", and # 4 to "Process" or "Approve"

tools/review_dependency_prs.py
```diff
@@ -0,0 +1,117 @@
...
+      "pr", "list", "--label", "dependencies", "--state", "open", "--json",
+      "number,title,headRefName,url"
+  ])
+  prs = json.loads(prs_output)
+
+  ready_prs = []
+  manual_prs = []
+
+  for pr in prs:
+    pr_num = pr["number"]
```
tobyhawker, line 39: Nit: this is a pretty long loop body. Consider factoring out into function(s).

tobyhawker (dismissed):

Ly-Joey (approved):

PR #5279, opened 2026-04-23
fix(ui): restore red accent on accordion headers

ashmod:
Follow up on #5252 

Adds back the red accent to accordion headers by using darker red text on a lighter gray.

**(Before)**
image

**(After)**
image

Also noticed an issue with the "Introduced" field value being white in light mode, so this was fixed too.


jess-lowe (approved):
LGTM :) Thanks!

Issue #2232, opened 2024-05-23
Bisection should not produce zero-length commit ranges


andrewpollock:
**Describe the bug**
#1984 surfaced that the bisection performed for OSS-Fuzz sometimes produces a zero-length range, which is invalid
See the list of records at https://github.com/google/osv.dev/issues/1984#issuecomment-2060518658

**Expected behaviour**
Such ranges should not be produced.

andrewpollock:
https://github.com/google/osv.dev/blob/master/docker/worker/oss_fuzz.py is where the action is

anishesg:
hey, added a PR to fix: #5280 please review and thanks in advance!

Eljees:
The zero-length ranges come out of `gcp/workers/worker/oss_fuzz.py`:

```python
introduced = regress_commits[0] if regress_commits else ''
fixed = fix_commits[-1] if fix_commits else ''
events = [osv.AffectedEvent(type='introduced', value=introduced)]
if fixed:
    events.append(osv.AffectedEvent(type='fixed', value=fixed))
```

When bisection lands on the same commit for regress and fix, `introduced == fixed` and the emitted GIT
range is empty. The same applies to the pairs iterated from `result.affected_ranges` further down, where
an `introduced_in == fixed_in` pair is appended without a check.

Two ways to handle it, and I'd rather ask than guess:

1. drop the `fixed` event when it equals `introduced` (the range stays open-ended, which reads as
   "still affected"), or
2. skip the whole range and log a warning, treating the bisection result as unusable.

I can send a patch for either, along with a small pure helper so the check is unit-testable without the
Datastore emulator. Which semantics do you want for these records?

PR #5280, opened 2026-04-24
fix(oss-fuzz): add defensive guards against zero-length commit range strings

anishesg:
**Note**: After reviewer feedback, this PR does *not* fix the core issue in #2232 (see comments). It adds two narrower defensive guards described below; the main fix is tracked for a follow-up.

---

Two defensive guards are added against zero-length `A:A` range strings:

1. **`_get_commit_range`**: when a stored range string has identical start and end (e.g. `"A:A"`, neither being the `unknown` sentinel), the git walker would hide the only commit and return an empty list, silently losing the commit. The guard returns `[end_commit]` instead.

2. **`process_bisect_task`**: when the bisector returns a `result.commit` that is already a colon-separated range with equal start and end, it would previously be written straight to the datastore. The guard treats it as a bisect failure and falls back to the request range with an error marker.

The root cause of #2232 — `RegressResult` and `FixResult` being independently bisected to the same single commit hash, causing `process_impact_task` to emit `introduced=X, fixed=X` — is **not** addressed here and requires a separate fix in `process_impact_task`.

### DIFF
--- a/gcp/workers/oss_fuzz_worker/oss_fuzz.py
+++ b/gcp/workers/oss_fuzz_worker/oss_fuzz.py
@@ -194,7 +194,7 @@ def process_bisect_task(oss_fuzz_dir, bisect_type, source_id, message):
 
   if result and result.commit:
     if ':' in result.commit:
-      range_start, range_end = result.commit.split(':')
+      range_start, range_end = result.commit.split(':', 1)
       if range_start == range_end:
         logging.warning(
             'Bisect produced a zero-length range for %s, ignoring result.',
--- a/gcp/workers/oss_fuzz_worker/oss_fuzz.py
+++ b/gcp/workers/oss_fuzz_worker/oss_fuzz.py
@@ -193,6 +193,15 @@ def process_bisect_task(oss_fuzz_dir, bisect_type, source_id, message):
   _set_result_attributes(oss_fuzz_dir, message, entity)
 
   if result and result.commit:
+    if ':' in result.commit:
+      range_start, range_end = result.commit.split(':')
+      if range_start == range_end:
+        logging.warning(
+            'Bisect produced a zero-length range for %s, ignoring result.',
+            source_id)
+        result = None
+
+  if result and result.commit:
     logging.info('Bisected to %s', result.commit)
     entity.commit = result.commit
     entity.repo_url = result.repo_url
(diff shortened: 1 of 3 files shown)


michaelkedar:
Hi,
It's been a while since anyone has looked at this code, and I don't exactly have a firm grasp on what it's doing.

But I don't think this fixes the issue? From what I can tell, the problem is not that individual RegressResults or FixResults may resolve to a range of a single commit (in fact, it's what we want).

The problem is when a RegressResult and a FixResult (which are computed independently) for a single bug are both pointing to the same commit. This doesn't look the right place to do it.

anishesg:
Thanks for taking a look — you're right, and I apologise for the misdiagnosis.

The two guards I added address narrower defensive cases:
- `_get_commit_range`: prevents `get_commit_and_tag_list` from returning an empty list when a *stored* range string already has identical start and end (e.g. `"A:A"` written by a previous imprecise path).
- `process_bisect_task`: prevents the bisector from writing an `A:A` range string when it returns an already-colon-separated range.

But as you point out, neither of those is the core scenario described in #2232. The actual problem is in `process_impact_task`: when the 'regressed' and 'fixed' bisections are run independently and both happen to resolve to the same single commit, the code at lines 412–418 produces `introduced=X, fixed=X` — an invalid zero-length GIT range — without any check.

The right fix belongs there, roughly after `regress_commits, fix_commits = _get_commits(...)`, detecting `len(regress_commits) == 1 and len(fix_commits) == 1 and regress_commits[0] == fix_commits[0]` and deciding how to handle it (clear the fix, mark the bug invalid, etc.). I'm not confident what the preferred behaviour should be (unfixed? invalid?) — happy to implement whichever approach you'd prefer, or to take a stab at it and ask for review.

I'll keep this PR as a draft until we settle on the right approach. The existing guards can stay as defensive belt-and-suspenders, but I'll update the PR description to stop claiming they fix #2232.

michaelkedar:
Hmm, I'm not sure 🫤
In `process_bisect_task`, you're setting the result to `None` if it's narrowed down to a single commit (that's formatted oddly) - this seems like it would cause otherwise valid cases to become invalid. The code seems to be capable of handling `A:A` ranges anyway.

> The right fix belongs there, roughly after `regress_commits, fix_commits = _get_commits(...)`, detecting `len(regress_commits) == 1 and len(fix_commits) == 1 and regress_commits[0] == fix_commits[0]` and deciding how to handle it (clear the fix, mark the bug invalid, etc.). I'm not confident what the preferred behaviour should be (unfixed? invalid?) — happy to implement whichever approach you'd prefer, or to take a stab at it and ask for review.

I think it'd be most correct to clear the affected range (though that would result in OSV records with no ranges, which is undesirable).

Honestly, this is a low-priority issue in code that is potentially brittle and not well maintained and difficult to test. I'm hesitant to change things or accept external contributions because we currently lack the bandwidth to ensure it has no unintended side-effects.

PR #5281, opened 2026-04-24
fix(security): 2 improvements across 2 files

tomaioo:
## Summary

fix(security): 2 improvements across 2 files

## Problem

**Severity**: `High` | **File**: `gcp/indexer/shared/shared.go:L57`

The tar extraction logic joins `tmpDir` with `hdr.Name` and then cleans the result, but does not verify that the final path is still داخل `tmpDir`. A crafted tar entry such as `../../etc/passwd` or an absolute path can escape the destination directory and overwrite arbitrary files on the host.

## Solution

Before writing, validate each target path: reject absolute paths, reject `..` traversal, and ensure `strings.HasPrefix(resolvedPath, tmpDir+string(os.PathSeparator))`. Also handle tar header types explicitly (regular files only) and reject symlinks/hardlinks.

## Changes

- `gcp/indexer/shared/shared.go` (modified)
- `go/cmd/exporter/writer.go` (modified)

### DIFF
--- a/gcp/indexer/shared/shared.go
+++ b/gcp/indexer/shared/shared.go
@@ -19,9 +19,11 @@ package shared
 import (
 	"archive/tar"
 	"context"
+	"fmt"
 	"io"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"cloud.google.com/go/storage"
 )
@@ -45,6 +47,7 @@ func CopyFromBucket(ctx context.Context, bucketHdl *storage.BucketHandle, name s
 	if err != nil {
 		return "", err
 	}
+	defer r.Close()
 	tarRdr := tar.NewReader(r)
 	for {
 		hdr, err := tarRdr.Next()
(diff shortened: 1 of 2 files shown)


another-rex:
But... aren't we the ones writing the tars to the bucket in the first place?

tomaioo:
Good question—today we *do* produce those tarballs, but this extractor is still parsing untrusted input at runtime (bucket contents can be replaced, stale artifacts can persist, or write credentials can be abused).  
Path traversal in archive extraction is a well-known high-impact class, so we should enforce safe extraction regardless of expected producer.  
This is a defense-in-depth hardening: low code cost, and it prevents a single compromised artifact from writing outside `tmpDir`.

tomaioo:
Good question—today we *do* produce those tarballs, but this extractor is still parsing untrusted input at runtime (bucket contents can be replaced, stale artifacts can persist, or write credentials can be abused).  
Path traversal in archive extraction is a well-known high-impact class, so we should enforce safe extraction regardless of expected producer.  
This is a defense-in-depth hardening: low code cost, and it prevents a single compromised artifact from writing outside `tmpDir`.

PR #5283, opened 2026-04-27
fix: return original tag from normalize_tag when no version components found

### LINKED ISSUE #5230: GIT ecosystem: digit-free version strings fuzzy-match digit-free tags
`normalize_tag` returns `""` for any input with no decimal digit, then [`_match_versions`](https://github.com/google/osv.dev/blob/master/gcp/api/server.py#L1058) does `"" in normalize_tags(vers)` — true whenever `vers` contains a digit-free tag.

```
$ curl -s -X POST https://api.osv.dev/v1/querybatch \
    -d '{"queries":[{"package":{"name":"https://github.com/unicode-org/icu","ecosystem":"GIT"},"version":"deadbeef"}]}' \
  | jq '.results[0].vulns | length'
7
```

icu has a `last-cvs-commit` tag. `normalize_tag("deadbeef") == normalize_tag("last-cvs-commit") == ""`, so `"deadbeef"` matches every CVE whose range encompasses that tag. `"deadbeef0"` returns 0.

Happy to send a PR — guard the check on `normalize_tag(version)` being non-empty?

andrew:
`normalize_tag` returned `''` for any input with no digit or rc/alpha/beta/preview component. In `_match_versions` that meant any digit-free query version matched any digit-free tag in a repo's affected versions, e.g. querying icu for `deadbeef` matched the `last-cvs-commit` tag and returned 7 vulns.

Now the original string is returned when normalisation finds no version components, so unrelated digit-free strings no longer collide.

Fixes #5230

### DIFF
--- a/osv/bug.py
+++ b/osv/bug.py
@@ -37,6 +37,10 @@ def normalize_tag(tag):
   if tag.startswith('.'):
     tag = '0' + tag
   components = VERSION_COMPONENT_REGEX.findall(tag)
+  if not components:
+    # Don't collapse tags with no version-like components to '', otherwise any
+    # digit-free query would fuzzy-match any digit-free tag.
+    return tag
   return '-'.join(components)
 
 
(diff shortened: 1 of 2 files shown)


michaelkedar (approved):
Thanks!

PR #5284, opened 2026-04-27
fix(api): inject CORS header via gRPC interceptor to cover 5xx error responses

### LINKED ISSUE #5262: CORS headers not present on API requests that 404 and maybe 5xx
~Support was added in #3222 but looks to have gone again:~

```
❯ curl -sSL -D - -o /dev/null -H 'Origin: localhost:3000' -d '{}'     "https://api.osv.dev/v1/query1"
HTTP/2 404
content-type: application/json
x-envoy-decorator-operation: ingress UnknownOperationName
x-cloud-trace-context: f1935a9e4c4ee53532c97f7297f20d17
date: Tue, 21 Apr 2026 21:51:56 GMT
server: Google Frontend
content-length: 73
```

anishesg:
ESPv2's `--cors_preset=basic` flag only adds CORS headers to responses that it
successfully routes and handles. When the backend itself generates a 5xx error
(e.g., an unhandled exception in a gRPC servicer), ESPv2's per-route CORS
filter is bypassed and the `Access-Control-Allow-Origin` header is missing from
the HTTP response, causing browsers to block the error details.

The fix adds a new `CORSInterceptor` class in `gcp/api/server.py` that wraps
every gRPC handler variant (unary-unary, unary-stream, stream-unary,
stream-stream) and calls `servicer_context.send_initial_metadata` with the
`access-control-allow-origin: *` header before invoking the real handler. Because
gRPC initial metadata is forwarded by ESPv2 as HTTP response headers, the CORS
header now appears on all responses, including application-generated errors.

The interceptor is registered by passing it to `grpc.server(interceptors=...)`
in the `serve` function, which requires no changes to any individual servicer
method.

Fixes #5262

### DIFF
--- a/gcp/api/server.py
+++ b/gcp/api/server.py
@@ -99,6 +99,11 @@ _TAG_PREFIX = "refs/tags/"
 
 _TEST_INSTANCE = 'oss-vdb-test'
 
+# CORS header injected via gRPC initial metadata so that ESPv2 transcoding
+# includes it in every HTTP response, including application-generated 5xx errors
+# that would otherwise bypass ESPv2's per-route CORS filter.
+_CORS_METADATA = (('access-control-allow-origin', '*'),)
+
 # ----
 
 _ndb_client = ndb.Client()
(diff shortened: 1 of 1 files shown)


another-rex:
Thank you! 

However I don't think the added complexity is worth it to support CORS for 5xx responses.

anishesg:
Hey — looks like the Cloud Build checks (PR-api-tests, PR-api-snapshot-tests) need to be manually triggered for external contributors. Could a maintainer approve the CI run? Happy to address anything that comes up.

Issue #5282, opened 2026-04-26
Echo support for secured language packages


orizerah:
We are introducing secured language packages in our product, in order to support it, scanners will now have to ingest an additional feed for these packages.

When we introduced the additional advisories to our ecosystem, it seemed to affect all PyPi advisories and showed false-positives in scans unrelated to our packages.

What we want to accomplish with OSV is to publish advisories for libraries that exist in our ecosystem without affecting unrelated ecosystems.

orizerah:
@another-rex 
Hey, Following the issue we introduced and later fixed last week, we want to consult you about how we can implement this.
Thank you

another-rex:
One option you can do is use suffixes for it. For example, have Ecosystem = "ECHO:Pypi", and then modify the ecosystem.py to perform sorting using the PyPI version sorting scheme when it finds that suffix.

orizerah:
@another-rex when do you think you will be able to review the PR?

PR #5286, opened 2026-04-28
feat(echo): Add echo pypi support

orizerah:
Introduces an Echo ecosystem class that supports PyPI versioning semantics via the Echo:PyPI suffix, allowing Echo to publish advisories for secured Python packages without affecting unrelated PyPI ecosystems.

#5282


osv/ecosystems/echo.py
```diff
@@ -0,0 +1,97 @@
...
+from .ecosystems_base import (
+    coarse_version_generic,
+    coarse_version_from_ints,
+    MAX_COARSE_PART,
+    OrderedEcosystem,
+)
+
+
+class Echo(OrderedEcosystem):
+  """Echo container security ecosystem.
```
michaelkedar, line 22: Is there a reason not to just defer the logic to the other helpers
e.g.
```python
from .debian import DPKG
from .pypi import PyPI
# ...
class Echo(OrderedEcosystem):
  def _sort_key(self, version: str):
    if self.suffix and self.suffix.lower() == 'pypi':
      return PyPI()._sort_key(version)  # might need to silence the linter here
    return DPKG()._sort_key(version)

# same for coarse_version
```

michaelkedar (commented):
We're in the middle of a migration of much of our backend to go
Could you also add equivalent changes to our [Go implementation](https://github.com/google/osv.dev/blob/a56d91d923181f80e5dc52d1dd1b31ca5e5bbfc5/go/osv/ecosystem/ecosystem.go#L55)

orizerah:
@michaelkedar I've addressed you comments

michaelkedar:
Looks good to me.
Just need to remove echo from [this test](https://github.com/google/osv.dev/blob/98659497915825355848b61f2fc0f0f6d216814f/osv/ecosystems/debian_test.py#L115)

and fix the lints:
```
osv/ecosystems/_ecosystems.py:22:0: W0611: Unused DPKG imported from debian (unused-import)

--- osv/ecosystems/_ecosystems_test.py	(original)
+++ osv/ecosystems/_ecosystems_test.py	(reformatted)
@@ -73,8 +73,7 @@
     # PyPI version ordering
     self.assertLess(echo_pypi.sort_key('1.0.0'), echo_pypi.sort_key('1.0.1'))
     self.assertLess(echo_pypi.sort_key('1.0.0a1'), echo_pypi.sort_key('1.0.0'))
-    self.assertLess(
-        echo_pypi.sort_key('1.0.0rc1'), echo_pypi.sort_key('1.0.0'))
+    self.assertLess(echo_pypi.sort_key('1.0.0rc1'), echo_pypi.sort_key('1.0.0'))
     self.assertLess(echo_pypi.sort_key('1.9'), echo_pypi.sort_key('1.10'))
 
   def test_echo_base_ecosystem(self):
   ```

orizerah:
@michaelkedar I've addressed your comments

michaelkedar (approved):
Thanks!

PR #5288, opened 2026-04-29
feat(go worker): deletions, gitter/affected commits, pypi

michaelkedar:
- Wire up gitter to go worker (and move the proto files to be accessible)
- Logic for handling deletions/withdrawn
- Logic for notifying PyPI function to send vulns to PyPI

I was going to implement the Datastore writes here, but it turns out to be a lot of code so it's probably better in its own PR.


go/internal/worker/affectedcommits.go
```diff
@@ -0,0 +1,191 @@
...
+		}
+	}
+
+	return models.AffectedCommitsResult{
+		Commits: allCommits,
+		Skip:    false,
+	}, nil
+}
+
+func callGitter(ctx context.Context, client *http.Client, gitterHost string, aRange *osvschema.Range, gitAnalysis *models.GitAnalysisConfig) (*gitterpb.AffectedCommitsResponse, error) {
```
another-rex: Can this be named something more specific? E.g. about getting the affected commits. This is still in just the worker package, so the naming can still be confusing even if it's under affectedcommits.go

go/internal/worker/affectedcommits.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+	resp := new(gitterpb.AffectedCommitsResponse)
+	if err := proto.Unmarshal(respBytes, resp); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal gitter response: %w", err)
+	}
+
+	return resp, nil
+}
+
+func applyGitterResponse(resp *gitterpb.AffectedCommitsResponse, affected *osvschema.Affected, aRange *osvschema.Range) {
```
another-rex: Same with this

go/internal/worker/affectedcommits.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+
+	return models.AffectedCommitsResult{
+		Commits: allCommits,
+		Skip:    false,
+	}, nil
+}
+
+func callGitter(ctx context.Context, client *http.Client, gitterHost string, aRange *osvschema.Range, gitAnalysis *models.GitAnalysisConfig) (*gitterpb.AffectedCommitsResponse, error) {
+	req, err := makeAffectedCommitsRequest(aRange, gitAnalysis)
```
another-rex: Nit: wouldn't it make more sense for  the req to be passed into this function?

go/internal/worker/engine.go
```diff
@@ -78,26 +91,151 @@ func (e *Engine) handleUpdate(ctx context.Context, task Task) error {
...
+
+	// Notify PyPI
+	if e.NotifyPyPI {
+		e.notifyPyPI(ctx, withdrawn, existing)
+	}
+
 	return nil
 }
+
+func (e *Engine) notifyPyPI(ctx context.Context, vuln, old *osvschema.Vulnerability) {
```
another-rex, line 183: nit: maybe add some comment about this is just sending to the message bus, and there's a separate worker that handles the actual notification from the message bus.

go/internal/worker/affectedcommits.go
michaelkedar: renamed to `fetchAffectedCommits`

go/internal/worker/affectedcommits.go
michaelkedar: renamed to `applyAffectedCommitsAndTags`

go/internal/worker/affectedcommits.go
michaelkedar: Since this is now `fetchAffectedCommits`, I'm electing to leave it in here.
If we're going to move the graph walking from the gitter to the worker, the request would change.

I like having this function as a more abstract 'get all the affected commits and tags' function

go/internal/worker/engine.go
michaelkedar: Added a comment

another-rex (approved):

PR #5293, opened 2026-05-04
feat(go worker): implement datastore/GCS interface

michaelkedar:
Implements the datastore/gcs writes needed for creating vulns in the go worker, including `ListedVulnerability` `AffectedVersions` and `AffectedCommits` entities.

### DIFF
--- a/go/cmd/importer/main.go
+++ b/go/cmd/importer/main.go
@@ -81,7 +81,7 @@ func main() {
 	}
 	config.SourceRepoStore = db.NewSourceRepositoryStore(datastoreClient)
 	// Needed for deletions only
-	config.VulnerabilityStore = db.NewVulnerabilityStore(datastoreClient)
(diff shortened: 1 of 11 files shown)


go/cmd/importer/main.go
```diff
@@ -81,7 +81,7 @@ func main() {
 	}
 	config.SourceRepoStore = db.NewSourceRepositoryStore(datastoreClient)
 	// Needed for deletions only
-	config.VulnerabilityStore = db.NewVulnerabilityStore(datastoreClient)
+	config.VulnerabilityStore = db.NewVulnerabilityStore(datastoreClient, nil)
```
another-rex, line 84: Feels like this could cause a panic that's hard to catch if we somehow accidently try to do a write in the importer.

another-rex (approved):
LGTM

go/cmd/importer/main.go
michaelkedar: IMO we shouldn't *accidentally* be writing in the importer

Issue #4913, opened 2026-02-25
Add TuxCare ecosystem support


Korulag:
- [x] Prepare your data \- refer to the [OSV Schema](https://ossf.github.io/osv-schema/) documentation for information on how to properly format the data so it can be accepted.
Our OSV data source is https://github.com/cloudlinux/tuxcare-osv.
        
- [x] Create a PR to [reserve an ID prefix and define a new ecosystem](https://ossf.github.io/osv-schema/#id-modified-fields) ([example](https://github.com/ossf/osv-schema/pull/219)). We review the records you start publishing for OSV Schema [correctness](https://github.com/ossf/osv-schema/tree/main/validation) and [quality](https://google.github.io/osv.dev/data_quality.html) as part of reviewing and merging this PR.
https://github.com/ossf/osv-schema/pull/520

- [x] Prepare and publish your records via a Git repository ([example](https://github.com/AlmaLinux/osv-database/tree/master)). If this method isn’t ideal, we also support publishing records from [REST API endpoints](https://google.github.io/osv.dev/data/new/rest-api) or through a GCS bucket([example](https://storage.googleapis.com/android-osv/)).
        
- [x] To support API querying, please create a PR to extend [purl\_helpers.py](https://github.com/google/osv.dev/blob/master/osv/purl_helpers.py) and create a new ecosystem in [\_ecosystems.py](https://github.com/google/osv.dev/blob/master/osv/ecosystems/_ecosystems.py). You can refer to existing examples showing how to implement support for [Semver](https://github.com/google/osv.dev/blob/139de7b69a2ea39e2113309b3a0a47aab920ddcf/osv/ecosystems/_ecosystems.py#L45) and [non-Semver](https://github.com/google/osv.dev/pull/3430) ecosystems.  
        
- [x] Create a PR to start [importing the records you are publishing into our test instance of OSV.dev](https://github.com/google/osv.dev/blob/master/source_test.yaml) and validate everything is working as intended there.
https://github.com/google/osv.dev/pull/4968

- [ ] Create a PR to start [importing the records you are publishing into our production environment](https://github.com/google/osv.dev/blob/master/source.yaml)

michaelkedar:
Hi, I made #5231 as a placeholder to allow Tuxcare to be ingested into OSV.dev, but without an implementation, we won't be able to match it on the API.
What versioning rules do Tuxcare packages use? From a quick glance it seems to me to be based on RPM?

michaelkedar:
[...]

PR #5296, opened 2026-05-04
feat: Add proper version parsing mapping for TuxCare advisories

Korulag:
Added changes to parse package versions properly if the ecosystem is TuxCare as per discussion in https://github.com/google/osv.dev/issues/4913


michaelkedar:
Thanks for addressing this.

I have concerns over the way this is being implemented as special logic just for handling TuxCare.
Could you create a `tuxcare.py` and implement an `OrderedEcosystem` that can call `get` itself?

something like
```python
class TuxCareEcosystem(OrderedEcosystem):
  def __init__(self, suffix: str | None):
    super().__init__(suffix)
    
    # Lazy import to avoid circular dependency
    from ._ecosystems import get
    # Add the checks for TuxCare:TuxCare here
    self._inner = get(suffix) if suffix else None
    
  def _sort_key(self, version):
    # check that _inner is set
    return self._inner._sort_key(version)

  # def coarse_version(...)
```

Also, we're in the process of migrating stuff to go - could you also add a similar implementation to the [equivalent go code](https://github.com/google/osv.dev/blob/a649fb377ffe534a09d88780b2fd0bd9d64fb14b/go/osv/ecosystem/ecosystem.go#L41)

Korulag:
@michaelkedar could you please take a look again? I'm not sure about correctness of Go part, but Python one was reworked as you requested

go/osv/ecosystem/tuxcare.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+func (e tuxcareEcosystem) IsSemver() bool {
+	inner, err := e.resolve()
+	if err != nil {
+		return false
+	}
+
+	return inner.IsSemver()
+}
```
michaelkedar, line 80: One small thing:
I'd prefer this always return `false`
`IsSemver()` is mostly only to be used to convert `affected[].ranges[].type` from `ECOSYSTEM` to `SEMVER` for those ecosystems, which I don't think we want to do with TuxCare.

michaelkedar (dismissed):
Thanks! Happy to merge this.

go/osv/ecosystem/tuxcare.go
Korulag: Did the change, please take a look

Korulag:
@michaelkedar I've fixed linting errors

michaelkedar (approved):
Thanks! Looks good

PR #5299, opened 2026-05-05
feat(gitter): Stop trying to make fetch happen

Ly-Joey:
Gitter will attempt to solve a few common git errors (index.lock file exists, and refname conflict) when git fetch and reset fail.
If that doesn't fix the git fetch, fallback to a fresh reclone instead of redoing fetch on a broken repo next time.

### DIFF
--- a/go/cmd/gitter/gitter.go
+++ b/go/cmd/gitter/gitter.go
@@ -433,6 +433,7 @@ func attemptGitRecovery(ctx context.Context, repoPath string, err error) bool {
 			logger.ErrorContext(ctx, "failed to remove index.lock", slog.Any("err", err))
 			return false
 		}
+
 		return true
 	}
 
@@ -444,6 +445,7 @@ func attemptGitRecovery(ctx context.Context, repoPath string, err error) bool {
 			logger.ErrorContext(ctx, "failed to prune origin", slog.Any("err", err))
 			return false
 		}
+		
 		return true
 	}
 
--- a/go/cmd/gitter/gitter.go
+++ b/go/cmd/gitter/gitter.go
@@ -312,15 +312,6 @@ func isAuthError(err error) bool {
 		(strings.Contains(strings.ToLower(errString), "repository") && strings.Contains(strings.ToLower(errString), "not found"))
 }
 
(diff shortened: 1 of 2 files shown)


jess-lowe:
fetch

go/cmd/gitter/gitter.go
```diff
@@ -407,6 +398,75 @@ func getFreshRepo(ctx context.Context, w http.ResponseWriter, repoURL string, fo
...
+	// We can try removing stale remote-tracking branches and retry
+	if isRefConflictError(err) {
+		logger.WarnContext(ctx, "ref conflict detected, running git remote prune origin")
+		if err := runCmd(ctx, repoPath, nil, "git", "remote", "prune", "origin"); err != nil {
+			logger.ErrorContext(ctx, "failed to prune origin", slog.Any("err", err))
+			return false
+		}
+		
+		return true
+	}
```
another-rex: I think this is even being too clever, now that we have graceful shutdowns, how often do these issues actually come up?

I'm leaning towards just always reclone at the first sign of trouble.

Ly-Joey: The index.lock fix has been in gitter since #4816 so I don't really have stats for it.

For the second case (refname conflict). We’ve seen this issue in 7 repos over the past week in our test env.

It happens when, for example:
1. A repo starts off with a `fix` branch
2. Gitter clones / fetches at this point (we now have a local ref `fix`).
3. Afterwards, the branch can be renamed upstream to `fix/some-other-name`. Similarly they could have deleted the `fix` branch and started some more branches under the `fix/` prefix.
4. When gitter fetches again, the refname conflict will happen as git doesn't differentiate between directory and files. So the local `fix` ref name will be in conflict with every branch upstream that starts with `fix/`.

That said, this is not a hill I'm dying on, so I'm happy to change to just using reclone as the fallback.

another-rex: Let's keep this second case. For the first case I think let's reclone, and log it so we know how often this happens.

Ly-Joey: Done :)

another-rex (approved):
LGTM

PR #5301, opened 2026-05-05
fix: add log so we know object has updated successfully

jess-lowe:
hard to know if upload is successful without logs...


Ly-Joey (approved):
🪵

PR #5307, opened 2026-05-06
fix(conv): Only fallback to legacy ref checks if response is a 500 error

jess-lowe:
The converters are doing double the work currently if the gitter fails to find the repo - not ideal.


vulnfeeds/git/repository.go
```diff
@@ -282,8 +294,11 @@ func RepoTags(repoURL string, repoTagsCache *RepoTagsCache) (tags Tags, e error)
 	if os.Getenv("GITTER_HOST") != "" {
 		refs, err = gitterRepoRefs(repoURL)
 		if err != nil {
-			logger.Warn("Failed to fetch tags from gitter, falling back to legacy enumeration", slog.String("repo", repoURL), slog.Any("error", err))
-			refs, err = RemoteRepoRefsWithRetry(repoURL, 3)
+			var gitterErr *GitterError
+			if errors.As(err, &gitterErr) && gitterErr.StatusCode == http.StatusInternalServerError {
```
Ly-Joey: Maybe use `gitterErr.StatusCode != http.StatusNoContent && gitterErr.StatusCode != http.StatusNotFound` instead? 
If gitter decides to do more granular 5xx codes in the future we won't have to change CVE conversion code again.

vulnfeeds/git/repository.go
```diff
@@ -424,7 +439,12 @@ func ValidRepo(repoURL string) (valid bool) {
 		if err == nil {
 			return true
 		}
-		logger.Warn("Failed to validate repo through gitter, falling back to legacy check", slog.String("repo", repoURL), slog.Any("error", err))
+		var gitterErr *GitterError
+		if errors.As(err, &gitterErr) && gitterErr.StatusCode == http.StatusInternalServerError {
```
Ly-Joey: Same as above. Use `StatusNoContent` and `StatusNotFound` instead?

vulnfeeds/git/repository.go
```diff
@@ -446,7 +466,12 @@ func ValidRepoAndHasUsableRefs(repoURL string) (valid bool) {
 		if err == nil {
 			return len(refs) > 0
 		}
-		logger.Warn("Failed to validate repo through gitter, falling back to legacy check", slog.String("repo", repoURL), slog.Any("error", err))
+		var gitterErr *GitterError
+		if errors.As(err, &gitterErr) && gitterErr.StatusCode == http.StatusInternalServerError {
```
Ly-Joey: Same as above :)

vulnfeeds/git/repository_test.go
```diff
@@ -501,12 +501,26 @@ func TestValidRepoWithGitter(t *testing.T) {
...
-			name:                    "StatusNotFound => Falls back to legacy check and returns false",
+			name:                    "StatusNotFound with Invalid Repo => Does NOT fall back, returns false",
 			status:                  http.StatusNotFound,
 			repo:                    "https://github.com/andrewpollock/mybogusrepo",
 			wantValidRepo:           false,
 			wantValidRepoAndHasRefs: false,
 		},
+		{
+			name:                    "StatusNotFound with Valid Repo => Does NOT fall back, returns false",
+			status:                  http.StatusNotFound,
```
Ly-Joey: Valid repo with no tags will have `StatusNoContent` (204) not `StatusNotFound`

Ly-Joey (dismissed):

another-rex (approved):

PR #5310, opened 2026-05-07
feat(go worker): implement the enrichers

michaelkedar:
Added Enrichers to the pipeline to cover what the python worker is currently doing.
There's quite a bit, but I've tried making individual tests to cover each enricher, and copied one of the Python test yaml files to go to make sure it's similar


go/internal/osvutil/event.go
```diff
@@ -0,0 +1,78 @@
...
+
+import (
+	"slices"
+	"strings"
+
+	"github.com/google/osv.dev/go/osv/ecosystem"
+	"github.com/ossf/osv-schema/bindings/go/osvschema"
+)
+
+type EventType int
```
another-rex, line 12: Should this just be a string type rather than an int type, and we can skip the switch statement  in affected_versions.go

go/internal/worker/engine.go
```diff
@@ -55,6 +58,19 @@ func (e *Engine) handleUpdate(ctx context.Context, task Task) error {
 		return errors.New("vuln not provided")
 	}
 
+	// Get the current state of the vuln to check against
+	current, err := e.Stores.Vulnerability.Get(ctx, task.Vuln.GetId())
+	currentNotFound := errors.Is(err, models.ErrNotFound)
```
another-rex: can you invert this bool? The double negation is a bit confusing.

another-rex (dismissed):
hmm for each enricher should it have a name, that way we can debug up to where each enrichment run actually got up to and which one errored.

Overall  LGTM!

go/internal/worker/pipeline/enumerateversions/enumerateversions.go
```diff
@@ -0,0 +1,169 @@
...
+			// 3. Cache parsed versions to avoid repetitive parsing.
+			type parsedEvent struct {
+				osvutil.Event
+
+				Parsed ecosystem.Version
+			}
+
+			var parsedEvents []parsedEvent
+			for _, evt := range events {
+				p, err := sys.Parse(evt.Version)
```
another-rex, line 100: Hmm do we have a linter or some sort of warning where we can get notified about  these invalid versions? E.g. maybe this is showing a edge case in our version parsing where we are wrong.

michaelkedar:
> hmm for each enricher should it have a name, that way we can debug up to where each enrichment run actually got up to and which one errored.

I actually use the [`%T` formatting here](https://github.com/google/osv.dev/blob/9cf4399fdbdb56aada5da800ba6d86751e44c32b/go/internal/worker/engine.go#L63) which should log e.g. `*namenormalize.Enricher`

I can still add a Name() method if you'd prefer

go/internal/osvutil/event.go
michaelkedar: I want this to be an int type to use for sorting, but I've added a `String()` method to the type.

go/internal/worker/pipeline/enumerateversions/enumerateversions.go
michaelkedar: I've made the SortVersions and this log an error and skip. Hopefully we don't have bad versions OSV.dev (though IIRC, most `semantic` parsers cannot actually error.

go/internal/worker/engine.go
michaelkedar: redid the logic here to avoid this variable altogether

another-rex (approved):

PR #5312, opened 2026-05-07
feat(go worker): copy over a bit more logic from python

michaelkedar:
- Clear the ImportFindings in Datastore when import is successful
- send a `gcs_retry` pub/sub message on (what will be) the failed-tasks topic when writing to GCS fails after writing datastore entities
- log the task latency that we have a log-based metric set up for

### DIFF
--- a/go/internal/database/datastore/import_findings.go
+++ b/go/internal/database/datastore/import_findings.go
@@ -1,0 +1,23 @@
+package datastore
+
+import (
+	"context"
+
+	"cloud.google.com/go/datastore"
+	"github.com/google/osv.dev/go/internal/models"
+)
+
+type ImportFindingsStore struct {
+	client *datastore.Client
+}
+
+var _ models.ImportFindingsStore = (*ImportFindingsStore)(nil)
+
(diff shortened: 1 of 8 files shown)


go/internal/database/datastore/import_findings.go
```diff
@@ -0,0 +1,23 @@
...
+	"github.com/google/osv.dev/go/internal/models"
+)
+
+type ImportFindingsStore struct {
+	client *datastore.Client
+}
+
+var _ models.ImportFindingsStore = (*ImportFindingsStore)(nil)
+
+func NewImportFindingsStore(client *datastore.Client) *ImportFindingsStore {
```
another-rex, line 16: Is this being created anywhere?

another-rex: Seems like it'll always be nil

michaelkedar: It will be created in the main function
[spoilers for upcoming PR](https://github.com/michaelkedar/osv.dev/blob/94456306195a63a1efdd9c9231527385c62bbbbe/go/cmd/worker/main.go#L90)

another-rex (approved):

PR #5328, opened 2026-05-10
fix(conv): repositories should be lowercase for deterministic behaviour

jess-lowe:
Noticed some records were flip flopping indeterminately. Root cause is that repos have sometimes uppercase letters version in references, but lower case letter version in CPEs, causing duplicate repos, and its picking the first one it comes across in the non-deterministicly ordered map. 

also put notes in the logs if the record errors


michaelkedar (dismissed):

another-rex (commented):
Not all repositories are case insensitive though, not sure if that will be a problem when it comes to resolving links

jess-lowe:
> Not all repositories are case insensitive though, not sure if that will be a problem when it comes to resolving links

changed to only do the hosts that are case-insensitive

michaelkedar (approved):

PR #5329, opened 2026-05-11
feat(vulnfeeds): Add persistent Redis caching layer for canonical repo links

jess-lowe:
This PR adds a persistent Redis caching layer to `vulnfeeds` to cache the mappings of resolved canonical repository links. Caching these lookups helps avoid making redundant external HEAD requests, preventing us from regularly getting rate limited (HTTP 429) during CVE processing.

* **Protocol**: Configured `go-redis` to force `Protocol: 2` (RESP2) to ensure full compatibility with older Memorystore Redis instances.
* **TTL**: Keys are stored with a randomized expiration of 1 week ± 1 day to avoid cache stampedes.
* **Fallback**: If Redis isn't configured or goes down, the cache gracefully falls back to in-memory caching.
* **Deployments**: Configured `REDISHOST` in both the test and prod GKE CronJob manifests (`nvd-cve-osv.yaml`).

### Testing
Added a new unit test `TestRedisCanonicalLinkCache` in `git/versions_test.go` that uses a mock TCP Redis server to test connect, get, set, and pipelined RESP2 parsing. Ran all other tests in `vulnfeeds/` and verified they all pass.

### DIFF
--- a/vulnfeeds/git/repository.go
+++ b/vulnfeeds/git/repository.go
@@ -54,8 +54,8 @@ var ErrRateLimit = errors.New("rate limit exceeded")
 
 // A GitTag holds a Git tag and corresponding commit hash.
 type Tag struct {
-	Tag    string // Git tag
-	Commit string // Git commit hash
+	Tag    string `json:"tag"`    // Git tag
+	Commit string `json:"commit"` // Git commit hash
 }
 
 type Tags []Tag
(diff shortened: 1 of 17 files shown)


vulnfeeds/git/repository.go
```diff
@@ -125,22 +165,51 @@ func (c *RepoTagsCache) IsInvalid(repo string) bool {
...
+		ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
+		defer cancel()
+
+		key := "canonical_link:" + repo
+		ttl := randomTTL()
+		c.redisClient.Set(ctx, key, canonicalLink, ttl)
+	}
 }
 
 func (c *RepoTagsCache) GetCanonicalLink(repo string) (string, bool) {
```
another-rex: Good start, though you should have two implementations. When running with a redis cache there is no need to have a local in memory cache. 

The functions on RepoTagsCache should be defined in an interface, and we just need to split out the redis code into it's own impl.

When running without the cache, implemented with the current RepoTagsCache, and when running with redis, implemented with a redis version.

Basically all the code should be able to remain the same, just no need to add these c.redisClient == nil checks everywhere. And we save some memory.

another-rex (commented):
Nice, logic wise basically what we're looking for, just need to change the impl a bit.

another-rex (dismissed):

another-rex (approved):

PR #5330, opened 2026-05-11
feat: migrate to new go worker

michaelkedar:
yay 👷
Set up the entry point and Dockerfile and kubernetes yaml for the go worker to supplant the python worker.
- The Go worker can concurrently process 10 pub/sub tasks (configurable), up from the Python's 1, so I've adjusted the horizontal scaler appropriately.
- The go worker builds the new `gcr.io/oss-vdb/osv-worker` image. The old python `gcr.io/oss-vdb/worker` image is still being built because a number of our other images use it as a base image.
- I haven't touched the memory/CPU limits (but I have removed unused configs), since I don't really know the memory footprint of this. Was planning to watch this on the test instance and adjust appropriately.

Merging this WILL start deploying the go worker, so best to do this right after the week's release.


go/cmd/worker/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+
+COPY ./go.mod /src/go.mod
+COPY ./go.sum /src/go.sum
+RUN go mod download && go mod verify
+
+
+COPY ./ /src/
+RUN CGO_ENABLED=0 go build -o worker ./cmd/worker
+
+FROM gcr.io/distroless/static-debian12@sha256:20bc6c0bc4d625a22a8fde3e55f6515709b32055ef8fb9cfbddaa06d1760f838
```
another-rex, line 27: Just double checking running this locally works fine? no problems with musl because we are compiling on alpine because CGO is 0?

deployment/clouddeploy/gke-workers/environments/oss-vdb/workers.yaml
another-rex, line 10: Should we set all the envs here, even if it's the default value, just to make it easier to see all the options?

another-rex (dismissed):
Nice

go/cmd/worker/Dockerfile
michaelkedar: yeah, running locally seems to work

deployment/clouddeploy/gke-workers/environments/oss-vdb/workers.yaml
michaelkedar: added a bunch of defaults to the base yaml file

another-rex (approved):

Issue #5285, opened 2026-04-28
Create Allowlist for datasource ecosystems (and aliases?)


michaelkedar:
We currently place a lot of trust in our data sources to not publish bad data outside of their jurisdiction.
Nearly all data sources should only actually publish for 1 or 2 ecosystems.

We should add a list of expected ecosystems to our SourceRepository / source.yaml and filter out `affected[]` for things that do not match when we ingest.
GHSA & Malicious Packages would need an exception, since we trust them to publish for most ecosystems anyway.

Similarly, we might want to restrict alias computation for some ecosystems that may be misusing the aliases field and causing unnecessary linkages between records.

another-rex:
We do want to do the aliases, though a bit more design is needed. (Broad alias allow/deny, or just specific ecosystem allow list)

I think it should be similar to ecosystems, where you have a list of allowed sources.

PR #5332, opened 2026-05-11
feat: create allowlist of accepted ecosystems per datasource

michaelkedar:
#5285
(merge after #5330)
Adding a new `accepted_ecosystems` list to `SourceRepository` to prevent data sources from publishing vulnerabilities to OSV.dev that are outside of their scopes.
Did my best to go through and add all the ecosystems to the source yaml files - I don't think I've missed any.


source.yaml
```diff
@@ -281,6 +300,7 @@
   type: 2
   rest_api_url: 'https://cna.erlef.org/osv/all.json'
   db_prefix: ['EEF-']
+  accepted_ecosystems: ['Hex']
```
jess-lowe, line 303: do we need a note that they also do Git or are we happy to exclude that?

go/internal/database/datastore/internal/validate/validate.py
```diff
@@ -132,6 +132,7 @@ def main() -> int:
       extension='.json',
       key_path='vulnerability',
       ignore_git=False,
+      accepted_ecosystems=['Go', 'PyPI'],
```
jess-lowe, line 135: Why is this Go and PyPi?

michaelkedar: Just wanted to pick two random ecosystems to make sure the repeated field gets parsed properly

source.yaml
michaelkedar: Git is kinda separate from ecosystems, and this filter doesn't really deal with it anyway.
I think `ignore_git` is probably meant to capture this (but why is it missing from eef?)

another-rex (dismissed):

jess-lowe (approved):

PR #5338, opened 2026-05-11
feat(exporter): skip GCS upload when object CRC32C is unchanged

### LINKED ISSUE #3513: Exporter should check the crc hash before uploading

SanskaarUndale21:
## Summary

Fixes #3513.

Before this change, the exporter uploaded every output file on every run regardless of whether the content had changed. Since all.zip and other generated files became reproducible in #3491, unchanged runs would still create new object generations in the bucket -- making it harder for downstream consumers to detect real updates and generating unnecessary storage churn.

**Approach:** before each GCS write the writer now fetches the existing object's attributes via `ReadObjectAttrs` and computes the CRC32C checksum of the outgoing data using the Castagnoli polynomial (the same algorithm GCS itself uses). If the checksums match, the upload is skipped and an info log is emitted. Three edge cases fall through to the normal upload path:
- Object does not exist yet (`ErrNotFound`) -- upload always proceeds
- Any transient error reading attrs -- upload proceeds, a warning is logged
- Checksums differ -- upload proceeds as usual

The local-write path (used for dev/test via `-upload-to-gcs=false`) is unchanged, since writes there carry no network cost.

## Changes

- `go/cmd/exporter/writer.go`: add `gcsContentUnchanged` helper and call it before `WriteObject`
- `go/cmd/exporter/writer_test.go`: new test file covering skip-on-unchanged, upload-on-changed, upload-of-new-object, and multi-file skip

## Test plan

- [x] `TestWriter_GCS_SkipsUnchangedContent` -- same data, generation must not increment
- [x] `TestWriter_GCS_UploadsChangedContent` -- different data, generation must increment
- [x] `TestWriter_GCS_UploadsNewObject` -- no pre-existing object, must be created at generation 1
- [x] `TestWriter_GCS_SkipsMultipleUnchanged` -- three unchanged files, all generations stable

### DIFF
--- a/go/.golangci.yaml
+++ b/go/.golangci.yaml
@@ -122,6 +122,7 @@ linters:
       - path: _test\.go
         linters:
           - dupl
+          - unparam
       - path-except: _test\.go
         text: use `testutility.GetCurrentWorkingDirectory`
     paths:
--- a/go/.golangci.yaml
+++ b/go/.golangci.yaml
@@ -122,7 +122,6 @@ linters:
       - path: _test\.go
         linters:
           - dupl
-          - unparam
       - path-except: _test\.go
         text: use `testutility.GetCurrentWorkingDirectory`
     paths:
--- a/go/cmd/exporter/writer.go
+++ b/go/cmd/exporter/writer.go
@@ -81,10 +81,12 @@ func gcsContentUnchanged(ctx context.Context, client clients.CloudStorage, path 
 		if !errors.Is(err, clients.ErrNotFound) {
 			logger.WarnContext(ctx, "failed to read object attrs, proceeding with upload", slog.String("path", path), slog.Any("err", err))
 		}
+
 		return false
 	}
 	if attrs.CRC32C == crc32.Checksum(data, crc32cTable) {
 		logger.InfoContext(ctx, "skipping upload, content unchanged", slog.String("path", path))
+
 		return true
 	}
 	return false
--- a/go/cmd/exporter/writer.go
+++ b/go/cmd/exporter/writer.go
@@ -2,6 +2,8 @@ package main
 
 import (
 	"context"
+	"errors"
+	"hash/crc32"
 	"log/slog"
 	"os"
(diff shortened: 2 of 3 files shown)


SanskaarUndale21:
Hi, I've pushed a follow-up commit fixing the lint failures from the previous CI run. 

Could someone please run /gcbrun and re-review when you get a chance? Thanks!

go/.golangci.yaml
```diff
@@ -121,6 +121,7 @@ linters:
       - path: _test\.go
         linters:
           - dupl
+          - unparam
```
michaelkedar: I'd prefer we just fix the linter complaint than add more exceptions - could we just have `runWriter` always use "out" instead of taking in `pathPrefix`? (or even just inline `//nolint:unparam`)

michaelkedar (commented):
This LGTM - just need to fix the lints again

go/.golangci.yaml
SanskaarUndale21: Ok i will look into

SanskaarUndale21:
@michaelkedar

Hi, addressed the feedback -- removed the `unparam` golangci.yaml exclusion and refactored `runWriter` to drop the `pathPrefix` parameter, hardcoding `"out"` directly. Also fixed the `nlreturn` blank line that came up in the latest lint run. Could you please run `/gcbrun` when you get a chance?

i synced my fork to latest changes and rebased it aswell

 Thanks!

michaelkedar (approved):
Thanks for this.
I'll merge this tomorrow after we do our weekly deployment just to give it some time to test on the testing environment

PR #5340, opened 2026-05-12
feat: add workflow to create go generate PRs

michaelkedar:
There is a `go generate` command to (for now) copy the osv-schema JSON file from the submodule into the `go/` folder (so it may be embedded).
We want to keep this up to date, so automatically make PRs to update this.

### DIFF
--- a/.github/workflows/go-generate.yml
+++ b/.github/workflows/go-generate.yml
@@ -1,0 +1,54 @@
+name: Go Generate
+
+on:
+  push:
+    branches: [ master ]
+    paths:
+      - 'go/**'
+      - 'osv/osv-schema'
+  workflow_dispatch:
+
+concurrency:
+  # Pushing new changes to a branch will cancel any in-progress CI runs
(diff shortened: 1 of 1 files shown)


.github/workflows/go-generate.yml
```diff
@@ -0,0 +1,54 @@
...
+      - name: Check out code
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          submodules: recursive
+          persist-credentials: false
+
+      - name: Set up Go
+        uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
+        with:
+          go-version: '1.26.2'
```
another-rex: Should this just be "stable"?

another-rex (dismissed):

.github/workflows/go-generate.yml
michaelkedar: sure

another-rex (approved):

PR #5344, opened 2026-05-12
chore(website): remove unused osv scanner ignore

ashmod:
Website has migrated away from spicy-sections so this ignore should be no longer needed.


another-rex (approved):
Thanks

PR #5364, opened 2026-05-13
fix: bare except clauses, noisy logging, and dead type annotation in API server and frontend handlers

### LINKED ISSUE #5363: bug: bare `except` clauses in API server silently swallow SystemExit/KeyboardInterrupt
### **Describe the bug**
_match_versions and _match_events in gcp/api/server.py use bare except: clauses which catch all exceptions including SystemExit and KeyboardInterrupt. This can silently swallow fatal signals, making the server appear to hang rather than exit cleanly. Additionally, the same blocks log at ERROR level for a case the existing TODO(michaelkedar): This log is noisy comment acknowledges is expected — and a dead source_path: str type annotation in frontend_handlers.py looks like an uninitialized variable but is actually already bound earlier in the function.

### **To Reproduce**

Trigger a KeyboardInterrupt or SystemExit while _match_versions or _match_events is executing in gcp/api/server.py
The exception is silently caught by the bare except: block and swallowed — the server does not exit as expected
Expected behaviour

Bare except: should be except Exception: so fatal signals are not caught
The ecosystem helper failure should log at DEBUG, not ERROR, since it is an expected/recoverable case
The dead source_path: str annotation in add_source_info (frontend_handlers.py ~line 532) should be removed to avoid misleading readers into thinking the variable is uninitialized

### **Additional context**

**_Affected locations:_**

gcp/api/server.py — _match_versions (~lines 1035, 1045) and _match_events (~lines 1082, 1092)
gcp/website/frontend_handlers.py — add_source_info (~line 532)

Saatvik-GT:
## Summary

- Replace 4 bare `except:` blocks in `_match_versions` and `_match_events` (`gcp/api/server.py`) 
  with `except Exception:` to prevent silently swallowing `SystemExit` and `KeyboardInterrupt`
- Downgrade `logging.error` → `logging.debug` for the ecosystem helper failure path, 
  addressing the existing `TODO(michaelkedar): This log is noisy` comments
- Remove dead `source_path: str` type annotation in `add_source_info` 
  (`gcp/website/frontend_handlers.py`) — variable is already bound via `partition(':')` 
  on line 517, making the annotation misleading

## Files Changed

- `gcp/api/server.py` — `_match_versions`, `_match_events`
- `gcp/website/frontend_handlers.py` — `add_source_info`

## Testing

No behavior change intended. The `except Exception:` narrowing is strictly safer and 
the log level change matches the intent described in the existing TODO comment.


michaelkedar (approved):
Thanks!

Saatvik-GT:
> Thanks!

Thank you for the approval! I don't have access to the GCB logs, is there anything I can do on my end to fix the snapshot test failure, or is it something that needs to be handled internally?

michaelkedar:
> > Thanks!
> 
> Thank you for the approval! I don't have access to the GCB logs, is there anything I can do on my end to fix the snapshot test failure, or is it something that needs to be handled internally?

Sorry this is an unrelated failure. I will merge this PR when we've dealt with it

PR #5367, opened 2026-05-14
chore(gitter): log revamps

Ly-Joey:
- Reduce log noise
- Add ref_id field for easier tracing and get vuln ID from worker

### DIFF
--- a/go/cmd/gitter/gitter.go
+++ b/go/cmd/gitter/gitter.go
@@ -105,7 +105,6 @@ func separateEvents(events []*pb.Event) (*SeparatedEvents, error) {
 	}
 
 	if len(se.Limit) > 0 && (len(se.Fixed) > 0 || len(se.LastAffected) > 0) {
-
 		return nil, errors.New("limit and fixed/last_affected shouldn't exist in the same request")
 	}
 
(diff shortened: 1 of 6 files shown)


go/cmd/gitter/gitter.go
```diff
@@ -701,18 +700,36 @@ func main() {
...
+	logger.DebugContext(ctx, "Received request: /git", slog.Bool("forceUpdate", forceUpdate), slog.String("remoteAddr", r.RemoteAddr))
 
 	// Fetch repo first
 	if err := doFetch(ctx, w, repoURL, forceUpdate); err != nil {
+		if isAuthError(err) {
+			statusCode = http.StatusForbidden
+		} else if isNotFoundError(err) {
+			statusCode = http.StatusNotFound
+		} else {
+			statusCode = http.StatusInternalServerError
```
another-rex, line 720: Are these status codes being used in the response? I.e. I'm not seeing a http.Error() here.

Ly-Joey: They were in doFetch() before but I've moved them to the handlers.

another-rex (approved):

PR #5370, opened 2026-05-14
fix: send protos from recoverer

michaelkedar:
The go worker now requires the vuln to be sent over the pub/sub message.
The recoverer was sending messages without these, causing problems.
Upgrading python to get the compression.zstd was too annoying, so I'm allowing uncompressed Protos to be sent to the worker now.

### DIFF
--- a/gcp/workers/recoverer/recoverer.py
+++ b/gcp/workers/recoverer/recoverer.py
@@ -16,13 +16,16 @@
 
 import base64
 import datetime
+import functools
 import logging
 import os
+import requests
 import sys
 import time
 
 from google.cloud import ndb
 from google.cloud import pubsub_v1
+from google.cloud import storage
 
 import osv
 import osv.models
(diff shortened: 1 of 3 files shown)


gcp/workers/recoverer/recoverer.py
```diff
@@ -123,6 +141,69 @@ def handle_gcs_missing(message: pubsub_v1.types.PubsubMessage) -> bool:
...
+    raw_data = response.content
+  elif source_repo.type == osv.SourceRepositoryType.BUCKET:
+    bucket = storage_client().bucket(source_repo.bucket)
+    blob = bucket.blob(vuln_path)
+    raw_data = blob.download_as_bytes()
+  elif source_repo.type == osv.SourceRepositoryType.GIT:
+    # Cheeky workaround: all of our GIT repos are on GitHub
+    # So download the files directly from the GitHub API
+    # instead of cloning the repos.
+    org, _, repo = source_repo.repo_url.removeprefix(
```
another-rex, line 182: hmm... something something move logic to gitter..., in the future. we can share the functionality between importer and recoverer as well

another-rex (approved):

PR #5381, opened 2026-05-18
fix(vulnfeeds): use the cache for commits in refs too

jess-lowe:
it seems extracting commits from the refs was not utilising the repo tags cache for looking up canonical links. 

Also, we were not errorring and exiting upon 429 responses

### DIFF
--- a/vulnfeeds/conversion/versions.go
+++ b/vulnfeeds/conversion/versions.go
@@ -571,12 +571,7 @@ func ExtractGitCommit(link string, httpClient *http.Client, depth int, cache git
 	commit = c
 
 	// If URL doesn't validate, treat it as linkrot.
-	var possiblyDifferentLink string
-	if cache != nil {
-		possiblyDifferentLink, err = git.FindCanonicalLink(link, httpClient, cache)
-	} else {
-		possiblyDifferentLink, err = git.ValidateAndCanonicalizeLink(link, httpClient)
-	}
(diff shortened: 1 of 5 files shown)


vulnfeeds/conversion/versions.go
```diff
@@ -567,7 +571,12 @@ func ExtractGitCommit(link string, httpClient *http.Client, depth int) (string,
 	commit = c
 
 	// If URL doesn't validate, treat it as linkrot.
-	possiblyDifferentLink, err := git.ValidateAndCanonicalizeLink(link, httpClient)
+	var possiblyDifferentLink string
+	if cache != nil {
+		possiblyDifferentLink, err = git.FindCanonicalLink(link, httpClient, cache)
+	} else {
+		possiblyDifferentLink, err = git.ValidateAndCanonicalizeLink(link, httpClient)
```
another-rex: Why are these two separate functions? I would have expected this to be the same function that does something slightly different (save vs not save) when the cache passed in is nil or not?

jess-lowe: second one is the legacy one. I can make this clearer.

another-rex: Ah I see, looked at the impl of FindCanonicalLink. Let's move this if statement inside FindCanonicalLink.

another-rex (approved):

PR #5386, opened 2026-05-18
fix(combine-to-osv): remove last affected if fixed exists

jess-lowe:


another-rex (commented):
Please add a rational in the comments and PR description

PR #5387, opened 2026-05-18
feat(frontend): use URL search params to quickly load a record for triage

jess-lowe:
I wanna quickly look up a record triage output sometimes - this should make it a lot easier


another-rex (dismissed):
When the load button is clicked it should probably populate the url with the query+ID as well.

another-rex (approved):

PR #5389, opened 2026-05-18
feat(nvd): retry 429ed records after conversion is completed

jess-lowe:
- Updated RepoTags so that temporary 429 (rate limit / too many requests) errors only mark a repository as invalid for ~ 1 hr.


another-rex (commented):
Let's just do the first point. And rather than not setting it as invalid in the cache, set it as invalid, just with a shorter timeout  than the standard 1 week. (Maybe just 1 hour is good enough, just so we are not spamming requests when multiple records relies on 1 429 host)

It'll run again in 12 hours and pick it back up.

another-rex (approved):
Probably need to update the PR description. Otherwise LGTM!

PR #5331, opened 2026-05-11
fix(PyPi): nil pointer and index out-of-bounds panics in PyPI converter

jess-lowe:
Fix #5142  

#### Core bug issue
The PyPI converter crashed with a panic when running `loadExisting` if:
1. The YAML/JSON unmarshaler tried to decode into a newly initialized `Vulnerability` struct, because its embedded `*osvschema.Vulnerability` pointer remained `nil`.
2. An existing YAML file possessed an empty `affected` block, causing `vuln.GetAffected()[0]` to raise an index out-of-range panic.
3. The first element in `affected` had no package information, triggering a nil pointer dereference when fetching the package name.

#### Fix
- **Initialize Embedded Struct Pointer:** Updated `FromYAML` and `FromJSON` in `vulnfeeds/vulns/vulns.go` to explicitly initialize the inner `*osvschema.Vulnerability` pointer before parsing.
- **Safe Iteration and Nil Checks:** Refactored the walker loop in `loadExisting` inside `vulnfeeds/cmd/pypi/main.go` to iterate over `vuln.GetAffected()` and safely check for `nil` package pointers and empty package names.
- **Added Regression Tests:** Created `vulnfeeds/cmd/pypi/main_test.go` to cover parsing files with valid, empty, or incomplete `affected` blocks.

PR #5393, opened 2026-05-18
fix(vulnfeeds): output standard OSV YAML instead of wrapped format

### LINKED ISSUE #5390: `ids` command is failing to parse some advisories
**Describe the bug**
Since May 17, 11:44 PM EDT, the `github.com/google/osv/vulnfeeds/cmd/ids@latest` command has begun failing to parse some advisories in https://github.com/pypa/advisory-database/ with:

> `INFO:   Failed to assign IDs err=failed to assign ID: failed to parse yaml: proto: (line 1:2): unknown field "vulnerability"`

As a result, the workflow to assign new IDs is failing.

**To Reproduce**
See https://github.com/pypa/advisory-database/actions/workflows/automation.yaml

**Expected behaviour**
This workflow was passing until recently. Should the advisories wiht the `vulnerability` field be updated, or is the command broken?

Ref: https://github.com/pypa/advisory-database/issues/283

di:
Introduced in https://github.com/google/osv.dev/pull/5331, fixes #5390.

yaml.v2 treats the anonymous `*osvschema.Vulnerability` field in the `vulns.Vulnerability` wrapper struct as a named `Vulnerability` key. This caused new advisories to be written with a top-level `vulnerability:` field after a recent bugfix initialized the pointer.

This commit updates `ToYAML`, `FromYAML`, and `FromJSON` to encode and decode the inner `osvschema.Vulnerability` struct directly, restoring the standard OSV output format while maintaining the fix for nil pointer panics.

### DIFF
--- a/vulnfeeds/cmd/pypi/main_test.go
+++ b/vulnfeeds/cmd/pypi/main_test.go
@@ -25,14 +25,13 @@ func TestLoadExisting(t *testing.T) {
 
 	// 1. Write a valid vulnerability YAML
 	validYaml := `
-vulnerability:
-  id: PYSEC-2021-123
-  affected:
-    - package:
-        name: foo-pkg
-        ecosystem: PyPI
-  aliases:
-    - CVE-2021-12345
+id: PYSEC-2021-123
+affected:
+  - package:
+      name: foo-pkg
+      ecosystem: PyPI
+aliases:
+  - CVE-2021-12345
 `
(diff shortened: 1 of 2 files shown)


jess-lowe (approved):
LGTM

Note to self: I'm pretty sure only the PyPi converter uses FromYAML, and FromJSON doesn't seem to be used elsewhere at all.

PR #5394, opened 2026-05-18
feat(frontend): add collapsible fields to the json on the triage page

jess-lowe:
uses the json-formatter-js dependency to add collapsible fields to the json (uncollapsed by default)

Built on top of the changes in #5387, so probably merge that first.

image


another-rex (commented):
What if the data is not json?

another-rex (approved):

PR #5421, opened 2026-05-20
fix: importer-reconciler actually checkout branch

another-rex:
Actually checkout the main/master branch when there is no specific branch specified  in the reconciler.

Also fixing tests by:

Isolates `HOME` and `XDG_CONFIG_HOME` to a clean temporary directory package-wide for the `importer` package. This ensures host machine Git configurations (such as `commit.gpgSign`) don't interfere with tests, resolving failures in `TestGitSourceRecord_Open` and `TestHandleImportGit` on environments with those custom configs.

### DIFF
--- a/go/internal/importer/git.go
+++ b/go/internal/importer/git.go
@@ -285,20 +285,27 @@ func handleReconcileGit(ctx context.Context, ch chan<- WorkItem, config Config, 
 			return nil, err
 		}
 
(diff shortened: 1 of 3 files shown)


michaelkedar (approved):
Thoughts on having a helper function (instead of in TestMain) that calls `t.TempDir()` and `t.SetEnv()`?

PR #5424, opened 2026-05-20
fix(vulnfeeds): Modify valid versions output logic in versions.go

jess-lowe:
Add condition to output valid versions only if there are errors.

otherwise this prevents the PyPi advisory database from producing records with successful ranges...


Ly-Joey (approved):

woodruffw:
Is this related to https://github.com/pypa/advisory-database/issues/284? We're seeing some reports of invalid OSV records in PYSEC, via https://github.com/astral-sh/uv/issues/19492.

woodruffw:
https://github.com/google/osv.dev/pull/5399 also seems potentially related.

PR #5425, opened 2026-05-20
refactor: unify go Dockerfiles and begin setting up Go OSV API

michaelkedar:
Wanted  to set up the things needed for migrating the API to go (including porting code for running REST API locally).
Need to import the API grpc definitions from `bindings/go`, which required an `replace` directive in `go.mod` to keep them synced.
The `replace` directive means that every go image built from there needs the bindings folder in its Docker context.
To make this easier to maintain, unified nearly all `go/` Dockerfiles into a multi-target Dockerfile that they can all build from.


go/cmd/api-devserver/main.go
another-rex, line 1: Please catch ctrl C

go/cmd/api-devserver/main.go
```diff
@@ -0,0 +1,151 @@
...
+	if err := run(); err != nil {
+		os.Exit(1)
+	}
+}
+
+func run() error {
+	logger.InitGlobalLogger()
+	defer logger.Close()
+
+	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
```
another-rex: Please use syscall.SIGINT

AGENTS.md
```diff
@@ -160,44 +160,68 @@ Many tests use expected outputs saved directly in the source tree:
...
 
 ## Go Component Architecture (`go/`)
 The Go component contains the active and migrated services for the OSV database. It is structured with executables in `cmd/` and shared libraries in `internal/`.
 
+### Go Monorepo Docker Builds
+All Go microservices are compiled using a single, unified multi-target Dockerfile (`go/Dockerfile`) which shares the workspace compilation setup.
+- Due to the `replace` directive in `go/go.mod` pointing to the sibling `bindings/` library, Go Docker builds **must** mount the `bindings/` folder into the build context.
+- Since the production Cloud Build steps run inside `dir: 'go'`, they leverage BuildKit's `--build-context` flag to cleanly map the sibling folder without importing other monorepo files:
+  ```bash
+  docker build -t osv/importer --target importer --build-context bindings=../bindings -f Dockerfile .
```
another-rex, line 179: Can you add to the general contribution docs with these instructions so we know how to build the docker image.

(Also should we have a build docker image command in our makefile?)

go/cmd/api-devserver/main.go
another-rex, line 1: Hmm how fast does esp start up / stop?

One thing that might be useful (which we have for the python api) is to be able to startup the esp part separately from the server. (So I can do fast iterations on the server while keeping the ESP up)

michaelkedar: It seems to be fairly fast - about 2 seconds.

> One thing that might be useful (which we have for the python api) is to be able to startup the esp part separately from the server. (So I can do fast iterations on the server while keeping the ESP up)

You mean maybe having a [`--no-backend` flag](https://github.com/google/osv.dev/pull/5425/changes#diff-74e4c54795f213d9742eb60fcb7a786ec52bf5dbffee3ad4791c654ca2dd7e26R42) 🙂

AGENTS.md
michaelkedar: Added to CONTRIBUTING.md and also to the top of the Dockerfile itself.

We shouldn't really need to be building the images themselves regularly since they're all pretty lightweight - basically equivalent to just `go run`-ing the equivalent command. I don't really think we'd need a Makefile target dedicated to these cases.

another-rex (approved):

PR #5427, opened 2026-05-20
fix(vulnfeeds): ToYAML round-tripping

woodruffw:
This is an attempt to fix some downstream breakage that we've been observing in `uv audit`. The underlying cause appears to be a regression in YAML roundtripping within `vulnfeeds`.

TL;DR: vulnfeeds generates OSV records as protobuf-backed `osvschema.Vulnerability` structs. Those structs only carry `protobuf` and `json` struct tags, not `yaml` tags. `vulns.ToYAML` was serializing them by handing the struct straight to a generic YAML encoder (`gopkg.in/yaml.v2`), which doesn't understand protobuf naming and just lowercases the Go field names. So an event field `LastAffected` got written as `lastaffected:` instead of the OSV schema's `last_affected:`, among other things.

That didn't cause problems until #5397 changed `vulns.FromYAML` to decode via `protojson` with `DiscardUnknown: true`. `protojson` expects the real schema name `last_affected` and silently throws away the unrecognized `lastaffected` key. So on the round trip that the PyPI pipeline does (`cmd/pypi` writes with `ToYAML`, `cmd/ids` reads with `FromYAML`), the `last_affected` value was dropped. An event whose only key was `last_affected` decoded into a completely empty `Event{}`, which then re-serialized to `{}`, violating the schema.

I've added a test that demonstrates the fixed behavior, and I had an LLM produce a reproducer [here](https://gist.github.com/woodruffw/2c4a62b2bff3df6cb09cd9b743e9e7a5).

Note: I'm not particularly experienced in Go, so this could use a sanity check for idioms (and correctness with respect to the larger codebase). It would also be ideal (IMO) to have a stronger regression backstop here; this breakage was pretty disruptive to a slice of the Python ecosystem that depends on `uv audit` and other OSV-aware auditing tools 🙂 

Refs:

- https://github.com/pypa/advisory-database/issues/284
- https://github.com/astral-sh/uv/issues/19492

CC @jess-lowe @another-rex

### DIFF
--- a/vulnfeeds/vulns/vulns.go
+++ b/vulnfeeds/vulns/vulns.go
@@ -33,7 +33,6 @@ import (
 	goccyyaml "github.com/goccy/go-yaml"
 	"google.golang.org/protobuf/encoding/protojson"
 	"google.golang.org/protobuf/types/known/timestamppb"
-	"gopkg.in/yaml.v2"
 
 	"github.com/google/osv/vulnfeeds/models"
 	"github.com/google/osv/vulnfeeds/utility"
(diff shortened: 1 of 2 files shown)


jess-lowe (approved):

jess-lowe:
I've been working on this a few days and I have a similar fix locally, so I'll merge this to hopefully start fixing some of the proto problems. I think I'm going to have to retroactively generate the ranges for some of the records produced over the past few days, as well as fixing some bad mapping issues.

woodruffw:
Thank you @jess-lowe!

Yeah, if I understand correctly, this will require some manual fixups on PYSEC entries, right? If so I'll do that now.

jess-lowe:
@woodruffw I'm trying to run the converter locally to just regenerate the records from the last couple of days, and then I'll make a pull request to the repo if that helps? Hopefully will reduce the manual work required. 

After this is all happy again, I'm going to try move the PyPi conversion code in the OSV.dev repo into its own isolated folder so we can version it, and pin the dependencies properly.

woodruffw:
> @woodruffw I'm trying to run the converter locally to just regenerate the records from the last couple of days, and then I'll make a pull request to the repo if that helps? Hopefully will reduce the manual work required.

That would be great, thank you! I'm currently working on a little script that pulls the `lastaffected` fields out of history, but I think your PR is likely to be more correct than what I'm hacking together.

> After this is all happy again, I'm going to try move the PyPi conversion code in the OSV.dev repo into its own isolated folder so we can version it, and pin the dependencies properly.

Thanks, I think that would be great 🙂

woodruffw:
https://github.com/pypa/advisory-database/pull/285 is my approach to that, although I'm intentionally leaving it drafted since I think it's not going to be as good/high-quality as re-generating.

PR #5437, opened 2026-05-25
feat: create new work pools for reimports & cves

michaelkedar:
Set up new pub/sub subscriptions & workers for handling re-imports and CVE imports (though, this still needs to be configured on the `SourceRepository`) separate from regular imports.

The auto-scaling is the same as the regular workers, which isn't great. We should adjust these later.


another-rex (dismissed):
Please add to the PR description that we're keeping the scaling the same  for now and adjusting later

another-rex (approved):

PR #5440, opened 2026-05-26
fix(importer): clean untracked files after checkout in reconciler

another-rex:
In `handleReconcileGit`, after checking out the target branch, clean up any untracked files and directories to ensure a clean state and avoid processing stale/deleted files.

This extends commit 828810a9700f5da00193f8d3512e10758b688685.

### Key Changes:
*   **System Git Refactor**: Refactored `handleReconcileGit` to use system `git checkout` and `git clean -fdx` commands instead of `go-git` library calls. This is much more robust and avoids potential out-of-memory or timeout issues with `go-git`'s `Status()` on large repositories when deployed.
*   **Concurrency Fix**: Moved checkout and clean out of `repoGroup.Do` and protected them with `repo.mu` lock to avoid incorrect state when concurrent imports and reconciles occur.
*   **Testing**: Added `TestHandleReconcileGit_CleanUntracked` to verify that untracked files and directories are correctly cleaned up.

agy

### DIFF
--- a/go/internal/importer/git_test.go
+++ b/go/internal/importer/git_test.go
@@ -1,6 +1,8 @@
 package importer
 
 import (
+	"crypto/sha256"
+	"encoding/hex"
 	"io"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


go/internal/importer/git_test.go
another-rex, line 1: I'm not sure whether we should even have this test.

michaelkedar (approved):

PR #5441, opened 2026-05-26
test(importer): update CleanUntracked test to cover complex subdir structures

another-rex:
This PR updates the `CleanUntracked` test to cover complex subdirectory structures, ensuring that untracked files and directories inside both root and tracked/untracked subdirectories are correctly cleaned up.

This is a follow-up to #5440 to ensure thorough test coverage of the `go-git` `wt.Clean` implementation.

agy


another-rex:
This test passes, but it still doesn't fix the issue with it not doing what it's supposed to (cleaning up untracked items) in prod. I'm a bit confused about what is going on.

jess-lowe (approved):

PR #5442, opened 2026-05-26
docs: Clarify querybatch response mapping

### LINKED ISSUE #466: Clarify documentation of return format for /v1/querybatch API
According to the Swagger documentation at https://osv.dev/docs/#operation/OSV_QueryAffectedBatch the endpoint `https://api.osv.dev/v1/querybatch` should return a list of `osvVulnerability`.

Instead it returns a list of vulnerability IDs without any vulnerability details or the actual link to the package the vulnerability belongs to.

This leads to the following issues:
- The batch endpoint does not return the OSV format, leading to clients having a need to query those vulnerability IDs. This completely goes against the idea of having a batch endpoint where you get all info for a set of libraries in a single HTTP request so that you do not overload the service.
- It is unclear and undocumented how the current response from the batch endpoint should be used. How to link these vulnerability IDs to the actual packages that were provided in the request?
- The documentation and implementation are out of sync

Example:
```
cat <<EOF | curl -X POST -d @- "https://api.osv.dev/v1/querybatch"
{
  "queries": [
    {
      "package": {
        "ecosystem": "Maven",
        "name": "org.apache.logging.log4j:log4j-core"
      },
      "version": "2.13.0"
    },
    {
      "package": {
        "ecosystem": "Packagist",
        "name": "noumo/easyii"
      },
      "version": "0.8"
    }
  ]
}
EOF
```

Returns:
```
{
    "results": [
        {
            "vulns": [
                {
                    "id": "GHSA-7rjr-3q55-vv33"
                },
                {
                    "id": "GHSA-jfh8-c2jp-5v3q"
                },
                {
                    "id": "GHSA-p6xc-xr62-6r2g"
                },
                {
                    "id": "GHSA-vwqq-5vrc-xw9h"
                }
            ]
        },
        {}
    ]
}
```

With this approach, the current service is almost impossible to use without hitting rate limits.

Haihan-Jiang:
Clarifies the /v1/querybatch response shape:

- each result maps to the query at the same index
- an empty result object means that query has no matched vulnerabilities
- returned vuln entries are id/modified summaries, not full OSV records
- clients should fetch full records with GET /v1/vulns/{id} and can use id + modified for cache freshness

Tests:
- git diff --check
- test -f docs/api/get-v1-vulns.md

Not run:
- bundle exec jekyll build; the local system Ruby is 2.6 and the docs lockfile requires Bundler 2.4.20.

Fixes #466.

### DIFF
--- a/docs/api/post-v1-querybatch.md
+++ b/docs/api/post-v1-querybatch.md
@@ -8,6 +8,14 @@ nav_order: 3
 # POST /v1/querybatch
 Query for multiple packages (by either package and version or git commit hash) at once. Returns vulnerability ids and modified field only. The response ordering will be guaranteed to match the input.
 
+Each item in `results` corresponds to the query at the same index in `queries`.
+If a query has no matched vulnerabilities, its corresponding result entry may
+be an empty object. The `vulns` entries are intentionally summaries containing
+`id` and `modified`, not full vulnerability records. To retrieve the complete
+OSV record, call [GET /v1/vulns/{id}](get-v1-vulns.md) for the returned IDs;
+clients can use the `id` and `modified` pair to decide whether a cached record
+is still fresh.
+
 {: .no_toc }
 
 <details open markdown="block">


G-Rath:
Please [read and follow our contributing guide](https://github.com/google/osv.dev/blob/master/CONTRIBUTING.md#code-reviews) before opening pull requests

PR #5457, opened 2026-05-27
feat(nvd): differentiate between CPE ranges and string sources

jess-lowe:
Versions originating from CPE ranges vs the CPE match criteria string are of vastly different qualities. 

OSV favours ranges, whereas CPE strings usually represent a single affected version. Being able to differentiate between these two origins should help make better informed decisions in the combine-to-osv merging. 

New version sources constants: `CPE_RANGE` and `CPE_STRING` 
   - Version information extracted using NVD's explicit boundary fields (`VersionStartIncluding`, `VersionStartExcluding`, `VersionEndExcluding`, or `VersionEndIncluding`) is categorized under `CPE_RANGE`.
   - Versions extracted from the match criteria CPE string itself (`CPE.Version`) are categorized under `CPE_STRING`.

Refactored `ProcessRanges` in common.go now extracts and logs the actual, distinct sources of processed ranges dynamically rather than using the generic `"CPE_FIELD"` fallback parameter.

### DIFF
--- a/vulnfeeds/conversion/common.go
+++ b/vulnfeeds/conversion/common.go
@@ -654,7 +654,7 @@ func AddFieldToDatabaseSpecific(ds *structpb.Struct, field string, value any) er
 }
 
 // ProcessRanges attempts to resolve the given ranges to commits and updates the metrics accordingly.
-func ProcessRanges(ranges []models.RangeWithMetadata, repos []string, metrics *models.ConversionMetrics, cache git.RepoTagsCache, source models.VersionSource) ([]models.RangeWithMetadata, []models.RangeWithMetadata, []string) {
+func ProcessRanges(ranges []models.RangeWithMetadata, repos []string, metrics *models.ConversionMetrics, cache git.RepoTagsCache) ([]models.RangeWithMetadata, []models.RangeWithMetadata, []string) {
 	if len(ranges) == 0 {
 		return nil, nil, nil
 	}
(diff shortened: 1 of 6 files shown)


vulnfeeds/conversion/common.go
```diff
@@ -672,7 +672,24 @@ func ProcessRanges(ranges []models.RangeWithMetadata, repos []string, metrics *m
 		}
 	}
 
-	metrics.VersionSources = append(metrics.VersionSources, source)
+	// Dynamically record the precise sources from each processed range's metadata.
+	// This ensures that granular version sources (such as CPE-RANGE or CPE-STRING) are tracked in
+	// the final conversion metrics instead of a single generic fallback source.
+	for _, rng := range ranges {
```
another-rex: Can you rename rng?

vulnfeeds/conversion/common.go
```diff
@@ -672,7 +672,24 @@ func ProcessRanges(ranges []models.RangeWithMetadata, repos []string, metrics *m
...
+		if rng.Metadata.Source == "" {
+			continue
+		}
+		found := false
+		for _, s := range metrics.VersionSources {
+			if s == rng.Metadata.Source {
+				found = true
+				break
+			}
+		}
```
another-rex: I think this can be shortened to slices.Contains or something like that.

vulnfeeds/conversion/common.go
```diff
@@ -672,7 +672,24 @@ func ProcessRanges(ranges []models.RangeWithMetadata, repos []string, metrics *m
...
+		if !found {
+			metrics.VersionSources = append(metrics.VersionSources, rng.Metadata.Source)
```
another-rex: Since this is inside the ranges loop, wouldn't this be called a lot of times per loop and give you duplicate metadata.Sources ?

another-rex (approved):

PR #5478, opened 2026-05-27
chore(deps): manually update go deps

michaelkedar:
Fix all the osv-scanner-reported vulns. I can't actually find a job from renovate to do these updates.


another-rex (approved):
hmm you would think that's what lockfile maintenance  should be doing?

PR #5489, opened 2026-06-02
refactor(api): query and querybatch in go

michaelkedar:
Big PR for implementing Query / QueryBatch endpoints in go.

TODO in a follow-up PR:
- Attempt to parse page_token from python
- Send pubsub message when we find orphaned affected versions
- Add withdrawn field to AffectedCommits (or delete AffectedCommits of withdrawn vulns)
- Slight refactor of database stuff to allow consolidating queries in future postgres migration


go/internal/api/query_affected.go
```diff
@@ -0,0 +1,517 @@
...
+		logFields = append(logFields, slog.String("type", "invalid"))
+	} else {
+		if queryInfo.commit != "" {
+			logFields = append(logFields, slog.String("type", "commit"))
+		} else {
+			if queryInfo.fromPURL {
+				logFields = append(logFields, slog.String("type", "purl"))
+			} else {
+				logFields = append(logFields, slog.String("type", "ecosystem"))
+			}
```
another-rex: nit: set type to a var and add it at the end to make it easier to read.

go/internal/api/query_affected.go
```diff
@@ -0,0 +1,517 @@
...
+	if s.verboseLogs {
+		logger.InfoContext(ctx, "full query", slog.Any("query", params.GetQuery()))
+	}
+
+	if errors.Is(err, purl.ErrUnknownPURL) {
+		// All unsupported PURL queries would simply return a 200
+		// status code with an empty response.
+		// To avoid breaking existing behavior,
+		// we return an empty response here with no error.
+		// This needs to be revisited with a more considerate design.
```
another-rex, line 60: add a TODO with an issue here for this?

go/internal/api/query_affected.go
```diff
@@ -0,0 +1,517 @@
...
+		return &pb.VulnerabilityList{}, nil
+	}
+	if err != nil {
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+
+	estimatedSize := new(atomic.Int64)
```
another-rex: nit: Add the unit here too. (Bytes probably?)

go/internal/api/query_affected.go
```diff
@@ -0,0 +1,517 @@
...
+	queries := params.GetQuery().GetQueries()
+	queryInfos := make([]*parsedQueryInfo, len(queries))
+	for i, query := range queries {
+		info, err := s.parseQuery(query)
+		if errors.Is(err, purl.ErrUnknownPURL) {
+			// All unsupported PURL queries would simply return a 200
+			// status code with an empty response.
+			// To avoid breaking existing behavior,
+			// we return an empty response here with no error.
+			// This needs to be revisited with a more considerate design.
```
another-rex, line 119: Same as above

go/internal/api/query_affected.go
```diff
@@ -0,0 +1,517 @@
...
+			continue
+		}
+		if err != nil {
+			if firstErr == "" {
+				firstErr = fmt.Sprintf("error in query at index %d: %s", i, err.Error())
+			}
+			queryInfos[i] = nil
+			invalidQueries++
+
+			continue
```
another-rex, line 130: I don't recall python's API having this feature (it is really nice to have though), should we be documenting this in the API docs?

go/internal/api/query_affected.go
```diff
@@ -0,0 +1,517 @@
...
+	}
+	ecoGroups := make([]any, 0, len(ecosystemQueries))
+	for ecosystem, count := range ecosystemQueries {
+		ecoGroups = append(ecoGroups, slog.Int(ecosystem, count))
+	}
+	logger.InfoContext(ctx, "QueryAffectedBatch",
+		slog.Int("commit", commitQueries),
+		slog.Group("ecosystem", ecoGroups...),
+		slog.Int("purl", purlQueries),
+		slog.Int("invalid", invalidQueries))
```
another-rex: It would really help readability to separate the logging stuff out into a separate function.

another-rex: Same with queryAffected.

go/internal/api/query_affected.go
```diff
@@ -0,0 +1,517 @@
...
+		// we return an empty response here with no error.
+		// This needs to be revisited with a more considerate design.
+		return &pb.VulnerabilityList{}, nil
+	}
+	if err != nil {
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+
+	estimatedSize := new(atomic.Int64)
+	vulns, nextToken, err := s.queryAndHydrate(ctx, queryInfo, maxSingleQueryResponses, s.getSingleQueryTimeout(), s.vulnStore.Get, estimatedSize)
```
another-rex: I feel like we should rename vulnStore.Get to GetFull or something. It's a bit weird that GetModified get's less stuff than Get imo.

go/internal/api/query_affected.go
```diff
@@ -0,0 +1,517 @@
...
+
+		return func(ctx context.Context) iter.Seq2[models.MatchResult, error] {
+			return s.vulnStore.MatchCommits(ctx, commit, startTok)
+		}, nil
+	}
+	if qi.packageName != "" {
+		return func(ctx context.Context) iter.Seq2[models.MatchResult, error] {
+			return s.vulnStore.MatchPackages(ctx, qi.ecosystem, qi.packageName, qi.version, startTok)
+		}, nil
+	}
```
another-rex, line 406: Should this error if both commit and packageName is filled?

go/internal/api/query_affected.go
```diff
@@ -0,0 +1,517 @@
...
+
+	return vulns, matchResult.getPageToken(), nil
+}
+
+type matcherFunc func(context.Context) iter.Seq2[models.MatchResult, error]
+
+func (s *server) resolveMatcher(qi parsedQueryInfo) (matcherFunc, error) {
+	startTok := qi.pageToken
+	if startTok == startCursor {
```
another-rex, line 389: hmm I don't quite get what the point of startCursor is, is it just something left over from the python impl?

go/internal/api/query_affected.go
```diff
@@ -0,0 +1,517 @@
...
+	if err != nil {
+		return nil, "", err
+	}
+
+	return vulns, matchResult.getPageToken(), nil
+}
+
+type matcherFunc func(context.Context) iter.Seq2[models.MatchResult, error]
+
+func (s *server) resolveMatcher(qi parsedQueryInfo) (matcherFunc, error) {
```
another-rex, line 387: I know there's a private methods, but still probably good to add some docstrings to them (e.g. this just picks between package matcher and commit matcher.)

go/internal/api/query_affected.go
```diff
@@ -0,0 +1,517 @@
...
+	startTok string,
+	limit int,
+	timeout time.Duration,
+	cancel context.CancelCauseFunc,
+) (matcherResult, context.CancelFunc) {
+	resultIDs := make(chan matchVuln, numParallelHydration)
+	currentCursor := func() string { return startTok }
+	if startTok == "" {
+		currentCursor = func() string { return startCursor }
+	}
```
another-rex, line 432: Rather than using these inline functions, should this just be a pointer to a string? Then it'll always be correctly updated right?

go/internal/api/query_affected.go
```diff
@@ -0,0 +1,517 @@
...
+	go func() {
+		wg.Wait()
+		close(hydrated)
+	}()
+
+	return hydrated
+}
+
+func (s *server) collectAndSort(ctx context.Context,
+	hydrated <-chan hydratedResult,
```
another-rex, line 514: nit: A lot of the channel names are past tense, might be better to append ch to them so the user knows they are looping through a channel rather than an already collected loop more easily.

go/internal/api/query_affected.go
```diff
@@ -0,0 +1,517 @@
...
+				logger.ErrorContext(ctx, "failed to hydrate", slog.String("id", res.id), slog.String("error", res.err.Error()))
+			}
+			// continue to drain the channel
+			continue
+		}
+		unordered = append(unordered, res)
+		estimatedSize.Add(int64(proto.Size(res.v)))
+		if sz := estimatedSize.Load(); sz > s.getResponseSizeLimit() {
+			if !loggedSize && s.verboseLogs {
+				logger.InfoContext(ctx, "estimated response size limit reached", slog.Int64("estimatedSize", sz))
```
another-rex, line 540: do we want to keep going to see a log of the final size so we can know how much overshoot happens?

go/internal/api/query_affected.go
another-rex: Though, even with these as functions, is there a race condition potentially, I guess collect and sort prevents stops that by waiting until everything is done before grabbing the cursor, but still might be a good idea to abstract this cursor into its own struct with a mutex and pointer.

go/internal/database/datastore/vulnerability_matching.go
```diff
@@ -0,0 +1,536 @@
...
+
+	return 4
+}
+
+func sortAffectedEvents(events []AffectedEvent, ecosystem eco.Ecosystem) {
+	slices.SortFunc(events, func(a, b AffectedEvent) int {
+		pa, errA := ecosystem.Parse(a.Value)
+		pb, errB := ecosystem.Parse(b.Value)
+
+		if errA != nil || errB != nil {
```
another-rex, line 428: If any of them fail to parse, maybe it's better to re-sort the entire thing based on strings, rather than doing a single fallback, which could cause some weird results as it breaks transitivity.

go/internal/api/query_affected.go
michaelkedar: Do we *really* need to document this?
If they can read the documentation, they can read the error message.

go/internal/api/query_affected.go
michaelkedar: It's so if the API returns before even starting the query, it doesn't return an empty `next_page_token` (which would indicate the query is complete and returned 0 results).

It's still hypothetically possible in go, though very unlikely.

go/internal/api/query_affected.go
michaelkedar: The API docs actually say for `package`
> The package to query against. When a `commit` hash is given, this is optional.

IMO it probably feels better to query by repo + commit, rather than the commit itself.
I also feel like we've mentioned in using the repo in the query to filter out forks or something.

go/internal/api/query_affected.go
michaelkedar: This final size is logged in the `queryAndHydrate completed successfully` log in the handler methods.

go/internal/api/query_affected.go
michaelkedar: These are functions to avoid having to string encode every single query cursor when only the final one is needed (I don't think there's a clean way to avoid that with string pointers?).

I've changed it so that `getPageToken` blocks until the matcher is done to be a bit more defensive (though, it will deadlock if it is used incorrectly).

go/internal/api/query_affected.go
another-rex: I'm thinking they can programmatically remove the invalid queries if they want. Maybe we shouldn't encourage automatically parsing this I guess.

go/internal/api/query_affected.go
another-rex: Can you add this as a comment?

go/internal/database/datastore/vulnerability_matching.go
michaelkedar: Done - this is probably very rare given how lenient the semantic library is with comparisons.
This was copied over from some other code that the worker uses, so I've changed it there too.

go/internal/api/query_affected.go
another-rex: Good point, let's stick with this. Though there's now so many layers of function calls it's really hard to track what's calling what now especially if I'm not viewing it in an IDE.

go/internal/api/query_affected.go
michaelkedar: added to the startCursor const

another-rex (approved):
LGTM

PR #5526, opened 2026-06-10
fix(ui): improve light-theme component rendering

ashmod:
Fixes some light-theme specific quirks:

On chromium-based browsers, scrollbar always rendered in dark styling, because the `color-scheme` property was missing, additionally the `$osv-background` was not visible. 

Before/After:
home_before
home_after

---

The "Load More" spinner on the list page had a white color making it invisible on light mode so that was fixed too by setting it to `$osv-text-color` instead:

spinner_light
 
---

Finally, non-CVSS severity chips had poor contrast on light mode, showing white text on a very light gray background (osv-grey-800):
ubuntu_before

Fixed by updating the text color:
ubuntu_after

### DIFF
--- a/gcp/website/frontend3/src/styles.scss
+++ b/gcp/website/frontend3/src/styles.scss
@@ -5,6 +5,7 @@
 @use '@material/data-table/data-table-theme';
 
 :root {
+  color-scheme: dark;
   --osv-bg: #292929;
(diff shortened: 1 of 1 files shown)


Ly-Joey (approved):
Looks good! Thank you! :D

PR #5527, opened 2026-06-11
refactor: Improvements to database querying, notify missing records

michaelkedar:
- Publish a `gcs_missing` message to the recoverer if a vuln matched in datastore, but no full records was found in GCS.
- Create batch query methods on the vulnerability store interface, to allow more efficient queries to be done in a future postgres migration
- Batch `GetModified` queries together across goroutines into a single datastore `GetMulti` request to save on network round trips.

### DIFF
--- a/go/internal/database/datastore/vulnerability.go
+++ b/go/internal/database/datastore/vulnerability.go
@@ -10,11 +10,13 @@ import (
 	"slices"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"cloud.google.com/go/datastore"
 	"cloud.google.com/go/pubsub/v2"
 	"github.com/google/osv.dev/go/internal/models"
+	"github.com/google/osv.dev/go/internal/osvutil"
 	"github.com/google/osv.dev/go/logger"
 	"github.com/google/osv.dev/go/osv/clients"
 	"github.com/google/osv.dev/go/osv/ecosystem"
@@ -28,6 +30,13 @@ type VulnerabilityStore struct {
 	gcsStore             clients.CloudStorage
 	failedWritePublisher clients.Publisher
 	ecosystems           *ecosystem.Provider
+
+	// Auto-batcher fields
+	batchMu          sync.Mutex
+	pendingBatch     []*getModifiedRequest
+	batchTimeout     time.Duration
+	batchMaxElements int
+	triggerChan      chan struct{}
 }
 
 var _ models.VulnerabilityStore = (*VulnerabilityStore)(nil)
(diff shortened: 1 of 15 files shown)


go/internal/osvutil/batcher/context.go
```diff
@@ -0,0 +1,56 @@
...
+	if len(subCtxs) == 0 {
+		cancel()
+		return mergedCtx, cancel
+	}
+
+	var activeCount atomic.Int64
+	activeCount.Store(int64(len(subCtxs)))
+
+	for _, ctx := range subCtxs {
+		go func(c context.Context) {
```
another-rex: Apparently there's this function called context.AfterFunc() which can be used here instead of us manually spawning a goroutine?

another-rex: This should result in a lot less goroutines because it'll all be managed by the parent context.

go/internal/database/datastore/vulnerability.go
```diff
@@ -37,6 +40,8 @@ type VulnStoreConfig struct {
 	GCS                  clients.CloudStorage
 	FailedWritePublisher clients.Publisher
 	Ecosystems           *ecosystem.Provider
+	BatchTimeout         time.Duration
+	BatchMaxElements     int
```
another-rex, line 44: Do we want to make these configurable in main.go? (Probably as env vars?)

go/internal/osvutil/batcher/batcher_test.go
```diff
@@ -0,0 +1,231 @@
...
+
+	// Give it a tiny bit of time to get queued
+	time.Sleep(5 * time.Millisecond)
+
+	// Cancel the first request
+	cancel1()
+
+	// Request 1 should return immediately because it was cancelled
+	// We check if the goroutine for req1 finished quickly
+	c1Done := make(chan struct{})
```
another-rex: The comment say we check, but does this check? It only seems to be checking if c1Done is finished (both requests finished)

go/internal/database/datastore/vulnerability.go
```diff
@@ -131,17 +148,49 @@ func (s *VulnerabilityStore) GetFull(ctx context.Context, id string) (*osvschema
...
-			return time.Time{}, models.ErrNotFound
-		}
+	return s.modifiedBatcher.Get(ctx, id)
+}
 
-		return time.Time{}, err
+func (s *VulnerabilityStore) batchGetModified(ctx context.Context, ids []string) []batcher.Result[time.Time] {
+	results := make([]batcher.Result[time.Time], len(ids))
+
+	keys := make([]*datastore.Key, len(ids))
```
another-rex, line 159: Is it worth deduping here?

another-rex: actually probably not worth the additional logic, though maybe add a comment to state that's something considered.

go/internal/database/datastore/vulnerability.go
michaelkedar: added some env var parsing

go/internal/database/datastore/vulnerability.go
michaelkedar: comment added

go/internal/osvutil/batcher/batcher_test.go
michaelkedar: test should be fixed now

go/internal/osvutil/batcher/context.go
michaelkedar: neat - added

another-rex (approved):
LGTM, nice!

PR #5534, opened 2026-06-15
feat: move away from github-pages gem and add ruby 4 support for docs

G-Rath:
Based off https://github.com/google/osv-scanner/pull/2588 and #4972

---

This PR reworks the doc page deployment by removing the github pages gem and deploying using a github actions workflow. This was done to enable support for ruby 4, which is unsupported by github pages. As a result of the docs being deployed by github actions rather than pages, the deployment process for docs has changed, this PR also has an update to the contribution docs to reflect these changes. The github action workflow is set to deploy docs when there is a push to the master branch.

This PR also adds some plugins for Jekyll. These were ones that github pages added automatically, now that we have removed github pages, thee plugins are included here instead.

Once this PR is merged, the "Pages" settings will need to be updated. The build and deployment source will need to be changed to github actions

Closes #4972
Closes #5005


G-Rath:
Note I've purposely not really put a lot of attention into the specific versions being used, as I figure the bots can easily take care of bumping to latest once this is landed

docs/Gemfile.lock
```diff
@@ -279,16 +112,25 @@ PLATFORMS
...
+  jekyll-titles-from-headings
+  just-the-docs
   tzinfo (>= 1, < 3)
   tzinfo-data
   wdm (~> 0.2.0)
   webrick (~> 1.7)
 
 RUBY VERSION
-   ruby 3.4.8p72
+  ruby 4.0.4
```
another-rex, line 133: This is 4.0.4, but above in checkout is 4.0.5. Should we be relaxing one of these so there's only 1 thing we  need to update?

another-rex (dismissed):

docs/Gemfile.lock
G-Rath: pretty sure the only way to do that would be to have a `.ruby-version` file - otherwise one can always get out of sync 🤔

G-Rath: for now I've just updated CI to use "latest v4" which matches what's in the `Gemfile` 🤷

another-rex (approved):

PR #5552, opened 2026-06-17
fix(go api): Catch panics, and check health of databases

michaelkedar:
- Made all spawned goroutines recover from panics and return an error, so as to not crash the whole API if one request is bad.
- To accomplish this, make a `safe` package for the boilerplate of the defer recover statements.
- Made the health checks ping the database (Datastore and GCS) every so often so that critical unrecoverable errors do eventually get the server killed.

### DIFF
--- a/go/internal/osvutil/safe/safe.go
+++ b/go/internal/osvutil/safe/safe.go
@@ -12,12 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package safe provides simple utilities to spawn goroutines with panic recovery
-// and propagate those panics back to the main thread as error values.
+// Package safe provides simple utilities for panic recovery in goroutines and worker tasks,
+// propagating recovered panics back to callbacks or as error values.
 package safe
 
 import (
-	"context"
 	"fmt"
 	"runtime/debug"
 )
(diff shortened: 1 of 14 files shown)


go/internal/osvutil/safe/safe.go
```diff
@@ -0,0 +1,78 @@
...
+					Value: r,
+					Stack: debug.Stack(),
+				})
+			}
+		}()
+		f()
+	}()
+}
+
+// Func wraps any function with panic recovery, propagating it to a callback.
```
another-rex: Please extend comment to explain what this does.

go/internal/osvutil/safe/safe.go
```diff
@@ -0,0 +1,78 @@
...
+	Value any
+	Stack []byte
+}
+
+func (p *PanicError) Error() string {
+	return fmt.Sprintf("panic recovered: %v", p.Value)
+}
+
+// GoCancel spawns a goroutine. If it panics, it cancels the context with a PanicError.
+func GoCancel(cancel context.CancelCauseFunc, f func()) {
```
another-rex: Comment here with TODO, this can cause a race condition, but  we don't want to fix it

go/internal/osvutil/safe/safe.go
michaelkedar: commented further.

go/internal/osvutil/safe/safe.go
michaelkedar: I actually fixed the race condition by removing this function (which was a bit of a trap)

another-rex (approved):

PR #5553, opened 2026-06-17
refactor: final migration of API to go

michaelkedar:
- Update `build-and-stage` to build the new go API image
- Update Cloud Run configs to set new environment variables for the go API'
- Updated integration tests to run against go API
  - I might translate these wholly to go later, but it's a bigger project
  - The go API is a bit more lenient with somewhat invalid queries - it doesn't validate the package information if a commit is provided (since it does a commit query and ignores those fields anyway)
- Updated Makefile targets to do go API instead
- Added some missing checks/features to the go API
- Added package name normalisation to API queries (so in PyPI, `Jinja2` and `jinja2` match the same vulns)


gcp/api/fixtures/IntegrationTests_api_query_response.txt
```diff
@@ -1,25 +1,25 @@
 [   '400:{"commit": "d374094d8c49b6b7d288f307e11217ec5a502391", "package": '
     '{"ecosystem": "crates.io", "name": "crossbeam-utils", "purl": '
     '"pkg:cargo/crossbeam-utils"}, "version": "0.8.5"}\n',
-    '400:{"commit": "d374094d8c49b6b7d288f307e11217ec5a502391", "package": '
+    '200:{"commit": "d374094d8c49b6b7d288f307e11217ec5a502391", "package": '
```
another-rex, line 4: Probably mention in the PR description why this changed.

another-rex (dismissed):

another-rex (approved):

PR #5562, opened 2026-06-23
fix: mismatching repository casings between nvd and cve5 records

jess-lowe:
sometimes there are records where the case in the CVE5 OSV record and the NVD OSV record have the same repo, but different cases. if these are a non-case sensitive base (which is most cases - GitHub), this difference shouldnt matter. Currently, the record gets errored out instead.

### DIFF
--- a/vulnfeeds/conversion/versions.go
+++ b/vulnfeeds/conversion/versions.go
@@ -129,6 +129,18 @@ func repoGitWeb(parsedURL *url.URL) (string, error) {
 	return "", fmt.Errorf("unsupported GitWeb URL: %s", parsedURL.String())
 }
 
+// IsSameRepo returns true if repo1 and repo2 are for the same repository.
+func IsSameRepo(repo1, repo2 string) bool {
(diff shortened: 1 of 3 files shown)


vulnfeeds/conversion/versions.go
```diff
@@ -129,6 +129,19 @@ func repoGitWeb(parsedURL *url.URL) (string, error) {
 	return "", fmt.Errorf("unsupported GitWeb URL: %s", parsedURL.String())
 }
 
+// IsSameRepo returns true if repo1 and repo2 are for the same repository.
+func IsSameRepo(repo1, repo2 string) bool {
+	u1, err1 := url.Parse(repo1)
+	if err1 != nil {
+		return false
+	}
+	if isCaseInsensitiveHost(u1.Hostname()) {
```
another-rex, line 138: Why are repo1 and repo2 treated differently?

another-rex: actually this might be fine.

another-rex (approved):

PR #5567, opened 2026-06-25
fix: integration test new vuln appeared for ws

another-rex:
Add the new vuln that appeared.


michaelkedar (dismissed):
inb4 lint line too long

michaelkedar (approved):

Issue #1802, opened 2023-11-08
Put a Package/ID search box on the osv.dev main page


maxfisher-g:
**Is your feature request related to a problem? Please describe.**
There is no search box present on the front page of `osv.dev`. In order to search `osv.dev`, you have to click on "Search Vulnerability Database", then click on the search box, then type your query. This requires two extra mouse clicks plus a page / screen change, which is excessive for such a core function of the website.

**Describe the solution you'd like**
There should be a package / ID search box on the main page of `osv.dev`, which functions identically to the one on `osv.dev/list`. (i.e. when a search query is submitted, the page redirects to `osv.dev/list`).

The search box should be brought into focus immediately when the page is loaded. This means the user can type or paste a query into the search box without having to manually click (or tab) to bring it into focus. (An example is when you visit [google.com](https://www.google.com)).

As an extra bonus, the search box on `osv.dev/list` should also be autofocused when the page is loaded.

**Describe alternatives you've considered**
Implement URL redirection so that navigating to `osv.dev/<query>`  redirects to `osv.dev/list` and performs a search for `<query>`. This would be even more efficient, bypassing the main page completely, however this way of searching is  less discoverable for a new user.

**Additional context**
Having search functionality immediately available on the main page would greatly improve the usability and discoverability of the OSV search function, which I think is one of the primary functions of the osv.dev website.

andrewpollock:
> **Describe alternatives you've considered**
> Implement URL redirection so that navigating to `osv.dev/<query>` redirects to `osv.dev/list` and performs a search for `<query>`. This would be even more efficient, bypassing the main page completely, however this way of searching is less discoverable for a new user.

FYI, this is already the case, e.g. https://osv.dev/CVE-2023-4863

maxfisher-g:
> FYI, this is already the case, e.g. https://osv.dev/CVE-2023-4863

Oh cool! I didn't know that. But actually, I think these two features are complementary rather than true alternatives. One is a UI-based discoverability, the other more of a power-user feature. (The fact that I thought of this redirection but assumed it wasn't implemented so did not actually "discover it" kind of proves this).

maxfisher-g:
> FYI, this is already the case, e.g. https://osv.dev/CVE-2023-4863

[...]

PR #5569, opened 2026-06-25
feat: activate search input from keyboard

miketheman:
Refs: #1802


miketheman:
@another-rex Anything in particular I should consider before this can be shipped?

another-rex:
Should be good from your end, we just need to merge it.

Ly-Joey (approved):

PR #5586, opened 2026-06-30
fix(alpine): use NVD CPE version ranges to populate Introduced field

### LINKED ISSUE #5199: Data quality issue with CVE-2024-3094
**The CVE ID**
https://osv.dev/vulnerability/ALPINE-CVE-2024-3094

**Describe the data quality issue observed**
introducedAt is 0 which isn't the case for this. It looks like this is true for all alpine ingested CVEs as introducedAt isn't set in the alpine converter

**Suggested changes to record**
Appropriate introducedAt fields

**Additional context**
Note that the Alpine security tracker (https://security.alpinelinux.org/vuln/CVE-2024-3094) has correct info here but it is combining info from secdb and NVD in a richer way than is currently done with the OSV importer

katzj:
Follow the pattern used by the Alpine security tracker to use data from NVD to be able to give some information on Introduced versions for vulnerabilities. This includes the same rewriting rules used there.

This avoids over-reporting, for example CVE-2024-3094 should only show for xz 5.6.0 through 5.6.1-r2, not for earlier versions of xz

Fixes #5199

### DIFF
--- a/vulnfeeds/cmd/converters/alpine/main.go
+++ b/vulnfeeds/cmd/converters/alpine/main.go
@@ -15,6 +15,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/google/osv/vulnfeeds/conversion"
 	"github.com/google/osv/vulnfeeds/conversion/writer"
 	"github.com/google/osv/vulnfeeds/models"
 	"github.com/google/osv/vulnfeeds/utility/logger"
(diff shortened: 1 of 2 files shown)


jess-lowe:
Thanks for this! My two cents: This should be fine for most historical cases, but I am slightly concerned it may not work for future cases (due to NVD's recent decision to stop annotating CPEs to records unless explicitly asked). Still better than nothing though. 

I'll run this locally and see if there are any unexpected consequences of this change.

katzj:
> Thanks for this! My two cents: This should be fine for most historical cases, but I am slightly concerned it may not work for future cases (due to NVD's recent decision to stop annotating CPEs to records unless explicitly asked). Still better than nothing though.

Yeah. I can keep an eye on what Alpine does upstream to be able to follow suit with OSV as it seems like they may struggle with the same. And it's part of what their tooling uses to know when to tell people to update
 
> I'll run this locally and see if there are any unexpected consequences of this change.

Let me know if you see anything which needs improvements and happy to take a look

jess-lowe:
Ran it locally, and the only changes are the expected introduced value updates. Nice work!

jess-lowe (approved):

katzj:
@jess-lowe is it worth running the update across the history of all alpine vulns to ensure that they all have data if possible?

PR #5596, opened 2026-07-07
feat(vulnfeeds): treat ambiguious versions as enumerated by default

jess-lowe:
This pull request changes how "standalone" (unbounded/ambiguous) version information from CPE match criteria and CVE5 records is converted to OSV version ranges. 

Previously, when a single version string was extracted (e.g., `cpe:2.3:a:vendor:product:1.2.3` or `version: "1.2.3"` in CVE5 `affected` with no `lessThan`/`lessThanOrEqual` boundaries), the converter defaulted to assuming a range starting from `"0"` up to that version (`introduced: "0"`, `last_affected: "1.2.3"`).

With these changes, we treat these standalone/ambiguous versions as **enumerated** versions. They are converted to single-commit GIT ranges (e.g. `introduced: <commit of 1.2.3>`, `last_affected: <commit of 1.2.3>`) rather than starting from `"0"`. All enumerated version strings are added to the flat `"versions"` array in the OSV record.

### Standalone Range Simplification (Combining)
To avoid cluttering the OSV record's `Ranges` list with dozens of disjoint single-commit intervals when a CVE lists many standalone versions (e.g., FFmpeg in `CVE-2016-1897`), the combiner groups all standalone versions for a repository and represents them as a single continuous range from the earliest version commit to the latest version commit (`[earliest_commit, latest_commit]`) in first-appearance record order, while listing all intermediate version strings in the flat `"versions"` array.

---

### Pros (Improved Cleanliness & Matching)
* **Precision**: Eliminates incorrect `"introduced": "0"` range assumptions. This avoids falsely matching versions prior to the actual earliest defined vulnerable version.
* **Schema Cleanliness**: OSV JSON files remain clean and highly readable. Instead of containing dozens of `introduced`/`last_affected` pairs, a single consolidated range is created for the entire span of enumerated versions.
* **Version Matching**: Querying by version string (via package managers) remains fully accurate, as all intermediate versions are preserved inside the flat `"versions"` array.

### Cons (Git Commit Matching Granularity)
* **Range Approximation**: By simplifying the multiple single-version commits into a single `[earliest_commit, latest_commit]` range, we assume that all commits/versions between the earliest and latest versions are vulnerable. In some rare cases, this might lead to false positives if intermediate versions were not vulnerable (e.g., if a patch was backported to a version within that range). However, this matches the standard OSV range convention and is significantly cleaner than representing disjoint commits.

* **Matching only release commits**: Only release commits will be available in the enumerated affected commits dataset in some cases (especially when it's only one version affected), but we accept this, as most people would be using release tag version commits rather than something in between. 

* **Database Specific is still noisy**: all of the CPEs and versions used in the making of the final version will still be in the database_specific data, which can get a little noisy for cases with a lot of enumerated versions.

### DIFF
--- a/vulnfeeds/conversion/grouping.go
+++ b/vulnfeeds/conversion/grouping.go
@@ -302,6 +302,11 @@ func MergeRangesAndCreateAffected(
 				}
 			}
 
+			// Standalone ranges (where introduced == lastAffected) are assumed to represent a
+			// sequence of enumerated ranges. If there are multiple, we attempt to combine
+			// them into a single range spanning from the first introduced commit to the last
+			// affected commit. If both endpoints cannot be determined, we fall back to keeping
+			// them as separate standalone ranges.
 			if len(standaloneRanges) > 0 {
 				var standaloneVersions []string
 				for _, vrwm := range standaloneRanges {
--- a/vulnfeeds/conversion/grouping.go
+++ b/vulnfeeds/conversion/grouping.go
@@ -289,18 +289,70 @@ func MergeRangesAndCreateAffected(
 		slices.Sort(successfulRepos)
 		successfulRepos = slices.Compact(successfulRepos)
 		for _, repo := range successfulRepos {
-			var mergedRange *osvschema.Range
+			var standaloneRanges []models.RangeWithMetadata
+			var otherRanges []models.RangeWithMetadata
 			for _, vrwm := range resolvedRanges {
+				if vrwm.Range.GetRepo() != repo {
+					continue
+				}
+				if isStandaloneRange(vrwm) {
+					standaloneRanges = append(standaloneRanges, vrwm)
+				} else {
+					otherRanges = append(otherRanges, vrwm)
+				}
+			}
+
+			if len(standaloneRanges) > 0 {
+				var standaloneVersions []string
+				for _, vrwm := range standaloneRanges {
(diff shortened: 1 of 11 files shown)


vulnfeeds/conversion/grouping.go
```diff
@@ -289,18 +289,70 @@ func MergeRangesAndCreateAffected(
...
+					continue
+				}
+				if isStandaloneRange(vrwm) {
+					standaloneRanges = append(standaloneRanges, vrwm)
+				} else {
+					otherRanges = append(otherRanges, vrwm)
+				}
+			}
+
+			if len(standaloneRanges) > 0 {
```
another-rex, line 310: nit: Add a comment here about standalone ranges being assumed as enumerated ranges, and how this processes them.

vulnfeeds/conversion/versions.go
```diff
@@ -795,6 +796,8 @@ func ExtractVersionsFromCPEs(cve models.NVDCVE, validVersions []string, vpRepoCa
 						lastaffected += "-" + CPE.Update
 					}
 					source = models.VersionSourceCPEString
+					introduced = lastaffected
```
another-rex, line 802: hmm this doesn't seem to explain why introduced = lastaffected? Can you add a comment?

another-rex (dismissed):

another-rex (approved):

PR #5609, opened 2026-07-09
perf: improve speed of combine-to-osv cron

jess-lowe:
downloading via gcloud was taking 25+ minutes, this does the download in go, and in parallel with a bunch of workers


vulnfeeds/cmd/combine-to-osv/main_test.go
```diff
@@ -61,38 +60,26 @@ func loadOSV(osvPath string) map[models.CVEID]*osvschema.Vulnerability {
...
-
-		if len(baseOSV.GetAffected()) == 0 || !hasRanges(baseOSV.GetAffected()) {
-			// check if part exists.
-			if !slices.Contains(mandatoryCVEIDs, string(cveID)) {
-				continue
-			}
-		}
-		osvRecords[cveID] = baseOSV
+	// Collect all unique CVE IDs
+	allIDs := make(map[models.CVEID]bool)
```
another-rex: nit: map[models.CVEID]struct{}

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -51,50 +198,190 @@ func main() {
...
+		defer listWg.Done()
+		logger.Info("Listing NVD objects", slog.String("path", *nvdPath))
+		nvdFiles, nvdListErr = listObjects(ctx, client, *nvdPath)
+		if nvdListErr != nil {
+			logger.Fatal("Failed to list NVD objects", slog.Any("err", nvdListErr))
+		}
+	}()
+
+	listWg.Wait()
+
```
another-rex, line 236: Please add some logging here and above the listWg, so we can time how long each section takes.

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -51,50 +198,190 @@ func main() {
...
+		uploadWg.Add(1)
+		go func() {
+			defer uploadWg.Done()
+			writer.VulnWorker(ctx, validVulnChan, outBkt, overridesBkt, gcsHelper, *osvOutputPath, &successCount)
+		}()
+	}
+
+	// Interpose Collector to gather valid IDs
+	var validIDs []string
+	var idWg sync.WaitGroup
```
another-rex: Use some longer variable names, it's unclear what this waitgroup is for.

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -51,50 +198,190 @@ func main() {
...
+			overridesBkt = client.Bucket(*overridesBucketName)
+		}
+		gcsHelper, err = gcs.InitUploadPool(ctx, *numWorkers, *outputBucketName)
+		if err != nil {
+			logger.Fatal("Failed to initialize GCS upload pool", slog.Any("err", err))
+		}
+		defer gcsHelper.CloseAndWait()
+	}
+
+	// Start Channels
```
another-rex, line 277: Might be useful to comment out a diagram of how the different channels are working and passing the vulnerability.

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -51,50 +198,190 @@ func main() {
...
+	totalWork := len(workItems)
+	go func() {
+		defer idWg.Done()
+		count := 0
+		for v := range vulnChan {
+			count++
+			if len(v.GetAffected()) > 0 {
+				validIDs = append(validIDs, v.GetId())
+			}
+			validVulnChan <- v
```
another-rex: It seems like validVulnChan is named wrong? It's just all vulns right? Maybe you want to call it uploadVulnsChan?

vulnfeeds/cmd/combine-to-osv/run_combine_to_osv_convert.sh
```diff
@@ -16,31 +16,19 @@ trap  '__error_handing__ $? $LINENO' ERR
 
 set -eu
 
-INPUT_BUCKET="${INPUT_GCS_BUCKET:=cve-osv-conversion}"
-OUTPUT_BUCKET="${OUTPUT_GCS_BUCKET:=cve-osv-conversion}"
+INPUT_BUCKET="${INPUT_GCS_BUCKET:=osv-test-cve-osv-conversion}"
+OUTPUT_BUCKET="${OUTPUT_GCS_BUCKET:=osv-test-cve-osv-conversion}"
```
another-rex: Is this change intentional? I think this will break prod.

another-rex: The kubernetes files are not correctly setting the bucket env variables I believe. Please double check.

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -32,6 +36,149 @@ const (
...
+			readWg.Add(1)
+			go func() {
+				defer readWg.Done()
+				nvd, nvdErr = readVulnerability(ctx, client, work.NVDPath)
+				if nvdErr != nil {
+					logger.Warn("Failed to read NVD", slog.String("id", string(work.ID)), slog.Any("err", nvdErr))
+				}
+			}()
+		}
+
```
another-rex, line 170: If either errors, this should just skip the record right? And probably should error log instead of warn.

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -32,6 +36,149 @@ const (
...
 
+// CVEWorkItem represents a unit of work for a single CVE.
+type CVEWorkItem struct {
+	ID       models.CVEID
+	CVE5Path string
+	NVDPath  string
+}
+
+func cveIDFromPath(p string) models.CVEID {
+	base := filepath.Base(p)
```
another-rex: I think this needs to be path.Base()

Technically if we want to support windows local paths, we need to check whether p starts with `gs://`, but I think it's fine to not support windows.

another-rex (approved):

PR #5611, opened 2026-07-09
fix(api): don't error log context cancellations

michaelkedar:
it's noisy


another-rex (commented):
Err I feel like there should at least be a warning? or something logging the error. Otherwise we'll be getting error responses with no logs for why the  response is erroring.

michaelkedar:
> Err I feel like there should at least be a warning? or something logging the error. Otherwise we'll be getting error responses with no logs for why the response is erroring.

I dunno - I guess I'm trying to avoid flooding too many repeated logs on cancel of any kind - every single hydrator in each batch query logs the same thing - it's probably annoying.

another-rex (approved):

PR #5617, opened 2026-07-09
fix: prevent batcher from exceeding maxSize

michaelkedar:
The request batcher was using maxSize as a trigger to override the timeout, but wasn't capping the number of requests to that size (if a very large spike in requests happen at that exact moment).
This was causing the BatchQuery to attempt a GetMulti with over 1000 keys, which is an error in datastore.

If the pending slice is over maxSize, spawn a new batch loop to consume the remaining entities.


another-rex (approved):

another-rex (commented):
Was this datastore error logged?

PR #5632, opened 2026-07-13
perf: optimize combine-to-osv bucket listing and avoid partial files.txt upload

another-rex:
This PR optimizes the bucket listing operation in `combine-to-osv` by using a pre-uploaded `files.txt` list file generated by the `cve5` and `nvd-cve-osv` converters.

- Added upload of `files.txt` list file to `cve5` and `nvd-cve-osv` converters.
- Optimized `combine-to-osv` to read `files.txt` instead of listing the GCS bucket.
- Added fallback to bucket listing if `files.txt` is too old or fails to read.
- Made list file max age configurable via flag.
- Skipped `files.txt` upload on partial runs (start-year > default).
- Refactored converters to use channels instead of global mutexes for tracking output files.

agy

### DIFF
--- a/vulnfeeds/cmd/combine-to-osv/main.go
+++ b/vulnfeeds/cmd/combine-to-osv/main.go
@@ -4,6 +4,7 @@ package main
 import (
 	"cmp"
 	"context"
+	"errors"
 	"flag"
 	"fmt"
 	"io"
@@ -16,6 +17,7 @@ import (
 	"strings"
 	"sync"
 	"sync/atomic"
+	"time"
 
 	"cloud.google.com/go/storage"
 	"github.com/google/osv/vulnfeeds/conversion"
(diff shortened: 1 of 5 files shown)


vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -57,11 +60,67 @@ func cveIDFromPath(p string) models.CVEID {
 	return ""
 }
 
-func listObjects(ctx context.Context, client *storage.Client, pathStr string) ([]string, error) {
+func readListFile(ctx context.Context, listObj *storage.ObjectHandle, bucketName string) ([]string, error) {
```
jess-lowe, line 67: can you add a comment on what the expected list file includes? like is it just a list of ids? the full bucket path? ids with suffix?

jess-lowe (approved):

PR #5639, opened 2026-07-14
feat: add reimport TUI tool for manually triggering Datastore reimports

another-rex:
This PR adds a terminal user interface (TUI) tool under `go/cmd/tools/reimport-tui` to manually trigger a full reimport of vulnerability sources in either the Test (`oss-vdb-test`) or Prod (`oss-vdb`) Datastore.

**entirely vibecoded**

Screenshot From 2026-07-15 13-28-55
Screenshot From 2026-07-15 13-30-09

agy


jess-lowe:
got any screenshots we can look at?

go/cmd/tools/reimport-tui/go.mod
```diff
@@ -0,0 +1,97 @@
+module github.com/google/osv.dev/go/cmd/tools/reimport-tui
+
+go 1.26.5
+
+replace github.com/google/osv.dev/go => ../../../
```
jess-lowe, line 5: I thought the linter disliked replace?

another-rex: It only matters if we want to publish it as a public package for others to install. This is only used by us running with go run within the repo, so no issues here.

jess-lowe (approved):

PR #5664, opened 2026-07-15
perf: further increase resources for vanir-signatures

cuixq:
- Increase CPU to 4 and memory limit to 64GB.
- Reduce default max-workers to 4 to match CPU allocation and reduce memory pressure.
- Remove redundant JSON serialization/deserialization in batch processing.


deployment/clouddeploy/gke-workers/base/extra/vanir-signatures.yaml
```diff
@@ -18,11 +18,11 @@ spec:
             imagePullPolicy: Always
             resources:
               requests:
-                cpu: "2"
-                memory: "32G"
-              limits:
-                cpu: "2"
+                cpu: "4"
                 memory: "48G"
```
another-rex: ~~Let's just keep this the same as the limit to avoid odd issues with it being kicked off. Also use 63G rather than 64, since we want some free memory on the node for~~

Actually, let's just set this to 63G, and leave limit at 64G. This way we can fit 4 of these on the same node, and we avoid issues with it being kicked out.

cuixq: done

another-rex (approved):

PR #5680, opened 2026-07-15
perf(vulnfeeds): make vulnfeeds tests go faster

jess-lowe:
1.  **`TestReposFromReferences` (in `vulnfeeds/conversion`)**:
    *   **Issue**: Was taking ~25 seconds. One subtest involving a `cgit` repository was taking over 20 seconds, likely due to network timeouts or slow responses during `go-git` operations or canonical link validation.
    *   **Fix**: Modified the test to use the VCR client for `git.FindCanonicalLink` calls (which perform HTTP HEAD requests). This allows recording and replaying these requests.
    *   **Result**: Test time reduced to **~3 seconds**.

2.  **`cve5` Package Tests (`vulnfeeds/conversion/cve5`)**:
    *   **Issue**: `TestCVE5Snapshot` and `TestExtractVersions` were doing live network calls during version extraction and Git version-to-commit conversion.
    *   **Fix**: Integrated `testutils.SetupGitVCR(t)` into these tests to record and replay Git remote interactions.
    *   **Result**: While initial recording still takes time, playback significantly reduces execution time. `TestExtractVersions` reduced from ~8s to **~3.8s** (played back). `TestCVE5Snapshot` took around 10-12s, processing multiple files sequentially, previously was taking ~20s.


another-rex (commented):
Slightly hesitant on encoding git responses, but let's try it and see.

Can you extend/explain the final line though?
> Result: While initial recording still takes time, playback significantly reduces execution time. TestExtractVersions reduced from ~8s to ~3.8s (played back). TestCVE5Snapshot took around 10-12s, processing multiple files sequentially.

jess-lowe:
```
conversion          24.753s -> 14.710s
conversion/cve5     28.300s -> 12.098s
git                 27.007s -> 8.728s
```

vulnfeeds/conversion/testdata/TestReposFromReferences_A_CVE_with_a_repo_that_redirects_(docker_docker_-__moby_moby).yaml.bak
another-rex, line 1: Should this file be added?

vulnfeeds/conversion/common.go
another-rex: Should this still be using default client?

another-rex: Also please check for any other usages of default client

another-rex: Another one here.

vulnfeeds/git/versions_test.go
```diff
@@ -403,6 +403,10 @@ func TestValidateAndCanonicalizeLink_429(t *testing.T) {
 }
 
 func TestValidateAndCanonicalizeLink_Retries(t *testing.T) {
+	oldDelay := BackoffBaseDelay
+	BackoffBaseDelay = 1 * time.Millisecond
+	defer func() { BackoffBaseDelay = oldDelay }()
```
another-rex, line 409: Bit worried this could lead to some race conditions. It's probably fine since this doesn't have t.Parallel(). Maybe make a comment making the exclusion of t.Parallel explicit.

vulnfeeds/conversion/cve5/converter.go
```diff
@@ -153,7 +155,7 @@ func FromCVE5(cve models.CVE5, refs []models.Reference, metrics *models.Conversi
 }
 
 // CVEToOSV converts a models.CVE5 object into an OSV Vulerability record and returns it along with its conversion metrics.
-func CVEToOSV(cve models.CVE5, sourceLink string) (*vulns.Vulnerability, *models.ConversionMetrics) {
+func CVEToOSV(cve models.CVE5, sourceLink string, cache git.RepoTagsCache, httpClient *http.Client) (*vulns.Vulnerability, *models.ConversionMetrics) {
```
another-rex, line 158: nit: Add some checks here for if httpClient is nil or cache is nil?

vulnfeeds/conversion/nvd/converter.go
```diff
@@ -23,7 +23,10 @@ var ErrNoRanges = errors.New("no ranges")
 var ErrUnresolvedFix = errors.New("fixes not resolved to commits")
 
 // CVEToOSV Takes an NVD CVE record and returns an OSV Vulnerability object, ConversionMetrics, and the outcome.
-func CVEToOSV(cve models.NVDCVE, repos []string, vpRepoCache *c.VPRepoCache, cache git.RepoTagsCache, metrics *models.ConversionMetrics) (*vulns.Vulnerability, *models.ConversionMetrics, models.ConversionOutcome) {
+func CVEToOSV(cve models.NVDCVE, repos []string, vpRepoCache *c.VPRepoCache, cache git.RepoTagsCache, metrics *models.ConversionMetrics, httpClient *http.Client) (*vulns.Vulnerability, *models.ConversionMetrics, models.ConversionOutcome) {
+	if httpClient == nil {
```
another-rex, line 27: You have a check here, but not in CVE5

another-rex: I'm actually thinking this should just panic, rather than silently setting the default http client.

another-rex (commented):
Pretty good! Let's try it.

another-rex (approved):

PR #5681, opened 2026-07-15
refactor(website): stub out go website endpoints

michaelkedar:
Set up all the endpoint stubs for the Go website migration.
Static files are to be `//go:embed`ed into a filesystem, which should make the website speedier (though the binary bigger).
I haven't set up the datastore connections, nor dealt with rendering the Jinja templates in Go yet.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -90,20 +90,28 @@ build-api-protos:
 
 build-protos: build-osv-protos build-api-protos
 
-run-website:
-	cd gcp/website/frontend3 && pnpm install && pnpm run build
-	cd gcp/website/blog && hugo --buildFuture -d ../dist/static/blog
+build-website-frontend:
+	cd website/frontend3 && pnpm install && pnpm run build
+	cd website/blog && hugo --buildFuture -d ../dist/static/blog
+
+run-website: build-website-frontend
 	cd gcp/website && $(install-cmd) && GOOGLE_CLOUD_PROJECT=oss-vdb OSV_VULNERABILITIES_BUCKET=osv-vulnerabilities $(run-cmd) python main.py
 
-run-website-staging:
-	cd gcp/website/frontend3 && pnpm install && pnpm run build
-	cd gcp/website/blog && hugo --buildFuture -d ../dist/static/blog
+run-website-staging: build-website-frontend
 	cd gcp/website && $(install-cmd) && GOOGLE_CLOUD_PROJECT=oss-vdb-test OSV_VULNERABILITIES_BUCKET=osv-test-vulnerabilities $(run-cmd) python main.py
 
-run-website-emulator:
-	cd gcp/website/frontend3 && pnpm install && pnpm run build
-	cd gcp/website/blog && hugo --buildFuture -d ../dist/static/blog
+run-website-emulator: build-website-frontend
 	cd gcp/website && $(install-cmd) && DATASTORE_EMULATOR_PORT=5002 $(run-cmd) python frontend_emulator.py
+
+run-go-website: build-website-frontend
+	cd go && go build -o ./website ./cmd/website && GOOGLE_CLOUD_PROJECT=oss-vdb OSV_VULNERABILITIES_BUCKET=osv-vulnerabilities ./website -static-dir ../website/dist -docs-dir ../docs
+
+run-go-website-staging: build-website-frontend
+	cd go && go build -o ./website ./cmd/website && GOOGLE_CLOUD_PROJECT=oss-vdb-test OSV_VULNERABILITIES_BUCKET=osv-test-vulnerabilities ./website -static-dir ../website/dist -docs-dir ../docs
+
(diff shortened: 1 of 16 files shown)


go/cmd/website/main.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	port := flag.Int("port", defaultPort, "port for the website server")
+	staticDir := flag.String("static-dir", "dist", "directory containing static assets")
+	docsDir := flag.String("docs-dir", "docs", "directory containing API docs")
+	flag.Parse()
+
+	var staticFS website.Config
+	if *staticDir != "" {
+		staticFS.StaticFS = os.DirFS(*staticDir)
```
another-rex: hmm maybe use os.OpenRootFS? Another layer of protection against any path traversal stuff.

go/cmd/website/main.go
```diff
@@ -0,0 +1,90 @@
...
+		}
+	}
+
+	port := flag.Int("port", defaultPort, "port for the website server")
+	staticDir := flag.String("static-dir", "dist", "directory containing static assets")
+	docsDir := flag.String("docs-dir", "docs", "directory containing API docs")
+	flag.Parse()
+
+	var staticFS website.Config
+	if *staticDir != "" {
```
another-rex: Do we even need this check? I assume this should just not launch / panic if the flags are set to ""

go/cmd/website/main.go
```diff
@@ -0,0 +1,90 @@
...
+			serverErrors <- err
+		}
+	}()
+
+	select {
+	case err := <-serverErrors:
+		logger.ErrorContext(ctx, "Server error", slog.Any("error", err))
+		return err
+	case <-ctx.Done():
+		logger.InfoContext(ctx, "Shutting down website server...")
```
another-rex: Is it fine to log with a ctx that's done?

go/internal/website/server.go
```diff
@@ -0,0 +1,162 @@
...
+	s.mux.HandleFunc("GET /healthz", s.handleHealthz)
+
+	// Go vanity imports & root check
+	s.mux.HandleFunc("GET /{$}", s.handleRoot)
+	s.mux.HandleFunc("GET /bindings/go", s.handleGoBindingsVanity)
+
+	// Simple redirects & static pages
+	s.mux.HandleFunc("GET /v2/{subpath...}", func(w http.ResponseWriter, r *http.Request) {
+		http.Redirect(w, r, "/"+r.PathValue("subpath"), http.StatusFound)
+	})
```
another-rex: Maybe we can finally deprecate the v2 path...

Also I think the redirect is missing passing any query strings.

another-rex (dismissed):
LGTM, some minor comments

Makefile
```diff
@@ -90,21 +90,29 @@ build-api-protos:
...
 	cd gcp/website && $(install-cmd) && DATASTORE_EMULATOR_PORT=5002 $(run-cmd) python frontend_emulator.py
 
+run-go-website: build-website-frontend
+	cd go && go build -o ./website ./cmd/website && GOOGLE_CLOUD_PROJECT=oss-vdb OSV_VULNERABILITIES_BUCKET=osv-vulnerabilities ./website -static-dir ../website/dist -docs-dir ../docs
+
+run-go-website-staging: build-website-frontend
+	cd go && go build -o ./website ./cmd/website && GOOGLE_CLOUD_PROJECT=oss-vdb-test OSV_VULNERABILITIES_BUCKET=osv-test-vulnerabilities ./website -static-dir ../website/dist -docs-dir ../docs
+
+run-go-website-emulator: build-website-frontend
+	cd go && go build -o ./website ./cmd/website && DATASTORE_EMULATOR_HOST=localhost:5002 ./website -static-dir ../website/dist -docs-dir ../docs
```
another-rex: Can you add the output of this to the gitignore?

(also why is this not just go run, but go build)

go/internal/website/server.go
michaelkedar: Got rid of it.
Sorry to the 150 AI agents per month who are hitting the redirect 🫡

go/cmd/website/main.go
michaelkedar: made this fail if no fs exists

go/cmd/website/main.go
michaelkedar: I've added an embedded fs switch for prod.
This DirFS should only be used for local dev - I don't think path traversal is a big issue there.

Makefile
michaelkedar: it's go run now

go/cmd/website/main.go
michaelkedar: It is mostly okay (logging doesn't check if the context is done), but I've made this log on the shutdownCtx and used `context.WithoutCancel` there instead

another-rex (approved):

PR #5684, opened 2026-07-16
perf: optimize combine-to-osv bucket listing 

another-rex:
Undo the previous optimisation and optimise better


jess-lowe (dismissed):
LGTM 🚀

jess-lowe (approved):

PR #5690, opened 2026-07-20
feat(gitter): importer endpoints

Ly-Joey:
Adds:
- `/file-diffs`:  List file diffs from last_synced_commit to HEAD of target branch (defaults to remote's origin/HEAD if not provided in request)
- `/file-content`: Given commit hash and file path, `git cat-file` the content of the file

Also did some small refactoring to move code that handles repo fetching, cloning, etc into `git.go`.

There will be a follow-up PR to update importer to use these new endpoints.

### DIFF
--- a/go/cmd/gitter/gitter.go
+++ b/go/cmd/gitter/gitter.go
@@ -55,6 +55,8 @@ var endpointHandlers = map[string]http.HandlerFunc{
 	"POST /cache":            cacheHandler,
 	"GET /tags":              tagsHandler,
 	"POST /affected-commits": affectedCommitsHandler,
+	"POST /diffs":            diffsHandler,
+	"POST /file-content":     fileContentHandler,
 }
 
 var (
@@ -445,6 +447,16 @@ func getFreshRepo(ctx context.Context, repoURL string, forceUpdate bool) (*Repos
 	return repo, nil
 }
 
+// getRepoForDisk fetches/updates the repository on disk and returns a Repository struct with repoPath set,
+// skipping the expensive LoadRepository operation (commit graph building and patch ID calculation).
+func getRepoForDisk(ctx context.Context, repoURL string, forceUpdate bool) (*Repository, error) {
+	if err := doFetch(ctx, repoURL, forceUpdate); err != nil {
+		return nil, err
+	}
+
+	return NewRepository(repoURL), nil
+}
+
 func isIndexLockError(err error) bool {
 	if err == nil {
 		return false
@@ -497,6 +509,12 @@ func fetchAndReset(ctx context.Context, repoPath string) error {
 	err := runCmd(ctx, repoPath, nil, "git", "fetch", "origin")
 	if err != nil {
 		return fmt.Errorf("git fetch failed: %w", err)
+	}
+
+	// Make sure origin/HEAD points to the latest default branch from remotes
+	err = runCmd(ctx, repoPath, nil, "git", "remote", "set-head", "origin", "--auto")
+	if err != nil {
+		logger.WarnContext(ctx, "git remote set-head failed: ", slog.Any("err", err))
 	}
 
 	err = runCmd(ctx, repoPath, nil, "git", "reset", "--hard", "origin/HEAD")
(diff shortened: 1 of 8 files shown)


go/cmd/gitter/gitter.go
```diff
@@ -1110,3 +1128,166 @@ func tagsHandler(w http.ResponseWriter, r *http.Request) {
 		logger.ErrorContext(ctx, "Error writing tags response", slog.Any("error", err))
 	}
 }
+
+func diffsHandler(w http.ResponseWriter, r *http.Request) {
```
another-rex: nit: rename r to req, as we have resp as well.

go/cmd/gitter/gitter.go
```diff
@@ -1110,3 +1128,166 @@ func tagsHandler(w http.ResponseWriter, r *http.Request) {
...
+	out, err := marshalResponse(r, resp)
+	if err != nil {
+		logger.ErrorContext(ctx, "Error marshaling diff response", slog.Any("error", err))
+		statusCode = http.StatusInternalServerError
+		http.Error(w, fmt.Sprintf("Error marshaling response: %v", err), statusCode)
+
+		return
+	}
+
+	w.Header().Set("Content-Type", r.Header.Get("Content-Type"))
```
another-rex: Nit: I feel like this value should be returned by marshalResponse, rather than just setting whatever is sent.

go/cmd/gitter/gitter.go
```diff
@@ -1110,3 +1128,166 @@ func tagsHandler(w http.ResponseWriter, r *http.Request) {
...
+	w.WriteHeader(http.StatusOK)
+	if _, err := w.Write(out); err != nil {
+		logger.ErrorContext(ctx, "Error writing diff response", slog.Any("error", err))
+	}
+}
```
another-rex: infact this entire section feels like it's basically the same across all endpoints, should this just be a single function shared amongst them?

go/cmd/gitter/repository.go
```diff
@@ -969,3 +971,206 @@ func (r *Repository) GetRemoteTags(ctx context.Context) (map[string]SHA1, error)
...
+	return strings.TrimSpace(string(out)), nil
+}
+
+var (
+	emptyTreeOnce sync.Once
+	emptyTreeHash string
+)
+
+// getEmptyTreeHash calculates and caches the empty tree hash
+func getEmptyTreeHash(ctx context.Context, repoPath string) string {
```
another-rex: huh is this meant to be difference from the fallback value?

go/cmd/gitter/repository.go
```diff
@@ -969,3 +971,206 @@ func (r *Repository) GetRemoteTags(ctx context.Context) (map[string]SHA1, error)
...
+	emptyTreeOnce.Do(func() {
+		// Use `git mktree` with empty stdin for the empty tree hash
+		cmd := prepareCmd(ctx, repoPath, nil, "git", "mktree")
+		cmd.Stdin = bytes.NewReader(nil)
+		out, err := cmd.CombinedOutput()
+		if err != nil {
+			// Fallback to the empty tree SHA1 magic number.
+			logger.ErrorContext(ctx, "Failed to get empty tree SHA from `git mktree`, falling back to well-known empty tree SHA1", slog.Any("err", err))
+			emptyTreeHash = "4b825dc642cb6eb9a060e54bf8d69288fbee4904"
```
another-rex: Make this a const at the top of the file.

go/cmd/gitter/repository.go
```diff
@@ -969,3 +971,206 @@ func (r *Repository) GetRemoteTags(ctx context.Context) (map[string]SHA1, error)
...
+// parseNameStatusLine parses a single tab-delimited line of git diff-tree --name-status output.
+// Format is: <status>\t<path> OR <status>\t<old-path>\t<new-path>
+func parseNameStatusLine(line string) (*FileChange, error) {
+	fields := strings.Split(line, "\t")
+	if len(fields) < 2 || len(fields[0]) == 0 {
+		return nil, fmt.Errorf("invalid diff-tree line: %q", line)
+	}
+
+	/**
+	Possible status letters are:
```
another-rex, line 1091: Link to the docs here?

another-rex: Also, I think this is actually copied from the wrong part of the doc, so is missing the B status.

See https://git-scm.com/docs/git-diff-tree (--name-status)

> Show only the name(s) and status of each changed file. See the description of the --diff-filter option on what the status letters mean. Just like --name-only the file names are often encoded in UTF-8.

then see --diff-filter:

> --diff-filter=[(A|C|D|M|R|T|U|X|B)...[*]]
Select only files that are Added (A), Copied (C), Deleted (D), Modified (M), Renamed (R), have their type (i.e. regular file, symlink, submodule, …​) changed (T), are Unmerged (U), are Unknown (X), or have had their pairing Broken (B). Any combination of the filter characters (including none) can be used. When * (All-or-none) is added to the combination, all paths are selected if there is any file that matches other criteria in the comparison; if there is no file that matches other criteria, nothing is selected.
> 
> Also, these upper-case letters can be downcased to exclude. E.g. --diff-filter=ad excludes added and deleted paths.
> 
> Note that not all diffs can feature all types. For instance, copied and renamed entries cannot appear if detection for those types is disabled.

But this part of the doc is from the raw git-diff-files command.

go/cmd/gitter/gitter.go
another-rex: hmm I think we should name this something better than just semaphore, (just name it reqConcurrencySemaphore or something) Same with all the functions mentioning this. Right now it's a bit unclear for someone who's not familiar with what "SkipSemaphore" does for example.

go/cmd/gitter/repository.go
Ly-Joey: Added doc link, it's actually from https://git-scm.com/docs/diff-format.
The `diff-*` (index / tree / files) commands all share this raw output format, so their respective docs link back to the same `diff-format` section via an `include::`. So, this status list would apply to `git diff-tree` as well.

Re: the "B" status: 
TL;DR: In terms of output, "B" is never a real status that will show up.

It is associated with the -B (--break-rewrites) flag.
If enabled, it breaks large modifications into separate add (A) and delete (D) pairs. But later in the diffcore transformation, these broken-up pairs will either resolve as a rename (R) or merge back into a modified (M).
(ref: https://git-scm.com/docs/gitdiffcore)

While the `B` option in `--diff-filter` can be used to filter files with large modifications (thus "broken" at one point), the transformations within `git diff` will not be printed it as a "B" status in either the `--name-status` or `--raw` output.

go/cmd/gitter/repository.go
Ly-Joey: The empty tree hash for SHA256 will be different, but since gitter isn't supporting SHA256 yet, I've updated this bit to just use the SHA1 hash as a constant for now.

go/internal/gitter/pb/repository/repository.proto
```diff
@@ -62,3 +62,29 @@ message AffectedCommitsRequest {
...
 }
+
+message FileChange {
+    string from_path = 1;
+    string to_path = 2;
+}
+
+message FileDiffsRequest {
+    string url = 1;
+    string last_synced_commit = 2;
```
michaelkedar: nit: perhaps comment here that an empty last_synced_commit nets you all the files in the repo

michaelkedar (dismissed):

go/internal/gitter/pb/repository/repository.proto
Ly-Joey: done!

another-rex (approved):

PR #5691, opened 2026-07-20
feat: Exporter 12x faster hopefully

another-rex:
Made a few changes to vastly increase the speed of the exporter.

- Increased each channel buffer up to 100 from 0.
- 5x ed the worker count from 600 to 3000
- Added a fast object list GCS library (mostly copied over from vulnfeeds), and split up the initial object listing loop at various places.
  -  This is mostly guesswork at the moment, ideally we should do a review of where to roughly evenly split the work every now and then. But good enough for now. 
  - This does mean the exporter will slow down near the end.
- Refactored message receiving to use the `for` syntax. This should be a slight behavior change where we won't cancel based on the context, but only when the channel is closed. (Or when the context is cancelled on the sending side, just never on the receiving side)
- Fixed README typos
- Use logger.ErrorContext instead of logger.Error
-

### DIFF
--- a/go/cmd/exporter/downloader.go
+++ b/go/cmd/exporter/downloader.go
@@ -2,10 +2,10 @@ package main
 
 import (
 	"context"
-	"errors"
 	"log/slog"
 	"sync"
 
+	"github.com/google/osv.dev/go/internal/osvutil"
 	"github.com/google/osv.dev/go/logger"
 	"github.com/google/osv.dev/go/osv/clients"
 	"github.com/ossf/osv-schema/bindings/go/osvschema"
@@ -21,7 +21,7 @@ func downloader(ctx context.Context, client clients.CloudStorage, inCh <-chan st
 		// Process object.
 		data, err := client.ReadObject(ctx, path)
 		if err != nil {
-			if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
+			if osvutil.IsContextError(err) {
 				return
 			}
 			logger.ErrorContext(ctx, "failed to read vulnerability", slog.String("obj", path), slog.Any("err", err))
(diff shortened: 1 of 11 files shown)


go/cmd/exporter/downloader.go
```diff
@@ -16,29 +17,22 @@ import (
...
-		case <-ctx.Done():
-			return
-		}
-
+	for path := range inCh {
 		// Process object.
 		data, err := client.ReadObject(ctx, path)
 		if err != nil {
-			logger.Error("failed to read vulnerability", slog.String("obj", path), slog.Any("err", err))
+			if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
```
michaelkedar: I don't know if this happens with GCS, but with Datastore I had run into an issue that the context being cancelled during a gRPC operation returned a gRPC error that was not a `context.Canceled` or `DeadlineExceeded` error.

I made [this helper function](https://github.com/google/osv.dev/blob/4bf647cb347ce37e85427035b2f4f26a6398cad2/go/internal/osvutil/errors.go#L13) to capture gRPC context cancels as well, might be safer to use it here.

go/cmd/exporter/downloader.go
```diff
@@ -16,29 +17,22 @@ import (
...
+				return
+			}
+			logger.ErrorContext(ctx, "failed to read vulnerability", slog.String("obj", path), slog.Any("err", err))
 			continue
 		}
 		vuln := &osvschema.Vulnerability{}
 		if err := proto.Unmarshal(data, vuln); err != nil {
-			logger.Error("failed to unmarshal vulnerability", slog.String("obj", path), slog.Any("err", err))
+			if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
```
michaelkedar: likewise

go/internal/gcs/gcs.go
```diff
@@ -0,0 +1,184 @@
...
+		Prefix:      globalPrefixWithSlash,
+		StartOffset: globalPrefixWithSlash + sortedBreakdowns[len(sortedBreakdowns)-1],
+	})
+
+	return queries
+}
+
+// ListObjectsFast lists object names in parallel by partitioning the prefix space.
+// It partitions the space using StartOffset and EndOffset based on breakdownPrefixes.
+func ListObjectsFast(ctx context.Context, bucket *storage.BucketHandle, globalPrefix string, breakdownPrefixes []string) ([]string, error) {
```
michaelkedar: Why not have this function just wrap `ObjectsFastStream`?

another-rex: Good point, actually this entire function is unused, it's a byproduct of being copied over from what I did in combine-to-osv.

Removed the whole thing now.

another-rex:
Moved the breakdowns into a flag and put it in our kubernetes config.

michaelkedar (approved):

PR #5694, opened 2026-07-22
refactor(website): begin migrating Jinja templates to Go

michaelkedar:
Start migrating many of the website templates from Jinja templates to use go's native `html/template` syntax.
This does most pages except the individual vulnerability page and the list page (because those templates are more involved, and were going to make the PR too big).

The first commit on this branch is me copying over the original Python templates to a new folder - you may want to compare against that commit to see the actual changes to the templates

### DIFF
--- a/gcp/website/frontend3/src/go/templates/linter/index.html
+++ b/gcp/website/frontend3/src/go/templates/linter/index.html
@@ -1,4 +1,5 @@
-{% set active_section = 'linter' %} {% block content %}
+{{ define "top_bar" }}<!-- empty top_bar -->{{ end }}
+{{ define "content" }}
 <!DOCTYPE html>
 <html lang="en">
 
@@ -95,4 +96,4 @@
 </body>
 
 </html>
-{% endblock %}
+{{ end }}
--- a/gcp/website/frontend3/src/go/templates/linter/index.html
+++ b/gcp/website/frontend3/src/go/templates/linter/index.html
@@ -1,0 +1,98 @@
+{% set active_section = 'linter' %} {% block content %}
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Vulnerability Linter Report</title>
+    <link rel="icon" href="https://google.github.io/osv.dev/assets/icon.png" type="image/x-icon" />
+    <link href="https://fonts.googleapis.com/css?family=Overpass" rel="stylesheet" />
+    <link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons" />
+    <script src="https://cdn.jsdelivr.net/npm/js-yaml@4.1.0/dist/js-yaml.min.js"></script>
+</head>
+
+<body>
+    <div class="container">
+        <header>
+            <div id="header-left">
+                <img src="https://raw.githubusercontent.com/google/osv-scanner/refs/heads/main/docs/images/osv-scanner-OSV-logo-darkmode.png"
+                    alt="OSV-Scanner Logo" class="logo" />
+                <div class="vl"></div>
+                <h1>Open Source Vulnerabilities</h1>
+                <div id="global-loader" class="loader">
+                    <i class="material-icons">sync</i>
+                </div>
+            </div>
+            <div id="header-right">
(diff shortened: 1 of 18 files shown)


go/internal/website/linter.go
```diff
@@ -29,5 +35,10 @@ func (s *Server) handleLinterFindings(w http.ResponseWriter, r *http.Request) {
 
 	// TODO: Download linter-result/<source>/result.json from GCS bucket
 	w.Header().Set("Content-Type", "application/json")
-	http.Error(w, fmt.Sprintf(`{"error": "Linter findings handler stub", "source": %q}`, source), http.StatusNotImplemented)
+	//nolint:errchkjson
+	_ = json.NewEncoder(w).Encode(map[string][]map[string]string{
```
another-rex, line 39: Are these just placeholders?

gcp/website/frontend3/src/go/templates/triage.html
```diff
@@ -1,7 +1,11 @@
-{% extends 'base.html' %}
-{% set active_section = 'triage' %}
+{{ define "extra_head" }}
 
-{% block content %}
+<head>
+  <!-- Allow webpack to insert extra head content -->
+</head>
```
another-rex: Isn't this `<head>` part nested inside another `<head>`

gcp/website/frontend3/src/go/templates/linter/index.html
another-rex, line 12: Should all this wrapping stuff be removed, since this is being rendered inside base.html now.

another-rex: Or maybe it just shouldn't be rendered inside base.html

go/internal/website/blog.go
```diff
@@ -1,45 +1,117 @@
...
+		Content: postHTML,
+	}
+
+	s.render(w, r, "blog_post.html", http.StatusOK, data)
 }
 
 // handleBlogPostFile handles serving static assets inside blog post directories /blog/posts/{blog_name}/{file_name}.
 func (s *Server) handleBlogPostFile(w http.ResponseWriter, r *http.Request) {
 	blogName := r.PathValue("blog_name")
 	fileName := r.PathValue("file_name")
```
another-rex, line 94: Can you clean fileName to make sure it doesn't have ".." and stuff as well?

go/internal/website/server.go
```diff
@@ -101,6 +105,19 @@ func loggingMiddleware(next http.Handler) http.Handler {
 	})
 }
 
+// notFoundMiddleware returns an http.Handler that renders 404.html for unmatched routes.
+func (s *Server) notFoundMiddleware(next http.Handler) http.Handler {
```
another-rex, line 109: This only covers unmatched routes, can we also cover routes that return 404?

go/internal/website/linter.go
michaelkedar: yeah - the real data comes from GCS buckets that I didn't want to set up just yet

go/internal/website/blog.go
michaelkedar: We only use `fileName` to call `path.Base` on it, so I've just added a check if the base is `.` or `..`

go/internal/website/server.go
michaelkedar: It's not really possible. Handlers write directly to the `*http.Request` object, so I don't think we can easily intercept that.
Pages that need to return 404s themselves can call `s.RenderNotFound` instead

gcp/website/frontend3/src/go/templates/triage.html
michaelkedar: Yeah - it did seem to be working (on Chrome, at least) but it is technically invalid HTML.
I've changed it to use a specific webpack tag thing which should be more correct now.

gcp/website/frontend3/src/go/templates/linter/index.html
michaelkedar: made a new `renderStandalone` to render this without `base.html`

go/internal/website/server.go
another-rex: Apparently you are meant to do this by wrapping the http.ResponseWriter that you pass to next.ServeHTTP. Not sure if it's worth it for 404 handling

another-rex (approved):

PR #5700, opened 2026-07-24
ci: attempt to get serviceContext properly showing in gcp

jess-lowe:
We want the cron job name at least showing in the "seen in" section of error reporting in GCP - Gemini thinks this will do that.

### DIFF
--- a/go/logger/init.go
+++ b/go/logger/init.go
@@ -38,18 +38,26 @@ func InitGlobalLogger() {
 			return
 		}
 
+		serviceName := os.Getenv("K_SERVICE")
+		if serviceName == "" {
+			serviceName = filepath.Base(os.Args[0])
+		}
+
 		projectID = os.Getenv("GOOGLE_CLOUD_PROJECT")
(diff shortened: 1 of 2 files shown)


another-rex (dismissed):
let's try it I guess

go/logger/init.go
```diff
@@ -38,18 +38,26 @@ func InitGlobalLogger() {
 			return
 		}
 
+		serviceName := os.Getenv("K_SERVICE")
```
michaelkedar, line 41: I'm pretty sure `K_SERVICE` is only set for Cloud Run, and not GKE

michaelkedar: Oh right it's falling back - please keep the original comment here
```
// Fallback to binary name for GKE services where K_SERVICE is not set.
```

michaelkedar (approved):
If the binary name isn't working out, I suspect something in the metadata server can give us what we want:
https://pkg.go.dev/cloud.google.com/go/compute/metadata@v0.9.0

PR #5712, opened 2026-07-27
feat(website): custom collapsible JSON viewer with finding path highlighting

rjerrems:
Note this was heavily assisted by Antigravity to paper over my failings as a frontend developer.

Right now the code is probably a little more complex than it needs to be, which we can refactor if we provide more specific information where the linting errors are occurring in the file.

You can collapse/expand JSON, there is a JSON copy button now and also you can click on a finding and it will be highlighted in yellow. See attached screenshots.

Screenshot 2026-07-27 11 39 21 AM
Screenshot 2026-07-27 11 39 04 AM

### DIFF
--- a/gcp/website/frontend3/src/linter.js
+++ b/gcp/website/frontend3/src/linter.js
@@ -104,7 +104,10 @@ document.addEventListener("DOMContentLoaded", function () {
             const url = `/linter-findings/${sourceName}`;
             return fetch(url)
               .then((res) => (res.ok ? res.json() : {}))
-              .catch(() => ({}));
+              .catch((e) => {
+                  console.warn(`Failed to fetch linter source: ${sourceName}`, e);
+                  return {};
+              });
           });
           const linterResults = await Promise.all(linterPromises);
           linterResults.forEach((data) => {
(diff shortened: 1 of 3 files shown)


gcp/website/frontend3/src/linter.js
```diff
@@ -90,20 +95,29 @@ document.addEventListener("DOMContentLoaded", function () {
...
+      try {
+        const rootRes = await fetch("/linter-findings/");
+        if (rootRes.ok) {
+          const activeSources = await rootRes.json();
+          const sourcesToFetch = sourceNames.filter((name) => activeSources.includes(name));
+          const linterPromises = sourcesToFetch.map((sourceName) => {
+            const url = `/linter-findings/${sourceName}`;
+            return fetch(url)
+              .then((res) => (res.ok ? res.json() : {}))
+              .catch(() => ({}));
```
jess-lowe:
```suggestion
              .catch((e) => {
                  console.warn(`Failed to fetch linter source: ${sourceName}`, e);
                  return {};
              });
```
nit: this is currently failing silently if there is a network parsing error - not super important but probably still good to know for debugging.

jess-lowe (dismissed):
Looks great! Small nit about catching network/parsing errors, but otherwise I've tested it locally, and it works great too :)

rjerrems:
> Looks great! Small nit about catching network/parsing errors, but otherwise I've tested it locally, and it works great too :)

Oh cool, thanks for picking that up. I have applied but seems I need your approval again :)

jess-lowe (approved):

PR #5713, opened 2026-07-27
perf(debian-cve-convert): reduce memory usage and streamline Debian CVE converter

jess-lowe:
tl;dr: add a lightweight nvd struct for creating debian records and do gcs download in go

#### **Changes**
- Added `vulns.LoadTargetCVEMetadata` to decode only essential fields (`ID`, `Published`, `LastModified`, `Metrics`) from NVD JSONs, bypassing memory-heavy CPE configuration trees, descriptions, and reference lists.
- Refactored vulnerability generation into `buildDebianVulnerabilities()` and added `runtime.GC()` prior to uploading to allow immediate reclamation of intermediate package maps.
- Added `-download-from-gcs` flag to download NVD data directly in Go via `gcs.DownloadBucket`, removing reliance on `gcloud storage cp` in `run_debian_convert.sh`.
- Populated `v.Modified` timestamps on generated OSV records using NVD `lastModified` metadata (with fallback to `published`).

basically, the debian-cve-convert cron job is going to continue to need more and more memory as it loads all of the nvd cve records into memory, and there's been massive dumps of vulns recently (and they will continue to grow) it makes more sense to try and be more conservative about how we handle the memory problems.


vulnfeeds/cmd/converters/debian/run_debian_convert.sh
another-rex: Can we set all the variables in kubernetes config

vulnfeeds/cmd/converters/debian/main.go
```diff
@@ -28,33 +31,64 @@ const (
...
+	inputBucketDefault       = "osv-test-cve-osv-conversion"
 	outputBucketDefault      = "debian-osv"
 )
 
 func main() {
 	logger.InitGlobalLogger()
 	defer logger.Close()
 
+	cvePath := flag.String("cve-path", defaultCvePath, "Path to CVE JSON files.")
+	downloadFromGCS := flag.Bool("download-from-gcs", false, "If true, download NVD CVE data from GCS bucket before processing.")
```
another-rex: I don't think we need 2 flags. If inputBucketName is set, download from GCS, otherwise don't.

vulnfeeds/vulns/vulns.go
```diff
@@ -863,6 +866,96 @@ func CheckQuality(text string) QualityCheck {
...
+
+	err := filepath.WalkDir(cvePath, func(filePath string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+		if d.IsDir() || !strings.HasSuffix(d.Name(), ".json") {
+			return nil
+		}
+
+		wg.Add(1)
```
another-rex, line 910: 1 goroutine per file? This will need some comments saying that there's only 1 file per year, so this is not a huge number.

vulnfeeds/cmd/converters/debian/main.go
```diff
@@ -28,33 +31,64 @@ const (
...
+	if *downloadFromGCS {
+		logger.Info("Downloading NVD CVE data from GCS bucket", slog.String("bucket", *inputBucketName), slog.String("dest", *cvePath))
+		storageClient, err := storage.NewClient(ctx)
+		if err != nil {
+			logger.Fatal("Failed to create GCS client", slog.Any("err", err))
+		}
+		defer storageClient.Close()
+
+		bkt := storageClient.Bucket(*inputBucketName)
+		if err := gcs.DownloadBucket(ctx, bkt, "nvd/", *cvePath); err != nil {
```
another-rex, line 64: It's a bit weird that we are writing all of this to a folder, then reading it from the folder again. It feels like it can just be 1 operation, and nothing should need to touch disk.

another-rex: Please add a TODO here for this, and let's fix it in the fixit week.

another-rex (commented):
Nice!

another-rex (dismissed):

another-rex (approved):

PR #4388, opened 2025-11-22
Add DHI ecosystem

cdupuis:
Adding new ecosystem for Docker Hardened Images. relates to https://github.com/google/osv.dev/issues/4387.

PR #5715, opened 2026-07-27
fix(purl): resolve Docker Hardened Images apk/dhi and deb/dhi PURLs

aubm:
DHI advisories publish release-lineage PURLs — `pkg:apk/dhi/<name>` (Alpine lineage) and `pkg:deb/dhi/<name>` (Debian lineage) — but the reverse PURL map only knew `pkg:dhi/<name>`, so PURL-form `/v1/query` requests never resolved to the `Docker Hardened Images` ecosystem.

This adds `(apk, dhi)` and `(deb, dhi)` to the reverse lookup map. The `dhi` namespace disambiguates from Alpine `(apk, alpine)` and Debian `(deb, debian)`, so there is no collision.

Forward generation (`package_to_purl`) is intentionally left as `pkg:dhi/…`: it receives the ecosystem name after `ecosystems.normalize()` has stripped the `:Alpine:`/`:Debian:` lineage, so it cannot choose apk vs deb — and DHI feed records always ship their own PURL, so that path is not exercised for DHI in practice.

Tests added in `osv/purl_helpers_test.py`. Relates to #4388.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/go/purl/ecosystems_simple.go
+++ b/go/purl/ecosystems_simple.go
@@ -67,14 +67,6 @@ func init() {
 	registerSimple(osvconstants.EcosystemBitnami, "bitnami", "", nil)
 	registerSimple(osvconstants.EcosystemCRAN, "cran", "", nil)
 	registerSimple(osvconstants.EcosystemConanCenter, "conan", "", nil)
-	registerSimple(osvconstants.EcosystemDockerHardenedImages, "dhi", "", nil)
-	// DHI advisories also publish release-lineage PURLs -- pkg:apk/dhi/<name>
-	// for the Alpine lineage and pkg:deb/dhi/<name> for Debian -- so map both
-	// back to the ecosystem for PURL queries. Generation stays pkg:dhi/ (the
-	// registerSimple above); the dhi namespace disambiguates from Alpine
-	// (apk/alpine) and Debian (deb/debian), so there is no parser collision.
-	registerParser("apk", "dhi", simpleParser{ecosystem: osvconstants.EcosystemDockerHardenedImages})
-	registerParser("deb", "dhi", simpleParser{ecosystem: osvconstants.EcosystemDockerHardenedImages})
 	registerSimple(osvconstants.EcosystemHackage, "hackage", "", nil)
 	registerSlash(osvconstants.EcosystemHex, "hex")
 	registerSimple(osvconstants.EcosystemJulia, "julia", "", nil)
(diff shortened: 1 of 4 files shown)


Ly-Joey:
Hi @aubm, thanks for picking up the onboarding of Docker Hardened Images!
We have migrated our logic for purl conversion to Go, so please add the PURL conversion logic and tests to `go/purl` instead.

aubm:
> Hi @aubm, thanks for picking up the onboarding of Docker Hardened Images! We have migrated our logic for purl conversion to Go, so please add the PURL conversion logic and tests to `go/purl` instead.

Thanks @Ly-Joey! Should be good to test now.

go/purl/ecosystems_simple.go
```diff
@@ -68,6 +68,13 @@ func init() {
 	registerSimple(osvconstants.EcosystemCRAN, "cran", "", nil)
 	registerSimple(osvconstants.EcosystemConanCenter, "conan", "", nil)
 	registerSimple(osvconstants.EcosystemDockerHardenedImages, "dhi", "", nil)
+	// DHI advisories also publish release-lineage PURLs -- pkg:apk/dhi/<name>
+	// for the Alpine lineage and pkg:deb/dhi/<name> for Debian -- so map both
+	// back to the ecosystem for PURL queries. Generation stays pkg:dhi/ (the
+	// registerSimple above); the dhi namespace disambiguates from Alpine
+	// (apk/alpine) and Debian (deb/debian), so there is no parser collision.
+	registerParser("apk", "dhi", simpleParser{ecosystem: osvconstants.EcosystemDockerHardenedImages})
+	registerParser("deb", "dhi", simpleParser{ecosystem: osvconstants.EcosystemDockerHardenedImages})
```
michaelkedar: nit: just to make this easier to visually scan, can we move all the dhi stuff later in the init function (so all the regular non-namespace ecosystems stay together, and this comment doesn't break it up)

michaelkedar (commented):
Happy to merge this, just one tiny thing

go/purl/ecosystems_simple.go
aubm: Sure thing! You got it in https://github.com/google/osv.dev/commit/d2fdf30e31ad45fee162b09ad38e3a1140ccda23

michaelkedar (approved):
LGTM

PR #5718, opened 2026-07-28
refactor(website): vulnerability details page in go

michaelkedar:
Migrate all the logic required to render the vulnerability details page in the website into Go.

Still no database connections, so the rendered vuln is just a placeholder downloaded directly from the prod GCS bucket.


go/internal/website/vulnerability_models.go
```diff
@@ -0,0 +1,481 @@
...
+		case osvschema.Severity_CVSS_V3:
+			isCVSS = true
+			if strings.HasPrefix(score, "CVSS:3.1") {
+				parsed, err := gocvss31.ParseVector(score)
+				if err != nil {
+					logger.Error("failed to parse CVSS V3.1 vector", slog.Any("error", err))
+					continue
+				}
+				cvssScore = parsed.BaseScore()
+				cvssLevel, _ = gocvss30.Rating(cvssScore)
```
another-rex: This is incorrect?

Also nit, I think we should just swap the if strings.HasPrefix(score, "CVSS:3.1") around as well. so as you read down in the code it goes up in version number

go/internal/website/vulnerability_models.go
```diff
@@ -0,0 +1,481 @@
...
+				calcURL = firstCVSSCalculatorBaseURL + "/3.1#" + score
+			} else {
+				parsed, err := gocvss30.ParseVector(score)
+				if err != nil {
+					logger.Error("failed to parse CVSS V3.0 vector", slog.Any("error", err))
+					continue
+				}
+				cvssScore = parsed.BaseScore()
+				cvssLevel, _ = gocvss31.Rating(cvssScore)
```
another-rex: same here

go/internal/website/vulnerability_models.go
```diff
@@ -0,0 +1,481 @@
...
+		if ranges := entry.GetRanges(); len(ranges) > 0 {
+			totalPackageLen += len(ranges[0].GetRepo())
+		}
+	}
+
+	return max(totalEcosystemLen, totalPackageLen) > 100
+}
+
+// Severities calculates and returns display data for all severity entries on the vulnerability.
+func (v VulnerabilityPageData) Severities() []SeverityDisplay {
```
another-rex: It does feel a bit weird that all this logic is in models? Maybe these functions should be in a separate file.

another-rex (commented):
Nice, just some nits

go/internal/website/vulnerability_models.go
michaelkedar: fixed, and re-ordered

go/internal/website/vulnerability_models.go
michaelkedar: moved these functions into `vulnerability_view.go` (and added a few tests)

another-rex (dismissed):

another-rex (approved):

PR #5720, opened 2026-07-28
perf(relations): parallelize relations worker using Datastore key range sharding

another-rex:
## Description
- Add `-breakdown-prefixes` CLI flag to relations worker to enable parallel key range sharding across `errgroup` worker goroutines.
- Implement generic `runShardedQuery[E any]` executor in `go/cmd/relations/sharding.go` streaming entities directly to an `ItemHandler` callback without intermediate slice allocations.
- Enable parallel key range sharding across `upstream`, `alias`, and `related` group computations.
- Update CronJob deployment configs (`oss-vdb` and `oss-vdb-test`) with `--breakdown-prefixes=...` container arguments.

agy


go/cmd/relations/sharding.go
```diff
@@ -0,0 +1,119 @@
...
+	"google.golang.org/api/iterator"
+)
+
+// KeyShard represents a Datastore key range [Start, End) for sharded queries.
+type KeyShard = sharding.KeyShard
+
+// ParseBreakdownPrefixes parses a comma-separated string of key prefix breakdowns into KeyShards.
+func ParseBreakdownPrefixes(str string) []KeyShard {
+	return sharding.ParseBreakdownPrefixes(str)
+}
```
michaelkedar: why do we have these instead of using the `internal/sharding` types directly?

michaelkedar (dismissed):

go/cmd/relations/sharding.go
another-rex: Yeah good point, this was left over from a refactor. Removed and moved the sharding code into an internal library that's more easily share-able. I'll have a look at abstracting away datastore too

michaelkedar (approved):

PR #5742, opened 2026-07-29
feat: gitter client

Ly-Joey:
This PR adds the `go/internal/gitter` package containing the `Client` interface and HTTP implementation for interacting with Gitter.
This will also make it easier to swap to gRPC in the future.

Follow-up PR for importer integration: #5744

### DIFF
--- a/go/internal/gitter/http.go
+++ b/go/internal/gitter/http.go
@@ -99,6 +99,20 @@ func (c *httpClient) do(ctx context.Context, method, path string, query url.Valu
 	}
 }
 
+// Helper function to unmarshal the response body into target proto message
+func unmarshalProtoResponse(r io.Reader, msg proto.Message) error {
+	bodyBytes, err := io.ReadAll(r)
+	if err != nil {
+		return fmt.Errorf("failed reading response body: %w", err)
+	}
+
+	if err := proto.Unmarshal(bodyBytes, msg); err != nil {
+		return fmt.Errorf("failed decoding response: %w", err)
+	}
+
+	return nil
+}
+
 // GetGit handles GET /git (streaming tarball)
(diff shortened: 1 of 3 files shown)


go/internal/gitter/http.go
```diff
@@ -0,0 +1,221 @@
...
+func (c *httpClient) Cache(ctx context.Context, repoURL string) error {
+	req := &pb.CacheRequest{
+		Url: repoURL,
+	}
+
+	resp, err := c.do(ctx, http.MethodPost, "/cache", nil, req)
+	if err != nil {
+		return err
+	}
+	resp.Body.Close()
```
another-rex, line 148: Not sure if this is necessary, but apparently not draining the body makes it fail to reuse the connection:

```suggestion
	_, _ = io.Copy(io.Discard, resp.Body)
	resp.Body.Close()
```

I think the cache endpoint doesn't  return anything anyway.

go/internal/gitter/http.go
```diff
@@ -0,0 +1,221 @@
...
+
+	bodyBytes, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("failed reading tags response body: %w", err)
+	}
+
+	var tagsResp pb.TagsResponse
+	if err := proto.Unmarshal(bodyBytes, &tagsResp); err != nil {
+		return nil, fmt.Errorf("failed decoding tags response: %w", err)
+	}
```
another-rex: We probably can put all these into a helper function? Since it's repeated so often.

go/internal/gitter/http.go
Ly-Joey: Added!

go/internal/gitter/http.go
Ly-Joey: done :)

another-rex (approved):

PR #5744, opened 2026-07-29
feat: gitter importer integration

Ly-Joey:
Make importer use gitter's new /file-diffs and /file-content endpoints via gitter client, instead of cloning the source repos and reading from local.

~Stack PR isn't a thing yet so this PR also includes changes from #5742. But once we merge and rebase, this PR should be nicer to look at.~ GitHub released stack PR today just to spite me I guess 😔

### DIFF
--- a/go/internal/importer/git.go
+++ b/go/internal/importer/git.go
@@ -89,9 +89,6 @@ func handleImportGit(ctx context.Context, ch chan<- WorkItem, config Config, sou
 
 	resp, err := fetchGitterFileDiffs(ctx, config.GitterClient, sourceRepo, sourceRepo.Git.LastSyncedCommit)
 	if err != nil {
-		logger.ErrorContext(ctx, "Failed to get file diffs from gitter",
-			slog.Any("error", err), slog.String("source", sourceRepo.Name))
-
 		return err
 	}
 
--- a/go/internal/importer/git.go
+++ b/go/internal/importer/git.go
@@ -3,113 +3,49 @@ package importer
 import (
 	"bytes"
 	"context"
-	"crypto/sha256"
-	"encoding/hex"
 	"errors"
 	"io"
-	"io/fs"
 	"log/slog"
-	"os"
 	"path"
-	"path/filepath"
 	"strings"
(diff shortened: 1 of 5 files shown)


michaelkedar:
> Stack PR isn't a thing yet

You spoke too soon, I guess

go/internal/importer/git.go
```diff
@@ -149,8 +122,10 @@ func handleImportGit(ctx context.Context, ch chan<- WorkItem, config Config, sou
 			case ch <- WorkItem{
 				Context: ctx,
 				SourceRecord: gitSourceRecord{
-					path: from,
-					repo: repo,
+					client:  config.GitterClient,
+					repoURL: sourceRepo.Git.URL,
+					commit:  latestCommit,
+					path:    from,
 				},
```
michaelkedar: This should probably be nil.
The handle for the `ActionWithdraw` doesn't open the old record, and also the file path at this commit is definitionally non-existent here.

go/internal/importer/git.go
```diff
@@ -330,52 +194,37 @@ func handleReconcileGit(ctx context.Context, ch chan<- WorkItem, config Config,
...
-	err = filepath.WalkDir(searchDir, func(p string, d fs.DirEntry, err error) error {
-		if err := ctx.Err(); err != nil {
-			return err
-		}
-		if err != nil {
-			return err
+	// Query Gitter with empty lastSyncedCommit to get all files in the repository
+	resp, err := fetchGitterFileDiffs(ctx, config.GitterClient, sourceRepo, "")
+	if err != nil {
+		if !errors.Is(err, context.Canceled) {
```
michaelkedar: can we use this helper here. https://github.com/google/osv.dev/blob/af55634a4dd610669e80265e0e652c6f310628f8/go/internal/osvutil/errors.go#L13

I am paranoid that if we move to grpc, it won't wrap the context errors correctly.

michaelkedar (dismissed):
LGTM, a couple of small things

michaelkedar (approved):

PR #5746, opened 2026-07-29
fix: standardize default pub/sub pool name

michaelkedar:
> [!IMPORTANT]
> Before merging / deploying this, make sure the importers are paused and the Pub/Sub queues are empty!
> Renaming the Pub/Sub subscriptions forces them to be destroyed and re-created

After being confused for a while on the private instance, I discovered that having the name of the default Pub/Sub subscription be `default-pool` while having the filter that applies to it be `default` (which is different to how all our extra pools are done, which have the same name/field filter) is a confusing footgun.

I'm renaming things so these are both `default` so I don't spend an hour wondering why my Pub/Sub messages are disappearing again.

### DIFF
--- a/deployment/clouddeploy/gke-workers/base/core/importer-reconciler.yaml
+++ b/deployment/clouddeploy/gke-workers/base/core/importer-reconciler.yaml
@@ -27,8 +27,8 @@ spec:
                 value: "1.0"
               - name: TRACE_SAMPLE_RATE # for the individual vulnerability entries
                 value: "0.05"
-              - name: REIMPORT_TASK_POOL
-                value: "reimport"
+              - name: DEFAULT_TASK_POOL
+                value: "default"
             resources:
               requests:
                 cpu: "1"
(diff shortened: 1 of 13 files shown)


deployment/clouddeploy/gke-workers/base/core/importer-reconciler.yaml
```diff
@@ -27,8 +27,8 @@ spec:
                 value: "1.0"
               - name: TRACE_SAMPLE_RATE # for the individual vulnerability entries
                 value: "0.05"
-              - name: REIMPORT_TASK_POOL
-                value: "reimport"
+              - name: DEFAULT_TASK_POOL
```
another-rex: Please add this to the PR description

another-rex (dismissed):

deployment/clouddeploy/gke-workers/base/core/importer-reconciler.yaml
michaelkedar: This was actually wrong - the reconciler does use the `REIMPORT_TASK_POOL`

Yet another reason to not trust AI.

another-rex (approved):

PR #5747, opened 2026-07-29
fix(purl): keep the package repository in generated PURLs

### LINKED ISSUE #5683: DRUPAL generated PURLs incorrect
DRUPAL vulnerabilities are often not on the default packagist.org repository, but this information is dropped when we generate the PURLs for them.
e.g. [DRUPAL-CONTRIB-2026-079](https://osv.dev/vulnerability/DRUPAL-CONTRIB-2026-079) affects ecosystems `Packagist:https://packages.drupal.org/8`, but the generated PURL uses the generic composer PURL type without any qualifiers - `pkg:composer/drupal/commerce_guest_registration`.

From my reading of the packageurl spec, the unqualified [`composer` type refers to `packagist.org`](https://packageurl.org/docs/purl-spec/types-doc/composer-definition#repository-information), but we can use a [`repository_url` qualifier to point to a specific repository](https://packageurl.org/docs/purl/common-qualifiers)

In that example, the generated PURL should probably be `pkg:composer/drupal/commerce_guest_registration?repository_url=https://packages.drupal.org/8`

We also would want to make the PURL -> OSV conversion grab this field, so API queries with PURLs work as expected.

Eljees:
Fixes #5683

`package_to_purl` was called with `ecosystems.normalize(...)`, which drops everything after the colon —
so `Packagist:https://packages.drupal.org/8` became `Packagist` and the repository was lost from the
generated PURL.

Both callers now pass the full ecosystem, and `package_to_purl` turns a repository suffix into the
`repository_url` qualifier:

```
Packagist:https://packages.drupal.org/8 + drupal/commerce_guest_registration
  -> pkg:composer/drupal/commerce_guest_registration?repository_url=https://packages.drupal.org/8
```

Only a suffix that starts with `http://` or `https://` is treated as a repository, so release suffixes
keep behaving exactly as before (`Debian:11`, `Alpine:v3.17`). Qualifiers are emitted sorted by key, so
an ecosystem that has both still produces a valid PURL
(`?arch=source&repository_url=...`).

`parse_purl` does the reverse: a `repository_url` qualifier is appended back to the ecosystem name, so an
API query with the generated PURL resolves to `Packagist:https://packages.drupal.org/8` — that is the
second half of the issue.

### Verified

The repo's own `osv/purl_helpers_test.py` imports through `osv/__init__.py`, which pulls the Cloud
dependencies, so I could not run it as-is locally; I ran the same expectations against the module loaded
directly in `python:3.12`:

```
== package_to_purl ==
  OK  Packagist:https://packages.drupal.org/8 -> pkg:composer/drupal/commerce_guest_registration?repository_url=https://packages.drupal.org/8
  OK  Packagist                               -> pkg:composer/symfony/http-kernel
  OK  Debian:11                               -> pkg:deb/debian/nginx?arch=source
  OK  Debian:https://example.com/deb          -> pkg:deb/debian/nginx?arch=source&repository_url=https://example.com/deb
  OK  Alpine:v3.17                            -> pkg:apk/alpine/postgresql14?arch=source
  OK  Maven                                   -> pkg:maven/org.apache.struts/struts2-core
== parse_purl ==
  OK  ...?repository_url=https://packages.drupal.org/8  -> ('Packagist:https://packages.drupal.org/8', 'drupal/commerce_guest_registration', None)
  OK  pkg:composer/symfony/http-kernel@6.0.0            -> ('Packagist', 'symfony/http-kernel', '6.0.0')
  OK  pkg:deb/debian/nginx?arch=source                  -> ('Debian', 'nginx', None)
```

The same cases are added to `purl_helpers_test.py` as `test_package_to_purl_repository_url` and
`test_parse_purl_repository_url`, so CI runs them with the real dependencies.

One thing I deliberately did not do: the qualifier value is written verbatim, matching the example in the
issue. If you'd rather have it percent-encoded per the PURL spec, say so and I'll change it — that also
affects what existing consumers would need to match on.

AI-assisted (LLM used for drafting); the change and the runs above are mine.

### DIFF
--- a/gcp/workers/worker/worker.py
+++ b/gcp/workers/worker/worker.py
@@ -552,9 +552,8 @@ class TaskRunner:
         source_link = source_repo.link + relative_path
       for affected in vulnerability.affected:
         if not affected.package.purl:
-          if purl := purl_helpers.package_to_purl(
-              osv.ecosystems.normalize(affected.package.ecosystem),
-              affected.package.name):
+          if purl := purl_helpers.package_to_purl(affected.package.ecosystem,
+                                                  affected.package.name):
             affected.package.purl = purl
         if source_link:
           affected.database_specific.update({'source': source_link})
--- a/osv/models.py
+++ b/osv/models.py
@@ -381,8 +381,8 @@ class Bug(ndb.Model):
     for pkg in self.affected_packages:
       # Set PURL if it wasn't provided.
       if not pkg.package.purl:
-        pkg.package.purl = purl_helpers.package_to_purl(
-            ecosystems.normalize(pkg.package.ecosystem), pkg.package.name)
+        pkg.package.purl = purl_helpers.package_to_purl(pkg.package.ecosystem,
+                                                        pkg.package.name)
 
     self.project = list({
         pkg.package.name for pkg in self.affected_packages if pkg.package.name
--- a/osv/purl_helpers.py
+++ b/osv/purl_helpers.py
@@ -126,7 +126,11 @@ def _url_encode(package_name):
 
 def package_to_purl(ecosystem: str, package_name: str) -> str | None:
   """Convert a ecosystem and package name to PURL."""
-  purl_data = ECOSYSTEM_PURL_DATA.get(ecosystem)
+  # Some ecosystems name a non-default package repository after a colon, e.g.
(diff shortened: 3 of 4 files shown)


Eljees:
﻿@googlebot I signed it!

G-Rath:

PR #5757, opened 2026-07-31
perf: Move to new n4 nodes to avoid scaling issues and save on costs

another-rex:
- Faster
- Cheaper
- More reliable (Because we actually have more instances of n4 that can be scaled up to.

Why not.

Apparently emptyDir is shares/uses the boot drive now. Not that we really need it. Adding a extra 500 for highend for uses like the indexer.

### DIFF
--- a/deployment/terraform/modules/osv/workers_gke.tf
+++ b/deployment/terraform/modules/osv/workers_gke.tf
@@ -67,13 +67,11 @@ resource "google_container_node_pool" "default_pool" {
 
 
   node_config {
-    machine_type    = "n1-highmem-2"
-    disk_type       = "pd-ssd"
-    disk_size_gb    = 64
-    local_ssd_count = 1
+    machine_type = "n4-standard-8"
+    disk_type    = "hyperdisk-balanced"
+    disk_size_gb = 64
 
     oauth_scopes = ["https://www.googleapis.com/auth/cloud-platform"]
-
   }
 }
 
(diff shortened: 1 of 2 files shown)


deployment/terraform/modules/osv/workers_gke.tf
```diff
@@ -67,13 +67,11 @@ resource "google_container_node_pool" "default_pool" {
 
 
   node_config {
-    machine_type    = "n1-highmem-2"
-    disk_type       = "pd-ssd"
-    disk_size_gb    = 64
-    local_ssd_count = 1
+    machine_type = "n4-standard-8"
```
michaelkedar, line 70: any reason to go from `n1-highmem-2` (2 CPU / 13 GB) to `n4-standard-8` (8 CPU / 32GB)
rather than, say, `n4-highmem-2` (2 CPU / 16GB)?

deployment/terraform/modules/osv/workers_gke.tf
```diff
@@ -179,7 +172,7 @@ resource "google_container_node_pool" "worker_pool_temp" {
   }
 
   node_config {
-    machine_type = "n4-highcpu-8"
+    machine_type = "n4-standard-8"
```
michaelkedar: Why change this?
iirc, I chose `highcpu` to optimize the packing of the workers on these machines (which, like, limit 1.5CPU, 1.3 GiB)

deployment/terraform/modules/osv/workers_gke.tf
another-rex: I want the workers pool to be  actually usable for general things and not have to rely on the highend pool any time we want to bump up the CPU count a bit. This way it's a lot more flexible with what type of work can go on the pool. (e.g. some job can have 6 cpus and still be on this pool).

another-rex: WIth basically no work happening at the moment, there's 4 nodes allocated, so it's not too wasteful to be going in increments of 8 cpus at a time as well.

michaelkedar (approved):

PR #5761, opened 2026-08-03
feat(vulnfeeds): handle rejected CVEs and propagate withdrawn status

### LINKED ISSUE #4610: Withdraw CVE5 records that have changed status to "rejected"

jess-lowe:
If a CVE is marked as Rejected upstream (in NVD or CVEList), OSV will now handle it properly instead of ignoring the change or keeping stale data.

**How it works**

- If we already had the record in cve5 or nvd OSV records, it gets updated with a "Withdrawn" timestamp.
- If a rejected CVE is brand new to us, we make a tombstone record so it gets withdrawn properly.
- In the final step (combine-to-osv), if either source is rejected, the combined OSV file is deleted to keep the feed clean.
- We skip merging logic in combine-to-osv if we know a record is withdrawn.

(Tested and verified with local scripts and unit tests).

Closes #4610


vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -136,18 +136,21 @@ func readVulnerability(ctx context.Context, client *storage.Client, fullPath str
...
 func combineIntoOSV(cve5 *osvschema.Vulnerability, nvd *osvschema.Vulnerability) *osvschema.Vulnerability {
-	var baseOSV *osvschema.Vulnerability
-	if cve5 != nil && nvd != nil {
-		baseOSV = combineTwoOSVRecords(cve5, nvd)
-	} else if cve5 != nil {
-		baseOSV = cve5
-	} else if nvd != nil {
-		baseOSV = nvd
-	} else {
+	if (cve5 != nil && cve5.GetWithdrawn() != nil) || (nvd != nil && nvd.GetWithdrawn() != nil) {
```
another-rex: I don't think the cve5 != nil part is necessary. I think .GetWithdrawn works fine on nil items.

vulnfeeds/cmd/combine-to-osv/main_test.go
```diff
@@ -1075,6 +1075,58 @@ func TestCombineTwoOSVRecords(t *testing.T) {
...
+			name: "CVE5 not withdrawn, NVD withdrawn",
+			cve5: &osvschema.Vulnerability{
+				Id: "CVE-2023-1234",
+			},
+			nvd: &osvschema.Vulnerability{
+				Id:        "CVE-2023-1234",
+				Withdrawn: timestamppb.New(nvdWithdrawnTime),
+			},
+		},
+		{
```
another-rex, line 1112: Should this also test that one of them is nil?

vulnfeeds/cmd/combine-to-osv/main.go
```diff
@@ -136,18 +136,21 @@ func readVulnerability(ctx context.Context, client *storage.Client, fullPath str
...
-	return baseOSV
+	if cve5 != nil && nvd != nil {
+		return combineTwoOSVRecords(cve5, nvd)
+	}
+	if cve5 != nil {
+		return cve5
+	}
+	if nvd != nil {
+		return nvd
+	}
```
another-rex: can just return nvd

vulnfeeds/cmd/converters/cve/nvd-cve-osv/main.go
```diff
@@ -217,7 +217,7 @@ func worker(wg *sync.WaitGroup, jobs <-chan models.NVDCVE, gcsHelper *gcs.Helper
 
 		if outcome != models.Successful {
 			logger.Info("Failed to generate a successful OSV record", slog.String("cve", cveID), slog.String("outcome", outcome.String()))
-			if *rejectFailed {
+			if *rejectFailed && (vuln == nil || vuln.Withdrawn == nil) {
```
another-rex: is vuln here the proto? If so we can just use vuln.GetWithdrawn() == nil

another-rex: If not, probably just implement the vuln.GetWithdrawn() function to make it more readable.

another-rex (commented):
> If a rejected CVE is brand new to us, we ignore it.

Should we also just be publishing withdrawn entires for them rather than ignoring it?
At least, should we be doing this in the cve and nvd versions, and then in combine-to-osv decide whether to publish it. (To avoid cases where it's withdrawn in nvd, but there in cve, and we're just publishing it as if cve is the only source and not correctly marking it as withdrawn.)

vulnfeeds/cmd/converters/cve/cve5/bulk-converter/main.go
another-rex: Do we also need to change here? (I.e. all the places we do the outcome != successful check)

another-rex: Also here., maybe other places as well. Might be a good idea to package `metrics.Outcome != models.Successful` into a function so we can tweak the "not emit this file" logic it in one place.

vulnfeeds/conversion/cve5/converter.go
```diff
@@ -70,7 +70,18 @@ func getCWEs(cna models.CNA, metrics *models.ConversionMetrics) []string {
 // It populates the main fields of the OSV record, including ID, summary, details,
 // references, timestamps, severity, and version information.
 func FromCVE5(cve models.CVE5, refs []models.Reference, metrics *models.ConversionMetrics, sourceLink string) *vulns.Vulnerability {
```
another-rex, line 72: Can you add tests for this and nvd for "REJECTED" as well?

vulnfeeds/conversion/cve5/converter.go
```diff
@@ -70,7 +70,18 @@ func getCWEs(cna models.CNA, metrics *models.ConversionMetrics) []string {
...
+	var withdrawnTime *timestamppb.Timestamp
+	if cve.Metadata.State == "REJECTED" {
+		withdrawn, err := models.ParseCVE5Timestamp(cve.Metadata.DateUpdated)
+		if err != nil {
+			metrics.AddNote("[%s]: Withdrawn date failed to parse, setting time to now", cve.Metadata.CVEID)
+			withdrawn = time.Now()
```
another-rex: Hmm will this cause issues with the records changing non-deterministically? Every run it'll have a new withdrawn date. Maybe just set this to published or something

jess-lowe: bc they all kinda do that, I refactored it so that if published fails, use epoch time. If modified fails, use whatever published is, and use the same for withdrawn as modified.

vulnfeeds/conversion/cve5/converter.go
```diff
@@ -70,7 +70,25 @@ func getCWEs(cna models.CNA, metrics *models.ConversionMetrics) []string {
...
+
+	modified, err := models.ParseCVE5Timestamp(cve.Metadata.DateUpdated)
+	if err != nil {
+		metrics.AddNote("[%s]: Modified date failed to parse, falling back to Published time", cve.Metadata.CVEID)
+		modified = published
+	}
+
+	var withdrawnTime *timestamppb.Timestamp
+	if cve.Metadata.State == "REJECTED" {
+		withdrawnTime = timestamppb.New(modified)
```
another-rex: It looks like there's also "dateRejected", which might fit better here?

another-rex: https://test.osv.dev/triage?id=CVE-2026-51233&s1=test-nvd&s2=cve-org

vulnfeeds/conversion/nvd/converter.go
```diff
@@ -29,6 +29,17 @@ func CVEToOSV(cve models.NVDCVE, repos []string, vpRepoCache *c.VPRepoCache, cac
...
 
+	if cve.VulnStatus != nil && *cve.VulnStatus == "Rejected" {
+		metrics.SetOutcome(models.Rejected)
+		v := vulns.FromNVDCVE(cve.ID, cve)
+		databaseSpecific, err := utility.NewStructpbFromMap(make(map[string]any))
+		if err == nil {
+			v.DatabaseSpecific = databaseSpecific
+		}
+
+		return v, metrics, models.Rejected
```
another-rex, line 40: Hmm this returns early while the CVE one doesn't. Is this because NVD removes most of the details once it rejects a record?

another-rex (commented):
Some questions, but otherwise everything else LGTM.

vulnfeeds/conversion/nvd/converter.go
jess-lowe: I've made them both return early now - information outside the basic stuff isn't really relevant after rejection

vulnfeeds/conversion/cve5/converter.go
jess-lowe: good point - updated for cve5. nvd doesn't seem to have an equivalent so I'll keep it as modified date

another-rex (approved):

PR #5764, opened 2026-08-03
feat: Update gitter and importer_reconciler_git_cache to use hyperdisk

another-rex:
Update gitter and importer_reconciler_git_cache to use hyperdisk.

```
  lifecycle {
    ignore_changes = [
      type,
    ]
  }
```

should allow the changes to be ignored by terraform.


michaelkedar:
hmm
```
- snapshot                       = "https://www.googleapis.com/compute/v1/projects/oss-vdb-test/global/snapshots/importer-reconciler-git-cache-snapshot" -> null # forces replacement
```

(but why doesn't the gitter disk have this??)

another-rex:
I think it is just getting confused because I manually created the hyperdisk for the reconciler. I might have to apply locally and then rerun this

michaelkedar (approved):

PR #5769, opened 2026-08-04
feat: Tell me why you are a coward

another-rex:
Actually print out the entries that it's trying to delete. Allow a flag to override the number it shows.

### DIFF
--- a/go/internal/importer/bucket.go
+++ b/go/internal/importer/bucket.go
@@ -160,7 +160,7 @@ func handleDeleteBucket(ctx context.Context, ch chan<- WorkItem, config Config, 
 		logger.ErrorContext(ctx, "Cowardly refusing to delete missing records (threshold exceeded)",
 			slog.String("source", sourceRepo.Name),
 			slog.Int("to_delete", len(toDelete)),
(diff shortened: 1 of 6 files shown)


go/internal/importer/bucket.go
```diff
@@ -160,6 +160,7 @@ func handleDeleteBucket(ctx context.Context, ch chan<- WorkItem, config Config,
 		logger.ErrorContext(ctx, "Cowardly refusing to delete missing records (threshold exceeded)",
 			slog.String("source", sourceRepo.Name),
 			slog.Int("to_delete", len(toDelete)),
+			slog.String("to_delete_entries", formatEntries(toDelete, config.MaxCowardEntriesToShow)),
```
michaelkedar: Do you think this would be better as a `slog.Any` of `[]string`, so GCP logging can render it as a list rather than a long string?

jess-lowe:
200w

michaelkedar (commented):
> feat: Tell me why you are a coward

I'm feeling intimidated

Ly-Joey:
WomanYellingAtACat_meme

go/internal/importer/bucket.go
another-rex: Done.

michaelkedar (approved):

PR #5774, opened 2026-08-04
feat(sources): separate TuxCare test ingestion

Entropius:
Point the TuxCare entry in `source_test.yaml` at the `osv-test` branch so the test instance can ingest advisories separately from production.

The source link now points to the same branch. `source.yaml` remains unchanged.

Validated the source configuration with `tools/sourcerepo-sync/source_sync.py --validate`.


Entropius:
Hi @another-rex! Can I kindly ask for your review?

PR #5793, opened 2026-08-05
fix(importer): init gitter client in main but not for deleter  

Ly-Joey:
#5792 was a temporary fix, here's the real one.


michaelkedar (commented):
tbh, I think it's more correct to do it in main (but skip if it's the deleter ig)

michaelkedar (approved):

PR #5794, opened 2026-08-05
refactor(website): Migrate /list page to go

michaelkedar:
Final template migration for the Go website, for the /list page.

- Implemented a bunch of rendering functions needed (e.g. for the "just now", "x minutes ago", etc.
- Added new ListedVulnerability model for database abstrtactions to return.
- Did a little bit of refactoring to share some CVSS code between list and vulnerability page.
- Buncha placeholder data

---

### DIFF
--- a/gcp/website/frontend3/src/go/templates/list.html
+++ b/gcp/website/frontend3/src/go/templates/list.html
@@ -1,65 +1,36 @@
-{% extends 'base.html' %}
-{% set active_section = 'vulnerabilities' %}
-{% set disable_turbo_cache = 'true' %}
-
-{% macro table_header_cell(column_id, column_name, is_sortable, is_sorted, is_descending) %}
-<span class="vuln-table-cell mdc-data-table__header-cell vuln-table-header
-    {% if is_sortable %}mdc-data-table__header-cell--with-sort__DISABLED{% endif %}
-    {% if is_sorted %}mdc-data-table__header-cell--sorted{% endif %}
-    {% if is_descending %}mdc-data-table__header-cell--sorted-descending{% endif %}" role="columnheader" scope="col"
-  aria-sort="{% if is_sorted %}{% if is_descending %}descending{% else %}ascending{% endif %}{% else %}none{% endif %}"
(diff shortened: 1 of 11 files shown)


gcp/website/frontend3/src/go/templates/list.html
```diff
@@ -1,166 +1,155 @@
...
+                {{ .Summary }}
+              {{ else }}
                 See record for full details
-              {% endif %}
+              {{ end }}
             </span>
             <span role="cell" class="vuln-table-cell mdc-data-table__cell">
-              <relative-time datetime="{{ vulnerability.published }}">
-                {{ vulnerability.published | relative_time }}
+              <relative-time datetime="{{ .FormattedPublished }}">
```
another-rex: Can you just make this a span with a title="original date". We can look at this again in the future if the web component make sense to use.

another-rex: Actually, it looks like we are not even installing the relative-time web component

another-rex: It was removed here: 25bc7c661e2151eeb5ca1c317de25a57d0b4daf2

go/internal/website/list_models.go
```diff
@@ -0,0 +1,178 @@
...
+		pkg := v.Packages[i]
+		if pkg.Repo != "" {
+			result[i] = stripScheme(pkg.Repo)
+		} else if pkg.Package != nil {
+			if pkg.Package.GetEcosystem() == "" {
+				result[i] = pkg.Package.GetName()
+			} else {
+				result[i] = pkg.Package.GetEcosystem() + "/" + pkg.Package.GetName()
+			}
+		}
```
another-rex: Maybe just do append so there won't be empty entries if both parts of the if statement fails (That's probably pretty rare though)

gcp/website/frontend3/src/go/templates/list.html
michaelkedar: changed `<relative-time>` to a `span` with a title.

another-rex (approved):

PR #5797, opened 2026-08-05
feat: repo-based allowlist for git enumeration flags

Ly-Joey:
Add repo-based git analysis flags logic in worker before we send the request to gitter to enumerate affected commits.

List of configurable bools:
- consider_all_branches
- detect_cherrypicks_intro
- detect_cherrypicks_fixed
- detect_cherrypicks_limit

The allowlist is in datastore. 
#5800 is the tool to sync the allowlist (in yaml) into datastore entities.

PR #5800, opened 2026-08-06
feat: sync tool for repo-based git enumeration flags allowlist

Ly-Joey:
Add a tool to sync repo-based git analysis flags allowlist entries from yaml to datastore.

Related: #5797
(Stack PR makes me sad)


tools/repo-allowlist-sync/main.go
```diff
@@ -0,0 +1,245 @@
...
+// RepoAllowListEntity represents a repository allowlist entity stored in Cloud Datastore
+type RepoAllowListEntity struct {
+	Key                   *datastore.Key `datastore:"__key__"`
+	Type                  string         `datastore:"type"`
+	Value                 string         `datastore:"value"`
+	ConsiderAllBranches   bool           `datastore:"consider_all_branches"`
+	CherrypicksIntroduced bool           `datastore:"cherrypicks_introduced"`
+	CherrypicksFixed      bool           `datastore:"cherrypicks_fixed"`
+	CherrypicksLimit      bool           `datastore:"cherrypicks_limit"`
+}
```
another-rex: This should be using the go models. (Which I know you can't really do right now because it's not a stacked PR)

tools/repo-allowlist-sync/main.go
```diff
@@ -0,0 +1,245 @@
...
+			}
+			if _, err := regexp.Compile(raw.Value); err != nil {
+				log.Printf("Warning: Skipping invalid regex pattern %q: %v", raw.Value, err)
+				continue
+			}
+		default:
+			log.Printf("Warning: Skipping unrecognized entry type %q for value %q", raw.Type, raw.Value)
+			continue
+		}
+
```
another-rex: Let's immediately fail if something is invalid.

tools/repo-allowlist-sync/main.go
```diff
@@ -0,0 +1,245 @@
...
+	}
+
+	dsEntitiesMap := make(map[string]RepoAllowListEntity)
+	for _, entity := range dsEntities {
+		dsEntitiesMap[entity.Value] = entity
+	}
+
+	localEntriesMap := make(map[string]RepoAllowListEntity)
+	for _, item := range entries {
+		localEntriesMap[item.Value] = item
```
another-rex: We probably should say any duplicate Values (even with different type) is invalid, and prevent that from happening.

tools/repo-allowlist-sync/main.go
```diff
@@ -0,0 +1,245 @@
...
+			Type:                  item.Type,
+			Value:                 item.Value,
+			ConsiderAllBranches:   item.ConsiderAllBranches,
+			CherrypicksIntroduced: item.CherrypicksIntroduced,
+			CherrypicksFixed:      item.CherrypicksFixed,
+			CherrypicksLimit:      item.CherrypicksLimit,
+		}
+
+		if !exists {
+			key := datastore.IncompleteKey("RepoAllowList", nil)
```
another-rex: Can we use Value as the key (maybe have to base64 encode it, idk what value key values are), since that's the dedup key we use in map anyway.

tools/repo-allowlist-sync/main.go
```diff
@@ -0,0 +1,245 @@
...
+					return fmt.Errorf("failed deleting entity for %s: %w", val, err)
+				}
+			}
+		}
+	}
+
+	if dryRun {
+		log.Println("[DRY RUN] Sync completed successfully.")
+	} else {
+		log.Println("Sync completed successfully.")
```
another-rex: Can we log a bit more in the non verbose mode as well (e.g. summarize how many have entries added/deleted/updated)

Though I  think we should just always be running in verbose mode, since it shouldn't be updating that much.

tools/repo-allowlist-sync/main.go
```diff
@@ -0,0 +1,245 @@
...
+	Value                 string `yaml:"value"`
+	ConsiderAllBranches   bool   `yaml:"consider_all_branches"`
+	Cherrypicks           *bool  `yaml:"cherrypicks"`
+	CherrypicksIntroduced *bool  `yaml:"cherrypicks_introduced"`
+	CherrypicksFixed      *bool  `yaml:"cherrypicks_fixed"`
+	CherrypicksLimit      *bool  `yaml:"cherrypicks_limit"`
+}
+
+func (e RepoAllowListEntity) matches(other RepoAllowListEntity) bool {
+	return e.Type == other.Type &&
```
another-rex: missing Value here? I know you already use a dict to dedup, but feels important to check here incase it's called elsewhere.

tools/repo-allowlist-sync/main.go
```diff
@@ -0,0 +1,245 @@
...
+	normalized = strings.TrimRight(normalized, "/")
+	normalized = strings.TrimSuffix(normalized, ".git")
+
+	return normalized
+}
+
+// parseYAMLEntries parses and validates allowlist YAML content, expanding shorthand fields and normalizing values.
+func parseYAMLEntries(data []byte) ([]RepoAllowListEntity, error) {
+	var rawEntries []rawYAMLEntry
+	if err := yaml.Unmarshal(data, &rawEntries); err != nil {
```
another-rex: Good place to use strict yaml unmarshalling, i.e. error if we see unknown fields.

another-rex (commented):
Maybe also add a --validate flag, similar to how source_sync.py has for validating the yaml keys and regexes are correct

tools/repo-allowlist-sync/main.go
Ly-Joey: Added a TODO for now, will get it done after I merge #5797 today.

tools/repo-allowlist-sync/main.go
Ly-Joey: Added more stat reporting at the end, also made verbose flag default to true.

michaelkedar (approved):

PR #5815, opened 2026-08-07
fix(purl): preserve Packagist repository URLs

### LINKED ISSUE #5683: DRUPAL generated PURLs incorrect

kappa9999:
## Overview

Preserve non-default Packagist repositories when converting between OSV ecosystems and Package URLs.

Fixes #5683

## Details

The active Go PURL registry treated Packagist as a generic slash-separated ecosystem. That preserved the Composer namespace and package name, but discarded the repository suffix from ecosystems such as `Packagist:https://packages.drupal.org/8`. The reverse parser likewise ignored the `repository_url` qualifier.

This change:

- registers a dedicated Packagist generator/parser;
- writes the ecosystem suffix as the standard `repository_url` qualifier;
- restores that qualifier to the OSV ecosystem during PURL parsing;
- keeps unqualified `Packagist` behavior unchanged; and
- relies on `packageurl-go` for canonical qualifier escaping.

The production worker enrichment path has a focused Drupal regression case, and the PURL package has a generation/parse round-trip test.

## Testing

```console
$ cd go
$ go test ./purl ./internal/worker/pipeline/purl
ok  github.com/google/osv.dev/go/purl
ok  github.com/google/osv.dev/go/internal/worker/pipeline/purl
```

`gofmt -d` reports no differences for the changed Go files.

AI-assisted contribution: Codex was used for code and test drafting, repository inspection, diff review, and local validation.

### DIFF
--- a/go/internal/worker/pipeline/purl/purl_test.go
+++ b/go/internal/worker/pipeline/purl/purl_test.go
@@ -61,6 +61,13 @@ func TestEnricher_Enrich(t *testing.T) {
 			initialPurl:  "",
 			expectedPurl: "pkg:deb/debian/curl?arch=source&distro=bullseye",
 		},
+		{
+			name:         "Generate Packagist PURL with repository",
+			ecosystem:    "Packagist:https://packages.drupal.org/8",
+			pkgName:      "drupal/commerce_guest_registration",
+			initialPurl:  "",
+			expectedPurl: "pkg:composer/drupal/commerce_guest_registration?repository_url=https:%2F%2Fpackages.drupal.org%2F8",
+		},
 	}
 
 	for _, tc := range tests {
--- a/go/purl/ecosystem_packagist.go
+++ b/go/purl/ecosystem_packagist.go
@@ -1,0 +1,51 @@
+package purl
+
+import (
+	"strings"
+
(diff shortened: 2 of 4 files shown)


michaelkedar (approved):
Looks good, thanks

PR #5817, opened 2026-08-07
fix(docs): fix 404 on FAQ API links

ashmod:
The API links in the FAQ [404](https://google.github.io/osv.dev/faq/#:~:text=By%20querying%20OSV.dev%E2%80%99s%20API) at the moment. The path is relative, so it resolves against the FAQ's own URL instead of the site root.

This PR links it to `api/index.md` instead to be resolved into the correct permalink.


another-rex (approved):
Thanks!

PR #5837, opened 2026-08-17
perf(vanir): optimize signature job concurrency and node resource allocation

cuixq:
- Increase CPU (30 vCPUs) and memory (180Gi request / 220Gi limit) in CronJob to utilize the highend node
- Set default max_workers to 10 and batch_size to 100 for parallel batch processing
- Use isolated per-batch temporary directories for thread safety and cleanup


another-rex (commented):
Is this change necessary? I'm looking at the past week for both test and prod, and it's barely using any CPU or it's current memory. I thought we might want to decrease these values...

another-rex (approved):

PR #5838, opened 2026-08-17
refactor(go-website): connect vulnerability page to database

michaelkedar:
Populate the individual vulnerability pages from the GCS/datastore.

- Plumbed database Stores to server for reading vuln data from GCS.
- Re-implementent upstream/downstream computation in go.
- Since human links are using jinja template strings, I had to import a library to render these in go. I want to migrate these to go-native templates, but have to wait until the python website is no longer used to change the fields on the SourceRepository.

---


go/internal/website/vulnerability.go
```diff
@@ -24,62 +32,187 @@ func (s *Server) handleVulnerabilityDetails(w http.ResponseWriter, r *http.Reque
...
+			for _, c := range children {
+				candidateIDs[c] = struct{}{}
+			}
+		}
 	}
 
-	mockDownstream := ComputedHierarchy{}
+	var mu sync.Mutex
+	knownIDs := make(map[string]bool)
+	g, ctx := errgroup.WithContext(r.Context())
```
another-rex, line 145: Set a limit here, we have some very big alias/upstream/related groups

go/internal/website/vulnerability.go
```diff
@@ -1,15 +1,23 @@
...
-	"google.golang.org/protobuf/proto"
+	"github.com/google/osv.dev/go/internal/models"
+	"github.com/google/osv.dev/go/logger"
+	"golang.org/x/sync/errgroup"
 )
 
+var vulnIDRegex = regexp.MustCompile(`^[a-zA-Z0-9:_.-]+$`)
+
 // handleVulnerabilityDetails handles rendering the vulnerability details page or raw JSON for /vulnerability/{vuln_id}.
 func (s *Server) handleVulnerabilityDetails(w http.ResponseWriter, r *http.Request) {
```
another-rex, line 22: No vulnIDRegex check here?

go/internal/database/datastore/relations.go
```diff
@@ -67,14 +68,24 @@ func (s *RelationsStore) GetRelated(ctx context.Context, id string) (*models.Get
...
+	return &groups[0], nil
+}
+
+func (s *RelationsStore) GetUpstream(ctx context.Context, id string) (*models.GetUpstreamResult, error) {
+	upstreamGroup, err := s.getUpstreamGroup(ctx, id)
 	if err != nil {
-		return nil, fmt.Errorf("failed to get upstream group: %w", err)
+		return nil, err
 	}
 	upstream := make([]string, len(upstreamGroup.UpstreamIDs))
```
another-rex: nit: I think we can use slices.Clone here. I think this pattern is also used elsewhere so can replace those too.

another-rex (commented):
LGTM with some nits.

go/internal/website/vulnerability.go
michaelkedar: it's probably not super necessary to have it here - this specifically for the `osv.dev/vulnerability/[ID]` page, and it might be hypothetically possible we end up with an ID that doesn't match the regex in the database (maybe some old withdrawn vuln or something) and we probably don't want it to be impossible to navigate to.

go/internal/database/datastore/relations.go
michaelkedar: replaced most instances of that pattern (even if it's unrelated to this PR 🤷)

go/internal/website/vulnerability.go
michaelkedar: arbitrarily chose a limit of 20 here

another-rex (approved):

PR #5839, opened 2026-08-17
refactor(go-website): connect list page to database

michaelkedar:
Connect the list/search page to Datastore.
- Made a new VulnerabilitySearchStore to provide needed methods for ecosystem counts, search results, and autocomplete.
  - have not connected Redis to the ecosystem counts, so these take a long time to populate (but they are locally cached). Is done in follow-up PR (next in stack)
- Changed how the next-page logic works to use a `after=timestamp_ID` instead of a page number - page numbers require re-iterating through all previous results in the database to arrive at the next page, which becomes very inefficient later pages. Swapping to a timestamp/id lets us use the database more correctly.
  - Maintained `page=N` functionality (with a warning log) so we can see if people have been manually scraping our website

---

### DIFF
--- a/gcp/website/frontend3/src/go/templates/list.html
+++ b/gcp/website/frontend3/src/go/templates/list.html
@@ -73,7 +73,7 @@
         </div>
       </div>
       <div role="rowgroup" class="vuln-table-rows mdc-data-table__content">
-        <turbo-frame id="vulnerability-table-page{{ .Page }}" data-turbo-action="advance" target="_top">
+        <turbo-frame id="vulnerability-table-page{{ if .CurrentAfter }}-{{ .CurrentAfter }}{{ end }}" data-turbo-action="advance" target="_top">
           {{ range .Vulnerabilities }}
           <div role="row" class="vuln-table-row mdc-data-table__row">
             <span role="cell" class="vuln-table-cell mdc-data-table__cell">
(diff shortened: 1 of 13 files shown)


gcp/website/frontend3/src/go/templates/list.html
```diff
@@ -122,15 +122,11 @@ <h1 class="title">Vulnerabilities</h1>
           {{ if eq (len .Vulnerabilities) 0 }}
           <span class="no-results">No results (check our <a href="https://google.github.io/osv.dev/faq/">FAQ</a> if this is unexpected)</span>
           {{ end }}
-          {{ if lt .Page .TotalPages }}
-          <turbo-frame id="vulnerability-table-page{{ add .Page 1 }}" data-turbo-action="advance" target="_top" class="next-page-frame">
+          {{ if .NextAfter }}
+          <turbo-frame id="vulnerability-table-page-{{ .NextAfter }}" data-turbo-action="advance" target="_top" class="next-page-frame">
             <div class="next-page-container">
-              <a class="next-page-button link-button" data-turbo-frame="_self" href="/list?page={{ add .Page 1 }}{{ if $.Query }}&q={{ $.EncodedQuery }}{{ end }}{{ if $.SelectedEcosystem }}&ecosystem={{ $.EncodedEcosystem }}{{ end }}">
+              <a class="next-page-button link-button" data-turbo-frame="_self" href="/list?after={{ .NextAfter }}{{ if $.Query }}&q={{ $.Query }}{{ end }}{{ if $.SelectedEcosystem }}&ecosystem={{ $.SelectedEcosystem }}{{ end }}">
```
another-rex, line 128: Is it fine to not encode the query anymore? Why was this changed?

another-rex: Please add a comment if this is intentional.

go/internal/database/datastore/vulnerability_search.go
```diff
@@ -0,0 +1,352 @@
...
+				}
+			}()
+		}
+		s.mu.Unlock()
+
+		return cached, nil
+	}
+
+	// 3. Cold start fallback: fetch synchronously
+	return s.refreshEcosystemCounts(ctx)
```
another-rex, line 287: This feels like a good place for singleflight, so we don't have a bunch of refreshes all running at the start of the server.

another-rex (commented):
Some initial comments

gcp/website/frontend3/src/go/templates/list.html
michaelkedar: `html/template` actually [automatically encodes](https://pkg.go.dev/html/template#hdr-Contexts) these, and we don't want to double-encode them.

go/internal/database/datastore/vulnerability_search.go
michaelkedar: Made the singleflight change in #5840 (to avoid merge conflicts)

another-rex (approved):

PR #5840, opened 2026-08-17
refactor(go-website): implement redis caching of ecosystem counts

michaelkedar:
Use the website redis to cache the ecosystem counts across Cloud Run instances.
- Store the ecosystem counts (and the time last fetched) in a new redis key that doesn't collide with the old Python ones.
- Added another key to track when the counts are being recomputed by any of the instances, to avoid every instance attempting to recompute the counts when it gets stale.
- Did NOT re-implement the rate-limiting through Redis. We're running on GCLB now, we should configure it at that level (if we even want it at all)

---


another-rex (dismissed):
LGTM! No rate limits! <sub>just in gclb</sub>

another-rex (approved):

PR #5842, opened 2026-08-17
refactor(go-website): Implement auth & triage page

michaelkedar:
Implemented the google login page & cookies & connections needed for the triage page.

Auth stuff implements AES-256-GCM (instead of Flask's unencrypted signed cookies) to store the auth session client-side. I was thinking of using [scs](https://pkg.go.dev/github.com/alexedwards/scs/v2), but that is server side (would need to connect to redis), and I'd rather keep it similar to Flask's client-side stuff.

---

### DIFF
--- a/go/cmd/website/main.go
+++ b/go/cmd/website/main.go
@@ -11,6 +11,7 @@ import (
 	"os"
 	"os/signal"
 	"strconv"
+	"strings"
 	"syscall"
 	"time"
 
@@ -117,6 +118,10 @@ func run() error {
 		redisClient = rdb
 	}
 
+	bypassOAuth := strings.EqualFold(os.Getenv("BYPASS_OAUTH_FOR_LOCAL_DEV"), "true") ||
+		os.Getenv("BYPASS_OAUTH_FOR_LOCAL_DEV") == "1" ||
+		strings.EqualFold(os.Getenv("BYPASS_OAUTH_FOR_LOCAL_DEV"), "t")
+
 	stores := website.Stores{
 		Vuln: db.NewVulnerabilityStore(db.VulnStoreConfig{
 			Client: dbClient,
@@ -126,6 +131,7 @@ func run() error {
 		SourceRepo: db.NewSourceRepositoryStore(dbClient),
 		VulnSearch: db.NewVulnerabilitySearchStore(dbClient, redisClient),
 		Linter:     gcs.NewLinterStore(gcsClient.Bucket(linterBucket), "linter-result/"),
+		Triage:     gcs.NewTriageStore(gcsClient),
 	}
 
 	apiURL := os.Getenv("OSV_API_URL")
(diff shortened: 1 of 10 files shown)


michaelkedar:
@jess-lowe can you make sure the Go triage tool still does what you need it to do

jess-lowe (commented):
still does the thing 👍🏻

go/cmd/website/main.go
```diff
@@ -136,11 +142,25 @@ func run() error {
...
 	}
 
+	secretKey := os.Getenv("SESSION_SECRET_KEY")
+	if secretKey == "" {
+		secretKey = os.Getenv("FLASK_SECRET_KEY")
+	}
+	if secretKey == "" {
+		secretKey = os.Getenv("SECRET_KEY")
+	}
+
```
another-rex, line 155: Can we log a warning if secretkey is not set? Also,  what situation would we not have a secret key? Just local dev?

go/internal/gcs/triage.go
```diff
@@ -0,0 +1,117 @@
...
+		seq := matches[2]
+		seqPrefix := "0"
+		if len(seq) > 3 {
+			seqPrefix = seq[:len(seq)-3]
+		}
+		url = fmt.Sprintf("https://raw.githubusercontent.com/CVEProject/cvelistV5/refs/heads/main/cves/%s/%sxxx/%s.json", year, seqPrefix, cveID)
+	case "nvd":
+		url = "https://services.nvd.nist.gov/rest/json/cves/2.0?cveId=" + cveID
+	default:
+		return nil, fmt.Errorf("invalid source: %s", source)
```
another-rex: If source is invalid, it should be models.ErrNotFound right?

go/internal/website/auth.go
```diff
@@ -1,34 +1,332 @@
...
+	if err != nil || statePayload.State == "" {
+		s.renderJSON(w, r, http.StatusBadRequest, map[string]string{
+			"error": "Invalid state parameter",
+		})
+
+		return
+	}
+
+	queryState := r.URL.Query().Get("state")
+	if queryState == "" || queryState != statePayload.State {
```
another-rex: Technically state is a secret, so use constant time comparison to compare them.

go/internal/gcs/triage.go
```diff
@@ -0,0 +1,117 @@
...
+	defer resp.Body.Close()
+
+	if resp.StatusCode == http.StatusNotFound {
+		return nil, models.ErrNotFound
+	}
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("external API returned status %d", resp.StatusCode)
+	}
+
+	return io.ReadAll(resp.Body)
```
another-rex: Maybe add a limit reader here incase nvd responds with an dos attack.

another-rex (commented):
Just some nits

go/cmd/website/main.go
michaelkedar: Made a warning. I'm pretty sure prod doesn't set this either.

go/internal/gcs/triage.go
michaelkedar: Made a `ErrInvalidArgument` for this, which feels more correct for an invalid source.

go/internal/website/auth.go
michaelkedar: changed to constant time comparison

go/internal/gcs/triage.go
michaelkedar: Added a 10MB limit in case the Nefarious Vulnerability Database decides to attack us

go/cmd/website/main.go
another-rex: wait... if prod doesn't set this, how is auth working at all? Every time a request gets routed to a different instance they'll have a different secret.

another-rex (approved):

PR #5843, opened 2026-08-17
refactor(go-website): implement non-datastore dev server for testing

michaelkedar:
This serves the same function as `run-website-emulator`, but doing it with implementations of the `XyzStore` interfaces to read and serve vulnerability JSON files from disk.

This way, we can test the website without needed to spin up the datastore (and gcs) emulator, and without having to populate every datastore tables.

The vuln json files (and metadata yaml files) should be able to be hot-loaded, which should make testing easier.

The code assumes that aliases, upstream, related are all fully computed in the vulnerability JSON - it does not the the transitive alias/upstream calculations.

---


go/cmd/website-devserver/store.go
```diff
@@ -0,0 +1,635 @@
...
+		return nil
+	}
+
+	var records []*osvschema.Vulnerability
+	for _, entry := range entries {
+		if entry.IsDir() || filepath.Ext(entry.Name()) != ".json" {
+			continue
+		}
+
+		vuln, _, err := readRecordFromFile(filepath.Join(ds.dataDir, entry.Name()))
```
another-rex, line 181: If this errors, can we log an error?

go/cmd/website-devserver/store.go
```diff
@@ -0,0 +1,635 @@
...
+
+		return strings.Compare(b.ID, a.ID)
+	})
+
+	pageSize := query.PageSize
+	if pageSize <= 0 {
+		pageSize = 16
+	}
+
+	// Cursor and legacy pagination
```
another-rex: Hm? why do we need legacy pagination here. Is this just the best way to paginate locally?

another-rex: If so please update the comment explaining why.

go/cmd/website-devserver/store.go
```diff
@@ -0,0 +1,635 @@
...
+	}
+
+	return repo, nil
+}
+
+// =========================================================================
+// models.VulnerabilitySearchStore Implementation
+// =========================================================================
+
+func (ds *DevStore) Search(_ context.Context, query models.VulnerabilitySearchQuery) (*models.VulnerabilitySearchResult, error) {
```
another-rex, line 362: as far as I can  tell this doesn't search for repos names (this is something datastore indexes)

go/cmd/website-devserver/store.go
michaelkedar: We do the new pagination first, the legacy pagination is just 1 line of code which we might as well add for feature parity.
Added a small comment.

go/cmd/website-devserver/store.go
michaelkedar: Added simple matching to repo names.
Really, this search matching is a database implementation detail, and will probably be done differently again with postgres, so I'm not too concerned on 1-to-1 matching the behaviour.

another-rex (approved):

PR #5869, opened 2026-08-19
chore: remove Python API code

michaelkedar:
Getting rid of all the Python API implementation, since our shiny new Go implementation works.
- Removed the `gcp/api/` directory
  - Moved related Protos into `proto/` directory
    - Removed the `googleapis` submodule and vendored the two required protos directly
  - Moved ESPv2 Dockerfile to `docker` directory
  - Re-linked terraform symlinks to point to new locations of files
- Replaced API tester python script with bash script
- Replaced `build_swagger.py` script with bash script


docs/osv_service_v1.swagger.json
```diff
@@ -329,30 +440,35 @@
...
     },
     "osvSeverityType": {
       "type": "string",
       "enum": [
         "UNSPECIFIED",
-        "CVSS_V4",
         "CVSS_V3",
-        "CVSS_V2"
+        "CVSS_V2",
+        "CVSS_V4",
```
jess-lowe, line 456: this ordering annoys me but I'm assuming this is generated?

proto/google/api/http.proto
```diff
@@ -0,0 +1,370 @@
...
+    string put = 3;
+
+    // Maps to HTTP POST. Used for creating a resource or performing an action.
+    string post = 4;
+
+    // Maps to HTTP DELETE. Used for deleting a resource.
+    string delete = 5;
+
+    // Maps to HTTP PATCH. Used for updating a resource.
+    string patch = 6;
```
jess-lowe, line 332: at what point would we need to support the QUERY method too?

docs/osv_service_v1.swagger.json
```diff
@@ -309,15 +415,20 @@
...
         "NONE",
-        "WEB",
         "ADVISORY",
-        "REPORT",
+        "ARTICLE",
+        "DETECTION",
+        "DISCUSSION",
+        "EVIDENCE",
         "FIX",
+        "GIT",
```
jess-lowe, line 424: are we keeping 'git' around even though it was a mistake from like 5 years ago?

jess-lowe (dismissed):
bye bye python

docs/osv_service_v1.swagger.json
michaelkedar: Yeah this file is generated.
This is the order of the enum in the osv-schema proto 🙃

docs/osv_service_v1.swagger.json
michaelkedar: It's technically in the schema (and hypothetically can be returned by a record), so I guess it stays 🤷

proto/google/api/http.proto
michaelkedar: if ESPv2 decides to support it

jess-lowe (approved):

PR #5872, opened 2026-08-19
feat: OSS FUZZ Syncer

another-rex:
Added a cronjob that'll run the syncer script  that will check the last 14 days (default).


osv/osv-schema
michaelkedar, line 1: 🤔

another-rex:
Permissions might be too thorny, closing for now.

PR #5878, opened 2026-08-20
refactor(terraform): migrate worker gke to `osv_pipeline` module

michaelkedar:
- Removed `worker-pool` / `worker-pool-temp` node pools and tolerations. Workers now run on `default-pool`.
- Extracted Datastore, GCS export/backup buckets, GKE worker cluster, networkin, and Pub/Sub task queues from `modules/osv` into `modules/osv_pipeline` for `oss-vdb` and `oss-vdb-test`.
  - Added `moved` blocks in both environments to migrate without having to recreate resources.
- Node pools in `osv_pipeline` now run with dedicated `osv-worker` service account, which should have only the permissions it actually needs (instead of, like, editor).
  - Changing service accounts requires recreating node pools 😞

Might make a new service account for the oss-fuzz workers, but I need to look into whether they also need more permissions or not.


michaelkedar:
The [terraform plan check](https://github.com/google/osv.dev/pull/5878/checks?check_run_id=96342883240) is worth reviewing, but it looks fine to me

deployment/terraform/environments/oss-vdb-test/moved.tf
another-rex, line 1: Interesting, I'm guessing this can be removed after everything's moved?

another-rex (approved):
We probably want to compartmentalize more than just 1 "worker" service account in the future. E.g. vulnfeeds might want a different account than the core pipeline.

Can leave to later though.

deployment/terraform/environments/oss-vdb-test/moved.tf
michaelkedar: It can, but I don't think there's much harm leaving it in

Issue #5692, opened 2026-07-21
Flaky `build-importer` step in Cloud Build


cuixq:
The Go `build-importer` step in `deployment/build-and-stage.yaml` occasionally fails with the following error:

```
ERROR: failed to solve: rpc error: code = Unknown desc = failed to solve with frontend dockerfile.v0: failed to solve with frontend gateway.v0: frontend grpc server closed unexpectedly
```

This might be a memory issue, but bumping the memory seems not fix it.

Also the Cloud Build docker image is a bit old, but not sure if we are able to upgrade it.

PR #5882, opened 2026-08-21
build: bump Cloud Build machineType to E2_HIGHMEM_32

cuixq:
Another try for https://github.com/google/osv.dev/issues/5692

Cloud Build steps building Go images (such as `build-importer`) have been intermittently failing with the following error:
```text
ERROR: failed to solve: rpc error: code = Unknown desc = failed to solve with frontend dockerfile.v0: failed to solve with frontend gateway.v0: frontend grpc server closed unexpectedly
```

The current machine type `E2_HIGHCPU_32` provides 32 vCPUs but only 32 GB of RAM (1 GB per core). During concurrent Go builds and module downloads, sudden memory spikes may trigger Linux OOM kills on ephemeral Docker BuildKit helper containers (`gateway.v0`), causing the gRPC connection to abruptly terminate.

Bump Cloud Build `machineType` in `deployment/build-and-stage.yaml` from `E2_HIGHCPU_32` (32 GB RAM) to `E2_HIGHMEM_32` (256 GB RAM).


michaelkedar:
Looking at the [Cloud Build pricing page](https://cloud.google.com/build/pricing), it doesn't seem like `E2_HIGHMEM_32` is an available option without setting up a private pool

PR #5885, opened 2026-08-21
refactor(vulnfeeds): migrate debian-copyright-mirror to Go

jess-lowe:
Migrates the `debian-copyright-mirror` service from Python to Go and extracts reusable networking, archive, and decompression utilities for other `vulnfeeds` tools.

### Key Changes
- **Go Migration**: Replaced `debian-copyright-mirror.py` and `debian-copyright-mirror.sh` with a native Go service (`cmd/mirrors/debian-copyright-mirror/main.go`).
- **Streaming Pipeline**: Streams `filelist.yaml.xz` over HTTP through an `xz -dc` pipe directly into a custom streaming line parser, extracting ~44k package paths without disk writes.
- **Worker Pool**: Added a concurrent worker pool with HTTP/2 keep-alive connection pooling, per-request timeouts, retry logic, and atomic failure rate tracking (`-max-failure-rate`).
- **Native Archiving & GCS Upload**: Built-in POSIX `.tar` packaging and direct streaming to Google Cloud Storage (`cloud.google.com/go/storage`).
- **Shared Utilities**:
  - `utility/decompress.go`: Generic XZ streaming decompression helpers.
  - `utility/download.go`: Resilient atomic file downloader with exponential backoff.
  - `utility/archive.go`: Tar archive creation and extraction with traversal/bomb protections.
  - `gcs-tools/gcs.go`: Added `ParseGCSPath` URI helper.
- **Dockerfile & Deployment**: Switched to a multi-stage Go build on a minimal `alpine:3.21` image with the Go binary as direct `ENTRYPOINT`, removing the need for `google-cloud-cli` and wrapper shell scripts.

## Testing Done
- `go test ./...` in `vulnfeeds` (all tests passing).
- `golangci-lint run` (0 issues).
- Verified full download and local `.tar` generation with live Debian changelog metadata.


another-rex (commented):
This is a lot of unnecessary code imho. We don't need to reimplement the whole archiving stack in go. Just shell out to tar to extract the files like it did in the original script.

 This should be about 200 lines of go code, not 1500

vulnfeeds/cmd/mirrors/debian-copyright-mirror/Dockerfile
```diff
@@ -12,11 +12,22 @@
...
-COPY ./debian-copyright-mirror.py /
+COPY ./go.mod /src/go.mod
+COPY ./go.sum /src/go.sum
+RUN go mod download
 
-ENTRYPOINT ["/debian-copyright-mirror.sh"]
+COPY ./ /src/
+RUN CGO_ENABLED=0 go build -o debian-copyright-mirror ./cmd/mirrors/debian-copyright-mirror
+
+FROM alpine:3.21
```
michaelkedar:
```suggestion
FROM alpine:3.21@sha256:ce64758a109eb420d874a118f87920e625e12d3634e03b4a5573fd9f6e5d3507
```

PR #5892, opened 2026-08-24
fix: better error logging in gitter

### LINKED ISSUE #5805: Reduce Gitter Logging Noise
1) Errors are logged at a request level so when a singleflight group fails for whatever reason, all requests in that group will log the same error and cause an explosion.
Ideally we just log one error.

2) Commit hash related errors (not found, invalid length, etc) should all be warnings

3) Funky Repo URLssss

4) Large repo patch ID timeout

Ly-Joey:
Previously, gitter errors are logged at request level so when a singleflight group fails for whatever reason, all requests in that group will log the same error and make the logs unnecessarily noisy. 
This PR moves the logging inside the singleflight groups' Do() so one group only emit one set of logs.
 
Partially fixes: #5805

---

### DIFF
--- a/go/cmd/gitter/errors.go
+++ b/go/cmd/gitter/errors.go
@@ -184,7 +184,7 @@ func errorToHTTPStatusCode(err error) int {
 	case isRateLimitError(err):
 		return http.StatusTooManyRequests // 429
 	case isRemoteHostError(err):
-		return http.StatusFailedDependency // 424
+		return http.StatusBadGateway // 502
 	default:
 		return http.StatusInternalServerError // 500
 	}
--- a/go/cmd/gitter/errors.go
+++ b/go/cmd/gitter/errors.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"net/http"
 	"regexp"
 	"strconv"
 	"strings"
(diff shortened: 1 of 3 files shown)


go/cmd/gitter/errors.go
```diff
@@ -149,3 +150,23 @@ func isFileNotFoundError(err error) bool {
...
+	}
+
+	switch {
+	case isNotFoundError(err):
+		return http.StatusNotFound // 404
+	case isAuthError(err) || isForbiddenError(err):
+		return http.StatusForbidden // 403
+	case isRateLimitError(err):
+		return http.StatusTooManyRequests // 429
+	case isRemoteHostError(err):
```
cuixq, line 186: How do you feel about using `502 Bad Gateway` (`http.StatusBadGateway`) instead of `424 Failed Dependency` here?
- [`424`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/424) is a 4xx code (client error), which indicates that the caller made an invalid request or provided invalid state, but `isRemoteHostError` seems focus more on the upstream server errors, where the caller's request was valid.
- `gitter` acts as an intermediary/proxy to upstream remote git hosts, [`502 Bad Gateway`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/502) feels like the more suitable status code for upstream server failures.

Ly-Joey: Good point! Updated :D

cuixq (dismissed):

cuixq (approved):

PR #5919, opened 2026-08-26
perf: reduce exporter memory footprint with disk staging and streaming generation

another-rex:
Rather than keeping everything in memory, then writing it out at the end in order. Save the records into a temporary location on disk in JSON, then io.copy them into a zip in order. Streaming the output to GCS.


go/cmd/exporter/downloader.go
```diff
@@ -38,9 +42,70 @@ func downloader(ctx context.Context, client clients.CloudStorage, inCh <-chan st
...
+		ecosystems := make(map[string]struct{})
+		for _, aff := range vuln.GetAffected() {
+			eco := aff.GetPackage().GetEcosystem()
+			eco, _, _ = strings.Cut(eco, ":")
+			if eco != "" {
+				ecosystems[eco] = struct{}{}
+			}
+			for _, ref := range aff.GetRanges() {
+				if ref.GetType() == osvschema.Range_GIT {
+					ecosystems["GIT"] = struct{}{}
```
michaelkedar, line 82: break?

go/cmd/exporter/worker.go
```diff
@@ -30,75 +31,61 @@ const (
...
 
+// vulnMeta holds the ID and modified time for a vulnerability.
+type vulnMeta struct {
+	id       string
+	modified time.Time
+}
+
+// csvEntry holds the unix timestamp in nanoseconds and the relative entry path.
+type csvEntry struct {
+	modified int64
```
michaelkedar: changing this to a time.Time (or the RFC string) will probably make logic around this neater

go/cmd/exporter/writer.go
```diff
@@ -29,37 +31,114 @@ func writer(ctx context.Context, cancel context.CancelFunc, inCh <-chan writeMsg
...
+
+			break
+		}
+	}
+}
+
+// writeFromFile handles writing or uploading a file by streaming from a local file path.
+// It removes the source file when finished.
+func writeFromFile(ctx context.Context, client clients.CloudStorage, path, filePath, mimeType string) error {
+	defer os.Remove(filePath)
```
michaelkedar: Deleting the file in this function seems unnecessarily foot-gun-ny.
Can we just have the calling function call `os.Remove`

michaelkedar (commented):
Seems okay.
I wonder how much slower this will be with the disk reading/writing?
Hopefully it doesn't push the exporter above 15 minutes...

michaelkedar (approved):

PR #5933, opened 2026-08-28
fix(website): display severity source

### LINKED ISSUE #5922: Investigate whether severity.source is supported on the frontend
Since merging the 1.9.0 release including the added severity[].source field, I'm not sure whether our osv.dev frontend supports the new field or not. 

We should investigate this, and fix it if it is not supported yet.

It should be as simple as adding a severity + source on one of the test files for the local emulator, and running it with `make run-website-emulator`. If this is picked up after the python > go transition, it'll be `make run-website-devserver`

ZedingZhang:
## Overview

Display the optional severity source on vulnerability detail pages so users can distinguish NVD, CNA, and self-attributed assessments.

Fixes #5922

## Details

- Preserve severity source when building the Go website display model.
- Render the source in both the Go and legacy Jinja vulnerability templates when present.
- Add an emulator fixture containing a CVSS severity attributed to NVD.
- Add regression coverage for source propagation while leaving source-less records unchanged.

## Testing

- go test ./internal/website
- go vet ./internal/website
- pnpm --dir gcp/website/frontend3 exec webpack --config webpack.dev.js --mode development
- Validated the updated fixture against osv/osv-schema/validation/schema.json.
- Parsed the Jinja template with Jinja2.

This contribution was created with AI assistance (OpenAI Codex); the changes and test results were reviewed before submission.


jess-lowe:
As the Go website stuff hasn't been merged yet, I'll wait to review this until after that has happened (supposedly Wednesday)

jess-lowe (commented):
This works but the testcase may need to be updated in `go/cmd/website-devserver/testdata` or it's not really visible :) - and also this should probably be rebase/merged with master

jess-lowe (approved):

PR #5939, opened 2026-08-31
feat: Optimize linting by checking changed files

another-rex:
Trying to do quick fixes really make you find out where the pain points in the CI.

Turned the lint and format script into a python script (with no dependencies). We now tune a few settings to make it run a lot faster.

1. Make use of the parallel processing flags for python lints
3. Only run it on files that changed (In CI it's against the master branch).


deployment/terraform/modules/osv_pipeline/gke.tf
michaelkedar, line 88: I don't think this is related 🤔

tools/lint_and_format.py
```diff
@@ -0,0 +1,543 @@
...
+        if mod:
+          affected_mods.add(mod)
+
+    # Dependency propagation for Go
+    if "bindings/go" in affected_mods and "go" in all_go_mods:
+      affected_mods.add("go")
+    if "go" in affected_mods:
+      for dep in ("go/cmd/tools/reimport-tui", "tools/repo-allowlist-sync"):
+        if dep in all_go_mods:
+          affected_mods.add(dep)
```
michaelkedar: This hard-coded propagation is annoying to maintain, and I don't really understand its purpose - is it saying that if a dependency changes, then also lint the thing depending on it?

I don't really see why that would be necessary?

tools/lint_and_format.py
```diff
@@ -0,0 +1,595 @@
...
+  - Go (golangci-lint, gofmt):
+      * Tracked: '*.go', 'go.mod', 'go.sum',
+        '*.golangci.yaml', '*.golangci.yml'.
+      * Module resolution: Walks up parent directories to find the nearest
+        'go.mod' (excluding 'docs/').
+      * Inter-module dependencies:
+          - If 'bindings/go' changes -> both 'bindings/go' and 'go' are linted.
+          - If 'go' changes -> 'go', 'go/cmd/tools/reimport-tui', and
+            'tools/repo-allowlist-sync' are linted.
+      * Skipped (0.0s) when no in-scope Go modules changed.
```
michaelkedar, line 52: *why?*

michaelkedar (commented):
seems marginally over-engineered.

<sub>(also increasing the amount of Python in the repo 😔)</sub>

michaelkedar:
Also, generally you're using double quotes for strings, when the python style guide says single quotes should be used

deployment/terraform/modules/osv_pipeline/gke.tf
another-rex: Huh... shouldn't this not be a problem since it's already merged in? Maybe merging master into this might fix this.

tools/lint_and_format.py
another-rex: Yeah idk, removed now.

another-rex:
Simplified it a bit, now just finds all the targets (e.g. go.mod files), figures out the diffs and which belongs to which target, and then run the linter against the appropriate target.

michaelkedar (approved):

PR #5941, opened 2026-09-01
feat: set up a rough grafana monitoring dashboard

jess-lowe:
**Metrics in code:** Added Prometheus /metrics endpoints to worker (port 9090), gitter (port 8888), and vulnfeeds converters to track task speed, failures, disk usage, and CVE conversion success.
**Grafana on GKE:** Added a lightweight Grafana deployment to GKE (monitoring.osv.dev) protected by Google login (IAP) with read-only access to Prometheus and Cloud logs.

**4 Core Dashboards:**
**Executive SLAs:** API latency/uptime and total database records.
**Pipeline Ops:** Pub/Sub queue backlogs, worker processing speed, and errors.
**Data Quality:** Linter issues and Datastore vs. GCS sync drift.
**CVE Conversions:** Success and failure rates for upstream CVE conversion.
**Alerts:** Added Prometheus alert rules for high queue backlogs, task failure spikes, and data drift.

will all of these work immediately? probably not


deployment/clouddeploy/gke-workers/base/monitoring/deployment.yaml
```diff
@@ -0,0 +1,88 @@
...
+      app.kubernetes.io/name: grafana
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: grafana
+    spec:
+      serviceAccountName: grafana-sa
+      containers:
+      - name: grafana
+        image: grafana/grafana:10.4.0
```
michaelkedar, line 20: can we pin this image (and other images) to a specific sha

deployment/clouddeploy/gke-workers/base/monitoring/deployment.yaml
```diff
@@ -0,0 +1,88 @@
...
+          mountPath: /etc/grafana/provisioning/dashboards
+          readOnly: true
+        - name: dashboards-storage
+          mountPath: /etc/grafana/dashboards
+          readOnly: true
+      - name: gmp-frontend-proxy
+        image: gke.gcr.io/prometheus-engine/frontend:v0.19.0-gke.2
+        args:
+        - "--web.listen-address=0.0.0.0:9090"
+        - "--project-id=oss-vdb"
```
michaelkedar, line 65: You'll need to make an override in `environments` to set this project-id to the relevant project

deployment/clouddeploy/gke-workers/base/monitoring/managed_cert.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: networking.gke.io/v1
+kind: ManagedCertificate
+metadata:
+  name: grafana-managed-cert
+spec:
+  domains:
+    - monitoring.osv.dev
```
michaelkedar, line 7: do we need to do something with DNS for this?
also, override this for test instance with `monitoring.test.osv.dev`?

deployment/clouddeploy/gke-workers/base/monitoring/service_account.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v1
+kind: ServiceAccount
```
michaelkedar, line 2: can this be done in terraform, or is this something else more specific?

go/internal/worker/subscriber.go
```diff
@@ -83,15 +85,27 @@ func (s *Subscriber) handleMessage(ctx context.Context, m *pubsub.Message) {
...
+		metrics.WorkerTasksProcessed.WithLabelValues(task.Type.String(), "success", eco, task.SourceID).Inc()
 		logTaskLatency(taskCtx, task)
 		m.Ack()
 	}
 }
 
+func getTaskEcosystem(task Task) string {
+	if task.Vuln != nil && len(task.Vuln.GetAffected()) > 0 && task.Vuln.GetAffected()[0].GetPackage() != nil {
+		return task.Vuln.GetAffected()[0].GetPackage().GetEcosystem()
+	}
```
michaelkedar, line 104: Do we want some logic for "GIT" here, or is it fine to just have them as UNKNOWN?

deployment/clouddeploy/gke-workers/base/core/pod_monitoring.yaml
michaelkedar, line 1: IMO, this file should also be in `monitoring/`

deployment/clouddeploy/gke-workers/base/monitoring/rules.yaml
another-rex, line 1: Alerts should be a separate PR.

deployment/clouddeploy/gke-workers/base/monitoring/ingress.yaml
another-rex, line 1: Please name the files with its service name.

deployment/clouddeploy/gke-workers/base/monitoring/deployment.yaml
another-rex, line 1: Should grafana be running on GKE?

go/cmd/gitter/git.go
```diff
@@ -66,7 +67,14 @@ func runCmd(ctx context.Context, dir string, env []string, name string, args ...
 
 // cloneRepo clones a git repository into repoPath.
 func cloneRepo(ctx context.Context, repoURL string, repoPath string) error {
-	return runCmd(ctx, "", []string{"GIT_TERMINAL_PROMPT=0"}, "git", "clone", "--", repoURL, repoPath)
+	start := time.Now()
+	err := runCmd(ctx, "", []string{"GIT_TERMINAL_PROMPT=0"}, "git", "clone", "--", repoURL, repoPath)
+	metrics.GitterOperationDuration.WithLabelValues("clone").Observe(time.Since(start).Seconds())
```
michaelkedar, line 72: Do we want to always note this time, even if it fails?

If so, does using a defer statement here (and with other gitter stuff) make it neater?
```go
start := time.Now()
defer func() {
  metrics.GitterOperationDuration.WithLabelValues("clone").Observe(time.Since(start).Seconds())
}()
// rest of function
```

(you could probably even also capture the error that should be returned and also do the `clone_error` stuff in the deferred function)

Honestly, since this logic is used in a few places, it might be nice to have it as a helper function

deployment/clouddeploy/gke-workers/base/monitoring/deployment.yaml
```diff
@@ -0,0 +1,88 @@
...
+          value: "true"
+        - name: GF_AUTH_PROXY_HEADER_NAME
+          value: "X-Goog-Authenticated-User-Email"
+        - name: GF_AUTH_PROXY_HEADER_PROPERTY
+          value: "email"
+        - name: GF_AUTH_PROXY_AUTO_SIGN_UP
+          value: "true"
+        - name: GF_INSTALL_PLUGINS
+          value: "googlecloud-monitoring-datasource,googlecloud-logging-datasource,yesoreyeram-infinity-datasource"
+        - name: GF_PLUGINS_ALLOW_LOADING_UNSIGNED_PLUGINS
```
another-rex, line 40: Why is this unsigned... Would be good to avoid this if possible.

another-rex (commented):
This PR is doing a lot, can we try to get monitoring for 1 metric (Maybe just 1 number even) first (E.g. seconds since last vulnerability published), then in follow up PRs add the additional pieces.

michaelkedar (commented):
I only have a tenuous understanding of what's going on here.

It might be safer to deploy this only to the test instance (putting all the yaml files into `environments/oss-vdb-test` temporarily), just so this doesn't block the next release if something's not quite right.

Also, there seems to be some setup that needs/needed to be done on GCP that's not quite captured here (e.g. the `grafana-iap-secret`) is that documented somewhere?

jess-lowe:
made a simpler version

PR #5606, opened 2026-07-08
chore(sources): update chainguard test source to v3 feed URL

tdunlap607:
Point the Chainguard entry in `source_test.yaml` at the new Chainguard OSV v3 feed ([advisories.cgr.dev/chainguard/v3/osv/all.json](https://advisories.cgr.dev/chainguard/v3/osv/all.json)) for testing migration on [test.osv.dev](https://test.osv.dev/).

Chainguard v3 feed docs -> https://github.com/chainguard-dev/vulnerability-scanner-support/blob/main/docs/osv_v3_feed.md

PR #5944, opened 2026-09-01
chore(sources): update chainguard source to v3 feed URL

xnox:
Point the Chainguard entry in `source.yaml` at the new Chainguard OSV
v3 feed, as was previously already done in source_test.yaml. See:
- https://github.com/google/osv.dev/pull/5606

Chainguard v3 feed docs are in:
- https://github.com/chainguard-dev/vulnerability-scanner-support/blob/main/docs/osv_v3_feed.md

This has had a few weeks of testing now in the test instance, and behaviour observed there is the desired behaviour for the CGA- entries.

Signed-off-by: Dimitri John Ledkov <dimitri.ledkov@chainguard.dev>

### DIFF
--- a/source.yaml
+++ b/source.yaml
@@ -141,7 +141,7 @@
 
 - name: 'chainguard'
   versions_from_repo: False
-  rest_api_url: 'https://packages.cgr.dev/chainguard/osv/all.json'
+  rest_api_url: 'https://advisories.cgr.dev/chainguard/v3/osv/all.json'
   type: 2
   ignore_patterns: ['^(?!CGA-).*$']  # NOTE: Not currently supported for REST sources
   directory_path: 'chainguard/osv'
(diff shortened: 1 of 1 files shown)


xnox:
cc @another-rex

crosleyzack (approved):
secondary approval from Chainguard for this update

xnox:
cc @michaelkedar

another-rex (approved):

xnox:
> Needs /gcbrun from a collaborator

can someone please add a comment to run github action CI?

another-rex:
This is pretty much good to go, we're just timing the merge a bit later because it's a huge number of records :).

xnox:
> This is pretty much good to go, we're just timing the merge a bit later because it's a huge number of records :).

we generate them in the 64GB RAM instances in the batch GCP job and store in a bucket.... 

...maybe we can grant you authenticated bucket access if that at all helps for the initial load?

xnox:
Are weekends better to land this, or inverse?

Note this is the same data as in the test.osv.dev - thus possibly can backfill from there as well.

another-rex:
It's all imported now! Turned out to be relatively straight forward :). See https://osv.dev/list?ecosystem=Chainguard

The total count on the website might still take up to a day to refresh.

PR #5959, opened 2026-09-02
chore(dep): update go to v1.27, golangci-lint to v2.13, and fix all the lint

Ly-Joey:
Update Go module directives and Docker base images to Go `1.27.0`.

Go 1.27 introduced a new compiler export data format (v4) for compiled standard library packages. Older versions of `golangci-lint` fail during typechecking when encountering Go 1.27 standard library export data, so we are updating it to `v2.13.0` as well.

Also added a golangci config file to disable `exhaustruct_v5` and fixed ALL the new lint errors found.


another-rex (dismissed):

another-rex (commented):
also update the go version in lint.yaml in .workflows

cuixq (approved):

PR #5965, opened 2026-09-02
fix(go-website): minor fixes

michaelkedar:
A couple of fixes for the go website
- Sort the ecosystem names on the `/list` page alphabetically and case-insensitively
- Handle weird double-marshalled upstream hierarchies that can be found in the Datastore
- Don't check for existence of downstream vulns in the website - the fact they exist in the database implies that they should exist
- Cancel the request's context when the `writeTimeout` has been exceeded (so we don't waste time on work that can't be sent). And increase the `writeTimeout` to 60s
- Also, adjusted one of the tests to hopefully be less flaky (I think there was a race condition where ecosystem counts were being cached at the same time the entities were being written to the datastore emulator)

### DIFF
--- a/go/internal/database/datastore/relations.go
+++ b/go/internal/database/datastore/relations.go
@@ -5,10 +5,12 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"log/slog"
 	"slices"
 
 	"cloud.google.com/go/datastore"
 	"github.com/google/osv.dev/go/internal/models"
+	"github.com/google/osv.dev/go/logger"
 )
 
 type RelationsStore struct {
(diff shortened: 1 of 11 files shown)


go/internal/database/datastore/relations.go
```diff
@@ -106,11 +108,23 @@ func (s *RelationsStore) GetUpstreamHierarchy(ctx context.Context, id string) (*
...
+	// Attempt to unmarshal it into a string, then into the map
+	var str string
+	err2 := json.Unmarshal(upstreamGroup.UpstreamHierarchy, &str)
+	if err2 == nil {
+		err3 := json.Unmarshal([]byte(str), &rawHierarchy)
+		if err3 == nil {
+			logger.WarnContext(ctx, "double unmarshal needed for upstream hierarchy", slog.String("id", id))
+			return ComputeUpstreamHierarchy(id, rawHierarchy)
+		}
 	}
```
another-rex, line 125: This feels like a test would be appropriate.

go/internal/website/vulnerability.go
```diff
@@ -114,6 +114,7 @@ func (s *Server) handleVulnerabilityDetails(w http.ResponseWriter, r *http.Reque
 
 	// Collect all the vuln IDs mentioned by this vuln to check the database (in parallel) for existence.
 	candidateIDs := make(map[string]struct{})
+	knownIDs := make(map[string]bool)
```
another-rex: Why not struct{} like the one above?

go/internal/website/vulnerability.go
```diff
@@ -132,16 +133,18 @@ func (s *Server) handleVulnerabilityDetails(w http.ResponseWriter, r *http.Reque
...
+				knownIDs[c] = true
 			}
 		}
 	}
 
 	var mu sync.Mutex
-	knownIDs := make(map[string]bool)
 	g, ctx := errgroup.WithContext(r.Context())
 	g.SetLimit(20) // some vulns can have lots of related/alias/upstream/downstream
 	for id := range candidateIDs {
```
another-rex, line 152: Will candidate ID have overlap with knownIDs? Feels like we can skip checking if it's already in knownID

another-rex (dismissed):

go/internal/database/datastore/relations.go
michaelkedar: test added

go/internal/website/vulnerability.go
michaelkedar: I don't remember. I think I was maybe going to use it in the template, but that  doesn't seem to happen.
Changed to a `map[string]struct{}`

go/internal/website/vulnerability.go
michaelkedar: It's probably unlikely, but I've deleted the `knownIDs` from the `candidateIDs` map anyway.

another-rex (approved):

PR #5971, opened 2026-09-03
refactor(vulnfeeds): utilise a strategy pattern on a CNA basis to reduce carveout logic

jess-lowe:
Refactors CVE5 version extraction from nested fallback branches into a modular, priority-based strategy pipeline. This will be so much easier to extend as we get better insights on CNA quirks.

### Highlights

- **Strategy Pipeline**: Introduced [`VersionStrategy`](file:///usr/local/google/home/jesslowe/Documents/osv.dev/vulnfeeds/conversion/cve5/strategies.go) with configurable priority levels (`PriorityFirst`, `PriorityStandard`, `PriorityLastResort`) and CNA-specific presets (Default, GitHub, MITRE, Linux).
- **Extracted Strategies**: Isolated parsing for standard bounds, split/unspecified ranges, `changes` blocks, string range expressions, git commit hashes, CPE applicability, and single-version heuristics.
- **Direct Git Ranges**: Bypass tag-to-commit normalization when ranges already contain commit SHAs via [`IsDirectGitRange`](file:///usr/local/google/home/jesslowe/Documents/osv.dev/vulnfeeds/conversion/common.go#L746).
- **Early Exit**: If repos are unavailable, cleanly record unresolved ranges to `database_specific.unresolved_ranges` without running redundant fallbacks.
- **Tests**: Added comprehensive unit tests for each strategy, priority ordering, and split/git ranges.


vulnfeeds/conversion/cve5/version_extraction_test.go
```diff
@@ -615,6 +734,22 @@ func TestExtractVersions(t *testing.T) {
 					},
 				}},
 		},
+		{
+			name:        "CVE-2026-67185",
+			cve:         loadTestCVE(t, filepath.Join("..", "..", "cmd", "converters", "cve", "cve5", "bulk-converter", "cvelistV5", "cves", "2026", "67xxx", "CVE-2026-67185.json")),
```
another-rex: Please fix.

vulnfeeds/conversion/cve5/strategies/strategy.go
```diff
@@ -0,0 +1,58 @@
...
+const acceptableQuality = vulns.Spaces
+
+// VersionStrategy defines the contract for an individual version extraction strategy.
+// Strategies are evaluated in the sequential order they appear in the configured slice.
+type VersionStrategy interface {
+	// Name returns a human-readable identifier for the strategy.
+	Name() string
+	// Extract attempts to extract OSV version ranges from a CVE5 Versions entry.
+	// Returns the extracted ranges, the detected VersionRangeType, and true if this strategy handled the entry.
+	Extract(vers models.Versions, affected models.Affected, metrics *models.ConversionMetrics) (ranges []models.RangeWithMetadata, vrt VersionRangeType, handled bool)
```
another-rex: I like the strategy system, but the contract seems to be in the wrong place? 

I.e. it looks like you are being called per version, but you also have strategies that extract at the "affected" level.

So you end up with the same thing extracted multiple times (e.g. if versions extraction always fails, but CPE extraction succeeds).

another-rex (commented):
I like the refactor! I think we might need to adjust it a bit though so we're not dealing with all these manual checks within each strategy.

PR #5980, opened 2026-09-04
fix(alpine): do not derive introduced from an unrelated NVD CPE branch

### LINKED ISSUE #5634: Data quality issue with ALPINE-CVE-2018-18311 / -18312 (inverted version range, false positives)
### Affected records
- `ALPINE-CVE-2018-18311`
- `ALPINE-CVE-2018-18312`

### Problem
Both records carry an **inverted version range** on every Alpine branch:

```json
"events": [ { "introduced": "5.28.0" }, { "fixed": "5.26.3-r0" } ]
```

The `fixed` version (`5.26.3-r0`) sorts *below* the `introduced` version (`5.28.0`), so the
range `x >= 5.28.0 AND x < 5.26.3-r0` is unsatisfiable. Scanners drop the invalid `fixed`
event and flag **all** perl `>= 5.28.0` as vulnerable — including current `perl 5.42.2-r0`
on Alpine `v3.24` — for a 2018 CVE that was fixed in perl 5.26.3.

Source JSON:
- https://storage.googleapis.com/cve-osv-conversion/alpine/ALPINE-CVE-2018-18311.json
- https://storage.googleapis.com/cve-osv-conversion/alpine/ALPINE-CVE-2018-18312.json

### Evidence the range is wrong
Alpine's own secdb records the fix at `5.26.3-r0`, and the current branch ships `5.42.2-r0`:

```
# https://secdb.alpinelinux.org/v3.24/main.json  (pkg: perl)
"secfixes": {
  "5.26.3-r0": ["CVE-2018-18311", "CVE-2018-18312", "CVE-2018-18313", "CVE-2018-18314"],
  ...
  "5.42.2-r0": [...]        # current version on v3.24
}
```

The advisory text itself states *"Perl before 5.26.3 and 5.28.x before 5.28.1"*, so
perl 5.42.2 is not affected.

### Suggested correction (all Alpine branches)
```json
"events": [ { "introduced": "0" }, { "fixed": "5.26.3-r0" } ]
```

### Likely root cause
The alpine converter (`vulnfeeds/cmd/converters/alpine`) appears to have taken `introduced`
from the NVD CPE range (`5.28.x`) instead of emitting `introduced: 0` for a secfix-only
advisory, then paired it with the Alpine secfix `5.26.3-r0` — two different perl branches
(5.26.x vs 5.28.x) merged into one inverted range. The sibling records `ALPINE-CVE-2018-18313`
and `ALPINE-CVE-2018-18314` share the same secfix line and are worth checking for the same
defect.

This is the same class of malformed-Alpine-range false positive as #4165.

katzj:
findIntroducedVersion picked the first CPE match with a versionStartIncluding, regardless of which vulnerable branch it described. When a CVE was fixed independently on multiple branches (e.g. perl CVE-2018-18311, fixed at 5.26.3 and separately at 5.28.1), this could pair an introduced version from one branch with the Alpine secfix from a different, unrelated branch, producing an inverted and unsatisfiable range (introduced > fixed). Scanners then treated every higher version as affected, causing false positives long after the real fix.

Only use a candidate introduced version when it sorts before the Alpine fixed version, picking the closest qualifying one and falling back to "0" otherwise.

Fixes #5634

### DIFF
--- a/vulnfeeds/cmd/converters/alpine/main.go
+++ b/vulnfeeds/cmd/converters/alpine/main.go
@@ -202,7 +202,7 @@ func generateAlpineOSV(allAlpineSecDb map[string][]VersionAndPkg, allCVEs map[mo
 				continue
 			}
 
-			introduced := findIntroducedVersion(cve.CVE, verPkg.Pkg)
+			introduced := findIntroducedVersion(cve.CVE, verPkg.Pkg, verPkg.Ver)
 			pkgInfo := vulns.PackageInfo{
 				PkgName: verPkg.Pkg,
 				VersionInfo: models.VersionInfo{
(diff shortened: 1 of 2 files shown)


G-Rath:

PR #5981, opened 2026-09-05
fix(website): display per-package severity

ashmod:
Small follow-up to the Go website migration. The per-package severity section on vulnerability pages was hidden because `Severities` was not populated for each affected package. The section was already rendered in the template. 

This reuses the existing severity conversion for each `affected[]` entry.

Before/After on `BIT-activemq-2026-33227`:

before

after


michaelkedar (approved):
Thanks for catching this!

PR #5987, opened 2026-09-08
fix(importer): close REST response body on HTTP error

### LINKED ISSUE #5985: REST source record leaks response body on HTTP error
**Describe the bug**

`go/internal/importer/rest.go` does not close the HTTP response body when `restSourceRecord.Open` receives a 4xx/5xx response.

The method calls `r.cl.Do(req)`, then returns an error immediately when `resp.StatusCode >= http.StatusBadRequest`. On the success path it returns `resp.Body` to the caller, which can then close it. On the error path, however, the body is never returned and is not closed locally.

That means repeated REST-source failures can leave response resources unreleased and interfere with HTTP connection reuse until cleanup happens elsewhere.

**To Reproduce**

1. Use a test HTTP server whose handler returns a 4xx or 5xx response with a response body.
2. Construct a `restSourceRecord` pointing at that server.
3. Call `Open`.
4. Observe that `Open` returns an error without closing the response body.

A regression test can use a custom `RoundTripper`/`ReadCloser` that records whether `Close` was called on the error path.

**Expected behaviour**

`restSourceRecord.Open` should close the response body before returning an HTTP-status error. The successful path should continue returning the body to the caller unchanged.

The smallest fix appears to be a local `resp.Body.Close()` (or equivalent safe handling) before the status-error return, plus a focused regression test.

**Additional context**

I searched existing issues and PRs for `restSourceRecord`, `failed to fetch REST API`, and response-body/REST importer terms. I did not find an existing report for this leak. Issue #3327 was a separate Chainguard 403 availability incident.

I have not started implementation, per the repository's issue-first/assignment requirement. I am happy to implement the narrow fix and regression test if this issue is assigned to me.

keeltrace:
## Overview

Close the REST source response body before returning an HTTP status error from `restSourceRecord.Open`.

Fixes #5985

## Details

`Open` returns `resp.Body` to callers on success, but previously returned early on 4xx/5xx responses without closing it. This closes the body on the error path and adds a regression test using a tracking `ReadCloser`.

## Testing

- `go test -v ./internal/importer -run 'TestRestSourceRecord_Open'`
- `go test ./internal/importer`
- `git diff --check`

AI-assisted contribution: I used an AI coding agent during investigation and implementation. I reviewed the final diff and ran the verification listed above. I am responsible for the submitted change and follow-up.

### DIFF
--- a/go/internal/importer/rest.go
+++ b/go/internal/importer/rest.go
@@ -39,6 +39,7 @@ func (r restSourceRecord) Open(ctx context.Context) (io.ReadCloser, error) {
 	}
 
 	if resp.StatusCode >= http.StatusBadRequest {
+		resp.Body.Close()
 		return nil, fmt.Errorf("failed to fetch REST API: %d %s for %s", resp.StatusCode, resp.Status, u)
 	}
 
--- a/go/internal/importer/rest_test.go
+++ b/go/internal/importer/rest_test.go
@@ -4,11 +4,28 @@ import (
 	"io"
 	"net/http"
 	"net/http/httptest"
+	"strings"
 	"testing"
 	"time"
 
 	"github.com/google/osv.dev/go/internal/models"
 )
+
+type trackingReadCloser struct {
+	io.Reader
+	closed bool
+}
+
+func (r *trackingReadCloser) Close() error {
+	r.closed = true
+	return nil
+}
+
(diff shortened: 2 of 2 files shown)


michaelkedar (commented):
Thanks, happy to merge if you could fix the lint error

michaelkedar (approved):

keeltrace:
Ty!

PR #6009, opened 2026-09-10
feat(gitter): New endpoint for commit history and patch extraction

Ly-Joey:
Adds `/commit-diffs` endpoint to return structured commit data (SHA, timestamp, message, diff patches, and modified files) relative to a target branch and last sync point.

The last-sync reference can be a commit SHA or timestamp. 
When both are provided, the commit hash is prioritized, with the timestamp acting as a fallback if the hash is invalid (e.g., from a git commit --amend or rebase).

### DIFF
--- a/go/cmd/gitter/git.go
+++ b/go/cmd/gitter/git.go
@@ -42,8 +42,8 @@ func prepareCmd(ctx context.Context, dir string, env []string, name string, args
 }
 
 // runCmd executes a command with context cancellation handled by sending SIGINT.
-// It logs cancellation errors separately as requested.
-func runCmd(ctx context.Context, dir string, env []string, name string, args ...string) error {
+// It logs cancellation errors separately as requested
+func runCmd(ctx context.Context, dir string, env []string, name string, args ...string) ([]byte, error) {
 	cmd := prepareCmd(ctx, dir, env, name, args...)
 	out, err := cmd.CombinedOutput()
 
@@ -51,10 +51,10 @@ func runCmd(ctx context.Context, dir string, env []string, name string, args ...
 		if ctx.Err() != nil {
 			// Log separately if cancelled
 			logger.DebugContext(ctx, "Command cancelled", slog.String("cmd", name), slog.Any("err", ctx.Err()))
-			return fmt.Errorf("command %s cancelled: %w", name, ctx.Err())
+			return out, fmt.Errorf("command %s cancelled: %w", name, ctx.Err())
 		}
 
-		return fmt.Errorf("command %s failed: %w, output: %s", name, err, out)
+		return out, fmt.Errorf("command %s failed: %w, output: %s", name, err, out)
 	}
 
 	logger.DebugContext(ctx, "Git command executed",
@@ -63,12 +63,13 @@ func runCmd(ctx context.Context, dir string, env []string, name string, args ...
 		slog.String("output", string(out)),
 	)
 
-	return nil
+	return out, nil
 }
 
 // cloneRepo clones a git repository into repoPath.
 func cloneRepo(ctx context.Context, repoURL string, repoPath string) error {
-	return runCmd(ctx, "", []string{"GIT_TERMINAL_PROMPT=0"}, "git", "clone", "--", repoURL, repoPath)
+	_, err := runCmd(ctx, "", []string{"GIT_TERMINAL_PROMPT=0"}, "git", "clone", "--", repoURL, repoPath)
+	return err
 }
 
 // Attempt to recover from git fetch errors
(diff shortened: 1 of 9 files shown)


go/cmd/gitter/repository.go
```diff
@@ -1195,3 +1226,260 @@ func (r *Repository) GetFileContent(ctx context.Context, ref, path string) ([]by
...
+		hash := string(metaLines[0])
+		timestampSec, err := strconv.ParseInt(string(metaLines[1]), 10, 64)
+		if err != nil {
+			logger.WarnContext(ctx, "Invalid timestamp in git log output", slog.Any("error", err))
+			continue
+		}
+
+		var msg string
+		if len(metaLines) == 3 {
+			msg = string(bytes.TrimSpace(metaLines[2]))
```
another-rex: A lot of these places where  we are converting bytes to string might need: `strings.ToValidUTF8()`. Especially if we're cutting by length.a

go/cmd/gitter/repository.go
```diff
@@ -1195,3 +1226,260 @@ func (r *Repository) GetFileContent(ctx context.Context, ref, path string) ([]by
...
+					continue
+				}
+				filesChanged = append(filesChanged, change)
+			} else if bytes.HasPrefix(line, []byte("diff --")) {
+				// Section 2: First line of the unified diff patch reached.
+				// Everything from lineStart to the end of diffContent belongs to the unified diff patch.
+				patchBytes := diffContent[lineStart:]
+				if len(patchBytes) > maxPatchBytes {
+					// Truncate the patch if it exceeds maxPatchBytes.
+					patch = string(patchBytes[:maxPatchBytes]) + fmt.Sprintf("\n\n[Diff truncated: exceeded max patch size of %d bytes]", maxPatchBytes)
```
another-rex: E.g. definitely here.

go/cmd/gitter/repository.go
```diff
@@ -1195,3 +1226,260 @@ func (r *Repository) GetFileContent(ctx context.Context, ref, path string) ([]by
...
+		toCommit, err = r.resolveCommit(ctx, targetBranch)
+		if err != nil {
+			return "", "", nil, fmt.Errorf("failed to resolve target branch %q: %w", targetBranch, err)
+		}
+	}
+
+	// Step 2: Validate lastScanCommit
+	var fromCommit string
+	if lastScanCommit != "" {
+		if lastScanCommit == toCommit || strings.HasPrefix(toCommit, lastScanCommit) {
```
another-rex: why the prefix match?

go/internal/gitter/pb/repository/repository.proto
```diff
@@ -88,3 +90,30 @@ message FileContentRequest {
...
+    string message = 3;
+    string patch = 4;
+    repeated FileChange files_changed = 5;
+    bool patch_truncated = 6;
+}
+
+message CommitDiffsRequest {
+    string url = 1;
+    string branch = 2;                            // Optional, defaults to remote default branch (origin/HEAD)
+    string last_scan_commit = 3;                  // Optional, SHA of last scanned commit, preferred over last_scan_time when provided
```
another-rex: Maybe rename this to last sync commit to match FilesDiffs endpoint.

go/internal/gitter/pb/repository/repository.proto
```diff
@@ -88,3 +90,30 @@ message FileContentRequest {
...
+    google.protobuf.Timestamp last_scan_time = 4; // Optional, timestamp of last scan. Used as fallback if last_scan_commit is not provided or is no longer on the branch.
+    bool newest_first = 5;                        // Optional, if true, newest commits first. Defaults to false (chronological order).
+    repeated string include_paths = 6;            // Optional, paths/globs to include in diffs
+    repeated string exclude_paths = 7;            // Optional, paths/globs to exclude from diffs
+}
+
+message CommitDiffsResponse {
+    string url = 1;
+    string branch = 2;
+    string head_commit = 3;
```
another-rex: Also here could match latest_commit of FIlesDiffs

go/cmd/gitter/repository_test.go
another-rex, line 20: Btw, can you update this to add
```
cmd.Env = append(os.Environ(), "GIT_CONFIG_GLOBAL=/dev/null", "GIT_CONFIG_NOSYSTEM=1")
```

Otherwise when I run the tests my system config causes issues.

another-rex (dismissed):
Just some nits

go/cmd/gitter/repository.go
cuixq: +1 to valid UTF-8

go/cmd/gitter/git.go
```diff
@@ -63,12 +63,13 @@ func runCmd(ctx context.Context, dir string, env []string, name string, args ...
 		slog.String("output", string(out)),
```
cuixq: `runCmd` will now also be used for `git log`, the whole output could be huge. shall we consider truncating the logged output?

go/internal/gitter/pb/repository/repository.proto
cuixq: also probably rename `last_scan_time` to `last_sync_time`

go/cmd/gitter/git.go
Ly-Joey: I've changed to using `prepareCmd` (like in `buildCommitGraph`) so we have more flexibility with the output and logging. 
That said, it's still a good idea to truncate large output in general, so I've added that as well :D

another-rex (approved):

cuixq (approved):

PR #6036, opened 2026-09-17
ecosystem: normalize NuGet, crates.io, npm, and Packagist package names for matching

kanokwut:
## Overview

Implements `NormalizePackageName` for NuGet, crates.io, npm, and Packagist so that a valid non-canonical spelling of a package name matches the same stored name. These ecosystems previously left names unchanged, because only PyPI and Echo implemented `PackageNameNormalizer`.

I did not open a separate issue before this PR. Happy to open one if you would prefer to track it that way.

## Details

- NuGet, npm, and Packagist names are case-insensitive, and crates.io treats a dash and an underscore as equivalent, so a non-canonical spelling did not match the stored name.
- npm and crates.io get their own types, since they previously shared the generic `semverEcosystem` with case-sensitive ecosystems such as Go, which must stay case-sensitive.
- Both the query path (`vulnerability_matching.go`) and the `namenormalize` enricher call `NormalizePackageName`, so this fixes both sides symmetrically.
- Rollout note: existing stored `AffectedVersions` records are keyed on their un-normalized names, so this change should be paired with reprocessing the stored corpus through the `namenormalize` enricher. Otherwise a normalized query would stop matching the still-unnormalized stored names.

## Testing

- Updated unit tests in `go/osv/ecosystem/wrapper_test.go` and `go/internal/worker/pipeline/namenormalize/namenormalize_test.go`.
- Ran `go test ./osv/ecosystem/...` and the `namenormalize` package tests; both pass.
- Verified case-sensitive ecosystems (Go, Maven, RubyGems) are left unchanged and that PyPI normalization is unaffected.
- The API query snapshots may need regenerating with `make update-api-snapshots`, which I was not able to run without GCP project access.

---

This pull request was AI-assisted.


G-Rath:

PR #6044, opened 2026-09-22
feat: fallback to truncated description if advisory does not have a summary

G-Rath:
Resolves https://github.com/DrupalSecurityTeam/drupal-advisory-database/pull/232


G-Rath:
@another-rex I think I've done this right so hopefully the failures are unrelated, but afaik I cannot access the logs to see for myself 😅

